PR #1, opened 2019-10-01
tfbridge: Allow a field to be marked as Computed

stack72:
Fixes: https://github.com/pulumi/pulumi-aws/issues/752

This will override the TF schema and mark it as computed

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -149,6 +149,9 @@ type SchemaInfo struct {
 
 	// to remove empty object array elements
 	SuppressEmptyMapElements *bool
+
+	// this will make the parameter as computed and not allow the user to set it
+	MarkAsComputedOnly *bool
 }
 
 // ConfigInfo represents a synthetic configuration variable that is Pulumi-only, and not passed to Terraform.
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -734,7 +734,7 @@ func (g *generator) gatherDataSource(rawname string,
 		cust := info.Fields[arg]
 
 		// Remember detailed information for every input arg (we will use it below).
-		if input(args[arg]) {
+		if input(args[arg]) && !isComputedOverride(cust) {
 			argvar := propertyVariable(arg, sch, cust, parsedDocs.Arguments[arg], "", "", false /*out*/)
 			fun.args = append(fun.args, argvar)
 			if !argvar.optional() {
@@ -862,6 +862,12 @@ func (g *generator) emitProjectMetadata(pack *pkg) error {
 // input checks whether the given property is supplied by the user (versus being always computed).
 func input(sch *schema.Schema) bool {
 	return sch.Optional || sch.Required
+}
+
+// isComputedOverride checks whether the given property is marked to be a computed only value. This
+// ignores any sign of it being an input value
+func isComputedOverride(info *tfbridge.SchemaInfo) bool {
+	return info.MarkAsComputedOnly != nil && *info.MarkAsComputedOnly
 }
 
 // propertyName translates a Terraform underscore_cased_property_name into a JavaScript camelCasedPropertyName.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,6 @@ CHANGELOG
 * Update Terraform bridge to be based on v1.0.0 of the Terraform Plugin SDK
 * Add support to specify a custom package name for NodeJS package
 * Add ability to pass the TF provider version that the pulumi provider was generated against
+* Allow a schema variable to be overridden as being `Computed`
 
 ---
-


pkg/tfbridge/info.go
```diff
@@ -148,6 +148,9 @@ type SchemaInfo struct {
 
 	// to remove empty object array elements
 	SuppressEmptyMapElements *bool
+
+	// this will make the parameter as computed and not allow the user to set it
+	MarkAsComputed *bool
```
lukehoban: Computed doesn’t normally mean “output only”. It means “also output”. Right?

Per the issue this is trying to address - the property in question is computed + optional. It’s the optional piece that causes the problem I believe.

stack72: We can’t remove the optional - TF scheme requires something to be optional + computer, required or computed and some combination

This new override will drop anything else on that specific field and make that parameter computed only - it effectively forgets that the optional is there (or rather, that’s the intent)

This gives us control where these optional + computed happens. There are a lot of data sources that do this

pkg/tfbridge/schema.go
```diff
@@ -365,6 +365,14 @@ func MakeTerraformInput(res *PulumiResource, name string,
 		v = nv
 	}
 
+	// If we want to mark the property as Computed then we need to do so before we process the value
+	// this will override any specific usage of Optional. This is useful in scenarios where the data source
+	// has marked the parameter as Optional + Computed but they actually just mean that it's computed
+	// e.g. tags https://github.com/pulumi/pulumi-aws/issues/752
+	if ps != nil && ps.MarkAsComputed != nil && *ps.MarkAsComputed {
```
pgavlin: Do we actually need this? Seems like we should never get here...

pkg/tfbridge/info.go
pgavlin: The way we interpret these values is like so:
1. `Computed` only: this property only exists in the resource's state. It is not an input property.
2. `Optional` or `Required`: this property is an input, and its input value (if present) is also copied verbatim to the state.
3. `Optional` or `Required` _plus_ `Computed`: this property might be an input or might only be present in the state. If the property is an input, it is possible that the value in the state may differ (though in general this should not be the case w/o `CustomizeDiff`, as differing values will cause spurious changes in `Diff`).

What we want here is the ability to treat a property as if it was marked `Computed` only.

stack72: 👍 exactly that!

pkg/tfbridge/schema.go
stack72: I will happily remove the `&& *ps.MarkAsComputed` - it was purely there to check to make sure that someone didn't make this false - we can assume that if it's there, it will always be true :)

pgavlin: Sorry, I meant that I don't think we need this condition at all. It might even be incorrect, as we do call `MakeTerraformInputs` with values from the state, where these fields may be populated.

stack72: Oh interesting, the intent of this check was to say:

1. Is this a value that we are trying to override in the schema to be Computed
2. If so, then actually mark the property value as Computed

Do you not think this is the way to do this?

stack72: Oh interesting, the intent of this check was to say:

1. Is this a value that we are trying to override in the schema to be Computed
2. If so, then actually mark the property value as Computed

Do you not think this is the way to do this?

pgavlin: No, I don't think we need this here. We really only use `Computed` in the schema to guide `tfgen`'s code generation w.r.t. inputs/state.

stack72: Ah interesting - so the only work that this needs outside of a change to `info.go` is that we make the changes in tfgen to pick up that it's no longer an optional property but now only a computed property?

If so, can you point me in the direction of where that would live in tfgen?

stack72:
We expect downstream deps to fail on this as they have not been migrated to the new SDK yet

pkg/tfgen/generate.go
```diff
@@ -864,6 +864,16 @@ func input(sch *schema.Schema) bool {
...
 }
 
+// isComputedOverride checks whether the given property is marked to be a computed only value. This
+// ignores any sign of it being an input value
+func isComputedOverride(info *tfbridge.SchemaInfo) bool {
+	if info.MarkAsComputedOnly != nil {
+		return *info.MarkAsComputedOnly
+	}
+
+	return false
```
```suggestion
	return info.MarkAsComputedOnly != nil && *info.MarkAsComputedOnly
```

stack72: done!

pgavlin: Looks like this change didn't actually get made?

stack72: My bad - I didn't push the change!

pkg/tfgen/generate.go
```diff
@@ -734,7 +734,7 @@ func (g *generator) gatherDataSource(rawname string,
 		cust := info.Fields[arg]
 
 		// Remember detailed information for every input arg (we will use it below).
-		if input(args[arg]) {
+		if input(args[arg]) && !isComputedOverride(cust) {
```
pgavlin: Probably better to have `input` make this decision in case there are other callers: presumably those callers will also need to know about `MarkAsComputedOnly`.

stack72: You mean like change the input func to look like this?

```
func input(sch *schema.Schema, info *tfbridge.SchemaInfo) bool {
	return sch.Optional || sch.Required && !(info.MarkAsComputedOnly != nil && *info.MarkAsComputedOnly)
}
```

pgavlin: Needs some parens, but yes:

```
func input(sch *schema.Schema, info *tfbridge.SchemaInfo) bool {
	return (sch.Optional || sch.Required) && !(info.MarkAsComputedOnly != nil && *info.MarkAsComputedOnly)
}
```

stack72:
@pgavlin I had to continue this in https://github.com/pulumi/pulumi-terraform-bridge/pull/48 as I deleted the fork this had come from :(

stack72:
Closed via #48

PR #2, opened 2019-10-02
Update bridge to depend on Pulumi v1.2.0

jen20:
The work in pulumi/pulumi#3183 changes the way that secrets are returned - they are now pointers instead of `resource.Secret` instances directly. This makes it necessary to expect pointers when constructing expected outputs from `MakeTerraformOuptuts`.

Besides that, this is a simple dependency upgrade.


stack72:
We expect this to fail on downstream generation until they are updated to use the new terraform sdk

jen20:
The downstream generation changes look to be related to the `pulumi` CLI not being found on `PATH` in each repo - despite it being installed successfully by each! This is unrelated to this change, however we also expect downstream testing to fail since none of the tested repos are using the SDK-based bridge yet. Consequently merging this and then investigating whatever has changed in the Travis environment to not properly reflect the `PATH` is probably best.

PR #37, opened 2019-10-02
Add support for custom package name for nodejs

stack72:
I migrated this PR from https://github.com/pulumi/pulumi-terraform/pull/364


stack72:
I still expect this to fail in downstream generation until the main 3 providers change to use the new sdk

stack72:
@fernandocarletti thanks so much for this work - apologies it took so long to merge it

PR #38, opened 2019-10-04
Ability to pass TF Provider Version in tfbridge.ProviderInfo

stack72:
This will allow us to inform the user which version of the TF provider
that the current pulumi provider was generated against

It will format in the package metadata as:

```
A Pulumi package for creating and managing Microsoft Azure cloud resources.
Based on terraform-provider-azurerm: version v1.3.5
```

Related to https://github.com/pulumi/pulumi-azure/issues/377

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -24,22 +24,23 @@ import (
 // ProviderInfo contains information about a Terraform provider plugin that we will use to generate the Pulumi
 // metadata.  It primarily contains a pointer to the Terraform schema, but can also contain specific name translations.
 type ProviderInfo struct {
-	P              *schema.Provider           // the TF provider/schema.
-	Name           string                     // the TF provider name (e.g. terraform-provider-XXXX).
-	ResourcePrefix string                     // the prefix on resources the provider exposes, if different to `Name`.
-	GitHubOrg      string                     // the GitHub org of the provider. Defaults to `terraform-providers`.
-	Description    string                     // an optional descriptive overview of the package (a default will be given).
-	Keywords       []string                   // an optional list of keywords to help discovery of this package.
-	License        string                     // the license, if any, the resulting package has (default is none).
-	Homepage       string                     // the URL to the project homepage.
-	Repository     string                     // the URL to the project source code repository.
-	Config         map[string]*SchemaInfo     // a map of TF name to config schema overrides.
-	ExtraConfig    map[string]*ConfigInfo     // a list of Pulumi-only configuration variables.
-	Resources      map[string]*ResourceInfo   // a map of TF name to Pulumi name; standard mangling occurs if no entry.
-	DataSources    map[string]*DataSourceInfo // a map of TF name to Pulumi resource info.
-	JavaScript     *JavaScriptInfo            // optional overlay information for augmented JavaScript code-generation.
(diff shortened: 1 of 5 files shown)


pkg/tfbridge/info.go
```diff
@@ -24,22 +24,23 @@ import (
...
+	Homepage          string                     // the URL to the project homepage.
+	Repository        string                     // the URL to the project source code repository.
+	Config            map[string]*SchemaInfo     // a map of TF name to config schema overrides.
+	ExtraConfig       map[string]*ConfigInfo     // a list of Pulumi-only configuration variables.
+	Resources         map[string]*ResourceInfo   // a map of TF name to Pulumi name; standard mangling occurs if no entry.
+	DataSources       map[string]*DataSourceInfo // a map of TF name to Pulumi resource info.
+	JavaScript        *JavaScriptInfo            // optional overlay information for augmented JavaScript code-generation.
+	Python            *PythonInfo                // optional overlay information for augmented Python code-generation.
+	Golang            *GolangInfo                // optional overlay information for augmented Golang code-generation.
+	TFProviderVersion string                     // the version of the TF provider on which this was based
```
pgavlin: Would you mind adding this to the serializable version of `ProviderInfo` as well? This could be useful for `tf2pulumi`.

stack72: Added as requested!

pkg/tfbridge/info.go
```diff
@@ -546,10 +547,11 @@ func (m *MarshallableDataSourceInfo) Unmarshal() *DataSourceInfo {
...
 type MarshallableProviderInfo struct {
-	Provider    *MarshallableProvider                  `json:"provider"`
-	Config      map[string]*MarshallableSchemaInfo     `json:"config,omitempty"`
-	Resources   map[string]*MarshallableResourceInfo   `json:"resources,omitempty"`
-	DataSources map[string]*MarshallableDataSourceInfo `json:"dataSources,omitempty"`
+	Provider          *MarshallableProvider                  `json:"provider"`
+	Config            map[string]*MarshallableSchemaInfo     `json:"config,omitempty"`
+	Resources         map[string]*MarshallableResourceInfo   `json:"resources,omitempty"`
+	DataSources       map[string]*MarshallableDataSourceInfo `json:"dataSources,omitempty"`
+	TFProviderVersion *string                                `json:"tfProviderVersion,omitempty"`
```
pgavlin: This can be a plain `string`.

stack72: Updated!

pkg/tfbridge/info.go
```diff
@@ -590,10 +598,16 @@ func (m *MarshallableProviderInfo) Unmarshal() *ProviderInfo {
...
 
-	return &ProviderInfo{
+	info := ProviderInfo{
 		P:           m.Provider.Unmarshal(),
 		Config:      config,
 		Resources:   resources,
 		DataSources: dataSources,
 	}
+
+	if m.TFProviderVersion != "" {
```
pgavlin: Any reason to do this conditionally? just setting the field in the initializer will have the same effect. Same feedback above.

stack72: You are correct - just being over cautious here - moved it back to the initializer

Issue #7, opened 2019-08-22
Intellisense docs missing for resource types


hausdorff:
We had this problem in Kubernetes too. The culprit seems to be these [`__init__.py` files](https://github.com/pulumi/pulumi-aws/blob/d42f7d16700e2cd9b68264cd4135b7de6f37af25/sdk/python/pulumi_aws/cloudfront/__init__.py). We're doing something like this:

```python
# coding=utf-8
# *** WARNING: this file was generated by the Pulumi Kubernetes codegen tool. ***
# *** Do not edit by hand unless you're certain you know what you are doing! ***

# Export this package's modules as members:
from .DaemonSet import *
from .Deployment import *
```

but we should be doing something like this:

```python
# coding=utf-8
# *** WARNING: this file was generated by the Pulumi Kubernetes codegen tool. ***
# *** Do not edit by hand unless you're certain you know what you are doing! ***

# Export this package's modules as members:
from .DaemonSet import (DaemonSet)
from .Deployment import (Deployment)
```

cc @lukehoban @stack72 @jen20 also see our fix here https://github.com/pulumi/pulumi-kubernetes/pull/635

stack72:
Hi @hausdorff 

So I have been looking at this issue to try and solve it. I have managed go get the following diff:

```
diff --git a/sdk/python/pulumi_aws/acm/__init__.py b/sdk/python/pulumi_aws/acm/__init__.py
index 13427994..ad307745 100644
--- a/sdk/python/pulumi_aws/acm/__init__.py
+++ b/sdk/python/pulumi_aws/acm/__init__.py
@@ -3,6 +3,6 @@
 # *** Do not edit by hand unless you're certain you know what you are doing! ***

 # Export this package's modules as members:
-from .certificate import *
-from .certificate_validation import *
-from .get_certificate import *
+from .certificate import Certificate
+from .certificate_validation import CertificateValidation
+from .get_certificate import get_certificate
```

As you can see, we are actually importing real classes now. 

Certificate is a class in certificate.py that looks as follows:

```
class Certificate(pulumi.CustomResource):
```

CertificateValidation is a class in certificate_validation.py:

```
class CertificateValidation(pulumi.CustomResource):
```

Those 2 are easy and should be correct. But the last one, get_certificate.py. This file has the following:

```
class GetCertificateResult:

class AwaitableGetCertificateResult(GetCertificateResult):

[...]

PR #44, opened 2019-10-14
Export the python resource types for Intellisense

stack72:
Fixes: #7

### DIFF
--- a/pkg/tfgen/generate_python.go
+++ b/pkg/tfgen/generate_python.go
@@ -292,7 +292,7 @@ func (g *pythonGenerator) emitIndex(mod *module, exports, submods []string) erro
 		}
 		w.Writefmtln("# Export this package's modules as members:")
 		for _, exp := range exports {
-			w.Writefmtln("from .%s import *", pycodegen.PyName(exp))
+			w.Writefmtln("from .%s import %s", pycodegen.PyName(exp), pyClassName(exp))
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/tfgen/generate_python.go
```diff
@@ -645,7 +645,7 @@ func (g *pythonGenerator) emitResourceType(mod *module, res *resourceType) (stri
         return tables._SNAKE_TO_CAMEL_CASE_TABLE.get(prop) or prop
 `)
 
-	return name, nil
+	return res.name, nil
```
pgavlin: Why was this change made? Can you give an example of the generated output?

pkg/tfgen/generate_python.go
```diff
@@ -292,7 +292,7 @@ func (g *pythonGenerator) emitIndex(mod *module, exports, submods []string) erro
 		}
 		w.Writefmtln("# Export this package's modules as members:")
 		for _, exp := range exports {
-			w.Writefmtln("from .%s import *", pycodegen.PyName(exp))
+			w.Writefmtln("from .%s import %s", pycodegen.PyName(exp), pyClassName(exp))
```
pgavlin: Aren't there more types in the resource files besides just the resource class itself?

stack72:
@pgavlin I had questions on both of these things in the original issue :/

PR #48, opened 2019-10-14
tfbridge: Allow a field to be marked as Computed

stack72:
Fixes: pulumi/pulumi-aws#752

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -596,7 +596,7 @@ func (g *generator) gatherResource(rawname string,
 		}
 
 		// If an input, generate the input property metadata.
-		if input(propschema) {
+		if input(propschema, propinfo) {
 			inprop := propertyVariable(key, propschema, propinfo, doc, rawdoc, "", false /*out*/)
 			if inprop != nil {
 				res.inprops = append(res.inprops, inprop)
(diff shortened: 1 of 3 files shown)


pkg/tfgen/generate.go
```diff
@@ -596,7 +596,7 @@ func (g *generator) gatherResource(rawname string,
 		}
 
 		// If an input, generate the input property metadata.
-		if input(propschema) {
+		if input(propschema, info.Fields[key]) {
```
```suggestion
		if input(propschema, propinfo) {
```

pkg/tfgen/generate.go
```diff
@@ -734,7 +734,7 @@ func (g *generator) gatherDataSource(rawname string,
 		cust := info.Fields[arg]
 
 		// Remember detailed information for every input arg (we will use it below).
-		if input(args[arg]) {
+		if input(args[arg], info.Fields[arg]) {
```
```suggestion
		if input(args[arg], cust) {
```

pgavlin:
Can you show an example codegen diff?

stack72:
@pgavlin 

this is the diff that it would generate:

```
	modified:   sdk/nodejs/kinesis/firehoseDeliveryStream.ts
diff --git a/sdk/nodejs/elasticloadbalancing/getLoadBalancer.ts b/sdk/nodejs/elasticloadbalancing/getLoadBalancer.ts
index 4906dd9a..13bdcd02 100644
--- a/sdk/nodejs/elasticloadbalancing/getLoadBalancer.ts
+++ b/sdk/nodejs/elasticloadbalancing/getLoadBalancer.ts
@@ -41,7 +41,6 @@ export function getLoadBalancer(args: GetLoadBalancerArgs, opts?: pulumi.InvokeO
     }
     const promise: Promise<GetLoadBalancerResult> = pulumi.runtime.invoke("aws:elasticloadbalancing/getLoadBalancer:getLoadBalancer", {
         "name": args.name,
-        "tags": args.tags,
     }, opts);

     return pulumi.utils.liftProperties(promise, opts);
@@ -55,7 +54,6 @@ export interface GetLoadBalancerArgs {
      * The unique name of the load balancer.
      */
     readonly name: string;
-    readonly tags?: {[key: string]: any};
 }

 /**
```

This is correct as it's no longer settable in the get request

PR #50, opened 2019-10-21
Fix downstream testing using GitHub Actions

jen20:
This is currently a work in progress to investigate what is involved in using GitHub Actions to speed up downstream testing and get better cross-operating-system testing of the bridge.


jen20:
I think this is good to go for now - there is more work to do, but this doesn't lose us (much) functionality (only the formatting check) and gains us downstream testing again.

jen20:
Merging before the full run since it should be no different to the last several after some basic tidying. A full run will be done for #54.

PR #51, opened 2019-10-21
Allow passing a TF Provider LicenseType

stack72:
This will allow us to tailor our docs to not always assume MPL 2.0
We can now pass a License Type and the docs will be updated accordingly

The default will still be MPL 2.0

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -21,6 +21,12 @@ import (
 	"github.com/pulumi/pulumi/pkg/tokens"
 )
 
+const (
+	MPL20LicenseType    TFProviderLicense = "MPL 2.0"
+	MITLicenseType      TFProviderLicense = "MIT"
+	Apache20LicenseType TFProviderLicense = "Apache 2.0"
+)
+
 // ProviderInfo contains information about a Terraform provider plugin that we will use to generate the Pulumi
 // metadata.  It primarily contains a pointer to the Terraform schema, but can also contain specific name translations.
 type ProviderInfo struct {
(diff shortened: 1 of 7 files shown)


pkg/tfbridge/info.go
```diff
@@ -41,10 +47,14 @@ type ProviderInfo struct {
 	Python            *PythonInfo                // optional overlay information for augmented Python code-generation.
 	Golang            *GolangInfo                // optional overlay information for augmented Golang code-generation.
 	TFProviderVersion string                     // the version of the TF provider on which this was based
+	TFProviderLicense *TFProviderLicense         // license that the TF provider is distributed under. Default `MPL 2.0`.
 
 	PreConfigureCallback PreConfigureCallback // a provider-specific callback to invoke prior to TF Configure
 }
 
+// TFProviderLicense is a way to be able to pass a license type for the downstream Terraform provider.
```
jen20: Isn't the TF provider an upstream relative to the bridged Pulumi provider?

stack72: You are correct! Fixed

pkg/tfgen/generate.go
```diff
@@ -986,3 +986,16 @@ func generateManifestDescription(info tfbridge.ProviderInfo) string {
...
+func getLicenseTypeURL(license tfbridge.TFProviderLicense) string {
+	switch license {
+	case tfbridge.MITLicenseType:
+		return "https://mit-license.org/"
+	case tfbridge.MPL20LicenseType:
+		return "https://www.mozilla.org/en-US/MPL/2.0/"
+	case tfbridge.Apache20LicenseType:
+		return "https://www.apache.org/licenses/LICENSE-2.0.html"
+	default:
+		return ""
```
justinvp:
```suggestion
		contract.Failf("Unrecognized license: %v", license)
```

?

stack72: good idea - I added it above the return "", it needs to have return "" otherwise it won't build. But it will fail before it gets there :)

PR #54, opened 2019-10-23
Extend type information in tfgen.

pgavlin:
Create rich type information during the gather phase. This type
information can then be used to drive code generation with fewer special
cases.

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -234,6 +234,21 @@ type propertyType struct {
 
 func makePropertyType(sch *schema.Schema, info *tfbridge.SchemaInfo, out bool, parsedDocs parsedDoc) *propertyType {
 	t := &propertyType{}
+
+	var elemInfo *tfbridge.SchemaInfo
+	if info != nil {
+		t.typ = info.Type
+		t.nestedType = info.NestedType
+		t.altTypes = info.AltTypes
+		t.asset = info.Asset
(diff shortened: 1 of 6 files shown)


pkg/tfgen/generate.go
```diff
@@ -201,6 +201,125 @@ type moduleMember interface {
...
+	element    *propertyType
+	properties []*variable
+
+	typ        tokens.Type
+	nestedType tokens.Type
+	altTypes   []tokens.Type
+	asset      *tfbridge.AssetTranslation
+}
+
+func makePropertyType(sch *schema.Schema, info *tfbridge.SchemaInfo, out bool, parsedDocs parsedDoc) *propertyType {
```
pgavlin: This centralizes the handling of `MaxItemsOne`, `TypeMap` with a `*schema.Resource` element, and doc attachment. These changes provide the most important simplifications in the downstream code.

jen20:
I've rebased this on top of #50 now that has been merged - this should run the downstream tests and give us confidence in the generation changes.

pgavlin:
@jen20 how do I see the generated source code and check for diffs?

jen20:
@pgavlin Right now you look at each job in the actions tab, and the diff is presented at the end of the "Test Downstream" step. In the fairly short term it will post a link to the GitHub diff back to the PR.

pgavlin:
Thanks! Looks like this is clean, then. 👍

PR #60, opened 2019-10-29
Update to v2 of downstream testing action

jen20:


jen20:
This needs some more work, I'll look at it in the morning.

jen20:
This looks like it's working now, I'll merge after a full run of green CI.

PR #70, opened 2019-11-11
Move Config to the root namespace

mikhailshilkov:
The doubly nested `Config.Config` in `Pulumi.Gcp.Config.Config.Project` seems weird:

image

This PR moves the class to the root namespace.

cc @pgavlin @CyrusNajmabadi

### DIFF
--- a/pkg/tfgen/generate_csharp.go
+++ b/pkg/tfgen/generate_csharp.go
@@ -439,8 +439,8 @@ func (g *csharpGenerator) emitConfigVariableType(w *tools.GenWriter, typ *proper
 
 // emitConfigVariables emits all config vaiables in the given module, returning the resulting file.
 func (g *csharpGenerator) emitConfigVariables(mod *module, nestedTypes *csharpNestedTypes) (string, error) {
-	// Create a Variables.cs file into which all configuration variables will go.
-	w, config, err := g.openWriter(mod, "Variables.cs")
+	// Create a Config.cs file into which all configuration variables will go.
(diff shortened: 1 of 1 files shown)


mikhailshilkov:
Hmm, so for the downstream tests to succeed I would need to delete the old `Variables.cs` file... I guess there no real way to do that?

pgavlin:
> Hmm, so for the downstream tests to succeed I would need to delete the old Variables.cs file... I guess there no real way to do that?

That's correct. If you've validated this manually, I have no reservations about going ahead and merging.

mikhailshilkov:
@pgavlin Yes, I tested it with Azure.

Now we need to make sure that whenever somebody tfgens a provider, they would manually delete `Variables.cs`.

In general, wouldn't it better to generate SDKs from clean slate instead of adding files to the existing directory? Obviously, the mixins would still need to come from somewhere.

pgavlin:
> In general, wouldn't it better to generate SDKs from clean slate instead of adding files to the existing directory? Obviously, the mixins would still need to come from somewhere.

We leave the files checked-in so that we can easily do diffing and understand the evolution of the SDKs over time.

pgavlin:
But we could e.g. kill the entire sdk tree during the build and regenerate it from scratch--that is orthogonal to whether or not the files are checked in.

mikhailshilkov:
Yes, the latter is what I was trying to say, sorry for being unclear. Delete, regenerate, but keep the same checkin-wise.

PR #71, opened 2019-11-11
Prefix terraform reference links with proper domain

zchase:
We currently have around ~20 links being generated on our docs that are broken. This is because they are reference links in the Terraform docs and we are just leaving them as is. This PR detects these specific reference links and appends `https://terraform.io` to the front of the link.


lukehoban:
@stack72 note that each of the links that this PR fixes are likely also unhelpful in practice for Pulumi users. We may want to find all of these and fix the docs themselves to be more helpful.

Issue #20, opened 2018-12-05
Emit a warning when a default value is unprojected


joeduffy:
Because Terraform default values may be expressed as functions -- such as populating a default from an environment value -- we cannot automatically project them.

Instead, we have a way of manually doing so.

Unfortunately, this can very easily lead to us missing default values when bringing up a new provider. We should consider issuing a warning to trigger a manual audit (and remediation) of all of them.

PR #72, opened 2019-11-14
Ensure that provider config with Default Values is replicated in Pulumi

stack72:
Fixes: #20


pkg/tfgen/generate.go
```diff
@@ -361,6 +362,16 @@ func (v *variable) optional() bool {
 		return true
 	}
 
+
+
+	// We are checking to see if all of the properties have their Defaults set
+	if v.schema != nil && v.schema.DefaultFunc != nil && v.info != nil && v.info.Default == nil {
+		// we are hitting a condition that the TF provider has a default Func and pulumi isn't setting one
+		glog.V(1).Infof("")
```
jen20: This seems unnecessary?

pkg/tfgen/generate.go
```diff
@@ -16,6 +16,7 @@ package tfgen
 
 import (
 	"fmt"
+	"github.com/golang/glog"
```
jen20: This shouldn't be imported here?

PR #73, opened 2019-11-18
Centralise the logic for AutoNaming

stack72:
AWS and Azure have their own custom implementation for AutoNaming
therefore, we are going to centralize this logic into 3 main functions

1. AutoName
This just allows you to pass the name and the maxLength

2. AutoNameTransform
This allows you to pass the name, maxLength and a transform

3. AutoNameWithCustomOptions
This allows setting different combinations for sizes of random values,
different separators and transforms

This centralizes all of the custom logic that we have scattered for
AutoNaming.

**EDIT**
as expected, this is the failure we are receiving

```
BUILD:
go install -ldflags "-X github.com/pulumi/pulumi-aws/pkg/version.Version=v1.10.0-alpha.1574075843+g059ff10d" github.com/pulumi/pulumi-aws/cmd/pulumi-resource-aws
# github.com/pulumi/pulumi-aws
##[error]./resources.go:1710:31: too many arguments in call to tfbridge.FromName
	have (bool, number, nil)
	want (tfbridge.AutoNameOptions)
##[error]Makefile:29: recipe for target 'provider' failed
```

### DIFF
--- a/pkg/tfgen/generate_csharp.go
+++ b/pkg/tfgen/generate_csharp.go
@@ -748,7 +748,8 @@ func (rg *csharpResourceGenerator) generateResourceClass() {
 
 	rg.w.Writefmtln("        public %s(string name, %s args%s, %s? options = null)",
 		className, argsType, argsDefault, optionsType)
-	rg.w.Writefmtln("            : base(\"%s\", name, args ?? ResourceArgs.Empty, MakeResourceOptions(options, \"\"))", rg.res.info.Tok)
+	rg.w.Writefmtln("            : base(\"%s\", name, args ?? ResourceArgs.Empty, MakeResourceOptions(options, \"\"))",
+		rg.res.info.Tok)
 	rg.w.Writefmtln("        {")
 	rg.w.Writefmtln("        }")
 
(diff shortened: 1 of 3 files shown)


stack72:
I expect that some downstream generation may fail but that's because of the change of how the functions for autonaming work now

pkg/tfgen/generate_csharp.go
```diff
@@ -748,7 +748,8 @@ func (rg *csharpResourceGenerator) generateResourceClass() {
 
 	rg.w.Writefmtln("        public %s(string name, %s args%s, %s? options = null)",
 		className, argsType, argsDefault, optionsType)
-	rg.w.Writefmtln("            : base(\"%s\", name, args ?? ResourceArgs.Empty, MakeResourceOptions(options, \"\"))", rg.res.info.Tok)
+	rg.w.Writefmtln("            : base(\"%s\", name, args ?? ResourceArgs.Empty, MakeResourceOptions(options, \"\"))",
```
jen20: This looks unrelated?

stack72: This was just failing the build in linting was all :D

PR #79, opened 2019-11-21
Restore old autoname maxlen behavior.

pgavlin:
The prior logic did not check the actual length of the name against its
max length. Performing this check breaks back-compat, especially for
Maxlen=0 (the default).


pgavlin:
GCP failure is expected.

PR #83, opened 2019-11-25
Fix a test.

pgavlin:
This test was broken by the "null properties" change I made late last
week.

Also, define `test_fast` s.t. `make default` runs the tests in this
repo.


pgavlin:
cc @mikhailshilkov

PR #89, opened 2019-12-06
Fix paths in Node.js generation in Windows

mikhailshilkov:
Currently, tfgen on Windows generates some relative paths in `.ts` files with backslashes. (I think it's the first time I ran a tfgen on AWS).

This PR changes relative paths to slashes even on Windows.

### DIFF
--- a/pkg/tfgen/generate_nodejs.go
+++ b/pkg/tfgen/generate_nodejs.go
@@ -1269,7 +1269,8 @@ func (g *nodeJSGenerator) emitTypeScriptProjectFile(pack *pkg, files []string) e
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


stack72:
@mikhailshilkov do you think we also have a problem with the filepaths for Python on Windows as well?

mikhailshilkov:
@stack72 I don't think so... Generating python doesn't yield any changes for me.

mikhailshilkov:
@pgavlin  It is, but we generate Windows-style paths on Windows (in master version). So we I gen aws I get a bunch of changes from `/` to `\`

pgavlin:
Ah! I misunderstood which direction we were mapping here. LGTM.

PR #92, opened 2019-12-11
Require explicit namespace names

mikhailshilkov:
I opened PRs for all provider repositories to define explicit names for C# namespaces. Now we can enforce having them on tfgen level instead of falling back to upper-casing the first character.


mikhailshilkov:
The build fails because the PRs aren't merged yet. We may hold off with merging here.

mikhailshilkov:
This is ready for review

PR #93, opened 2019-12-11
Prevent panic when getting named Schema from the SchemaMap

stack72:
When the name was an empty string, this was causing a panic


stack72:
The generation will fail here until the new Pulumi nuget version is picked up in the providers to deal with InvokeArgs for DotNet provider

Issue #86, opened 2019-11-28
Deprecated note from terraform are lost


boillodmanuel:
### Problem description

- I saw a product documentation issue on the following URL: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/azure/keyvault/#KeyVaultArgs-sku

The `sku` description is: 
> ) A `sku` block as described below.

On [terraform documentation](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/docs/r/key_vault.html.markdown#argument-reference), the `sku` description is:
> (Optional Deprecated)) A sku block as described below.

=> The deprecated note is lost
=> The extra `)` is fine because it's in terraform doc

### Affected product version(s)

latest one: 1.6.1

justinvp:
This looks like an issue with tfgen (transferring the issue there). We don't appear to be doing anything special with the schema's `Deprecated` property.

https://github.com/terraform-providers/terraform-provider-azurerm/blob/adcea34a24de175b9e34b1e20ebc597dcfbbb3ae/azurerm/resource_arm_key_vault.go#L72

Ideally, for these, we'd also emit `@deprecated` doc comments.

@stack72, mind taking a further look?

PR #96, opened 2020-01-07
Ensuring that schema deprecations in TF are accounted for in Pulumi

stack72:
Fixes: #86

We can do this easily for TypeScript and DotNet.

It's not easy in any way to deprecate parameters in Python. We would
need to write our own decorator for that and that would seriously slow
down the performance of our python implementation

FYI, the output of this will change the providers to have something like
this:

```
diff --git a/sdk/nodejs/listChart.ts b/sdk/nodejs/listChart.ts
index e6f38b3..3a20209 100644
--- a/sdk/nodejs/listChart.ts
+++ b/sdk/nodejs/listChart.ts
@@ -244,6 +244,7 @@ export interface ListChartState {
     /**
      * List of properties that should not be displayed in the chart legend (i.e. dimension names). All the properties are visible by default. Deprecated, please use `legendOptionsFields`.
      */
+    /** @deprecated Please use legend_options_fields */
     readonly legendFieldsToHides?: pulumi.Input<pulumi.Input<string>[]>;
     /**
```

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -177,6 +177,9 @@ type SchemaInfo struct {
 
 	// this will make the parameter as computed and not allow the user to set it
 	MarkAsComputedOnly *bool
+
+	// the deprecation message for the property
+	DeprecationMessage string
 }
 
 // ConfigInfo represents a synthetic configuration variable that is Pulumi-only, and not passed to Terraform.
@@ -276,29 +279,31 @@ type PreConfigureCallback func(vars resource.PropertyMap, config *terraform.Reso
 
 // MarshallableSchema is the JSON-marshallable form of a Terraform schema.
 type MarshallableSchema struct {
-	Type          schema.ValueType  `json:"type"`
-	Optional      bool              `json:"optional,omitempty"`
-	Required      bool              `json:"required,omitempty"`
-	Computed      bool              `json:"computed,omitempty"`
-	ForceNew      bool              `json:"forceNew,omitempty"`
-	Elem          *MarshallableElem `json:"element,omitempty"`
-	MaxItems      int               `json:"maxItems,omitempty"`
-	MinItems      int               `json:"minItems,omitempty"`
-	PromoteSingle bool              `json:"promoteSingle,omitempty"`
+	Type               schema.ValueType  `json:"type"`
+	Optional           bool              `json:"optional,omitempty"`
+	Required           bool              `json:"required,omitempty"`
+	Computed           bool              `json:"computed,omitempty"`
+	ForceNew           bool              `json:"forceNew,omitempty"`
+	Elem               *MarshallableElem `json:"element,omitempty"`
+	MaxItems           int               `json:"maxItems,omitempty"`
+	MinItems           int               `json:"minItems,omitempty"`
+	PromoteSingle      bool              `json:"promoteSingle,omitempty"`
+	DeprecationMessage string            `json:"deprecated,omitempty"`
 }
 
 // MarshalSchema converts a Terraform schema into a MarshallableSchema.
 func MarshalSchema(s *schema.Schema) *MarshallableSchema {
 	return &MarshallableSchema{
-		Type:          s.Type,
-		Optional:      s.Optional,
-		Required:      s.Required,
-		Computed:      s.Computed,
-		ForceNew:      s.ForceNew,
-		Elem:          MarshalElem(s.Elem),
-		MaxItems:      s.MaxItems,
-		MinItems:      s.MinItems,
-		PromoteSingle: s.PromoteSingle,
+		Type:               s.Type,
+		Optional:           s.Optional,
+		Required:           s.Required,
+		Computed:           s.Computed,
+		ForceNew:           s.ForceNew,
+		Elem:               MarshalElem(s.Elem),
+		MaxItems:           s.MaxItems,
+		MinItems:           s.MinItems,
+		PromoteSingle:      s.PromoteSingle,
+		DeprecationMessage: s.Deprecated,
 	}
 }
 
@@ -314,6 +319,7 @@ func (m *MarshallableSchema) Unmarshal() *schema.Schema {
 		MaxItems:      m.MaxItems,
 		MinItems:      m.MinItems,
 		PromoteSingle: m.PromoteSingle,
+		Deprecated:    m.DeprecationMessage,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


stack72:
This brings the code gen changes in - I still need to add docs changes as well. But that will follow after speaking with @justinvp

pkg/tfgen/generate_nodejs.go
```diff
@@ -716,6 +715,9 @@ func (g *nodeJSGenerator) emitPlainOldType(w *tools.GenWriter, pot *propertyType
 		} else if prop.rawdoc != "" {
 			g.emitRawDocComment(w, prop.rawdoc, "    ")
 		}
+		if prop.deprecationMessage() != "" {
+			w.Writefmtln("    /** @deprecated %s */", prop.deprecationMessage())
```
justinvp: This is emitting a new doc comment for the property, which means any prior doc comment emitted by `g.emitDocComment` or `g.emitRawDocComment` won't be "associated" with the property as far as TypeDoc is concerned. This means the docs will show that the property is deprecated, but any other doc comments themselves will no longer show up in the docs for that property.

I could see how that might be OK -- after-all, the property is deprecated, so the doc comments are less relevant. But if we have docs for a property, I think it's still useful to show them, even if the property is deprecated.

Current (not ideal):

```typescript
/**
 * Important documentation.
 */
/** @deprecated use bar instead. */
readonly foo?: pulumi.Input<string>;
```

Proposed:

```typescript
/**
 * @deprecated use bar instead.
 *
 * Important documentation.
 */
readonly foo?: pulumi.Input<string>;
```

The fix would be to include the `@deprecated %s` as part of emitting the doc comment in `emitDocComment` and `emitRawDocComment`, possibly something like (passing the deprecation message to those functions):

```go
		if prop.doc != "" && prop.doc != elidedDocComment {
			g.emitDocComment(w, prop.doc, prop.docURL, prop.deprecationMessage(), "    ")
 		} else if prop.rawdoc != "" {
 			g.emitRawDocComment(w, prop.rawdoc, prop.deprecationMessage(), "    ")
 		} else if prop.deprecationMessage() != "" {
 			w.Writefmtln("    /** @deprecated %s */", prop.deprecationMessage())
 		}
```

And inside `emitDocComment` and `emitRawDocComment` if the passed-in deprecation message isn't `""`, emit it at the top with a blank comment line afterwards, so any further doc comments aren't included as part of the deprecation tag.

stack72: @justinvp I just pushed the change you requested

pkg/tfgen/generate_nodejs.go
```diff
@@ -671,14 +670,19 @@ func (g *nodeJSGenerator) emitDocComment(w *tools.GenWriter, comment, docURL, pr
 		}
 	}
 
+	if deprecationMessage != "" {
+		w.Writefmt("%v * ", prefix)
+		w.Writefmt("%v * @deprecated %s */", prefix, deprecationMessage)
+	}
+
```
justinvp: This should be emitted after the docURL below.

stack72: 👍

pkg/tfgen/generate_go.go
```diff
@@ -408,6 +408,9 @@ func (g *goGenerator) emitPlainOldType(w *tools.GenWriter, pot *propertyType) {
 		} else if prop.rawdoc != "" {
 			g.emitRawDocComment(w, prop.rawdoc, "\t")
 		}
+		if prop.deprecationMessage() != "" {
+			w.Writefmtln("// Deprecated: %s", prop.deprecationMessage())
```
justinvp: Nit: Looks like this is missing a `\t` at the beginning:

Screen Shot 2020-01-09 at 2 58 55 PM

```suggestion
			w.Writefmtln("\t// Deprecated: %s", prop.deprecationMessage())
```

pkg/tfgen/generate_csharp.go
```diff
@@ -1021,6 +1022,10 @@ func (rg *csharpResourceGenerator) generateInputProperty(prop *variable, nested
 		}
 		emitCSharpPropertyDocComment(rg.w, prop, "        ")
 
+		if prop.deprecationMessage() != "" {
+			rg.w.Writefmtln("        [Obsolete(\"%s\")]", prop.deprecationMessage())
```
justinvp: Do we need to do anything to sanitize the string placed in the attribute here? For example, what if the deprecation message happens to have a `"` (quote), then the resulting code won't compile. Or `/` needs to be escaped. etc.

stack72: You are correct here - I just spoke to @mikhailshilkov about this and we came up with the fix

pkg/tfbridge/info.go
```diff
@@ -177,6 +177,9 @@ type SchemaInfo struct {
 
 	// this will make the parameter as computed and not allow the user to set it
 	MarkAsComputedOnly *bool
+
+	// the deprecation message for the property
+	Deprecated string
```
pgavlin: Can we name this `DeprecationMessage` for consistency with `ResourceInfo.DeprecationMessage`?

pkg/tfgen/generate.go
```diff
@@ -275,6 +275,13 @@ func makePropertyType(sch *schema.Schema, info *tfbridge.SchemaInfo, out bool,
 		t.kind = kindSet
 	}
 
+	// We should carry across any of the deprecation messages, to Pulumi, as per Terraform schema
+	if sch.Deprecated != "" {
```
```suggestion
	if sch.Deprecated != "" && elemInfo != nil {
```

pkg/tfbridge/info.go
```diff
@@ -284,6 +287,7 @@ type MarshallableSchema struct {
 	MaxItems      int               `json:"maxItems,omitempty"`
 	MinItems      int               `json:"minItems,omitempty"`
 	PromoteSingle bool              `json:"promoteSingle,omitempty"`
+	Deprecated    string            `json:"deprecated,omitempty"`
```
pgavlin:
```suggestion
	DeprecationMessage    string            `json:"deprecated,omitempty"`
```

pgavlin:
@stack72 here are the changes I think we need in pulumi/pulumi: https://github.com/pulumi/pulumi/pull/3865

stack72:
@pgavlin I'd love to get this merged this week - I need to close some issues around deprecation

PR #100, opened 2020-01-21
Add generation for Python PEP 561 type defs

EvanBoyle:
See this PR for pulumi/pulumi https://github.com/pulumi/pulumi/pull/3704/files

We did some work to clean up and enable mypy in pulumi/pulumi in https://github.com/pulumi/pulumi/pull/#3758, and then merged the PR to publish the type spec files. 

This PR generates the same `py.typed` files and required config in `setup.py`. I tested this on the `pulumi-aws` package locally. Mypy runs clean on the examples repo, and I was able to solicit some mypy errors:

```py
from pulumi_aws import s33
```
```
Evans-MBP:aws-py-s3-folder evanboyle$ mypy __main__.py 
__main__.py:6: error: Module 'pulumi_aws' has no attribute 's33'
__main__.py:8: error: Name 's3' is not defined
__main__.py:16: error: Name 's3' is not defined
__main__.py:37: error: Name 's3' is not defined
Found 4 errors in 1 file (checked 1 source file)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v1
         with:
-          python-version: 3.6.9
+          python-version: 3.6.10
       - name: Install Pulumi CLI
         uses: pulumi/action-install-pulumi-cli@releases/v1
       - name: Install tf2pulumi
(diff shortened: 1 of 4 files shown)


EvanBoyle:
Thanks @stack72. What's the process for releasing these changes and regenerating the downstream libraries?

PR #106, opened 2020-02-04
Fix the initializer of C# config value types

mikhailshilkov:
Fixes the issue of `pulumi-azure` generation for the latest SDK


stack72:
:shipit: and tag it please :)

PR #108, opened 2020-02-07
Use the schema-based code generator for Python.

pgavlin:
- Track whether or not specific object properties should be case-mapped
  in Python
- Adopt the new schema for default property values


lukehoban:
One question though - I see a new `Provider.py` file showing up in the diffs above - two things:
1. Why is that new (vs. modifying existing file)?
2. Is that actually meant to be `provider.py`?  (and maybe that's the answer to (1))?

pgavlin:
> 1. Why is that new (vs. modifying existing file)?

I believe this is a casing difference.

> Is that actually meant to be provider.py? (and maybe that's the answer to (1))?

This should probably be `provider.py`, yes. That will require a small fix in `pulumi/pulumi`.

PR #111, opened 2020-02-17
Ensure the link to the Terraform documentation is correct

stack72:
Fixes: https://github.com/pulumi/docs/issues/2020

By default, we always assumed the documentation ended in
`.html.markdown` we now check to ensure we have the correct link by checking
that the documentation exists first. The potential file names are:

* .html.markdown
* .html.md
* .markdown

We check for the file in the same way we do when we want to parse the doc contents

Sample diff

```
     ///
-    /// &gt; This content is derived from https://github.com/terraform-providers/terraform-provider-vault/blob/master/website/docs/r/approle_auth_backend_login.html.markdown.
+    /// &gt; This content is derived from https://github.com/terraform-providers/terraform-provider-vault/blob/master/website/docs/r/approle_auth_backend_login.html.md.
     /// </summary>

```

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -55,25 +55,32 @@ const (
 	DataSourceDocs DocKind = "d"
 )
 
-// getDocsForProvider extracts documentation details for the given package from
-// TF website documentation markdown content
-func getDocsForProvider(g *generator, org string, provider string, resourcePrefix string, kind DocKind,
-	rawname string, info tfbridge.ResourceOrDataSourceInfo) (parsedDoc, error) {
-
+func getRepoPath(org string, provider string) (string, error) {
 	gomod, err := LoadGoMod()
 	if err != nil {
-		return parsedDoc{}, err
+		return "", err
 	}
 
 	calculatedImportPath := fmt.Sprintf("github.com/%s/terraform-provider-%s", org, provider)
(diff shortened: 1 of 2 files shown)


pkg/tfgen/docs.go
```diff
@@ -219,7 +228,29 @@ func getDocsBaseURL(org, p string) string {
...
+	repoPath, err := getRepoPath(org, p)
+	if err != nil {
+		return ""
+	}
+
+	possibleMarkdownNames := []string{
+		name + ".html.markdown",
+		name + ".markdown",
+		name + ".html.md",
+	}
```
lukehoban: Can we somehow unify this with the very similar logic we have for `possibleMarkdownNames` in this file?  It feels like there should be some common logic for finding the right names that we use across both of there (and other places as well - like the suspicious looking code in `RenameDataSource`).

stack72: Sure! I will try and refactor this more - I just didn't want to mess with any of the existing logic :)

stack72:
ok @lukehoban, this code now uses the same code as before, but when we find the actual markdown document, we pass the name that it was found in along with the output

stack72:
FWIW, I will follow up with the changes to the `RenameResourceWithAlias` and `RenameDataSource` as I need to give some more thought to this - the code in these functions are for the old docs - maybe we don't even need these links anymore

PR #112, opened 2020-02-17
Warn on <elided> document comments

stack72:
Related: https://github.com/pulumi/home/issues/642

Output from AWS gives us this

```
warning: Resource aws_api_gateway_rest_api contains an <elided> doc reference that needs update
warning: Resource aws_batch_compute_environment contains an <elided> doc reference that needs update
warning: Resource aws_dx_hosted_transit_virtual_interface_accepter contains an <elided> doc reference that needs update
warning: Resource aws_ecr_lifecycle_policy contains an <elided> doc reference that needs update
warning: Resource aws_ecr_repository_policy contains an <elided> doc reference that needs update
warning: Resource aws_ecs_capacity_provider contains an <elided> doc reference that needs update
warning: Resource aws_eks_cluster contains an <elided> doc reference that needs update
warning: Resource aws_eks_node_group contains an <elided> doc reference that needs update
warning: Resource aws_elasticsearch_domain contains an <elided> doc reference that needs update
warning: Resource aws_iam_group_policy contains an <elided> doc reference that needs update
warning: Resource aws_iam_policy contains an <elided> doc reference that needs update
......
```

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -402,7 +402,13 @@ func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind 
 		}
 	}
 
-	return cleanupDoc(g, info, ret), nil
+	doc, elided := cleanupDoc(g, info, ret)
+	if elided {
+		cmdutil.Diag().Warningf(diag.Message("",
+			"Resource %v contains an <elided> doc reference that needs updated"), rawname)
+	}
+
+	return doc, nil
 }
 
 var (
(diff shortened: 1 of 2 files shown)


stack72:
@lukehoban this will allow us to keep track of the elided comments we need to fix - I have left this as a warn for now until we are at the point of needing to fail when everything has been updated

stack72:
The diffs here don't actually relate to what is in this PR - they relate to what is in master of the bridge

pkg/tfgen/docs.go
```diff
@@ -402,7 +402,12 @@ func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind
 		}
 	}
 
-	return cleanupDoc(g, info, ret), nil
+	doc, err := cleanupDoc(g, info, ret)
+	if err != nil {
```
lukehoban: If we’re going to do this unconditionally on an error, why not just diag warn the error itself?

pkg/tfgen/docs.go
```diff
@@ -594,21 +599,34 @@ func convertHCL(hcl string) (string, string, error) {
...
 }
 
-func cleanupDoc(g *generator, info tfbridge.ResourceOrDataSourceInfo, doc parsedDoc) parsedDoc {
+func cleanupDoc(g *generator, info tfbridge.ResourceOrDataSourceInfo, doc parsedDoc) (parsedDoc, error) {
 	newargs := make(map[string]string, len(doc.Arguments))
 	for k, v := range doc.Arguments {
-		newargs[k] = cleanupText(g, info, v)
+		cleanupText, err := cleanupText(g, info, v)
+		if err != nil {
+			return parsedDoc{}, err
```
lukehoban: Is it correct to bail early and return an empty parseddoc?  Don’t we want to still compute this as normal and just pass back a bool indicating a that it contained elided?

We could even likely skip most of what’s added here and just compare the result the result of cleanupText with elidedDocComment in the places we want to check if we should warn?

stack72:
@lukehoban I reworked this to pass around a boolean if there is an elided comment - it's much better than passing an error as technically it's not an error... yet :)

PR #115, opened 2020-02-20
Custom resource options for provider

mikhailshilkov:
`ResourceOptions` becomes an abstract class in https://github.com/pulumi/pulumi/issues/3857.
Pass `CustomResourceOptions` to providers too.


mikhailshilkov:
Downstream failures look unrelated
https://github.com/pulumi/pulumi-terraform-bridge/pull/115/checks?check_run_id=457383175#step:8:721
https://github.com/pulumi/pulumi-terraform-bridge/pull/115/checks?check_run_id=457383109#step:8:591

PR #117, opened 2020-02-25
Return errors instead of panics where upstream provider errors may occur

stack72:
Fixes: #114

This will prevent panics being returned to the user and will return
a more accurate description of what has happened

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -644,11 +644,15 @@ func (p *Provider) Create(ctx context.Context, req *pulumirpc.CreateRequest) (*p
 
 	newstate, err := p.tf.Apply(info, state, diff)
 	if newstate == nil {
-		contract.Assertf(err != nil, "Expected non-nil error with nil state during Create of %s", urn)
+		if err == nil {
+			return nil, fmt.Errorf("expected non-nil error with nil state during Create of %s", urn)
+		}
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/provider.go
```diff
@@ -644,11 +644,15 @@ func (p *Provider) Create(ctx context.Context, req *pulumirpc.CreateRequest) (*p
...
-		contract.Assertf(err != nil, "Expected non-nil error with nil state during Create of %s", urn)
+		if err == nil {
+			return nil, errors.New(fmt.Sprintf("unspecified error during Create of %s", urn))
+		}
 		return nil, err
 	}
 
-	contract.Assertf(newstate.ID != "", "Expected non-empty ID for new state during Create of %s", urn)
+	if newstate.ID == "" {
+		return nil, errors.New(fmt.Sprintf("no ID returned by provider during Create of %s", urn))
```
lukehoban: Why this pattern instead of `fmt.Errorf`?

stack72:
@lukehoban both updated :) Both great points!

stack72:
The errors here are because what is in master of the bridge needs to go into the providers as there are incompatibilities

pkg/tfbridge/provider.go
```diff
@@ -644,11 +644,15 @@ func (p *Provider) Create(ctx context.Context, req *pulumirpc.CreateRequest) (*p
 
 	newstate, err := p.tf.Apply(info, state, diff)
```
gmesch, line 645: (link)

PR #118, opened 2020-02-26
Test new action options for Vault provider downstream

jen20:
This updates to use version 2 of the downstream action tester for Vault.


jen20:
This has the intended effect of fixing Vault's build, but the other failures are unrelated and known.

PR #121, opened 2020-02-27
Update pulumi dependency for initial docs gen

justinvp:
This adds a new `docs` language to `tfgen` for generating resource model docs from a Pulumi schema.

Note: It's unclear to me if `tfgen` is the right "place" for this long term, but it at least unblocks generating docs for TF-based providers in the short term.

I'm guessing we will likely _not_ be checking-in the generated docs in the various TF-based provider repos like we do for other languages, but rather, a script/tool in the docs repo will call `tfgen` (or whatever final tool we come up with) specifying the output directory where to emit the generated docs within the structure of the docs repo. The docs repo script/tool will need to be able to handle calling the appropriate doc gen tool for all of our providers, whether TF-based or native like Kubernetes (once Kubernetes generates a schema that we can gen docs from). Ultimately, this may look more like adding support to `tfgen` for emitting the schema, and then doing something similar for Kubernetes, and then having a single docs gen tool that can use the schema as input to generate docs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/openzipkin/zipkin-go v0.1.6 // indirect
 	github.com/pkg/errors v0.8.1
 	github.com/prometheus/client_golang v0.9.3-0.20190127221311-3c4408c8b829 // indirect
-	github.com/pulumi/pulumi v1.10.2-0.20200211223422-2b59d1405d72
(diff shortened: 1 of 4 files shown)


praneetloke:
Closing this in favor of https://github.com/pulumi/docs/pull/2641.

PR #126, opened 2020-03-09
Change the way to define invokes in .NET

### LINKED ISSUE #125: Reconsider Invokes class in .NET
"Invokes" are functions that can be called to retrieve information from cloud providers. For example, we may retrieve the default VPC with

```cs
var vpc = await Pulumi.Aws.Ec2.Invokes.GetVpc(
    new Pulumi.Aws.Ec2.GetVpcArgs { Default = true });
```

Functions in C# have to leave inside a class (i.e., be "methods"). Because of this, we generate a static class per namespace called `Invokes`, as one can see in the example above.

This is awkward in several ways:

1. Imports of multiple namespaces clash on `Invokes` name, so `Invokes` almost always has to be fully-qualified or aliased like `using Ec2 = Pulumi.Aws.Ec2;` and `Ec2.Invokes.`.

2. Argument classes still live in the parent namespace, so they don't have `Invokes` in the path. They are visible in IntelliSense if the namespace is imported, while the method itself is not.

3. Invokes are harder to discover with IntelliSense if a user doesn't know about the static class.

4. Typing `Invokes` in an IDE doesn't help to automatically import the namespace -- IDE can't decide which one to import.

5. If one wants to import the `Invokes` class with usings, they can do `using static Pulumi.Aws.Ec2.Invokes;`. However, they would also need to import the namespace with `using Pulumi.Aws.Ec2` for argument classes.

One possible solution is to name the class after the invoke method and its method to a fixed `InvokeAsync`:

``` cs
public static class GetVpc
{
    public static Task<GetVpcResult> InvokeAsync(GetVpcArgs? args = null, InvokeOptions? options = null)
        => Pulumi.Deployment.Instance.InvokeAsync<GetVpcResult>("aws:ec2/getVpc:getVpc", args ?? InvokeArgs.Empty, options.WithVersion());
}
```

@pgavlin Do you remember why we did not do that in the first place?

mikhailshilkov:
Resolves #125 by introducing class-based invokes to the .NET SDK. Function-based invokes are marked as obsolete and will be removed in #131.

### DIFF
--- a/pkg/tfgen/generate_csharp.go
+++ b/pkg/tfgen/generate_csharp.go
@@ -736,6 +736,7 @@ func (rg *csharpResourceGenerator) emit() (string, error) {
 	} else {
 		contract.Assert(rg.fun != nil)
 		rg.generateDatasourceFunc()
+		rg.generateDatasourceClass()
 		rg.generateDatasourceArgs()
 		rg.generateDatasourceResult()
 	}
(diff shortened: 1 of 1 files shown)


mikhailshilkov:
> Can you port this to the C# generator in pulumi/pulumi as well?

Opened https://github.com/pulumi/pulumi/pull/4114

PR #128, opened 2020-03-10
Add support for GetSchema.

pgavlin:
And add support for generating a schema from tfgen. The generated schema
can then be used in the provider to provide the JSON-encoded schema that
`GetSchema` will return.


pgavlin:
Here's an example of the downstream: https://gist.github.com/pgavlin/7213df7363997e49aa99576594d8951a

PR #130, opened 2020-03-12
Wrap examples in shortcode

### LINKED ISSUE #122: Add place to hold examples in schema
In the “old way” of generating docs, the examples are included as a doc comments on the resource class. We should have a way to inject examples into the docs that are generated using the new schema.

tasiah:
Fixes #122. Note this probably needs to be adjusted slightly once we support multi-language examples since it currently only wraps a single example.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -503,6 +503,8 @@ func parseExamples(language language, lines []string) ([]string, bool, error) {
 	for i := 0; i < len(lines); i++ {
 		line := lines[i]
 		if strings.Index(line, "```") == 0 {
+			result = append(result, "{{% examples %}}")
+
 			// If we found a fenced block, parse out the code from it.
 			if language == nodeJS || language == pulumiSchema {
 				var hcl string
@@ -553,6 +555,8 @@ func parseExamples(language language, lines []string) ([]string, bool, error) {
 				}
 				skippableExamples = true
 			}
+
+			result = append(result, "{{% /examples %}}")
 		} else if strings.Index(line, "<div") == 0 && strings.Contains(line, "oics-button") {
 			// Strip "Open in Cloud Shell" buttons.
 			for i = i + 1; i < len(lines); i++ {
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -410,6 +410,8 @@ func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind 
 					cmdutil.Diag().Warningf(diag.Message("", "Expected an H1 in markdown for resource %v"), rawname)
 				}
 			default:
+				subsection = wrapShortcode(subsection)
+
 				// For all other sections, append them to the description section.
 				if !wroteHeader {
 					ret.Description += fmt.Sprintf("## %s\n", header)
@@ -430,6 +432,12 @@ func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind 
 	}
 
 	return doc, nil
+}
+
+func wrapShortcode(lines []string) []string {
+	lines = append([]string{"{{% examples %}}"}, lines...)
+	lines = append(lines, "{{% /examples %}}")
+	return lines
 }
 
 var (
@@ -503,8 +511,6 @@ func parseExamples(language language, lines []string) ([]string, bool, error) {
 	for i := 0; i < len(lines); i++ {
 		line := lines[i]
 		if strings.Index(line, "```") == 0 {
-			result = append(result, "{{% examples %}}")
-
 			// If we found a fenced block, parse out the code from it.
 			if language == nodeJS || language == pulumiSchema {
 				var hcl string
@@ -555,8 +561,6 @@ func parseExamples(language language, lines []string) ([]string, bool, error) {
 				}
 				skippableExamples = true
 			}
-
-			result = append(result, "{{% /examples %}}")
 		} else if strings.Index(line, "<div") == 0 && strings.Contains(line, "oics-button") {
 			// Strip "Open in Cloud Shell" buttons.
 			for i = i + 1; i < len(lines); i++ {
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -283,6 +283,13 @@ func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind 
 			continue
 		}
 
+		// Add shortcode around each examples block.
+		var headerIsExampleUsage bool
+		if header == "Example Usage" {
+			headerIsExampleUsage = true
+			ret.Description += "{{% examples %}}\n"
+		}
+
 		// Now split the sections by H3 topics. This is done because we'll ignore sub-sections with code
 		// snippets that are unparseable (we don't want to ignore entire H2 sections).
 		var wroteHeader bool
@@ -410,8 +417,6 @@ func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind 
 					cmdutil.Diag().Warningf(diag.Message("", "Expected an H1 in markdown for resource %v"), rawname)
 				}
 			default:
-				subsection = wrapShortcode(subsection)
-
 				// For all other sections, append them to the description section.
 				if !wroteHeader {
 					ret.Description += fmt.Sprintf("## %s\n", header)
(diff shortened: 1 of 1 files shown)


pkg/tfgen/docs.go
```diff
@@ -503,6 +503,8 @@ func parseExamples(language language, lines []string) ([]string, bool, error) {
 	for i := 0; i < len(lines); i++ {
 		line := lines[i]
 		if strings.Index(line, "```") == 0 {
+			result = append(result, "{{% examples %}}")
```
praneetloke: @cnunciato just as an FYI (and in case you have any comments), this is the short-code that the examples in the resource docs would be wrapped-in.

cnunciato: Excellent, thanks for the mention. Having a look now.

cnunciato: Ok yeah, I think this'll work fine. Having the `{{% examples %}}` shortcode will allow us to render the internal content as markdown, and then from there, we can add interactivity with components. https://github.com/pulumi/docs/pull/2665 shows an example.

cnunciato: One nice-to-have would be if you could pass the language to the shortcode as well (so like `{{% examples typescript %}}`), if that made sense. That'd make it a bit easier to wrap them all in a language chooser without having to dig into the rendered markup to figure that out.

cnunciato: We'll also want to capture the example titles, when we have them -- probably just as markdown just above the backticked code samples, e.g.:

```
    {{% examples %}}

    ### Some title

    ```typescript
    import foo from "bar";
    ```

    ### Some other title

    ```typescript
    import bar from "baz";
    ```

    ### Some third title

    ```typescript
    import qux from "quux";
    ```

    {{% /examples %}}
```

praneetloke: For my understanding, what is the reason for passing a language to the short-code?

Also, the short-code will actually wrap _every_ single example, where each example will contain snippets in the supported languages. So your example above is actually like this:
```
    {{% examples %}}
    ### Some title for the example

    ```typescript
    import foo from "bar";
    ...
    ```

    ```python
    from "bar" import foo
    ...
    ```

    ```go
    import github.com/....
    ...
    ```

    ```csharp
    using Foo.Bar;
    ...
    ```

    {{% /examples %}}
```

cnunciato: Ah, okay -- I misread the code, then. It looked as though all of the code snippets within an `{{% examples %}}` block would be of the same language, but if that's not the case, passing the language at the top level wouldn't make sense, yeah. 

As I mentioned, the only reason would be to avoid having to discern the language of the whole lot by examining the HTML attributes of every snippet, which seemed unnecessary if they'd all be the same. But since they won't, just `{{% examples %}}` would be fine.

cnunciato: Although one question @praneetloke, based on your snippet --

Will there be more than one heading within an `{{% examples%}}` shortcode? Or just one, with multiple `{{% examples %}}` shortcodes within a single document? In other words, would there be this:

```
{{% examples %}}

### Heading 

[some typescript]

[some go]

### Another heading

[some typescript]

[some go]

{{% /examples %}}
```

or this?

```
{{% examples %}}
### Heading 

[some typescript]

[some go]

{{% /examples %}}

{{% examples %}}
### Another heading 

[some typescript]

[some go]

{{% /examples %}}
```

cnunciato: The reason I ask is, I think it's _most_ important to know where the _set_ of examples starts and ends, more than it is to have each individual example bookended, because I want to be able to wrap them all into a single component, so that each one can be expanded or collapsed individually. So wrapping them all in a single shortcode would be preferable, if that were possible.

praneetloke: @cnunciato it would be "Snippet 2" from this [comment](https://github.com/pulumi/pulumi-terraform-bridge/pull/130#discussion_r393287427). I actually would also like confirmation from @tasiah as well on this.

tasiah: I think the goal is Snippet 2, like Praneet said. I'll have to go back and make sure that that's actually true though 😅 .

cnunciato: Ah, then that's not ideal. How feasible is option 1?

tasiah: Shouldn't be too bad (🤞). I'll move the shortcodes around then unless @praneetloke or @pgavlin have any objections.

praneetloke: We had an offline conversation about this. We _could_ do option 1, but we would also want to see if Pat has any reservations for that per his original suggestion to do it the way this PR has it currently (i.e. option 2). We will sync-up with Pat and update this PR accordingly.

cnunciato: Something like this could also work (technically, at least):

```
{{% examples %}}

{{% example %}}

### Some example

[some typescript]

[some go]

{{% /example %}}

{{% example %}}

### Some OTHER example

[some typescript]

[some go]

{{% /example %}}

{{% /examples %}}
```

... but things do get a bit dicey with Hugo rendering in this case, as the tolerance for spacing between these nested shortcodes can be brittle and depending on positioning, produce inconsistent results. So from my perspective, it'd be ideal just to wrap all of the examples in one shortcode (as in my Snippet 1), and let that single shortcode and the client-side code figure out how to render them.

praneetloke: > So from my perspective, it'd be ideal just to wrap all of the examples in one shortcode (as in my Snippet 1), and let that single shortcode and the client-side code figure out how to render them.

Ok we'll confirm with Pat and do this.

tasiah:
Sample from the schema generated from `pulumi-tfgen-aws` for `aws:appautoscaling/target:Target`:

----

Provides an Application AutoScaling ScalableTarget resource. To manage policies which get attached to the target, see the [`aws.appautoscaling.Policy` resource](https://www.terraform.io/docs/providers/aws/r/appautoscaling_policy.html).

{{% examples %}}
## Example Usage

{{% example %}}
### DynamoDB Table Autoscaling

```typescript

const dynamodbTableReadTarget = new aws.appautoscaling.Target("dynamodb_table_read_target", {
    maxCapacity: 100,
    minCapacity: 5,
    resourceId: pulumi.interpolate`table/${aws_dynamodb_table_example.name}`,
    roleArn: aws_iam_role_DynamoDBAutoscaleRole.arn,
    scalableDimension: "dynamodb:table:ReadCapacityUnits",
    serviceNamespace: "dynamodb",
});
```

{{% /example %}}
{{% example %}}
### DynamoDB Index Autoscaling

```typescript

const dynamodbIndexReadTarget = new aws.appautoscaling.Target("dynamodb_index_read_target", {
    maxCapacity: 100,
    minCapacity: 5,
    resourceId: pulumi.interpolate`table/${aws_dynamodb_table_example.name}/index/${var_index_name}`,
    roleArn: aws_iam_role_DynamoDBAutoscaleRole.arn,
    scalableDimension: "dynamodb:index:ReadCapacityUnits",
    serviceNamespace: "dynamodb",
});
```

{{% /example %}}
{{% example %}}
### ECS Service Autoscaling

```typescript

const ecsTarget = new aws.appautoscaling.Target("ecs_target", {
    maxCapacity: 4,
    minCapacity: 1,
    resourceId: pulumi.interpolate`service/${aws_ecs_cluster_example.name}/${aws_ecs_service_example.name}`,
    roleArn: var_ecs_iam_role,
    scalableDimension: "ecs:service:DesiredCount",
    serviceNamespace: "ecs",
});
```

{{% /example %}}
{{% example %}}
### Aurora Read Replica Autoscaling

```typescript

const replicas = new aws.appautoscaling.Target("replicas", {
    maxCapacity: 15,
    minCapacity: 1,
    resourceId: pulumi.interpolate`cluster:${aws_rds_cluster_example.id}`,
    scalableDimension: "rds:cluster:ReadReplicaCount",
    serviceNamespace: "rds",
});
```

{{% /example %}}
{{% /examples %}}

> This content is derived from https://github.com/terraform-providers/terraform-provider-aws/blob/master/website/docs/r/appautoscaling_target.html.markdown.

praneetloke:
@tasiah that looks good to me! @pgavlin and @cnunciato ?

tasiah:
Do I need to fix the failing check? I don't think 

```
go mod edit -replace=github.com/pulumi/pulumi-terraform-bridge=../pulumi-terraform-bridge
go: cannot find main module, but found .git/config in /home/runner/work/pulumi-terraform-bridge/pulumi-vault
	to create a module there, run:
	go mod init
```
is related to my changes.

pgavlin:
No, you're good to merge. 👍

PR #134, opened 2020-03-22
update to use new modules structure from pulumi/pulumi

EvanBoyle:
Required to update the providers to the new modules structure, and move our examples and templates over to modules.


EvanBoyle:
Note the downstream providers will fail with ambiguous imports as they need to be updated to import from `pulumi/pulumi/go/sdk/common` for the packages that were moved.

EvanBoyle:
I've updated the AWS provider locally and everything is working.

PR #135, opened 2020-03-23
Ignore GCP User Project Overrides in the docs

stack72:


stack72:
vault is a known failure - I just merged a PR that will fix that in the future

pkg/tfgen/docs.go
```diff
@@ -277,7 +277,8 @@ func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind
 		if strings.Index(header, "## ") == 0 {
 			header = header[3:]
 		}
-		if header == "Import" || header == "Imports" || header == "Timeout" || header == "Timeouts" {
+		if header == "Import" || header == "Imports" || header == "Timeout" ||
+			header == "Timeouts" || header == "User Project Overrides" || header == "User Project Override" {
```
leezen: For `User Project Overrides` wouldn't it be preferable to actually point to our own documentation on the attribute? Otherwise, there's no way for the user to know which resources do/don't support this without referring to the upstream documentation. This seems different from Imports/Timeout where they're relatively obvious. If it's not possible, I think it's fine to skip for now.

PR #136, opened 2020-03-23
Revert "Wrap examples in shortcode"

stack72:
Reverts pulumi/pulumi-terraform-bridge#130

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -283,13 +283,6 @@ func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind 
 			continue
 		}
 
-		// Add shortcode around each examples block.
-		var headerIsExampleUsage bool
-		if header == "Example Usage" {
-			headerIsExampleUsage = true
-			ret.Description += "{{% examples %}}\n"
-		}
-
 		// Now split the sections by H3 topics. This is done because we'll ignore sub-sections with code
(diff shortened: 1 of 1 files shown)


pkg/tfgen/docs.go
```diff
@@ -425,20 +418,9 @@ func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind
...
 					}
 				}
-
-				description := strings.Join(subsection, "\n") + "\n"
-				if headerIsExampleUsage {
-					// Wrap each example in shortcode.
-					description = "{{% example %}}\n" + description + "{{% /example %}}\n"
-				}
-				ret.Description += description
+				ret.Description += strings.Join(subsection, "\n") + "\n"
```
praneetloke: Is this change intentional?

stack72: @praneetloke i just hit revert on the original PR - this wasnt manually reverted

stack72: Looks like it was done in the original PR https://github.com/pulumi/pulumi-terraform-bridge/pull/130/files#diff-9cad96fc5e12fbb60a4c1204a57f5035L421

stack72: this was just a GH based revert - not manual

praneetloke: Ah yes I see (it was moved a line below where it originally was.) Thanks for confirming, Paul!

PR #138, opened 2020-03-23
Improve documentation for nested types

### LINKED ISSUE #15: Generation of document and schema has problems if duplicate field names are found
In the Terraform AWS route53_record resource, we have the following in the schema:

```
"type": {
				Type:     schema.TypeString,
				Required: true,
				ValidateFunc: validation.StringInSlice([]string{
					route53.RRTypeSoa,
					route53.RRTypeA,
					route53.RRTypeTxt,
					route53.RRTypeNs,
					route53.RRTypeCname,
					route53.RRTypeMx,
					route53.RRTypeNaptr,
					route53.RRTypePtr,
					route53.RRTypeSrv,
					route53.RRTypeSpf,
					route53.RRTypeAaaa,
					route53.RRTypeCaa,
				}, false),
			},
```

We also have another use of the parameter `type` in the same schema:

```
"failover_routing_policy": {
				Type:     schema.TypeList,
				Optional: true,
				ConflictsWith: []string{
					"geolocation_routing_policy",
					"latency_routing_policy",
					"weighted_routing_policy",
					"multivalue_answer_routing_policy",
				},
				Elem: &schema.Resource{
						"type": {
							Type:     schema.TypeString,
							Required: true,
							ValidateFunc: func(v interface{}, k string) (ws []string, es []error) {
								value := v.(string)
								if value != "PRIMARY" && value != "SECONDARY" {
									es = append(es, fmt.Errorf("Failover policy type must be PRIMARY or SECONDARY"))
								}
								return
							},
						},
					},
				},
			},
```

When we are generating the documentation and schema for this, Pulumi has the following:

```
export interface RecordState {
    /**
     * An alias block. Conflicts with `ttl` & `records`.
     * Alias record documented below.
     */
    readonly aliases?: pulumi.Input<pulumi.Input<{ evaluateTargetHealth: pulumi.Input<boolean>, name: pulumi.Input<string>, zoneId: pulumi.Input<string> }>[]>;
    /**
     * Allow creation of this record in Terraform to overwrite an existing record, if any. This does not affect the ability to update the record in Terraform and does not prevent other resources within Terraform or manual Route 53 changes outside Terraform from overwriting this record. `false` by default. This configuration is not recommended for most environments.
     */
    readonly allowOverwrite?: pulumi.Input<boolean>;
    /**
     * A block indicating the routing behavior when associated health check fails. Conflicts with any other routing policy. Documented below.
     */
    readonly failoverRoutingPolicies?: pulumi.Input<pulumi.Input<{ type: pulumi.Input<string> }>[]>;
    /**
     * [FQDN](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) built using the zone domain and `name`.
     */
    readonly fqdn?: pulumi.Input<string>;
    /**
     * A block indicating a routing policy based on the geolocation of the requestor. Conflicts with any other routing policy. Documented below.
     */
    readonly geolocationRoutingPolicies?: pulumi.Input<pulumi.Input<{ continent?: pulumi.Input<string>, country?: pulumi.Input<string>, subdivision?: pulumi.Input<string> }>[]>;
    /**
     * The health check the record should be associated with.
     */
    readonly healthCheckId?: pulumi.Input<string>;
    /**
     * A block indicating a routing policy based on the latency between the requestor and an AWS region. Conflicts with any other routing policy. Documented below.
     */
    readonly latencyRoutingPolicies?: pulumi.Input<pulumi.Input<{ region: pulumi.Input<string> }>[]>;
    /**
     * Set to `true` to indicate a multivalue answer routing policy. Conflicts with any other routing policy.
     */
    readonly multivalueAnswerRoutingPolicy?: pulumi.Input<boolean>;
    /**
     * DNS domain name for a CloudFront distribution, S3 bucket, ELB, or another resource record set in this hosted zone.
     */
    readonly name?: pulumi.Input<string>;
    /**
     * A string list of records. To specify a single record value longer than 255 characters such as a TXT record for DKIM, add `\"\"` inside the Terraform configuration string (e.g. `"first255characters\"\"morecharacters"`).
     */
    readonly records?: pulumi.Input<pulumi.Input<string>[]>;
    /**
     * Unique identifier to differentiate records with routing policies from one another. Required if using `failover`, `geolocation`, `latency`, or `weighted` routing policies documented below.
     */
    readonly setIdentifier?: pulumi.Input<string>;
    /**
     * The TTL of the record.
     */
    readonly ttl?: pulumi.Input<number>;
    /**
     * `PRIMARY` or `SECONDARY`. A `PRIMARY` record will be served if its healthcheck is passing, otherwise the `SECONDARY` will be served. See http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/dns-failover-configuring-options.html#dns-failover-failover-rrsets
     */
    readonly type?: pulumi.Input<string>;
    /**
     * A block indicating a weighted routing policy. Conflicts with any other routing policy. Documented below.
     */
    readonly weightedRoutingPolicies?: pulumi.Input<pulumi.Input<{ weight: pulumi.Input<number> }>[]>;
    /**
     * Hosted zone ID for a CloudFront distribution, S3 bucket, ELB, or Route 53 hosted zone. See [`resource_elb.zone_id`](https://www.terraform.io/docs/providers/aws/r/elb.html#zone_id) for example.
     */
    readonly zoneId?: pulumi.Input<string>;
}
``` 

Notice there is only 1 `type` parameter found. It has chosen the documentation from the nested type in the Terraform schema NOT the parameter that it relates to in the actual schema

This has led to https://github.com/pulumi/pulumi-aws/issues/432

tasiah:
Fixes #15. 

I'll probably spend the next day cleaning it up a bit more and allowing for more ways to detect nested blocks. Would love suggestions on how to make this more readable.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -221,13 +221,15 @@ var (
 	argumentBulletRegexp = regexp.MustCompile(
 		"\\*\\s+`([a-zA-z0-9_]*)`\\s+(\\([a-zA-Z]*\\)\\s*)?[–-]?\\s+(\\([^\\)]*\\)\\s*)?(.*)")
 
-	argumentBlockRegexp = regexp.MustCompile("`([a-z_]+)`\\s+block[\\s\\w]*:")
-
 	nestedObjectRegexps = []*regexp.Regexp{
 		// For example:
 		// s3_bucket.html.markdown: "The `website` object supports the following:"
 		// ami.html.markdown: "When `virtualization_type` is "hvm" the following additional arguments apply:"
 		regexp.MustCompile("`([a-z_]+)`.*following"),
+
+		// For example:
+		// athena_workgroup.html.markdown: "#### result_configuration Argument Reference"
+		regexp.MustCompile("(?i)## ([a-z_]+).* argument reference"),
 	}
 
 	attributeBulletRegexp = regexp.MustCompile("\\*\\s+`([a-zA-z0-9_]*)`\\s+[–-]?\\s+(.*)")
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -180,7 +180,7 @@ func readMarkdown(repo string, kind DocKind, possibleLocations []string) ([]byte
 
 // mergeDocs adds the docs specified by extractDoc from sourceFrom into the targetDocs
 func mergeDocs(g *generator, info tfbridge.ResourceOrDataSourceInfo, org string, provider string,
-	resourcePrefix string, kind DocKind, docs parsedDoc, sourceFrom string, isTargetAttributes bool, isSourceAttributes bool) error {
+	resourcePrefix string, kind DocKind, docs parsedDoc, sourceFrom string, useTargetAttributes bool, useSourceAttributes bool) error {
 
 	if sourceFrom != "" {
 		sourceDocs, err := getDocsForProvider(g, org, provider, resourcePrefix, kind, sourceFrom, nil)
@@ -188,18 +188,18 @@ func mergeDocs(g *generator, info tfbridge.ResourceOrDataSourceInfo, org string,
 			return err
 		}
 
-		if isTargetAttributes && isSourceAttributes {
+		if useTargetAttributes && useSourceAttributes {
 			for k, v := range sourceDocs.Attributes {
 				docs.Attributes[k] = v
 			}
-		} else if isTargetAttributes && !isSourceAttributes {
+		} else if useTargetAttributes && !useSourceAttributes {
 			for k, v := range sourceDocs.Arguments {
 				docs.Attributes[k] = v.description
 				for kk, vv := range v.arguments {
 					docs.Attributes[kk] = vv
 				}
 			}
-		} else if !isTargetAttributes && !isSourceAttributes {
+		} else if !useTargetAttributes && !useSourceAttributes {
 			for k, v := range sourceDocs.Arguments { // string -> argument
 				arguments := sourceDocs.Arguments[k].arguments
 				docArguments := make(map[string]string)
@@ -385,8 +385,9 @@ func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind 
 							}
 							ret.Arguments[nested].arguments[matches[1]] = matches[4]
 
-							// Also record this as a top-level argument, since sometimes the recorded nested
+							// Also record this as a top-level argument just in case, since sometimes the recorded nested
 							// argument doesn't match the resource's argument.
+							// For example, see `cors_rule` in s3_bucket.html.markdown.
 							if ret.Arguments[matches[1]] == nil {
 								ret.Arguments[matches[1]] = &argument{
 									description: matches[4],
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -33,12 +33,19 @@ import (
 	"github.com/pulumi/pulumi-terraform-bridge/pkg/tfbridge"
 )
 
+// Either doc or arguments
+type argument struct {
+	doc       string
+	arguments map[string]string
+	isNested  bool
+}
+
 // parsedDoc represents the data parsed from TF markdown documentation
 type parsedDoc struct {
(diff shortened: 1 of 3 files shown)


tasiah:
Seems like the AWS, Azure, and GCP tests are failing because of the same thing:
```
build github.com/pulumi/pulumi-azure/cmd/pulumi-resource-azure: cannot load github.com/pulumi/pulumi/pkg/resource/provider: ambiguous import: found github.com/pulumi/pulumi/pkg/resource/provider in multiple modules:
	github.com/pulumi/pulumi v1.12.2-0.20200313044354-8111d33438b9 (/home/runner/go/pkg/mod/github.com/pulumi/pulumi@v1.12.2-0.20200313044354-8111d33438b9/pkg/resource/provider)
	github.com/pulumi/pulumi/pkg v0.0.0-20200322194843-61928f04e052 (/home/runner/go/pkg/mod/github.com/pulumi/pulumi/pkg@v0.0.0-20200322194843-61928f04e052/resource/provider)
```
Not quite sure what the best way to fix this is.

pkg/tfgen/docs.go
```diff
@@ -33,12 +33,23 @@ import (
...
+	isNested bool
+}
+
 // parsedDoc represents the data parsed from TF markdown documentation
 type parsedDoc struct {
 	// Description is the description of the resource
 	Description string
 	// Arguments includes the names and descriptions for each argument of the resource
-	Arguments map[string]string
+	Arguments map[string]*argument
```
justinvp: Just for my own understanding: This is now a map of "object name" to arguments associated with that object? And an empty string for the object name represents the top-level properties?

Maybe add some of these details to the comments?

praneetloke:
> Seems like the AWS, Azure, and GCP tests are failing because of the same thing:

Yeah I think this is related to moving `pulumi/pulumi` to the new Go modules structure. Those provider repos are yet to be updated. https://github.com/pulumi/pulumi-aws/pull/905, https://github.com/pulumi/pulumi-azure/pull/496 and https://github.com/pulumi/pulumi-gcp/pull/305 are the relevant PRs. So I don't think the failure you are seeing is because of your changes.

pkg/tfgen/docs.go
tasiah: Using some of the arguments from [`s3_bucket.html.markdown`](https://raw.githubusercontent.com/terraform-providers/terraform-provider-aws/master/website/docs/r/s3_bucket.html.markdown) as an example, I want to generate a map like this:

- bucket
    - description: "(Optional, Forces new resource) The name of the bucket. If omitted, Terraform will assign a random, unique name."
    - arguments: nil
- website
    - description: "(Optional) A website object (documented below)."
    - arguments:
        - index_document: "(Required, unless using `redirect_all_requests_to`) Amazon S3 returns this index document when requests are made to the root domain or any of the subfolders."
- index_document
    - description: "(Required, unless using `redirect_all_requests_to`) Amazon S3 returns this index document when requests are made to the root domain or any of the subfolders."
    - arguments: nil
   - isNested: true

where both `bucket` and `website` are top-level properties, but `website` is also an object with its own arguments. I'm *guessing* there's not many instances of repeated fields with the same parent name [1], so I only go a single nest deep.

I'll add a comment to (hopefully) clear this up a bit more.


[1] For example, suppose `website` has a `website` field that also has a `website` field; then `map["website"].arguments["website"]` could actually refer to both our top-level `website`'s `website` argument or that `website`'s `website` argument.

tasiah:
> So I don't think the failure you are seeing is because of your changes.

Does this mean I can/should merge even though those tests are failing?

justinvp:
> Does this mean I can/should merge even though those tests are failing?

@stack72, what would you recommend?

praneetloke:
> Does this mean I can/should merge even though those tests are failing?

Yes, you could. Although paging @EvanBoyle or @stack72  for confirmation. If the plan is to have those repos updated soon (as there are pending PRs in each of those already to address this), then I recommend waiting, but if not, merging this PR as-is should be fine.

stack72:
@tasiah if you think this is in a situation where it can be released then it’s fine to merge :) the diffs look fine so the broken build is unrelated

PR #140, opened 2020-03-24
Search for go.mod in the provider folder as well

praneetloke:
I think this is needed for the new module structure in the provider repos to work.

### DIFF
--- a/pkg/tfgen/gomod.go
+++ b/pkg/tfgen/gomod.go
@@ -35,7 +35,7 @@ func LoadGoMod() (*modfile.File, error) {
 	if moduleRoot == "" {
 		// Some provider repos have a "provider" module, rather than a
 		// module at the root of the repo.
-		moduleRoot = findModuleRoot(exePath + "provider")
+		moduleRoot = findModuleRoot(exePath + "/provider")
 		if moduleRoot == "" {
 			return nil, errors.New("cannot find module root")
 		}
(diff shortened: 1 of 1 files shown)


pkg/tfgen/gomod.go
```diff
@@ -33,7 +33,12 @@ func LoadGoMod() (*modfile.File, error) {
 
 	moduleRoot := findModuleRoot(exePath)
 	if moduleRoot == "" {
-		return nil, errors.New("cannot find module root")
+		// Some provider repos have a "provider" module, rather than a
+		// module at the root of the repo.
+		moduleRoot = findModuleRoot(exePath + "provider")
```
praneetloke:
```suggestion
		moduleRoot = findModuleRoot(exePath + "/provider")
```

pkg/tfgen/gomod.go
```diff
@@ -33,7 +33,12 @@ func LoadGoMod() (*modfile.File, error) {
 
 	moduleRoot := findModuleRoot(exePath)
 	if moduleRoot == "" {
-		return nil, errors.New("cannot find module root")
+		// Some provider repos have a "provider" module, rather than a
+		// module at the root of the repo.
+		moduleRoot = findModuleRoot(exePath + "/provider")
```
lukehoban: Should probably use filepath.Join.

praneetloke:
@lukehoban unless I am missing something, I think this is safe to land regardless, as this PR only falls-back to searching in the `/provider` repo, if a `go.mod` couldn't be found at the root, which is the old/current way.

EvanBoyle:
@lukehoban @praneetloke ideally these changes would all be made atomically, but that isn't really possible. Every provider needs to be updated to consume the new modules structure from pulum/pulumi, and the providers need to be broken down into multi-module repos which inclue top level go.mod files for /provider and /sdk.

PR #141, opened 2020-03-25
Ensure SchemaInfo is checked for MaxItemsOne before naming params

stack72:
It didn't matter what we had in our SchemaInfo, the naming was
always coming from the TF Schema which wasn't quite correct. It should
take into account what we wanted to do about pluralizing the names
based on our schema overrides

Fixes: https://github.com/pulumi/pulumi-aws/issues/900

### DIFF
--- a/pkg/tfbridge/names.go
+++ b/pkg/tfbridge/names.go
@@ -26,7 +26,7 @@ import (
 
 // PulumiToTerraformName performs a standard transformation on the given name string, from Pulumi's PascalCasing or
 // camelCasing, to Terraform's underscore_casing.
-func PulumiToTerraformName(name string, tfs map[string]*schema.Schema) string {
+func PulumiToTerraformName(name string, tfs map[string]*schema.Schema, ps map[string]*SchemaInfo) string {
 	var result string
 	for i, c := range name {
 		if c >= 'A' && c <= 'Z' {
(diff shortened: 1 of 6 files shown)


pkg/tfbridge/names.go
```diff
@@ -49,19 +49,28 @@ func PulumiToTerraformName(name string, tfs map[string]*schema.Schema) string {
...
 			result = singularResult
 		}
+
+		if ps != nil {
+			p, ok := ps[singularResult]
+			if ok && p != nil && p.MaxItemsOne != nil && !*p.MaxItemsOne &&
+				(sch.Type == schema.TypeList || sch.Type == schema.TypeSet) {
+				result = singularResult
+			}
+		}
```
lukehoban: Just to make this logic a little clearer, and to avoid duplicating a bunch of this - perhaps compute “isMaxItemsOne” first using both schema and schemainfo, then have just one test for the condition that would choose the singular result?

pkg/tfbridge/names.go
```diff
@@ -49,19 +49,28 @@ func PulumiToTerraformName(name string, tfs map[string]*schema.Schema) string {
...
 // to Pulumi's PascalCasing (if upper is true) or camelCasing (if upper is false).
-func TerraformToPulumiName(name string, sch *schema.Schema, upper bool) string {
+func TerraformToPulumiName(name string, sch *schema.Schema, ps *SchemaInfo, upper bool) string {
 	var result string
 	var nextCap bool
 	var prev rune
 
 	// Pluralize names that will become array-shaped Pulumi values
-	if sch != nil && sch.MaxItems != 1 && (sch.Type == schema.TypeList || sch.Type == schema.TypeSet) {
+	if (ps == nil || ps != nil && ps.MaxItemsOne == nil) && sch != nil && sch.MaxItems != 1 &&
```
lukehoban: Is this right? It looks different than the test above?

Also might make sense to do the same thing and compute `isMaxItemsOne` separately here as well just for readability.

pkg/tfbridge/provider.go
```diff
@@ -379,7 +380,7 @@ func (p *Provider) Configure(ctx context.Context,
 	var missingKeys []*pulumirpc.ConfigureErrorMissingKeys_MissingKey
 	for key, meta := range p.config {
 		if meta.Required && !config.IsSet(key) {
-			name := TerraformToPulumiName(key, meta, false)
+			name := TerraformToPulumiName(key, meta, nil, false)
```
lukehoban: We do have schemainfo for config, right?  Is it not used here because we don’t (yet?) support maxitemsone annotations there?

pkg/tfbridge/provider.go
```diff
@@ -431,7 +432,7 @@ func (p *Provider) formatFailureReason(res Resource, reason string) string {
 	if len(parts) == 2 {
 		schema := res.TF.Schema[parts[1]]
 		if schema != nil {
-			name := TerraformToPulumiName(parts[1], schema, false)
+			name := TerraformToPulumiName(parts[1], schema, nil, false)
```
lukehoban: Do we not have a way to get the schemainfo assed in here?  This does look like it will risk mis-singularizing names in error messages.

pkg/tfbridge/names.go
stack72: done!

pkg/tfbridge/provider.go
stack72: In config annotations, we *only* support primitive types as of now

pkg/tfbridge/provider.go
stack72: fixed - you are correct here!

stack72:
@lukehoban this is now ready for a second round of reviews - it feels much cleaner now

PR #142, opened 2020-03-26
Remove vendoring

EvanBoyle:
Removes the need for vendoring, uses `go mod download` instead.


EvanBoyle:
I'm going to go ahead and merge this to unblock my progress on updating the providers:

https://github.com/pulumi/pulumi-aws/pull/905
https://github.com/pulumi/pulumi-azure/pull/496
https://github.com/pulumi/pulumi-gcp/pull/305

Happy to revert, or open a PR with updates if we have concerns here (but I expect this to be non-controversial).

PR #143, opened 2020-03-26
Re-introduce the examples short code as well as strip non-relevant examples in the lang generators

praneetloke:
Related to issue pulumi/pulumi#4158 and the PR https://github.com/pulumi/pulumi/pull/4184.

This PR restores the examples short code that was originally introduced with #122, but then later reverted with #136 because the code generators weren't equipped to handle the inclusion of the short codes, which is what this PR does. It uses a function from the `pulumi/pulumi/pkg/codegen` package to strip out the short code as well as any code snippets that are not relevant to a specific language.

**Note**: I overrode the `go.mod` file locally to get the new function I added to the `pulumi/pulumi` repo. So this PR will fail until I merge https://github.com/pulumi/pulumi/pull/4184.


pkg/tfgen/docs.go
```diff
@@ -162,7 +162,7 @@ func getDocsForProvider(g *generator, org string, provider string, resourcePrefi
 
 	doc, err := parseTFMarkdown(g, info, kind, string(markdownBytes), markdownFileName, resourcePrefix, rawname)
 	if err != nil {
-		return parsedDoc{}, nil
+		return parsedDoc{}, err
```
praneetloke: I don't know why we never returned the error from parsing the markdown here, but this caused me to lose several hours while trying to understand why schema generation wasn't working properly for me. Because this error wasn't being bubbled-up, I missed out the fact that example parsing was failing due to `tf2pulumi` not existing in my `GOPATH` and therefore entire `Description` was being ignored.

praneetloke:
I opened https://github.com/pulumi/pulumi-terraform-bridge/issues/144 for the GH Actions failures in the testing of downstream providers, but those are not related to my changes here. I will go ahead and merge this PR. I have run a test build for the `pulumi-aws` provider and I don't see any failures. The diff looks as expected as well, except for a new-line character when there are no examples for a language, which I think is fine.

PR #155, opened 2020-04-08
Schema: generate distinct input/output/state types

### LINKED ISSUE #152: TF schema generator ignores type difference inputs vs outputs
I may be confused about how things are intended to work, so I'm opening this issue to get a confirmation that it's an actual problem.

I'll illustrate with an example. The current Azure schema contains this snippet:

```
"azure:storage/account:Account": {
    "description": "Manages an Azure Storage Account.\n\n{{% examples %}}\n{{% /examples %}}\n",
    "properties": {
        ...
        "identity": {
            "$ref": "#/types/azure:storage/AccountIdentity:AccountIdentity",
            "description": "A `identity` block as defined below.\n"
        },
        ...
    "inputProperties": {
        ...
        "identity": {
            "$ref": "#/types/azure:storage/AccountIdentity:AccountIdentity",
            "description": "A `identity` block as defined below.\n"
        },
```

As I read it, it points to the same type both from an input property `identity` and from the corresponding output property `identity`.

The problem is that those types should be different. The current (non-schema-based) Azure code in C# has a separate args (inputs) type -- note a nullable `PrincipalId`:

```csharp
public sealed class AccountIdentityArgs : Pulumi.ResourceArgs
{
    [Input("principalId")]
    public Input<string>? PrincipalId { get; set; }
```

https://github.com/pulumi/pulumi-azure/blob/ffeb2d2020298598f12fd7705bde6cf7ccf3cd36/sdk/dotnet/Storage/Account.cs#L1005-L1011

and a separeate output type (not a non-nullable `PrincipalId`):

```csharp
[OutputType]
public sealed class AccountIdentity
{
    public readonly string PrincipalId;
```

https://github.com/pulumi/pulumi-azure/blob/ffeb2d2020298598f12fd7705bde6cf7ccf3cd36/sdk/dotnet/Storage/Account.cs#L1694-L1699

I suspect that the schema generator loses information about required properties, which leads to the diff in generated non-schema vs. schema-based Azure C# code:

```diff
 [Pulumi.OutputType]
    public sealed class AccountIdentity
    {
-       public readonly string PrincipalId;
-       public readonly string TenantId;
+       public readonly string? PrincipalId;
+       public readonly string? TenantId;
        public readonly string Type;
    }
```

https://github.com/mikhailshilkov/api-compare/pull/1/files#diff-2eb546cb2cda8f85dd3888495e6511b4L47886

@pgavlin Does this make sense?

pgavlin:
If collisions occur between input, output, or state types with different
shapes, generate distinct object types for each.

Fixes #152.

### DIFF
--- a/pkg/tfgen/generate_schema.go
+++ b/pkg/tfgen/generate_schema.go
@@ -69,6 +69,8 @@ func (g *schemaGenerator) typeName(r *resourceType) string {
 
 type schemaNestedType struct {
 	typ       *propertyType
+	declarer  declarer
+	isInput   bool
 	pyMapCase bool
 }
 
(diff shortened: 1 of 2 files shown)


pkg/tfgen/generate_schema.go
```diff
@@ -97,39 +99,68 @@ func gatherSchemaNestedTypesForMember(member moduleMember) map[string]*schemaNes
...
+		// - input + state
+		// - output + state
+		contract.Assert(existing.isInput || isState)
+
+		if isState {
+			typeName = typeName + "State"
+			_, collision := nt.nameToType[typeName]
+			contract.Assert(!collision)
+		} else {
+			existing.typ.name = typeName + "Args"
```
pgavlin, line 144: Note that unless we decide on a different suffix, these changes will break Go codegen. The Go code generator uses `Args` as the suffix for its generated inputty types, and is guaranteed to cause collisions in these cases.

pgavlin: We can either pick a different suffix (e.g. `Inputs`, `Arg`, etc.) or fix the Go code generator.

mikhailshilkov: This change also breaks the C# code: the type I get is now called `Inputs.AccountIdentityArgsArgs` (and `Outputs.AccountIdentity` which is the same). Do we need to change the C# codegen now?

pkg/tfgen/generate_schema.go
```diff
@@ -97,39 +99,68 @@ func gatherSchemaNestedTypesForMember(member moduleMember) map[string]*schemaNes
...
-		baseName = typ.nestedType.Name().String()
+		typeName = typ.nestedType.Name().String()
 	}
 
-	typeName := baseName + strings.Title(nameSuffix)
-	typ.name = typeName
+	if existing, ok := nt.nameToType[typeName]; ok {
+		contract.Assert(existing.declarer == declarer)
+
+		if existing.typ.equals(typ) && pyMapCase == existing.pyMapCase {
```
mikhailshilkov, line 129: Can it happen that two types were equal in versions of a provider <= N but then a new version brings a property which makes them different. Will that cause the type name change and an unexpected breaking change?

mikhailshilkov: I guess I'm asking why don't we want to append those name suffixes all the time unconditionally?

pgavlin:
I did not realize this PR was still open when I re-used this branch name. Fixing now...

iwahbe:
Looks like this is out of date.

PR #156, opened 2020-04-10
dont get subattributes

ekrengel:
Fixes https://github.com/pulumi/docs/issues/2787

This fixes some formatting. I am still trying to figure out how to actually get this information from these sub- bullets. But for now this removes the mangeled sub-bullets.

Before:
image

After:
image

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -260,7 +260,8 @@ var (
 		regexp.MustCompile("(?i)## ([a-z_]+).* argument reference"),
 	}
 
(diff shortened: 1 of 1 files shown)


praneetloke:
> This fixes some formatting. I am still trying to figure out how to actually get this information from these sub- bullets. But for now this removes the mangeled sub-bullets.

I am wondering if we should hold off until the sub-bullets formatting is fixed as well rather than merge this without those. It feels like we will be losing some information that is currently there, even though the formatting isn't great. The reason I am a little hesitant is that the doc comments in each of the language SDKs will also lose this information.

ekrengel:
Okay will close for now

PR #157, opened 2020-04-13
allow no space after attribute name

ekrengel:
Problem here was the markdown does not have a space after the attribute name (and before the -) so it was not matching the regex.

image

Fixes https://github.com/pulumi/docs/issues/2919


justinvp:
I was looking at the AWS diff, and while there are definitely improvements, there also appears to be some regressions.

Some examples that don't look right:

https://github.com/pulumi-bot/pulumi-aws/compare/e5be582e9..fabbe8572#diff-5cb8ef4d7bc60986d86e6e2cdef12ffb

```diff
         /// <summary>
-        /// The priority associated with the rule.
+        /// is optional (with a default value of `0`) but must be unique between multiple rules
         /// </summary>
         [Input("priority")]
         public Input<int>? Priority { get; set; }
```

Here's a snippet from the [TF docs](https://raw.githubusercontent.com/terraform-providers/terraform-provider-aws/master/website/docs/r/s3_bucket.html.markdown):

```
The `rules` object supports the following:

* `id` - (Optional) Unique identifier for the rule.
* `priority` - (Optional) The priority associated with the rule.
* `destination` - (Required) Specifies the destination for the rule (documented below).
* `source_selection_criteria` - (Optional) Specifies special object selection criteria (documented below).
* `prefix` - (Optional) Object keyname prefix identifying one or more objects to which the rule applies.
* `status` - (Required) The status of the rule. Either `Enabled` or `Disabled`. The rule is ignored if status is not Enabled.
* `filter` - (Optional) Filter that identifies subset of objects to which the replication rule applies (documented below).

~> **NOTE on `prefix` and `filter`:** Amazon S3's latest version of the replication configuration is V2, which includes the `filter` attribute for replication rules.
With the `filter` attribute, you can specify object filters based on the object key prefix, tags, or both to scope the objects that the rule applies to.
Replication configuration V1 supports filtering based on only the `prefix` attribute. For backwards compatibility, Amazon S3 continues to support the V1 configuration.

* For a specific rule, `prefix` conflicts with `filter`
* If any rule has `filter` specified then they all must
* `priority` is optional (with a default value of `0`) but must be unique between multiple rules
```

Looks like it's now picking up that extra bullet at the end, rather than the first bullet for priority that it used to pick up at the top.


Also:

https://github.com/pulumi-bot/pulumi-aws/compare/e5be582e9..fabbe8572#diff-371b697c3717928f1cce98174a0f4dbd

```diff
         /// <summary>
-        /// Provide this to allow manage the identity provider configuration for the fleet. Fields documented below.
+        /// is cannot removed without force recreating.
         /// </summary>
         [Output("identityProvider")]
         public Output<Outputs.FleetIdentityProvider?> IdentityProvider { get; private set; } = null!;
```

and

```diff
         /// <summary>
-        /// Provide this to allow manage the company network configuration for the fleet. Fields documented below.
+        /// is cannot removed without force recreating.
         /// </summary>
         [Output("network")]
         public Output<Outputs.FleetNetwork?> Network { get; private set; } = null!;
```

etc.

pkg/tfgen/docs.go
```diff
@@ -247,7 +247,7 @@ func mergeDocs(g *generator, info tfbridge.ResourceOrDataSourceInfo, org string,
 // nolint:lll
 var (
 	argumentBulletRegexp = regexp.MustCompile(
-		"\\*\\s+`([a-zA-z0-9_]*)`\\s+(\\([a-zA-Z]*\\)\\s*)?[–-]?\\s+(\\([^\\)]*\\)\\s*)?(.*)")
+		"\\*\\s+`([a-zA-z0-9_]*)`\\s*(\\([a-zA-Z]*\\)\\s*)?[–-]?\\s+(\\([^\\)]*\\)\\s*)?(.*)")
```
lukehoban, line 250: Possibly worth adding tests to cover the desired behaviour of this regexp?  Potentially including a bunch of examples from real docs to ensure they are handled correctly?  (I realize this is existing test debt, but might be a good opportunity to improve on that while we are tweaking this).

ekrengel:
I made some adjustments to make sure the regex wasnt capturing mid sentence attributes. I also made a change that means we always take the first attribute description. (e.g. name is listed multiple times because of sub properties) This made a lot of changes and I am happy to change it back if we feel this is the wrong call.

justinvp:
Hmm, still no diffs?

ekrengel:
> Hmm, still no diffs?

Yeah I dont think this is right. Maybe I can open a brand new PR to see if that fixes it.

PR #161, opened 2020-04-16
Fix argument formatting / descriptions

ekrengel:
Trying to open a new PR to see if this fixes the diffs.


ekrengel:
Still no diffs 🤔 I am pretty confused at this point why theres no changes but the resource docs generation is showing lots of diffs.

ekrengel:
@justinvp James helped me fix the action so now you can see changes. This PR does two things:

1. use first occurrence of an argument as the description -> this is what has resulted in a ton of changes
1. Allow no space between argument name and the dash to fix some formatting issues like this


image

Fixes https://github.com/pulumi/docs/issues/2919

praneetloke:
> use first occurrence of an argument as the description

While this has worked well in a lot of cases, it seems there are some unintended effects to this. For example, https://github.com/pulumi-bot/pulumi-aws/compare/647f6f27c..e43f17a4d#diff-4e39899891863dc5be4fcda690a4004f. We could accept that and work on that as a separate issue. Earlier, Tasia worked on an [issue](https://github.com/pulumi/pulumi-terraform-bridge/issues/15) to fix issues like this, though, it is possible that we are somehow not doing that for the processing of datasources?

ekrengel:
> some unintended effects to this

yup -- I can take this out and leave it as is, which means the last one wins. I personally thing the first one should win since the first one the page is likely the one to be seen... either way things are wrong. I agree there is a lot of improvements that can be made here -- still trying to wrap my head around how these nested attributes work.

ekrengel:
@praneetloke this is what the diff looks like if we take the last -- this is because its now picking up some more lines. Let me see if i can tweak the regex to ignore those lines.

praneetloke:
Looked at the latest diff for aws and look like we regressed in quite a few places. Here are a few examples, but if you scroll past these you will see a lot more.
https://github.com/pulumi-bot/pulumi-aws/compare/884741cc8..c079f2b4b#diff-d813d54f3113aa49b74dbc6b0d001694
https://github.com/pulumi-bot/pulumi-aws/compare/884741cc8..c079f2b4b#diff-349e7951404c97711798b1652ee1bcdb
https://github.com/pulumi-bot/pulumi-aws/compare/884741cc8..c079f2b4b#diff-b2c7bfbc9e89ccb01cdbec802e53db5b

ekrengel:
Yes -- I have been looking into it. Starting to think that we may just need to edit our fork instead of making these changes. I could perhaps leave the tests and refactoring.

praneetloke:
I actually think that this [diff](https://github.com/pulumi-bot/pulumi-aws/compare/26997d122..7642fa00b) for AWS looked better. I did spot one case of a false positive, i.e. `*` without a space:
https://github.com/pulumi-bot/pulumi-aws/compare/26997d122..7642fa00b#diff-7c7d3a2485ab637f0841d806ff09cbd0

What do you think about reverting the regex to this [commit](https://github.com/pulumi/pulumi-terraform-bridge/pull/161/commits/0e69b122a4927b4abdbb4a8dc816bbbcc5b34804) and then updating the fork manually for other cases?

Azure wasn't looking too bad either with that older [diff](https://github.com/pulumi-bot/pulumi-azure/compare/e11433a3..09653a74). There are cases like this https://github.com/pulumi-bot/pulumi-azure/compare/e11433a3..09653a74#diff-c4356aca96c1bb0d49266c7597be719f (note the line 42) that I saw in the AWS diff as well. Those are because the bullets below them lack indentation and perhaps we could fix them manually in our fork?

ekrengel:
I think this last one looks better! Please take a look @praneetloke

ekrengel:
> I did spot one case of a false positive

This is actually an interesting one. `PATCH_SET` is not an attribute so the best we could do here is probably remove the indent and bullet in our fork so it still gets included in the line above.

praneetloke:
You are right, @ekrengel. The latest AWS diff much better! One thing with Azure, though, which I think we could look at fixing manually perhaps. For example, it looks like the resource name is incorrectly picked-up: https://github.com/pulumi-bot/pulumi-azure/compare/366b9c4e..2f750cc3#diff-c4356aca96c1bb0d49266c7597be719f. It should be "network flow logging" and not "traffic analytics."

I am continuing to look at the other diffs.

ekrengel:
For the azure one, its because `enabled` appears multiple times and so its being overriden by the last occurence of `enabled` on the [page](https://raw.githubusercontent.com/terraform-providers/terraform-provider-azurerm/4ea792fdd0748e27fc27a3756cd375caa82902ef/website/docs/r/network_watcher_flow_log.html.markdown) so this is just following the same heuristic we use all over (which is technically incorrect and we need to fix separately)

praneetloke:
Could you please post a screenshot of the GCP resource docs with the changes from this PR? It'd be good to see what the formatting now looks like there for the property comments.

ekrengel:
So some negatives, but also some added info.. I dont think its terrible and I can try to fix it now if you think this is a big problem!

Now:
image

image


Before:

image

image

praneetloke:
> So some negatives, but also some added info.. I dont think its terrible

Yeah I agree too!

> I can try to fix it now if you think this is a big problem!

I think we should tackle it separately. Would you mind filing a new issue for that and add it to the Docs Quality project for us to come back to?

Would love to see what @justinvp thinks of the latest as well? I think we should merge this.

ekrengel:
New issue: https://github.com/pulumi/pulumi-terraform-bridge/issues/165

PR #162, opened 2020-04-17
Update tf2pulumi

pgavlin:
- Support multi-language examples in the schema target
- Link in tf2pulumi


go.mod
```diff
@@ -8,17 +8,27 @@ require (
...
 )
 
-replace github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.4.3+incompatible
+replace (
+	github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.4.3+incompatible
+	github.com/coreos/etcd => github.com/pulumi/etcd v3.3.18+incompatible
+	github.com/hashicorp/terraform-plugin-sdk => github.com/pulumi/terraform-plugin-sdk v0.0.0-20200416232118-ec806f20dbeb
+)
+
+replace github.com/pulumi/tf2pulumi => ../tf2pulumi
```
EvanBoyle, line 32: Looks like these two replaces are leftover from development?

build/common.mk
```diff
@@ -147,7 +147,6 @@ all:: build install lint test_all
 
 ensure::
 	$(call STEP_MESSAGE)
-	echo "GO111MODULE=on go mod tidy"; GO111MODULE=on go mod tidy;
```
EvanBoyle, line 150: Just curious on the reason for removing this?

pgavlin:
Should've posted this as WIP--hence the replaces and the removal of `go mod tidy`, both of which are for local dev.

lukehoban:
From the diffs, looks like we are adding `name: ` in a lot of places we didn't previously - including where it has the same value as the logical name.  Intentional?

E.g. https://github.com/pulumi-bot/pulumi-azuread/compare/76048a1..0343e75#diff-653e69ce47175546644fc36c188fe9aeR26

pgavlin:
> From the diffs, looks like we are adding name: in a lot of places we didn't previously - including where it has the same value as the logical name. Intentional?

Nope. Digging in now.

PR #164, opened 2020-04-18
Make the comment for datasource `id` property slightly more generic

justinvp:
While it is `id` in the schema, it's not always `id` in each projected language. For example, in C# and Go it's `Id`. Instead of having the case mismatch in some languages, or having to fix-up the comment for each language, just remove the identifier from the comment.

This also matches the comment I added for the `id` resource output property that will be injected in the docs. https://github.com/pulumi/pulumi/pull/4443

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -903,7 +903,7 @@ func (g *generator) gatherDataSource(rawname string,
 			Computed: true,
 		}
 		cust := &tfbridge.SchemaInfo{}
-		rawdoc := "id is the provider-assigned unique ID for this managed resource."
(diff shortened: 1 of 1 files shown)


praneetloke:
I just realized that to really make use of this, we'd have to override the terraform-bridge dependency in all of the providers to the newer version that contains this change or use a local override while regenerating the docs locally.

justinvp:
I don’t think there’s any urgency to use this. It’ll happen naturally as providers are updated and released.

praneetloke:
We might have to do it anyway for the other PRs (quality improvement and the one that enables Python examples) in this repo that haven't been merged yet.

Issue #11, opened 2019-07-17
limits link dropped?


ericrudder:
[docs/reference/pkg/nodejs/pulumi/aws/lambda/_index.md](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/lambda/)

property timeout 
public timeout: pulumi.Output<number | undefined>;
The amount of time your Lambda Function has to run in seconds. Defaults to 3. See [Limits][5]

ekrengel:
@stack72 are you working on this or should i take it on?

PR #166, opened 2020-04-20
Fix links from footer section

ekrengel:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/11

**Now:**

image

**Before:**

image


ekrengel:
All of these changes appear to be from a previous PR I already merged.

PR #171, opened 2020-04-23
Ensure Renamed Resources and DataSources are marked as deprecated

stack72:
This just adds a deprecation message to these renamed resources / datasources

It doesn't change the functionality apart from export that DeprecationMessage as part of the schema


stack72:
looks perfect :)

```
https://github.com/pulumi-bot/pulumi-aws/compare/f30ea9b43..7a0253878#diff-a5f7f21569b4696de5fb36dcfd89be63R7
```

Issue #165, opened 2020-04-20
Fix GCP argument descriptions


ekrengel:
A [recent PR](https://github.com/pulumi/pulumi-terraform-bridge/pull/161) increased the amount of GCP argument descriptions we had, but has some formatting issues. 

For example:

![image](https://user-images.githubusercontent.com/8304499/79619848-e7408f00-80c2-11ea-8e08-92a402abafba.png)

![image](https://user-images.githubusercontent.com/8304499/79619939-2bcc2a80-80c3-11ea-81ac-6fa44d192551.png)

We should remove the extra `-` and `(Optional)` or `(Required)` that are getting included with these descriptions.

PR #174, opened 2020-04-24
cleanup desc that are multiline -- eg GCP

ekrengel:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/165


ekrengel:
See GCP diff no longer has the `- (Required)` or `- (Optional)`

Issue #176, opened 2020-04-27
terraform-provider-aws changed bool to string for some schema properties


stack72:
This is essentially a braking change for our users - this is due to Terraform wanting to have a nullable type that doesn't pass any default option to the AWS API when a user doesn't specify a value

We need a way to be able to deal with this in Pulumi and to maybe detect when this happens

lukehoban:
If we override the type in our layer to be "bool", does it work?  In particular (a) does the case of passing `true`/`false` work and (b) does the case of it being undefined correctly map to the undefined value instead of to `false`?

stack72:
So I overrode the schema as follows:

```
"aws_launch_template": {
				Tok: awsDataSource(ec2Mod, "getLaunchTemplate"),
				Fields: map[string]*tfbridge.SchemaInfo{
					"network_interfaces": {
						Elem: &tfbridge.SchemaInfo{
							Fields: map[string]*tfbridge.SchemaInfo{
								"associate_public_ip_address": {
									Type: "boolean",
								},
							},
						},
					},
				},
			},
```

When I run pulumi up to get the output, it still comes back as a string it seems:

```
          {
                "urn": "urn:pulumi:dev::launchtempalte::pulumi:pulumi:Stack::launchtempalte-dev",
                "custom": false,
                "type": "pulumi:pulumi:Stack",
                "outputs": {
                    "x": {
                        "associatePublicIpAddress": "true",
                        "deleteOnTermination": false,
                        "description": "",
                        "deviceIndex": 0,
                        "ipv4AddressCount": 0,
                        "ipv4Addresses": [],
                        "ipv6AddressCount": 0,
                        "ipv6Addresses": [],
                        "networkInterfaceId": "",
                        "privateIpAddress": "",
                        "securityGroups": [],
                        "subnetId": ""
                    }
                }
            },
```

even though the type is boolean...

```
interface GetLaunchTemplateNetworkInterface {
        associatePublicIpAddress: boolean;
        deleteOnTermination: boolean;
        /**
         * Description of the launch template.
         */
        description: string;
        deviceIndex: number;
        ipv4AddressCount: number;
        ipv4Addresses: string[];
        ipv6AddressCount: number;
        ipv6Addresses: string[];
        networkInterfaceId: string;
        privateIpAddress: string;
        securityGroups: string[];
        subnetId: string;
    }
```

lukehoban:
[...]

PR #180, opened 2020-04-29
Allow overriding TF Schema types to be optional

stack72:
Fixes: #176

When TF schema is set as a string
pulumi override with bool and TF passes "true" or "false" == true | false in the output
pulumi override with bool but with an empty type and TF passes "true" or "false" then we get "true" or "false" in the output as we don't have a type to override it with
pulumi override with bool and TF passes "" then we get nil in the output
For this to work, we need to do the following in resources.go in the provider:

```
			"aws_launch_template": {
				Tok: awsDataSource(ec2Mod, "getLaunchTemplate"),
					"network_interfaces": {
						Elem: &tfbridge.SchemaInfo{
								"associate_public_ip_address": {
									Type:           "boolean",
									MarkAsOptional: boolRef(true),
								},
							},
						},
					},
				},
			},
```

This gives us the correct output type:

```
	modified:   sdk/nodejs/elasticloadbalancingv2/listener.ts
diff --git a/sdk/nodejs/types/output.ts b/sdk/nodejs/types/output.ts
index 0d2c33364..c63569a08 100644
--- a/sdk/nodejs/types/output.ts
+++ b/sdk/nodejs/types/output.ts
@@ -5806,7 +5806,7 @@ export namespace ec2 {
     }
     export interface GetLaunchTemplateNetworkInterface {
-        associatePublicIpAddress: boolean;
+        associatePublicIpAddress?: boolean;
         deleteOnTermination: boolean;
         /**
          * Description of the launch template.
```

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -28,7 +28,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
 	"github.com/mitchellh/copystructure"
 	"github.com/pkg/errors"
-
 	"github.com/pulumi/pulumi/sdk/v2/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/resource/plugin"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
@@ -654,8 +653,9 @@ func MakeTerraformOutput(v interface{},
 
 			// Otherwise, it might be a string that needs to be coerced to match the Terraform schema type. Coerce the
 			// string to the Go value of the correct type and, if the coercion produced something different than the string
-			// value we already have, re-make the output.
-			coerced, err := CoerceTerraformString(tfs.Type, t)
+			// value we already have, re-make the output. We need to ensure that we take into account any Pulumi schema
+			// overrides as part of this coercion
+			coerced, err := CoerceTerraformString(tfs.Type, ps, t)
 			if err != nil || coerced == t {
 				return resource.NewStringProperty(t)
 			}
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/schema_test.go
```diff
@@ -990,6 +990,28 @@ func TestInvalidAsset(t *testing.T) {
...
+		},
+		nil,   /* assets */
+		false, /*useRawNames*/
+		true,
+	)
+	assert.Equal(t, resource.NewPropertyMapFromMap(map[string]interface{}{
+		"myStringPropertyValue": true,
+	}), result)
+}
+
```
lukehoban, line 1014: What happens with the empty string?  I believe the motivation for this change upstream is to support “undefined” as well as true and false as values. Is that true?  And if so - how does it manifest in practice in providers (and as a result - what do we need to support here)?

stack72: if we pass an override as an empty string then we just return the string - the same behaviour as the default case of the TF provider

pkg/tfbridge/schema.go
```diff
@@ -1022,33 +1022,61 @@ func CleanTerraformSchema(tfs map[string]*schema.Schema) map[string]*schema.Sche
 }
 
 // CoerceTerraformString coerces a string value to a Go value whose type is the type requested by the Terraform schema
-// type. Returns an error if the string can't be successfully coerced to the requested type.
-func CoerceTerraformString(schType schema.ValueType, stringValue string) (interface{}, error) {
+// type or the Pulumi SchemaInfo. We prefer the SchemaInfo overrides as it's an explicit call to action over the
+// Terraform Schema. Returns an error if the string can't be successfully coerced to the requested type.
+func CoerceTerraformString(schType schema.ValueType, schemaOverride *SchemaInfo, stringValue string) (interface{}, error) {
```
lukehoban, line 1027: Very minor - but we call this `*SchemaInfo` `ps` almost everywhere else in this file - might make sense to keep the naming the same here.  (whether that's the best name anywhere is certainly debatable - but consistency...)

pkg/tfbridge/schema.go
```diff
@@ -1022,33 +1022,61 @@ func CleanTerraformSchema(tfs map[string]*schema.Schema) map[string]*schema.Sche
...
+	// check for the override and use that over terraform if available
+	// we do this to ensure that we are following the explicit call to action of the override
+	if schemaOverride != nil && schemaOverride.Type != "" {
+		switch schemaOverride.Type {
+		case "boolean":
```
lukehoban, line 1032: Are the names of these types defined anywhere?  Like is there any existing sense in which `"boolean"` is meaningful as a value of this (vs. for example `"bool"`)?  Should we define the legal values of these types somewhere?

pkg/tfbridge/schema_test.go
lukehoban: Sorry - my question was about `""` coming from the provider as the value of the property.  So what happens if `""` is the value in the first argument to `TestOverridingTFSchema`. And ultimately will that show up sensibly in Pulumi SDKs (as an undefined value for an optional output I guess)?

Really I'm just interested in making sure this really works for the actual scenarios where we need to map into and out of Terraform resources that use this (across the range of values those provider can in practice accept or return).

pkg/tfbridge/schema.go
stack72: This is interesting actually. I only used primitive types for these as we define these as strings in our Pulumi SchemaInfo - if you think changing these to be an actual type then I'd be behind that

pkg/tfbridge/schema.go
stack72: done :)

pkg/tfbridge/schema.go
mikhailshilkov: Would it make sense to use the same types as in schema? AFAIK, those are `boolean`, `integer`, and `number`.

pkg/tfbridge/schema.go
```diff
@@ -1022,33 +1022,64 @@ func CleanTerraformSchema(tfs map[string]*schema.Schema) map[string]*schema.Sche
...
+		return convertTfStringToFloat(stringValue)
 	}
 
 	// Else it's just a string.
 	return stringValue, nil
 }
 
+func convertTfStringToBool(stringValue string) (interface{}, error) {
+	if stringValue == "" {
+		return false, nil // false is the default bool
```
mikhailshilkov, line 1058: Is `false` really the proper default for all cases? Why would they need to introduce the string to express 3 states then? Can we actually pass `undefined`/`null`/`nil` instead?

pkg/tfbridge/schema.go
```diff
@@ -1022,33 +1022,64 @@ func CleanTerraformSchema(tfs map[string]*schema.Schema) map[string]*schema.Sche
...
+	}
+	boolVal, err := strconv.ParseBool(stringValue)
+	if err != nil {
+		return nil, err
+	}
+	return boolVal, nil
+}
+
+func convertTfStringToInt(stringValue string) (interface{}, error) {
+	intVal, err := strconv.ParseInt(stringValue, 0, 0)
```
mikhailshilkov, line 1068: What should happen in the case of an empty string here?

pkg/tfbridge/schema.go
pgavlin: Yeah, I'm not sure that this is correct either.

The original behaviors used by this method were taken directly from the TF engine; we should continue this approach so we are consistent with TF's semantics.

stack72:
The Azure downstream failure is due to an issue with moment JS that was released into the npm ecosystem last night...

pkg/tfgen/generate.go
```diff
@@ -373,6 +373,11 @@ func (v *variable) optional() bool {
 		return true
 	}
 
+	//if we have an explicit marked as optional then let's return that
+	if v.info != nil && v.info.MarkAsOptional != nil {
```
lukehoban, line 377: Minor - but looks like we aren't actually checking the value of the bool?  Humorously, this is a case of confusing ternary logic similar to the one this issue is addressing :-).

PR #188, opened 2020-05-05
Update pulumi/pulumi

pgavlin:
And fix a couple lint issues that snuck in.


pgavlin:
@stack72 note that this doesn't necessarily need to be rolled out for downstream to pick up codegen fixes: those repos can also just `go get github.com/pulumi/pulumi/pkg/v2@master`

PR #191, opened 2020-05-11
Fix TransformJSONDoc when passing null due to regression in CoerceTerraformString

stack72:
The changes made in #180 introduced a regression. When a Pulumi override
was present and there was an empty string, the CoerceTerraformString func
returned nil, rather than the empty string. This was problematic when
there was a coupled TransformJSONDocument with that. We need to change
the functionality to do 2 things:

1. CoerceTerraformString - don't return nil for any empty string override
except boolean. This allows us to manage the Terraform changes to be strings
for bools that can be unset e.g. AssociatePublicIp

2. We now need to understand when a `nil` value is passed to the
TransformJSONDocument that we need to treat this as there is an empty
string being passed to it

Before:

```
▶ pulumi up
Previewing update (dev):
     Type                 Name        Plan     Info
     pulumi:pulumi:Stack  create-dev
     └─ aws:s3:Bucket     test                 1 error

Diagnostics:
  aws:s3:Bucket (test):
    error: preparing urn:pulumi:dev::create::aws:s3/bucket:Bucket::test's old property state: expected string or JSON map; got <nil>

```

After:

```
▶ pulumi up
Previewing update (dev):
     Type                 Name        Plan
     pulumi:pulumi:Stack  create-dev

Resources:
    2 unchanged

Do you want to perform this update? details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::create::pulumi:pulumi:Stack::create-dev]

```

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -1027,10 +1027,12 @@ func CleanTerraformSchema(tfs map[string]*schema.Schema) map[string]*schema.Sche
 func CoerceTerraformString(schType schema.ValueType, ps *SchemaInfo, stringValue string) (interface{}, error) {
 	// check for the override and use that over terraform if available
 	// we do this to ensure that we are following the explicit call to action of the override
+	// For now, we will only return nil when an override of the type is a boolean and there is no
+	// default value supplied - this will allow us to replicate the nullable-esquq bools that Terraform are
+	// creating by using strings in place of bools
+	// if we return nil for *all* override types when there is an empty string, then we can hit an edge case of
+	// breaking overrides where we have a string and a TransformJSONDocument (see pulumi/pulumi#4592)
 	if ps != nil && ps.Type != "" {
-		if stringValue == "" {
-			return nil, nil
-		}
 		switch strings.ToLower(ps.Type.String()) {
 		case "boolean":
 			if stringValue == "" {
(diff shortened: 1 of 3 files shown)


lukehoban:
Can we add test coverage?  In this repo and/or in pulumi/aws?

stack72:
> Can we add test coverage? In this repo and/or in pulumi/aws?

what do you want the test coverage to do - test for the nil or test that we don't regress again?

lukehoban:
> what do you want the test coverage to do - test for the nil or test that we don't regress again?

Both I suppose. Though I was thinking specifically of validation that `TransformJSONDoc` works to avoid future regressions.

stack72:
> > what do you want the test coverage to do - test for the nil or test that we don't regress again?
> 
> Both I suppose. Though I was thinking specifically of validation that `TransformJSONDoc` works to avoid future regressions.

great point - apologies for not adding this earlier! both missing test cases (empty string and null) have been added!

pkg/tfbridge/schema.go
```diff
@@ -1027,10 +1027,12 @@ func CleanTerraformSchema(tfs map[string]*schema.Schema) map[string]*schema.Sche
 func CoerceTerraformString(schType schema.ValueType, ps *SchemaInfo, stringValue string) (interface{}, error) {
 	// check for the override and use that over terraform if available
 	// we do this to ensure that we are following the explicit call to action of the override
+	// For now, we will only return nil when an override of the type is a boolean and there is no
```
pgavlin, line 1030: We should follow up with some investigation into whether or not this is actually safe. In particular, I'd like to know whether or not a non-existent property is treated identically to a property that exists and has the value `null`.

Issue #132, opened 2020-03-18
Add a place for extra types in the ProviderInfo type and update schema generator to use it


praneetloke:
Related to https://github.com/pulumi/pulumi/issues/4003.

This work involves adding support for typed JSON objects to the schema. The following tasks need to be completed:

* [ ] Update `ProviderInfo` type to include a new map types names and their object spec
* [ ] Update the schema generator to use the new property to generate nested types info from it
* [ ] Update the AWS provider repo to pass the information on policy documents and any other typed objects that are listed as "altTypes" in the `resources.go`

PR #192, opened 2020-05-11
Allow packages to define extra schema types.

pgavlin:
These types are only supported by the schema-based backends; NodeJS must
continue to use hand-written overlays.

This is part of #132.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,11 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
-- Populate default Tf timeouts. [#177](https://github.com/pulumi/pulumi-terraform-bridge/issues/177)
+- Allow providers to define additional object types in the schema.
+  [#192](https://github.com/pulumi/pulumi-terraform-bridge/pull/192)
+
+- Populate default TF timeouts.
+  [#177](https://github.com/pulumi/pulumi-terraform-bridge/issues/177)
 
 - Link in tf2pulumi for preliminary HCL2 support and Python codegen.
   [#162](https://github.com/pulumi/pulumi-terraform-bridge/pull/162)
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -17,6 +17,7 @@ package tfbridge
 import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+	pschema "github.com/pulumi/pulumi/pkg/v2/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/tokens"
 )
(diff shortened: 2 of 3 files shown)


pgavlin:
Here's an example of this as applied to `iam.PolicyDocument`: https://github.com/pulumi/pulumi-aws/compare/pgavlin/schemaOverlayExample

There's still some supporting work to do in the Go codegen before this is entirely usable, but this unblocks the ability to get to work writing the schema types.

lukehoban:
This change itself looks great to me :tada:.

>  Here's an example of this as applied to iam.PolicyDocument: https://github.com/pulumi/pulumi-aws/compare/pgavlin/schemaOverlayExample

A few notes/questions on this.  

```go
var iamPolicyDocument = pschema.ObjectTypeSpec{
```

I actually kind of like writing the schemas in Go instead of in JSON, as it lets us do more interesting programatic construction (and type-checking) as is taken advantage of in 

```go
string(awsType(iamMod, "documents", "PolicyDocument")):     iamPolicyDocument,
```

Is `documents` here just never used today?

```patch
-      public Input<string> Policy { get; set; } = null!;
+      public InputUnion<string, Pulumi.Aws.Iam.Inputs.PolicyDocumentArgs> Policy { get; set; } = null!;
```

I don't recall exactly - is this a breaking change?  One of the things I'm really worried about is that .NET and Go will cause it to be a breaking change to improve these typings.  Hopefully not the case for .NET?  Not sure what if anything we can do there, except have a lot more breaking changes in Go.


```cs
public sealed class PolicyDocumentArgs : Pulumi.ResourceArgs
```

Is `Args` the right name here?  I might have expected it to be just `PolicyDocument`.

```cs
public sealed class PolicyDocumentGetArgs : Pulumi.ResourceArgs
```

Why do we need to generate `GetArgs`?

```cs
public InputUnion<string, Pulumi.Aws.Iam.Inputs.PolicyDocumentArgs>? Policy { get; set; }
```

I'd also love to see an example of what the .NET usage of this PolicyDocument looks like.  Curious @mikhailshilkov if you think it will feel "natural"/"good"?

mikhailshilkov:
> I don't recall exactly - is this a breaking change? One of the things I'm really worried about is that .NET and Go will cause it to be a breaking change to improve these typings. Hopefully not the case for .NET?

It is somewhat of a breaking change. First, it's breaking on the binary level: if somebody compiles a library using the `Input<string>` version, a client with `InputUnion<string, x>` will probably fail.

Also, while we do have implicit conversions from `string` and `Output<string>` to union, we don't yet have one for `Input<string>`, so

```cs
string plain = "123";
Output<string> output = Output.Create("456");
Input<string> input = Output.Create("789");

InputUnion<string, PolicyDocumentArgs> policy = null;
policy = plain; //works
policy = output; //works
policy = input; //doesn't work
```

Lastly, all those assignments would fail in F# as it's more strict and has no implicit conversions.

> I'd also love to see an example of what the .NET usage of this PolicyDocument looks like. Curious @mikhailshilkov if you think it will feel "natural"/"good"?

A basic sanity check LGTM:

```cs
policy = new PolicyDocumentArgs
{
    Version = "2012-10-17",
    Statement =
    {
        new PolicyStatementArgs
        {
            Effect = "Allow",
            Principal = "*",
            Action = "s3:GetObject",
            Resource = $"arn:aws:s3:::{id}/*",
        }
    }
};
```

> `InputUnion<string, InputUnion<AWSPrincipalArgs, InputUnion<ServicePrincipalArgs, FederatedPrincipalArgs>>>`

This doesn't look very nice. Not only because of verbose shape, but also because this won't compile:

```cs
...
// C# can't figure out the conversion here
Principal = new FederatedPrincipalArgs(),
...
```

Also none of these work:

```cs
// C# doesn't understand that we ask for the immutable array option here
Resource = { $"arn:aws:s3:::{id}/*" },
Resource = new[] { $"arn:aws:s3:::{id}/*" },
```

I guess we may need to refine some types here, but that's not quite related to this PR.

pgavlin:
> Is documents here just never used today?

Correct.

> Is Args the right name here? I might have expected it to be just PolicyDocument.

FWIW, this is consistent with the codegen for the rest of the input types. To be very clear, the extra types operate exactly as do nested types: they use the same constructs in the schema. So we can't change one without changing the other unless we add lang-specific annotations or change the schema.

> Why do we need to generate GetArgs?

I think this happens by default in C# b/c there is an assumption that anything in the schema's `types` section might be part of the state inputs passed to `Get`. We don't really need to generate these (and indeed, I don't think that the Go code generator does).

lukehoban:
> FWIW, this is consistent with the codegen for the rest of the input types. To be very clear, the extra types operate exactly as do nested types: they use the same constructs in the schema. 

Right - I guess I was using this as a litmus test of whether that is the “right” way to represent these additional types - basically whether these overlayed types make sense to be treated the same a nested resource types, or really “should” be different. I *think* the answer is that this works well as-is and doesn’t justify a new kind of type generation. 

A related question- are there any usecases for this today beyond the JSON Transform capability?  I think the answer is no?

mikhailshilkov:
> A related question- are there any usecases for this today beyond the JSON Transform capability?

Can we use this for enum-like strings, e.g. regions/locations/tiers?

pgavlin:
> Can we use this for enum-like strings, e.g. regions/locations/tiers?

Not yet. We still need to add support for these in the schema itself.

lukehoban:
> > Can we use this for enum-like strings, e.g. regions/locations/tiers?
> 
> 
> 
> Not yet. We still need to add support for these in the schema itself.

Which is tracked in https://github.com/pulumi/pulumi/issues/4003.

pgavlin:
Just double-checking: is this good to go once the merge conflict is resolved?

lukehoban:
> Just double-checking: is this good to go once the merge conflict is resolved?

I think so.  The one other question that I forgot to ask previously - will the Node.js codegen for the IAM policies be compatible enough with the current manually managed types that we'll be able to switch to using it (once Node.js+tfgen+schema lands overall)?  For example - will it be able to live in the same place?

pgavlin:
> will the Node.js codegen for the IAM policies be compatible enough with the current manually managed types that we'll be able to switch to using it

I _believe_ so. I'll verify end-to-end before we start using it at all, but I think we can go ahead and land this PR regardless.

> For example - will it be able to live in the same place?

Do you mean in the same files?

lukehoban:
> Do you mean in the same files?

No - just namespaces.

pgavlin:
> No - just namespaces.

Ah, got it. Yeah, that should be possible.

PR #193, opened 2020-05-12
Populate default TF timeouts

komalali:
Fixes: https://github.com/pulumi/pulumi-terraform-bridge/issues/177

When creating a new resource, we start with an empty Meta object in state, so default timeout information has to be populated through the diff. For other operations, the timeout info is already in state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,10 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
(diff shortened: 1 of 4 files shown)


komalali:
@pgavlin I updated the tests that were using `tf.Diff()` (and therefore falsely passing because it handles timeouts) to use `tf.SimpleDiff()` like the actual code uses and added some more checks. 

Doing that identified an issue with the setTimeout function, which I fixed so it only overwrites the specific timeoutKey and not the entire timeouts object.

PR #194, opened 2020-05-13
Use the schema code generator for NodeJS.

pgavlin:
Mostly just what it says on the tin. There is one documentation fix that
adds an attribution line to doc comments in all languages.


pgavlin:
Note that this depends on https://github.com/pulumi/pulumi/pull/4618.

Diffs for pulumi-aws: https://github.com/pulumi/pulumi-aws/compare/pgavlin/nodeSchema

pgavlin:
The AWS is pretty noisy due to some whitespace differences. I'll clean those up with some preparatory changes in tfgen.

pgavlin:
The example AWS diff is now a bit cleaner, courtesy of pulumi/pulumi-terraform-bridge#195.

pgavlin:
AWS and GCP both fail because they need very small tweaks to their `resources.go` files to fix some malformed type tokens, but this is otherwise working. I'll generate diffs for those providers manually.

There is one major issue: https://github.com/pulumi/pulumi-terraform-bridge/issues/152. The shapes of many of the generated output types change because of this, making many nested properties optional where they are currently required. We can fix this by adding node-specific metadata to object types that indicates which properties are required on the output side of the API, or we can fix this by taking https://github.com/pulumi/pulumi-terraform-bridge/issues/152 and adjusting downstream code generators appropriately.

lukehoban:
> There is one major issue: #152. The shapes of many of the generated output types change because of this, making many nested properties optional where they are currently required. We can fix this by adding node-specific metadata to object types that indicates which properties are required on the output side of the API, or we can fix this by taking #152 and adjusting downstream code generators appropriately.

Indeed - that is definitely a breaking change.  

I do not think we could make that change without a major version bump, as it would cause a lot of existing code which currently works fine to start erroring.  Moreover, many/most cases of this would not be perceived by users as being an improvement.

I think we will need to introduce a Node.js compatibility mode here - and then open an issue to track deciding on whether to take a breaking change in 3.0 to drive consistency across languages here.

pgavlin:
Diff for [pulumi-aws](https://github.com/pulumi/pulumi-aws/compare/pgavlin/ptfb-194)
Diff for [pulumi-gcp](https://github.com/pulumi/pulumi-gcp/compare/pgavlin/ptfb-194)
Diff for [pulumi-azure](https://github.com/pulumi/pulumi-azure/compare/pgavlin/ptfb-194)

stack72:
can you explain why `Type: awsTypeNoFile("region", "Region")` => `Type: awsTypeDefaultFile(awsMod, "Region")`  is required for the AWS provider? 

These specific funcs have always been confusing as they are not used elsewhere

pgavlin:
> can you explain why Type: awsTypeNoFile("region", "Region") => Type: awsTypeDefaultFile(awsMod, "Region") is required for the AWS provider?

The schema code generator inspects the type token in order to generate an appropriate `import` statement. The module to import is derived from the module portion of the type token. `awsTypeNoFile` generates something like `aws:region:Region`, but the `Region` type is actually located in the root module. The `NoFile` type token is also out of line with the format otherwise used by the AWS package, which is `aws:module/file:Entity`. The use of `awsTypeDefaultFile(awsMod, "Region")` generates the type token `aws:index/region:Region`, which fixes both of these issues.

pgavlin:
There are two remaining diffs:
- Some comment lines break the 120-character limit. This is because the NodeJS code generator does not reflow comment lines, and the initial padding (e.g. `    * `) increases the length of the line. I propose we take this diff.
- Some files are cased differently. For example, the GCP SDK's `compute.SSLCertificate` is generated into the file `sdk/nodejs/compute/sslcertificate.ts` rather than `sdk/nodejs/compute/sSLCertificate.ts`. The new path seems like an improvement on the original, so I propose that we take this, even though it will involve deleting the currently-checked in versions of the generated files and generating new versions.

pgavlin:
cc @lukehoban for the casing changes

lukehoban:
> The new path seems like an improvement on the original, so I propose that we take this, even though it will involve deleting the currently-checked in versions of the generated files and generating new versions.

Agreed with this.  I am okay with breaking changes to the on-disk file naming.

pgavlin:
Alright. If nobody has any objections, I am going to go ahead and merge.

pgavlin:
I'll then work with @stack72 to handle the rollout of these changes to the major providers.

leezen:
Awesome! Very exciting!

PR #197, opened 2020-05-13
Remove a generated comment.

pgavlin:
This brings the legacy Node code generator in line with the schema
generators.


lukehoban:
Mostly unrelated - but do you know why we have so many blank lines in generated comments?

https://github.com/pulumi-bot/pulumi-random/compare/a879831..18f0eab#diff-77f5d574e4bc235c8a929b648a058987R18

pgavlin:
I don't, no. They're present in both the legacy and the schema-based code generator. We should clean those up later.

PR #198, opened 2020-05-13
Do not emit deprecation messages into descriptions.

pgavlin:
This should be left up to the code generators.


pgavlin:
Note that this has the side-effect of dropping deprecation messages from Go entirely. We'll want to fix the Go code generator so that it adds them.

pgavlin:
https://github.com/pulumi/pulumi/pull/4631 should fix the lack of deprecation comments for Go.

Issue #200, opened 2020-05-14
Ensure Version isn't set as part of generated schema


stack72:
The new flow will be to generate the schema.json. We will inject the version as part of the provider build before the language generation is taken care of

This can ensure we can check the schema.json file into github and it can be used as part of the schema diff checking

PR #201, opened 2020-05-14
Omit version from schema.PackageSpec

stack72:
This will be passed as part of provider builds so that we can
check the schema.json files in to allow schema diffs

Fixes: #200


stack72:
I believe this will break the codegeneration until this is ingested and the VERSION is passed in the makefile

pgavlin:
> I believe this will break the codegeneration until this is ingested and the VERSION is passed in the makefile

I don't think that will fix the code generators: they regenerate the schema in memory before codegen. I think you should instead remove the version immediately before writing the schema out in `schemaGenerator.emitPackage`.

stack72:
@pgavlin you were correct :) this does indeed stop the codegen being broken :)

PR #203, opened 2020-05-18
Refactor TF Markdown parsing.

pgavlin:
- Collect most of the context into a struct type
- Factor section and subsection parsing into methods
- Simplify subsection kinds

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -26,13 +26,10 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/pulumi/pulumi-terraform-bridge/v2/pkg/tfbridge"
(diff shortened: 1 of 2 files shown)


pgavlin:
Aside from general goodness, this also lays the groundwork for further refactoring:
- I intend to experiment with parallelizing Markdown parsing, which I'm hoping will speed up docs generation for providers with large numbers of examples
- I am also exploring the use of a reimplementing the Markdown parser on top of a proper AST (see https://pkg.go.dev/gopkg.in/russross/blackfriday.v2?tab=doc)

pgavlin:
> Looking at your changes, it doesn't look like this is due to your changes, though?

Correct. This is due to other changes that haven't yet been rolled out.

@stack72 @jen20 we should look into changing the downstream gen tests to generate once with the prior commit and once with the merge commit s.t. the results only represent the changes due to the PR.

PR #204, opened 2020-05-19
Add C# example generation

mikhailshilkov:


mikhailshilkov:
GCP build gets killed somehow but the reason is unclear. It works locally for me. Let's merge this and try directly in `pulumi-gcp`.

PR #211, opened 2020-05-27
Wrap all examples in shortcode.

pgavlin:
Including examples that are not nested inside an "Example Usage"
section.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -313,37 +313,6 @@ func getDocsIndexURL(org, p string) string {
 	return getDocsBaseURL(org, p) + "/index.html.markdown"
 }
 
-// fixExampleTitles looks at every line and fixes any title immediately preceding
-// a code snippet.
-func fixExampleTitles(lines []string) {
-	num := len(lines)
-	for i, line := range lines {
-		if strings.HasPrefix(line, "#### ") {
-			// Make sure we don't go beyond the number of lines we are dealing with.
-			if i >= num-1 {
-				break
-			}
-
-			// For every line that comes after this, process each line until we hit
-			// the beginning of a code-fence.
-			j := i + 1
-			for j < num {
-				if lines[j] == "\n" || lines[j] == "" {
-					j++
-					continue
-				}
-
-				if strings.HasPrefix(lines[j], "```") {
-					lines[i] = strings.ReplaceAll(line, "#### ", "### ")
-					// Stop searching further lines since we found the first code snippet.
-					break
-				}
-				j++
-			}
-		}
-	}
-}
-
 // parseTFMarkdown takes a TF website markdown doc and extracts a structured representation for use in
 // generating doc comments
 func parseTFMarkdown(g *generator, info tfbridge.ResourceOrDataSourceInfo, kind DocKind,
@@ -373,6 +342,14 @@ type tfMarkdownParser struct {
 	ret entityDocs
 }
 
+const (
+	sectionOther               = 0
+	sectionExampleUsage        = 1
+	sectionArgsReference       = 2
+	sectionAttributesReference = 3
+	sectionFrontMatter         = 4
+)
+
 func (p *tfMarkdownParser) parse() (entityDocs, error) {
 	p.ret = entityDocs{
 		Arguments:  make(map[string]*argumentDocs),
@@ -384,7 +361,12 @@ func (p *tfMarkdownParser) parse() (entityDocs, error) {
 	markdown := strings.Replace(p.markdown, "\r\n", "\n", -1)
 
 	// Split the sections by H2 topics in the Markdown file.
-	for _, section := range splitGroupLines(markdown, "## ") {
+	sections := splitGroupLines(markdown, "## ")
+
+	// Reparent examples that are peers of the "Example Usage" section (if any) and fixup some example titles.
+	sections = p.reformatExamples(sections)
+
+	for _, section := range sections {
 		if err := p.parseSection(section); err != nil {
 			return entityDocs{}, err
 		}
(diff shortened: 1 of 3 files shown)


pgavlin:
This is in support of https://github.com/pulumi/pulumi-terraform-bridge/issues/212.

While this matches my own expectations of what we should be generating, it does not match that of `pulumi/pkg/codegen/docs`: the example processing code in that package expects all examples in a piece of documentation to be nested inside a single `{{% examples %}}` shortcode. That doesn't seem correct to me, as examples may appear anywhere (and indeed do in the source TF). I see two options:

- Allow `{{% example %}}` to appear anywhere, including outside `{{% examples %}}`
- Allow multiple `{{% examples %}}` sections

These aren't exclusive, but either would be sufficient to address this problem.

pgavlin:
See e.g. https://github.com/pulumi-bot/pulumi-aws/compare/c19ae3e59..7ba332cb9#diff-5eacaeb84159f7448eba0923f1bbb27c for an example of the free-floating `{{% example %}}` markup this adds.

pkg/tfgen/docs.go
```diff
@@ -505,7 +502,7 @@ func (p *tfMarkdownParser) parseSection(section []string) error {
 				}
 			}
 			description := strings.Join(subsection, "\n") + "\n"
-			if sectionKind == sectionExampleUsage {
+			if hasExamples {
```
praneetloke, line 505: Instead of this, what do you think about identifying the example sections in `parse()` and making sure that the titles are nested under an h2? That is, identify all h2's (`##`) that have an example code-block and convert the title of those sections to an h3 (`###`) and insert an h2 called `Example Usage` at the beginning before the first h3 example section.

pgavlin:
> I am not sure I quite understand what you mean by "may appear anywhere". Do you have an example (no pun intended) of this?

See e.g. https://github.com/pulumi-bot/pulumi-aws/compare/c19ae3e59..7ba332cb9#diff-5eacaeb84159f7448eba0923f1bbb27cL11-L118, which I linked above. There are examples under `h2`s that are peers of the `## Example Usage` in the [original docs](https://www.terraform.io/docs/providers/aws/r/lambda_permission.html). This is very infrequent in the AWS and Azure providers, where fewer than 5% of entities exhibit this structure, but more prevalent in the GCP provider, where this structure occurs in ~50% of entities.

>  That is because the {{% examples %}} short-code is used to denote the start and end of all examples. It would be good to maintain that still.

Can you elaborate?

> Moreover, the `{{% examples %}}` is the mount-point for the lang chooser. If we have multiple of those, then we would have multiple lang choosers on the page.

Isn't the language a page-global setting? I would have expected things to work more like the .NET docs, e.g. https://docs.microsoft.com/en-us/dotnet/api/system.console?view=netcore-3.1. That structure is flexible in where code can appear. One of the reasons I originally proposed markdown as the docs language with shortcodes for additional semantics was to retain flexibility in authoring without putting too much burden on the frontend.

pgavlin:
FWIW, I do think that having a single `{{% examples %}}` section is useful--some canonical place to keep "examples" is great--but I think that it is also good to support code snippets outside the examples section.

In the case of tfgen, I think your earlier suggestion of collecting all examples under the `{{% examples %}}` shortcode will deliver good results, but I don't want to artificially constrain the structure of documentation solely based on what we can do here.

praneetloke:
> There are examples under h2s that are peers of the ## Example Usage in the original docs. This is very infrequent in the AWS and Azure providers, where fewer than 5% of entities exhibit this structure, but more prevalent in the GCP provider, where this structure occurs in ~50% of entities.

Ok yeah you are referring to the same thing. I wasn't sure if there was a different type of example structuring in the source docs that I had not seen before. I think we should be deliberate here and make those examples _not_ peers of the `## Example Usage` title. My suggestion is to always convert any peers as descendants of it.

> Can you elaborate?

Yep. Could you please check out this [example](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs/gen_test.go#L127) used in one of the docs generator [tests](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs/gen_test.go#L387)? Please let me know if it is not clear.

> Isn't the language a page-global setting? I would have expected things to work more like the .NET docs, e.g. https://docs.microsoft.com/en-us/dotnet/api/system.console?view=netcore-3.1. That structure is flexible in where code can appear.

Yes, the language setting is global in our docs. However, the language chooser component itself is not at the page-level like it is in the MS Docs link you shared. I don't think it is in the plans to make it that way right now. I could be wrong, though. (@cnunciato may have a better answer here.) But putting the page-level lang chooser aside, we _could_ support  multiple `{{% examples %}}` short-codes on the page. It is just a matter of looking at all matches of `{{% examples %}}` rather than only the first one [like we do today](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs.go#L130) (the linked function is used by both the lang SDKs to identify and strip non-relevant lang snippets, as well as by the docs generator to extract the examples section and do more processing on it to bring it more closer to our final markdown structure.)

> In the case of `tfgen`, I think your earlier suggestion of collecting all examples under the {{% examples %}} short-code will deliver good results, but I don't want to artificially constrain the structure of documentation solely based on what we can do here.

I don't think we'd be constraining the doc structure permanently, though. It would be for TF-based providers only. I believe once we make some more improvements to how the lang chooser component is rendered throughout the site, we should be able to support multiple `{{% examples %}}` short-codes. (Also note my earlier point above about this.) I just don't think the current content is laid out in a way that really warrants for this. Also, it is another way to solve this issue of examples sections that are peers to `## Example Usage`, but I think it is the way it is due to inconsistency in the upstream doc rather than a deliberate decision to have it be that way.

pgavlin:
@praneetloke the latest commit gathers all peer "Example Usage" sections under a single H2. It also logs when code snippets are encountered in other sections. Unfortunately, this sort of issue occurs frequently, especially in the GCP provider. In some cases, it may be possible to pull these examples into the canonical examples section, but in others the examples live in other important places (e.g. the front matter, argument reference sections, etc.). As such, I continue to believe that we need to support `{{% examples %}}` or `{{% example %}}` sections in multiple places.

stack72:
> @praneetloke the latest commit gathers all peer "Example Usage" sections under a single H2. It also logs when code snippets are encountered in other sections. Unfortunately, this sort of issue occurs frequently, especially in the GCP provider. In some cases, it may be possible to pull these examples into the canonical examples section, but in others the examples live in other important places (e.g. the front matter, argument reference sections, etc.). As such, I continue to believe that we need to support `{{% examples %}}` or `{{% example %}}` sections in multiple places.

FWIW, this is more than likely as there are multiple providers sharing the same docs. Some of the individual <resource>_iam_* docs are all gathered

praneetloke:
@pgavlin awesome! I'll take a look tomorrow.

praneetloke:
@stack72 I am not sure what you mean by this 👇 ?

> FWIW, this is more than likely as there are multiple providers sharing the same docs. Some of the individual iam* docs are all gathered

PR #217, opened 2020-06-08
Fix README generation for python

komalali:
Related to: https://github.com/pulumi/pulumi/pull/4759

### DIFF
--- a/pkg/tfgen/generate_schema.go
+++ b/pkg/tfgen/generate_schema.go
@@ -309,12 +309,13 @@ func (g *schemaGenerator) genPackageSpec(pack *pkg) (pschema.PackageSpec, error)
 	}
 	spec.Language["nodejs"] = rawMessage(nodeData)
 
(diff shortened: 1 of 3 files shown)


pkg/tfgen/generate_schema.go
```diff
@@ -312,6 +312,7 @@ func (g *schemaGenerator) genPackageSpec(pack *pkg) (pschema.PackageSpec, error)
 	if pi := g.info.Python; pi != nil {
 		spec.Language["python"] = rawMessage(map[string]interface{}{
 			"requires": pi.Requires,
+			"readme":   readme,
 		})
 	}
```
pgavlin, line 317: You should do something similar to the `nodeData` approach above, where the readme is set outside the check for a non-`nil` `g.info.Python`. Not all packages will specify Python-specific information, but all packages will need the Python-specific README, since we will generate Python code regardless.

PR #218, opened 2020-06-08
update to include go example generation

EvanBoyle:
Locally linked up with the aws provider to test: https://github.com/pulumi/pulumi-aws/compare/evan/goExamples

There are still a handful of issues with the go examples, but they are close enough to push out docs while we polish things up. 

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/216


pgavlin:
Took a glance through the generated code. Mostly this looks great! Just a couple things:

- https://github.com/pulumi/pulumi-aws/compare/evan/goExamples#diff-5d8be00c66f592ec4c052f1adf9ed76aR34-R39, https://github.com/pulumi/pulumi-aws/compare/evan/goExamples#diff-753f1c7ad589f4f70cebf932d8963d98R77-R86 Any idea what's up here?
- https://github.com/pulumi/pulumi-aws/compare/evan/goExamples#diff-adcc3cc51f7a96128c34d5a8f3600384R61 This looks like the output of the code generator. IMO these TODOs should be diagnostic errors for the time being s.t. we avoid generating invalid examples.

EvanBoyle:
@pgavlin looks like one case is handling StringMapOutput and the other might be a bug in the __input intrinsic recursion logic. I'm going to hold off on opening specific issues for these in favor of a blanket issue to do a quality sweep: https://github.com/pulumi/pulumi/issues/4783

Will work on turning TODOs into errors https://github.com/pulumi/pulumi/issues/4782

I don't think we need to block on either. From my conversations with @leezen we're willing to accept some quality issues in the first round of examples. We have control over whether or not we publicize their existence on the docs page and drive traffic.

pgavlin:
> I don't think we need to block on either. From my conversations with @leezen we're willing to accept some quality issues in the first round of examples. We have control over whether or not we publicize their existence on the docs page and drive traffic.

About 25% of the examples in AWS are currently hitting `TODO`s, and about 80% of the `TODO`s are multi-part template expressions. I'd prefer that we at least fix that issue before shipping this, as it should be as simple as a call to `pulumi.Sprintf`.

leezen:
FWIW, I don't think emitting "TODOs" into the examples makes a ton of sense. I'd rather just not emit them at all. My comment about "ship first" was around getting something incremental landed then continuing to iterate from there, but I should have clarified that by "incremental" I mean more like having coverage for a small percentage of examples (rather than more examples, but with TODOs in them).

Issue #212, opened 2020-05-27
Use the current schema when converting examples for schema.json.


pgavlin:
Today, the schema used when converting examples for inclusion in schema.json is sourced from the latest version of the plugin being built. Instead, the schema should be the sourced from the in-memory representation prior to example conversion.

PR #221, opened 2020-06-09
Do not fixup property refs in code blocks.

pgavlin:
Code blocks should not be reinterpreted as text. Changing the content of
a code block can alter its meaning (though this is rare).

This is part of #212.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Add JSON path links to properties when generating doc comments and deprecation messages for
+  a Pulumi schema.
+  [#202](https://github.com/pulumi/pulumi-terraform-bridge/pull/202)
+
 - Update README generation for python. 
   [#217](https://github.com/pulumi/pulumi-terraform-bridge/pull/217)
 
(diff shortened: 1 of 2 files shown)


praneetloke:
Hmm I guess this means we wouldn't convert property refs inside comments in the code blocks. See https://github.com/pulumi-bot/pulumi-aws/compare/298f98622..4c76dc95a#diff-97d9bece446a02a01e1f83d3e4d75032R117. I guess that's fine compared to accidentally changing a property ref that renders the whole code block incorrect?

Is there an example of the problem as it is today that prompted this change?

pgavlin:
> I guess that's fine compared to accidentally changing a property ref that renders the whole code block incorrect?

Right.

Here's a example that is no longer incorrect: https://github.com/pulumi-bot/pulumi-aws/compare/298f98622..4c76dc95a#diff-0daa3f4a80871a57ae6e2774ec93a7b4L55-L62

praneetloke:
The link https://github.com/pulumi-bot/pulumi-aws/compare/298f98622..4c76dc95a#diff-0daa3f4a80871a57ae6e2774ec93a7b4L55-L62 doesn't work for me. What is the filename? I'll look it up manually.

pgavlin:
sdk/nodejs/acm/certificate.ts

PR #224, opened 2020-06-11
Convert examples using an intermediate schema.

### LINKED ISSUE #212: Use the current schema when converting examples for schema.json.

pgavlin:
These changes convert examples using an intermediate, in-memory schema.
This removes the need to install a prior version of the resource plugin
before generation, and ensures that the generated examples use the
current version of the schema.

Fixes #212.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Convert examples using the current schema.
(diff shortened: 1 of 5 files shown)


stack72:
@pgavlin as this is in memory and we had issues before with GCP and memory issues due to the nested schemas, is there any way this may cause OOMs?

Issue #14, opened 2019-06-05
Implement DiffConfig/CheckConfig for tfbridged providers


ellismg:
In https://github.com/pulumi/pulumi/pull/2766 we added support to call DiffConfig/CheckConfig to detect cases where a provider's configuration has changed and we need to recreate the provider (and hence any resources it manages).

While we implemented `DiffConfig` and `CheckConfig` in the kubernetes provider, we have yet to implement it in any of our bridged providers.

We should do so.  We will probably need a way for providers to mark certain configuration properties as force new (for example, we would use this in our AWS provider if the `region` changes).

lukehoban:
Note - as part of this - we should either warn, or even force a replacement, if the AWS account being targeted by the provider changes. (And similar for other providers where this is a strong notion of account identity in the backing provider).

stack72:
I was able to see that DiffConfig has the following:

```
0.002s DiffConfig
       req=&pulumirpc.DiffRequest{
           Id:   "",
           Urn:  "urn:pulumi:dev::checks::pulumi:providers:aws::default_2_10_0_alpha_1592400278_ge84bbc90b",
           Olds: &structpb.Struct{
               Fields: {
                   "region": &structpb.Value{
                       Kind:                 &structpb.Value_StringValue{StringValue:"us-east-1"},
                       XXX_NoUnkeyedLiteral: struct {}{},
                       XXX_unrecognized:     nil,
                       XXX_sizecache:        0,
                   },
                   "version": &structpb.Value{
                       Kind:                 &structpb.Value_StringValue{StringValue:"2.10.0-alpha.1592400278+ge84bbc90b"},
                       XXX_NoUnkeyedLiteral: struct {}{},
                       XXX_unrecognized:     nil,
                       XXX_sizecache:        0,
                   },
               },
               XXX_NoUnkeyedLiteral: struct {}{},
               XXX_unrecognized:     nil,
               XXX_sizecache:        0,
           },
           News: &structpb.Struct{
               Fields: {
                   "region": &structpb.Value{
                       Kind:                 &structpb.Value_StringValue{StringValue:"us-east-2"},
                       XXX_NoUnkeyedLiteral: struct {}{},
                       XXX_unrecognized:     nil,
                       XXX_sizecache:        0,
                   },
                   "version": &structpb.Value{
                       Kind:                 &structpb.Value_StringValue{StringValue:"2.10.0-alpha.1592400278+ge84bbc90b"},
[...]

PR #225, opened 2020-06-19
Implement DiffConfig/CheckConfig for tfbridged providers

stack72:
Fixes:  #14

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -314,7 +314,74 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
 
 // DiffConfig diffs the configuration for this Terraform provider.
 func (p *Provider) DiffConfig(ctx context.Context, req *pulumirpc.DiffRequest) (*pulumirpc.DiffResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "DiffConfig is not yet implemented")
+	urn := resource.URN(req.GetUrn())
+	label := fmt.Sprintf("%s.DiffConfig(%s)", p.label(), urn)
+	glog.V(9).Infof("%s executing", label)
+
+	// There is no logic in the TF provider that suggests that provider level config
+	// should force a new provider. Therefore, we are going to do this based on our
+	// own schema overrides. We should use ForceNew as part of the SchemaInfo to do this
+
+	var olds resource.PropertyMap
+	var err error
+	if req.GetOlds() != nil {
+		olds, err = plugin.UnmarshalProperties(req.GetOlds(), plugin.MarshalOptions{
+			Label: fmt.Sprintf("%s.olds", label), KeepUnknowns: true})
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	news, err := plugin.UnmarshalProperties(req.GetNews(), plugin.MarshalOptions{
+		Label: fmt.Sprintf("%s.news", label), KeepUnknowns: true, SkipNulls: true})
+	if err != nil {
+		return nil, err
+	}
+
+	var diffs, replaces []string
+	for _, key := range olds.StableKeys() {
+		// check that the old key exists in the new keys - if not, then report as a diff
+		// check if the new key and the old key are difference
+		if news[key].IsNull() || news[key] != olds[key] {
+			// we need to find out what the schema value of this is and report it as expected
+			if p.info.Config != nil && p.info.Config[string(key)] != nil {
+				config := p.info.Config[string(key)]
+				// now is it a ForceNew or not
+				if config.ForceNew != nil && *config.ForceNew {
+					replaces = append(replaces, string(key))
+				} else {
+					diffs = append(diffs, string(key))
+				}
+			}
+		}
+	}
+
+	for _, key := range news.StableKeys() {
+		// Check to see if the new key is not present in the old keys
+		if olds[key].IsNull() {
+			if p.info.Config != nil && p.info.Config[string(key)] != nil {
+				config := p.info.Config[string(key)]
+				// now is it a ForceNew or not
+				if config.ForceNew != nil && *config.ForceNew {
+					replaces = append(replaces, string(key))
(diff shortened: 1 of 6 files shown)


pkg/tfbridge/provider.go
```diff
@@ -309,12 +307,148 @@ func (p *Provider) GetSchema(ctx context.Context,
...
+		Label: fmt.Sprintf("%s.news", label), KeepUnknowns: true, SkipNulls: true})
+	if err != nil {
+		return nil, err
+	}
+
+	var diffs, replaces []string
+	for _, key := range olds.StableKeys() {
+		// check that the old key exists in the new keys - if not, then report as a diff
+		// check if the new key and the old key are difference
+		if news[key].IsNull() || news[key] != olds[key] {
```
lukehoban, line 412: Is a reference equality check appropriate for all types of data that might be values of config variables?  For example, for arrays or nested structs? (I believe there are examples of both in AWS for example?)

pkg/tfbridge/provider.go
```diff
@@ -309,12 +307,148 @@ func (p *Provider) GetSchema(ctx context.Context,
...
+			return nil, err
+		}
+	}
+
+	if len(errs) > 0 {
+		return nil, errors.Wrap(multierror.Append(nil, errs...), "could not validate provider configuration")
+	}
+
+	return &pulumirpc.CheckResponse{Inputs: req.GetNews()}, nil
 }
```
lukehoban, line 380: @pgavlin do we have a protocol description of the provider RPC contract anywhere?  I *think* it is right to move this validation from Configure into CheckConfig - but not certain of the guarantees and requirements around timing of these two calls.

pkg/tfbridge/provider.go
```diff
@@ -309,12 +307,148 @@ func (p *Provider) GetSchema(ctx context.Context,
...
+				}
+			}
+		}
+	}
+
+	if len(diffs) > 0 || len(replaces) > 0 {
+		return &pulumirpc.DiffResponse{
+			Changes:  pulumirpc.DiffResponse_DIFF_SOME,
+			Diffs:    diffs,
+			Replaces: replaces,
```
lukehoban, line 445: Can/should we return a DetaildDiff here?

@pgavlin what is the current trade off in user experience if we don’t return a DetailedDiff?

pkg/tfbridge/provider.go
```diff
@@ -40,6 +36,8 @@ import (
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/rpcutil/rpcerror"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v2/proto/go"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc/codes"
```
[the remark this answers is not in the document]
stack72: It was not :) I will move them back!

pkg/tfbridge/provider.go
stack72: I am more than willing to change this to use reflect.DeepEqual if you think that's the better option here?

lukehoban: I’m somewhat nervous that that is not “right” either. Can we ask TF to do this diff for us?  That’s where the actual semantics of diffing these is defined, right?  And aligns with how we do non-config diffs in this provider?  Does the fact that we need to override ForceNew prevent us from doing that?

pkg/tfbridge/provider.go
pgavlin: I don't think that this code can move wholesale. If this check is never performed in `Configure`, older versions of the CLI that do not call `CheckConfig` will no longer get any validation of provider config. We should figure out some approach that allows us to perform validation in `Configure` if the inputs were not validated by `CheckConfig`. I think we can achieve this by having `Configure` only perform validation if the [`args` field of `ConfigureRequest`](https://github.com/pulumi/pulumi/blob/master/sdk/proto/provider.proto#L78) is empty.

pkg/tfbridge/provider.go
pgavlin: > Can we ask TF to do this diff for us? That’s where the actual semantics of diffing these is defined, right?

If we can do this, it does feel like the right approach. There's a chance that it's not straightforward--we need to build a state and a config to use as inputs to the diff--but I do think it would work. We should be able to override `ForceNew` by post-processing the diff and changing any updates/adds/deletes on force new properties to the corresponding replace. If we can get a TF diff, I _think_ we should also be able to return detailed diffs using the existing code that translates TF diffs to detailed diffs.

pkg/tfbridge/provider.go
pgavlin: The diff will be recomputed by the CLI and may show false diffs or omit actual diffs.

pkg/tfbridge/provider.go
```diff
@@ -375,43 +536,17 @@ func (p *Provider) Configure(ctx context.Context,
...
-			rpcerror.New(codes.InvalidArgument, "required configuration keys were missing"),
-			&pulumirpc.ConfigureErrorMissingKeys{MissingKeys: missingKeys})
-		return nil, err
-	}
-
-	// Perform validation of the config state so we can offer nice errors.
-	warns, errs := p.tf.Validate(config)
-	for _, warn := range warns {
-		if err = p.host.Log(ctx, diag.Warning, "", fmt.Sprintf("provider config warning: %v", warn)); err != nil {
+	if req.Variables == nil {
```
pgavlin, line 539: Can we factor out any of the code above? It looks like at least some of it can be shared between `CheckConfig`/`Configure`, and I think the `PreConfigureCallback` should only be called once.

stack72: Ok, I've refactored as much as I can here. I also moved the `PreConfigureCallback` to be inside the `if req.Varariables == nil` in the `Configure` func to ensure that we only call it if we are hitting the old (non CheckConfig) path via the CLI

PR #227, opened 2020-06-25
Do not drop all examples if one lang fails.

pgavlin:
When converting an HCL snippet to multiple languages for inclusion in a
Pulumi schema, only treat the example as skipped if _all_ languages fail
to convert.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -906,7 +906,9 @@ func (g *generator) convertHCL(hcl string) (string, string, error) {
 	convertHCL := func(languageName string) (err error) {
 		defer func() {
 			v := recover()
-			contract.Ignore(v)
+			if v != nil {
+				err = fmt.Errorf("panic converting HCL to %v: %v", languageName, v)
+			}
 		}()
 
 		files, diags, err := convert.Convert(convert.Options{
@@ -935,7 +937,9 @@ func (g *generator) convertHCL(hcl string) (string, string, error) {
 			err = diags.NewDiagnosticWriter(&stderr, 0, false).WriteDiagnostics(diags.All)
 			contract.IgnoreError(err)
 
-			return fmt.Errorf("failed to convert HCL to %v", languageName)
+			// Note that we intentionally avoid returning an error here. The caller will check for an empty code block
+			// before returning and translate that into an error.
+			return nil
 		}
 
 		contract.Assert(len(files) == 1)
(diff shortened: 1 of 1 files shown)


pkg/tfgen/docs.go
```diff
@@ -906,7 +905,9 @@ func (g *generator) convertHCL(hcl string) (string, string, error) {
 	convertHCL := func(languageName string) (err error) {
 		defer func() {
 			v := recover()
-			contract.Ignore(v)
+			if v != nil {
+				err = fmt.Errorf("panic converting HCL to %v: %v", languageName, v)
```
pgavlin, line 909: It looks like we're losing some Go examples due to this. It appears to be picking up failures from gofmt, e.g.

```
invalid Go source code:

package main

import (
	"github.com/pulumi/pulumi/sdk/v2/go/pulumi"
	"github.com/pulumi/pulumi-aws/sdk/v2/go/aws/cloudwatch"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		elasticsearch-log-publishing-policyPolicyDocument, err := iam.LookupPolicyDocument(ctx, &iam.LookupPolicyDocumentArgs{
			Statements: iam.getPolicyDocumentStatementArray{
				&iam.LookupPolicyDocumentStatement{
					Actions: []string{
						"logs:CreateLogStream",
						"logs:PutLogEvents",
						"logs:PutLogEventsBatch",
					},
					Principals: iam.getPolicyDocumentStatementPrincipalArray{
						&iam.LookupPolicyDocumentStatementPrincipal{
							Identifiers: []string{
								"es.amazonaws.com",
							},
							Type: "Service",
						},
					},
					Resources: []string{
						"arn:aws:logs:*",
					},
				},
			},
		}, nil);
		if err != nil {
			return err
		}
		elasticsearch-log-publishing-policyLogResourcePolicy, err := cloudwatch.NewLogResourcePolicy(ctx, "elasticsearch-log-publishing-policyLogResourcePolicy", &cloudwatch.LogResourcePolicyArgs{
			PolicyDocument: pulumi.String(elasticsearch-log-publishing-policyPolicyDocument.Json),
			PolicyName: pulumi.String("elasticsearch-log-publishing-policy"),
		})
		if err != nil {
			return err
		}
		return nil
	})
}
```

In a sense, this is good: this is clearly bad codegen. However, it does make the improvements less clear. Manual inspection confirms that new examples are being picked up in the AWS schema.

PR #233, opened 2020-07-02
Ensure that Python overlays write to the correct directory

stack72:


pkg/tfgen/generate.go
```diff
@@ -1290,6 +1289,15 @@ func getOverlayFilesImpl(overlay *tfbridge.OverlayInfo, extension, srcRoot, dir
 			if err != nil {
 				return err
 			}
+			// If we are in Python (and potentially Go) then we may need to strip the leading
+			// folder extension from the fp. Otherwise, when we write the overlay back
+			// it will write to a double nested structure
+			// eg. sdk/python/pulumi_provider/pulumi_provider/file.py
+			// We need to do this *after* we read then file so that we can assemble the package
```
komalali, line 1296:
```suggestion
			// We need to do this *after* we read the file so that we can assemble the package
```

PR #236, opened 2020-07-15
Clean codegen dir before regenerating

komalali:
Fixes: https://github.com/pulumi/home/issues/976

Didn't add Go since there is no handling of overlays for the lang yet.

Bumped python version because github actions removed 3.6.10


pkg/tfgen/generate.go
```diff
@@ -95,6 +96,19 @@ func (l language) emitSDK(pkg *pschema.Package, info tfbridge.ProviderInfo, outD
 				return nil, err
 			}
 		}
+
+		// We exclude the "tests" directory because some nodejs package dirs (e.g. pulumi-docker)
+		// store tests here. We don't want to include them in the overlays because we don't want it
+		// exported with the module, but we don't want them deleted in a cleanup of the directory.
+		exclusions := codegen.StringSet{}
+		exclusions.Add("tests")
```
pgavlin, line 104: You can use the `NewStringSet` factory function here:

```suggestion
		exclusions := codegen.NewStringSet("tests")
```

PR #241, opened 2020-08-19
Provide a way to set UsesIOClasses for TF-based providers

justinvp:
The `UsesIOClasses` option controls whether the generated Python examples and docgen should use input/output classes. This adds the option to the `PythonInfo` struct so it can be set by TF-based providers.

Here's an example of what it'd look like to update a TF-provider, using this option: https://github.com/pulumi/pulumi-aws/commit/e548fd9822f25690c4e37958de268f9c2689bcc3 (the commit does not include the necessary go.mod changes to reference this change in `pulumi-terraform-bridge` and the changes needed in `pulumi`).

Note: the alternative would be to just always set `pythonData["usesIOClasses"] = true` here in the bridge, and not expose the `UsesIOClasses` option on the struct, and a provider would opt-in implicitly when updating to a more recent `pulumi-terraform-bridge` commit. Any preference?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Add UsesIOClasses field to PythonInfo.
+  [#241](https://github.com/pulumi/pulumi-terraform-bridge/pull/241)
+
 - Add PluginDownloadURL field to package definition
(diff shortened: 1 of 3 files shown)


stack72 (approved):

stack72:
The build of the downstream providers will fail until this is rolled out so merging

PR #242, opened 2020-08-19
Temporarily revert the diff and config check due to nested object bug

stack72:
This feature was not yet in use anywhere so reverting this means we don't lose any functionality. It has a bug which is causing https://github.com/pulumi/pulumi-aws/issues/1075


stack72:
downstream codegen will fail due to new Python changes that are not yet rolled out

PR #246, opened 2020-08-25
Allow partial code conversions

komalali:
Related: #227

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -914,7 +914,8 @@ func (g *generator) convertHCL(hcl string) (string, string, error) {
 		defer func() {
 			v := recover()
 			if v != nil {
-				err = fmt.Errorf("panic converting HCL to %v: %v", languageName, v)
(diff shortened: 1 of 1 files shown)


pkg/tfgen/docs.go
```diff
@@ -914,7 +914,8 @@ func (g *generator) convertHCL(hcl string) (string, string, error) {
 		defer func() {
 			v := recover()
 			if v != nil {
-				err = fmt.Errorf("panic converting HCL to %v: %v", languageName, v)
+				cmdutil.Diag().Warningf(diag.Message("", fmt.Sprintf("failed to convert HCL example to %v", languageName)))
```
komalali, line 917: Is the warning needed or should this just be a debug? Also, how can I see the debug messages? I can't figure out what flag to use to enable them.

pgavlin: > Is the warning needed or should this just be a debug? 

I do think that this should be a warning, as it's something that requires action on our end.

> Also, how can I see the debug messages? I can't figure out what flag to use to enable them.

Honestly, I'm not certain myself.

PR #247, opened 2020-08-25
Replace tf2pulumi binary with pulumictl binary

stack72:
tf2pulumi is no longer required but pulumictl is


komalali:
No github token found?

PR #252, opened 2020-09-09
Refactor state/config unmarshaling.

pgavlin:
- Replace `MakeTerraformAttributes` with `MakeTerraformState`. This function
  now creates and returns an instance state instead of attributes and
  meta info.
- Rename `Make*FromRPC` to `Unmarshal*`, which better indicates what these
  functions do.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -69,7 +69,7 @@ type Resource struct {
 // resource ID, and returns a replacement input map if any resources are matched. A nil map
 // with no error should be interpreted by the caller as meaning the resource does not exist,
 // but there were no errors in determining this.
-func (res *Resource) runTerraformImporter(id resource.ID, provider *Provider) (resource.ID, map[string]string, error) {
+func (res *Resource) runTerraformImporter(id string, provider *Provider) (string, map[string]string, error) {
 	contract.Assert(res.TF.Importer != nil)
 
 	// Prepare a Terraform ResourceData for the importer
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/provider.go
```diff
@@ -674,16 +674,15 @@ func (p *Provider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*pulum
 
 	// To figure out if we have a replacement, perform the diff and then look for RequiresNew flags.
 	olds, err := plugin.UnmarshalProperties(req.GetOlds(),
-		plugin.MarshalOptions{Label: fmt.Sprintf("%s.olds", label), KeepUnknowns: false})
```
stack72, line 677: Any specific reason we not SkipNulls and no longer KeepUnknowns?

pkg/tfbridge/provider.go
```diff
@@ -778,7 +777,7 @@ func (p *Provider) Create(ctx context.Context, req *pulumirpc.CreateRequest) (*p
 	info := &terraform.InstanceInfo{Type: res.TFName}
 	state := &terraform.InstanceState{Meta: make(map[string]interface{})}
 	assets := make(AssetTable)
-	config, err := MakeTerraformConfigFromRPC(
+	config, err := UnmarshalTerraformConfig(
```
stack72, line 780: This feels much clearer now with the func name - thanks!!

stack72 (approved):
I had just one question but this is looking much clearer to the intent of what the funcs were doing - thank you!

stack72:
We just need to rebase this PR - it would be awesome to see if any code diffs appear

komalali (approved):

pkg/tfbridge/provider.go
pgavlin: The only addition here is `SkipNulls: true`; `KeepUnkonwns` is implicitly set to `false`. The addition of `SkipNulls` brings this call into alignment with other call sites that use the result to build TF state.

Issue #253, opened 2020-09-09
Provider does not implement pulumirpc.ResourceProviderServer (missing Construct method)


justinvp:
Downstream Codgen Tests are failing in pulumi/pulumi due to missing Construct method on Provider, which was added in https://github.com/pulumi/pulumi/pull/5280

```
2020-09-09T07:27:09.9277676Z (cd provider && go build -a -o /home/runner/work/pulumi/pulumi-aws/bin/pulumi-tfgen-aws -ldflags "-X github.com/pulumi/pulumi-aws/provider/v3/pkg/version.Version=3.3.0-alpha.1599636420+69ed3223" github.com/pulumi/pulumi-aws/provider/v3/cmd/pulumi-tfgen-aws)
2020-09-09T07:28:05.1599361Z # github.com/pulumi/pulumi-terraform-bridge/v2/pkg/tfbridge
2020-09-09T07:28:05.1616247Z ##[error]../../../../go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v2@v2.7.3/pkg/tfbridge/provider.go:178:5: cannot use (*Provider)(nil) (type *Provider) as type pulumirpc.ResourceProviderServer in assignment:
2020-09-09T07:28:05.1618500Z 	*Provider does not implement pulumirpc.ResourceProviderServer (missing Construct method)
2020-09-09T07:28:05.1620599Z ##[error]../../../../go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v2@v2.7.3/pkg/tfbridge/serve.go:30:21: cannot use NewProvider("context".TODO(), host, module, version, info.P, info, pulumiSchema) (type *Provider) as type pulumirpc.ResourceProviderServer in return argument:
2020-09-09T07:28:05.1621654Z 	*Provider does not implement pulumirpc.ResourceProviderServer (missing Construct method)
```

We'll need to add it here.

PR #255, opened 2020-09-09
Add Construct function to the provider

justinvp:
Adopt the latest commit from pulumi/pulumi master and add a `Construct` function to the provider (which just returns an error).

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/253

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -17,10 +17,11 @@ package tfbridge
 import (
 	"encoding/json"
 	"fmt"
-	"google.golang.org/grpc/status"
 	"log"
 	"regexp"
 	"strings"
+
+	"google.golang.org/grpc/status"
 
 	"github.com/golang/glog"
 	pbempty "github.com/golang/protobuf/ptypes/empty"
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/provider.go
```diff
@@ -1048,6 +1049,11 @@ func (p *Provider) Delete(ctx context.Context, req *pulumirpc.DeleteRequest) (*p
 	return &pbempty.Empty{}, nil
 }
 
+// Construct creates a new instance of the provided component resource and returns its state.
+func (p *Provider) Construct(context.Context, *pulumirpc.ConstructRequest) (*pulumirpc.ConstructResponse, error) {
+	return &pulumirpc.ConstructResponse{}, errors.New("resources may not be constructed")
```
stack72, line 1054: I think you should do something like:

```
return nil, status.Error(codes.Unimplemented, "Construct is not yet implemented")
```

rather than throw an error - thoughts?

justinvp: Ah, yes.

lblackstone (approved):

pgavlin (approved):

stack72:
:shipit:

Issue #240, opened 2020-08-17
Support Alternative Documentation Paths


stack72:
There are 2 problems that need addressed here:

1. Modules that do not follow the standard TF module structure and store docs in a location other `website/docs/{r,d}/resource.md`. 
	This alternative download path may require us to scrape the docs directly from GitHub rather than using the GOPATH locally

2. There are situations where a provider is moving from org X to org Y and the go.mod still shows the module name referencing org X. 
This leads us to not be able to load documentation for that specific module.

komalali:
I'm going to put this back in the input queue for now because it's not the actual culprit for the bug we're seeing in pulumi-random, but I did spend a little bit of time looking into this so adding that information here.

We already have a [solution](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfgen/docs.go#L128-L137) for (1) where we look at different possibilities for the file name. If there are examples where this isn't working, we can look into improving this logic.

I don't believe (2) is an actual problem. It doesn't matter if the repo has changed orgs, the module files will always follow the module name that is specified in go.mod.

PR #256, opened 2020-09-10
Allow for new docs structure

komalali:
Resolves: https://github.com/pulumi/pulumi-terraform-bridge/issues/240

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -87,9 +87,9 @@ type DocKind string
 
 const (
 	// ResourceDocs indicates documentation pertaining to resource entities.
-	ResourceDocs DocKind = "r"
+	ResourceDocs DocKind = "resources"
 	// DataSourceDocs indicates documentation pertaining to data source entities.
-	DataSourceDocs DocKind = "d"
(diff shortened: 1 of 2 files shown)


pkg/tfgen/docs.go
```diff
@@ -198,10 +200,36 @@ func getDocsForProvider(g *generator, org string, provider string, resourcePrefi
...
+	return path.Join(repo, "docs", kindString)
+}
+
 // readMarkdown searches all possible locations for the markdown content
 func readMarkdown(repo string, kind DocKind, possibleLocations []string) ([]byte, string, bool) {
+	locationPrefix := getDocsPath(repo, kind)
+
 	for _, name := range possibleLocations {
-		location := path.Join(repo, "website", "docs", string(kind), name)
+		location := path.Join(locationPrefix, name)
```
justinvp, line 232: Nit: I realize this is an existing use of `path.Join` and we run this code always on *nix so it doesn't really matter, but it'd be more correct to use `filepath.Join` when joining file paths.

justinvp (approved):
LGTM

stack72 (approved):

PR #258, opened 2020-09-10
Check if docs/resources exists instead of only docs

komalali:
The last PR https://github.com/pulumi/pulumi-terraform-bridge/pull/256 broke the AWS docs.

The AWS provider does have a docs folder at the root but it's not the true docs folder. Check if docs/resources exists instead.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -203,8 +203,8 @@ func getDocsForProvider(g *generator, org string, provider string, resourcePrefi
 // checkIfNewDocsExist checks if the new docs root exists
 func checkIfNewDocsExist(repo string) bool {
 	// Check if the new docs path exists
-	newDocsRoot := filepath.Join(repo, "docs")
-	_, err := os.Stat(newDocsRoot)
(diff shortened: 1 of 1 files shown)


pkg/tfgen/docs.go
```diff
@@ -203,8 +203,8 @@ func getDocsForProvider(g *generator, org string, provider string, resourcePrefi
 // checkIfNewDocsExist checks if the new docs root exists
 func checkIfNewDocsExist(repo string) bool {
 	// Check if the new docs path exists
-	newDocsRoot := filepath.Join(repo, "docs")
-	_, err := os.Stat(newDocsRoot)
+	newDocsPath := filepath.Join(repo, "docs", "resources")
```
komalali, line 206: I am assuming all TF providers have resources. If this isn't a safe assumption we can check if either `docs/resources` or `docs/data-sources` exist

stack72: all have resources! :)

justinvp (approved):

stack72 (approved):
:shipit:

PR #263, opened 2020-09-11
Move tf2pulumi into pulumi-terraform-bridge.

pgavlin:
Since the change to link tf2pulumi into tfgen, tf2pulumi and
pulumi-terraform-bridge have a nasty circular dependency. This circular
dependency prevents any changes to the `ProviderInfo` type in
particular, which makes it impossible to roll out the shim changes
necessary to support the v2 TF plugin SDK. These changes eliminate the
circularity by pulling tf2pulumi into pulumi-terraform-bridge.


stack72:
Does this kill off the tf2pulumi repo or will we just need to copy the files between them?

lukehoban (approved):
> These changes eliminate the
circularity by pulling tf2pulumi into pulumi-terraform-bridge.

Technically it moves the bulk of the tf2pulumi implementation into the bridge - but the actual `tf2pulumi` command still lives in the tf2pulumi repo - right?  Also - I assume we will follow up by removing the code from that repo so we don't have this all duplicated.

pgavlin:
> Technically it moves the bulk of the tf2pulumi implementation into the bridge - but the actual tf2pulumi command still lives in the tf2pulumi repo - right? Also - I assume we will follow up by removing the code from that repo so we don't have this all duplicated.

Correct on all counts.

PR #264, opened 2020-09-11
Consume the TF SDK shim.

### LINKED ISSUE #245: terraform-plugin-sdk/v2
The V2 of the terraform-plugin-sdk needs to be supported so that we can upgrade to v3.2.0 of the AWS Terraform Provider

There are some breaking changes in this version of the SDK so we need to run this as a branch alongside master for now

pgavlin:
These changes update tfbridge, tfgen, and tf2pulumi to consume the
Terraform SDK shim. Most of the changes are mechanical: field accesses
are replace with method calls, ranges are replaced with calls to Range,
etc.

Consumers of the bridge must now wrap their provider's schema in the
appropriate shim implementation prior to handing it off to
tfbridge.Main. Providers that use the v1 SDK should use the v1 shim;
those that use the v2 SDK should use the v2 shim.

Fixes #245.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Support terraform-plugin-sdk/v2. This is an API breaking change.
+  [#264](https://github.com/pulumi/pulumi-terraform-bridge/pull/264)
+
 - Support alternative documentation path.
   [#256](https://github.com/pulumi/pulumi-terraform-bridge/pull/256)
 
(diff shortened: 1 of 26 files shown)


pgavlin:
The API breaking nature of these changes is going to cause all of the downstream testing jobs to fail. I've manually validated these changes with `pulumi-random` and `pulumi-aws`; I'll cover the other big two as well.

stack72 (approved):

stack72:
@pgavlin this looks good - it's actually simplifying things in a few areas which is great!

Thanks!

justinvp (approved):

PR #267, opened 2020-09-16
Add a gRPC-backed shim implementation.

pgavlin:
This implementation of the shim uses the TF plugin gRPC interface to
implement provider operations. The gRPC interface exposes its own schema
system that is independent of the TF plugin SDK and communicates
configuration and state using `msgpack`-encoded `cty` values. The bulk
of the work is in translating the gRPC schema to the shim's schema,
encoding and decoding cty values to and from Go values, and computing
flatmap-style diffs.


pgavlin:
Gentle ping.

.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           GOPROXY: "https://proxy.golang.org"
         run: go mod download
       - name: Build
-        run: go build ./...
+        run: go build ./... && go install ./cmd/...
```
stack72, line 25: why are you needing to go install here?

stack72 (approved):

.github/workflows/ci.yml
pgavlin: These changes add a provider binary that needs to be on the `PATH` when running tests; this is the easiest way to achieve that.

pgavlin: Somewhat related: why does this repo no longer use the `Makefile`? IMO it's bad practice for CI and local development workflows to diverge at their core.

stack72: That's a pure error - I will fix that up with our CI scripts

pgavlin: Cool. Thanks!

PR #268, opened 2020-09-16
Expose tfgen APIs and use a virtual filesystem.

pgavlin:
These changes will allow for the creation of new tools that invoke the
generator via its API and that do not write to the filesystem.


justinvp (approved):
LGTM

PR #274, opened 2020-09-21
Add support for literal Markdown in doc info.

pgavlin:
Just what it says on the tin. This allows callers to implement their own
paths for finding documentation.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -119,6 +119,14 @@ func getRepoPath(org string, provider string, providerModuleVersion string) (str
 func getMarkdownDetails(g *Generator, org string, provider string, resourcePrefix string, kind DocKind,
 	rawname string, info tfbridge.ResourceOrDataSourceInfo, providerModuleVersion string) ([]byte, string, bool) {
 
+	var docinfo *tfbridge.DocInfo
+	if info != nil {
+		docinfo = info.GetDocs()
+	}
+	if docinfo != nil && len(docinfo.Markdown) != 0 {
+		return docinfo.Markdown, "", true
+	}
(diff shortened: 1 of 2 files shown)


pkg/tfgen/docs.go
```diff
@@ -299,7 +302,8 @@ var (
 
 	attributeBulletRegexp = regexp.MustCompile("^\\s*[*+-]\\s+`([a-zA-z0-9_]*)`\\s+[–-]?\\s+(.*)")
 
-	docsBaseURL    = "https://github.com/%s/terraform-provider-%s/blob/master/website/docs"
+	providerURL    = "https://github.com/%s/terraform-provider-%s"
+	docsBaseURL    = providerURL + "/blob/master/website/docs"
```
komalali, line 306: The docs base URL might vary. This is the older pattern, while [some providers](https://github.com/rancher/terraform-provider-rancher2/tree/master/docs) are now following a `/blob/master/docs` pattern.

pgavlin: That's a good point--am I remembering right that you're working on supporting those other patterns? Either way, I think that we can add support for those patterns outside of this PR.

komalali: Yeah I can clean up this bit. Up until now, these URLs weren't being used anymore so I had a [PR](https://github.com/pulumi/pulumi-terraform-bridge/pull/260) out to remove them. Still unclear if we actually use them anywhere, but I can either fix them or remove them depending on what the preference is.

pgavlin: Sounds good to me.

komalali (approved):

Issue #251, opened 2020-09-02
AssertionError: Unexpected type. Expected 'list' got '<class 'pulumi_datadog.outputs.MonitorThresholds'>'


justinvp:
The following Python program results in an error during preview:

CLI version: 2.9.2

```
$ venv/bin/python -m pip list
Package        Version
-------------- -------
Arpeggio       1.9.2
attrs          20.1.0
dill           0.3.2
grpcio         1.31.0
parver         0.3.0
pip            20.2.2
protobuf       3.13.0
pulumi         2.9.2
pulumi-datadog 2.7.0
semver         2.10.2
setuptools     50.0.3
six            1.15.0
wheel          0.35.1
```

```python
import pulumi_datadog as datadog

foo = datadog.Monitor("foo",
    name="Name for monitor foo",
    type="metric alert",
    message="Monitor triggered.",
    escalation_message="Escalation message",
    query="avg(last_1h):avg:aws.ec2.cpu{environment:foo,host:foo} by {host} > 4",
    thresholds=datadog.MonitorThresholdsArgs(
        ok=0,
        warning=2,
        warning_recovery=1,
        critical=4,
        critical_recovery=3,
    ),
    notify_no_data=False,
    renotify_interval=60,
    notify_audit=False,
    timeout_h=60,
)
```

```
Previewing update (dev)

View Live: <redacted>

     Type                      Name                Plan       Info
 +   pulumi:pulumi:Stack       datadogmonitor-dev  create     3 errors
 +   └─ datadog:index:Monitor  foo                 create     
 
Diagnostics:
  pulumi:pulumi:Stack (datadogmonitor-dev):
    error: Program failed with an unhandled exception:
    error: Traceback (most recent call last):
      File "/opt/pulumi/bin/pulumi-language-python-exec", line 85, in <module>
        loop.run_until_complete(coro)
      File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
        return future.result()
      File "/Users/user/Desktop/dd/datadogmonitor/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 83, in run_in_stack
        await run_pulumi_func(lambda: Stack(func))
      File "/Users/user/Desktop/dd/datadogmonitor/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 51, in run_pulumi_func
[...]

PR #276, opened 2020-09-22
Support provider-side preview.

pgavlin:
Read the proposed state out of the diff.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -549,7 +549,9 @@ func (p *Provider) Configure(ctx context.Context,
 		return nil, err
 	}
 
-	return &pulumirpc.ConfigureResponse{}, nil
+	return &pulumirpc.ConfigureResponse{
+		SupportsPreview: true,
+	}, nil
 }
 
 // Parse the TF error of a missing field:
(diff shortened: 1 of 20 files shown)


pgavlin:
These changes also include a local implementation of the field reader used to read values out of the Terraform diff.

The field reader implementations in the plugin SDK do not properly handle nested unknown values, instead treating such values as empty. The local implementation of the field reader correctly leaves nested unknown values as unknown.

A local implementation was chosen instead of an implementation in the plugin SDK fork so that it could be easily scoped to reading values from an `*InstanceDiff`, which only happens during previews. Naively changing the upstream field reader would affect all field reads, including those that happen during updates. This seems unacceptably risky. As an alternative, we could add a new implementation of the field reader in our fork, which may simplify any licensing issues.

pgavlin:
I have verified that this fixes https://github.com/pulumi/pulumi-azure/issues/192.

pkg/tfbridge/provider.go
```diff
@@ -954,23 +964,30 @@ func (p *Provider) Update(ctx context.Context, req *pulumirpc.UpdateRequest) (*p
...
+		}
+		if newstate.ID() == "" {
+			return nil, fmt.Errorf("expected non-empty ID for new state during Update of %s", urn)
+		}
 		if err != nil {
-			return nil, err
+			reasons = append(reasons, errors.Wrapf(err, "updating %s", urn).Error())
+		}
+	} else {
+		newstate, err = diff.ProposedState(res.TF, nil)
```
leezen, line 987: I haven't spent a lot of time in this code, but given the way this is structured, I would've expected to pass in the resource's prior state here as opposed to `nil` for previewing an update?

leezen (commented):
Any chance you had an opportunity to validate that this fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/251?

pkg/tfbridge/provider.go
pgavlin: Indeed, the prior state should be passed in. Still working on tests for this PR, clearly :)

leezen: Sure thing. Ping me when you're ready for another pass.

Issue #279, opened 2020-09-28
`required configuration keys were missing`


stack72:
Since migrating to the new shim functionality, we are getting an error when trying to validate the configuration for a provider

```
azuread:index:User me  error: required configuration keys were missing
```

it feels like the provider isn't respecting Environment variables for configuration anymore

PR #280, opened 2020-09-29
Ensure we applyDefaults when making TerraformConfig

stack72:
Fixes: #279

Without this parameter being set, we will never apply Defaults from our schema - this means we will always get errors that suggest `missing configuration keys`

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -351,10 +351,12 @@ func buildTerraformConfig(p *Provider, vars resource.PropertyMap) (shim.Resource
 		}
 	}
 
-	// Make a Terraform config map out of the variables. We do this before checking for missing properties
-	// s.t. we can pull any defaults out of the TF schema.
-	config, _, err := MakeTerraformConfig(p, tfVars, p.config, p.info.Config)
-	return config, err
+	inputs, _, err := MakeTerraformInputs(nil, tfVars, nil, nil, p.config, p.info.Config)
+	if err != nil {
+		return nil, err
+	}
+
+	return MakeTerraformConfigFromInputs(p.tf, inputs), nil
 }
 
 func validateProviderConfig(ctx context.Context, p *Provider, config shim.ResourceConfig) (


pgavlin:
I don't think that this is the right fix. `MakeTerraformConfig` expects defaults to have already been applied. I think that we're only missing that case when dealing with provider configuration (indeed, `MakeTerraformInputs` sets `ApplyDefaults` to true). Can we scope these changes to the provider config case?

stack72:
> I don't think that this is the right fix. `MakeTerraformConfig` expects defaults to have already been applied. I think that we're only missing that case when dealing with provider configuration (indeed, `MakeTerraformInputs` sets `ApplyDefaults` to true). Can we scope these changes to the provider config case?

The only way we can do this is to pass through a variable to suggest we need to apply the defaults if this variable is true, then we add it to the `conversionContext` - otherwise we just default as normal

thoughts?

pgavlin:
> The only way we can do this is to pass through a variable to suggest we need to apply the defaults if this variable is true, then we add it to the conversionContext - otherwise we just default as normal

We can call `MakeTerraformInputs` and `MakeTerraformConfigFromInputs` in tandem instead of calling `MakeTerraformConfig`.

stack72:
> > The only way we can do this is to pass through a variable to suggest we need to apply the defaults if this variable is true, then we add it to the conversionContext - otherwise we just default as normal
> 
> We can call `MakeTerraformInputs` and `MakeTerraformConfigFromInputs` in tandem instead of calling `MakeTerraformConfig`.

👌 PR update on way!

pkg/tfbridge/provider.go
```diff
@@ -351,10 +351,18 @@ func buildTerraformConfig(p *Provider, vars resource.PropertyMap) (shim.Resource
...
-	// s.t. we can pull any defaults out of the TF schema.
-	config, _, err := MakeTerraformConfig(p, tfVars, p.config, p.info.Config)
-	return config, err
+	ctx := conversionContext{
+		ProviderConfig: p.configValues,
+		Assets:         AssetTable{},
+		ApplyDefaults:  true,
+	}
+
+	inputs, err := ctx.MakeTerraformInputs(nil, tfVars, p.config, p.info.Config, false)
```
pgavlin, line 360: You should be able to just call the top-level `MakeTerraformInputs` function, which will take care of creating the `conversionContext`.

stack72: Nice! updated and tested that it works as expected with the Keycloak and AzureAD providers (where I had issues with this)

pgavlin (approved):

stack72:
Thank you @pgavlin!!!

PR #283, opened 2020-10-01
Fix nil pointer panic

komalali:
`opts.ProviderInfoSource` in [this change](https://github.com/pulumi/pulumi-terraform-bridge/pull/273/files#diff-1e9fb3ec2c3be8da026575d86e7b5694R655) can be nil, leading to a panic.


stack72 (approved):
Great fine!! Thank you

Issue #282, opened 2020-09-30
error: Preview failed: required configuration keys were missing


stack72:
We know what the missing keys are - we should ensure they are added to the output

the message should read something like:

```
The following required configuration keys were missing:

- x
- y
- z
```

PR #284, opened 2020-10-08
Supply user error about which config values were missing

stack72:
Fixes: #282

A regression removed this and it meant that when a user didn't
supply the correct config values, that they got the following error:

```
  aws:s3:Bucket (my-bucket):
    error: required configuration keys were missing
```

That had no context. Fixing the regression now shows this error:

```
  aws:s3:Bucket (my-bucket):
    error: 1 error occurred:
    	* missing required configuration key "aws:region": The region where AWS operations will take place. Examples are us-east-1, us-west-2, etc.
    Set a value using the command `pulumi config set aws:region <value>`.
```


mikhailshilkov (approved):
Looks reasonable

PR #285, opened 2020-10-08
Fix a bug building TF config.

pgavlin:
Recent changes to the way we build Terraform configuration
unintentionally dropped the user input and only applied defaults. These
changes ensure that the user's inputs are taken into account.


stack72 (approved):
<3

Issue #289, opened 2020-10-19
Provider a way to fail build when unmapped resource is found


stack72:
AS this will be our first foray into this territory, this needs to be opt-in so we can do it with an environment variable

`PULUMI_FAIL_BUILD_UNMAPPED_RESOURCE`

PR #290, opened 2020-10-19
Add the ability to fail the build on missing resource mapping

stack72:
Fixes: #289

As we start to move toward a more automated creation of the providers,
we want to ensure that missing resources and data sources fail the build
so that we don't release a provider without them

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -676,6 +676,10 @@ func NewGenerator(opts GeneratorOptions) (*Generator, error) {
 	}, nil
 }
 
+func (g *Generator) error(f string, args ...interface{}) {
+	g.sink.Errorf(diag.Message("", f), args...)
+}
+
 func (g *Generator) warn(f string, args ...interface{}) {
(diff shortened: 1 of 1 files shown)


pkg/tfgen/generate.go
```diff
@@ -879,14 +883,19 @@ func (g *Generator) gatherResources() (moduleMap, error) {
 	}
 	modules := make(moduleMap)
 
+	_, failBuildOnProviderMapError := os.LookupEnv("PULUMI_PROVIDER_MAP_ERROR")
```
komalali, line 886: Can we call this once and add it as a flag to `Generator`? Or even just in `gatherPackage()` and pass it as an arg to these two functions?

komalali (approved):
Some suggestions, but LGTM otherwise.

pkg/tfgen/generate.go
stack72: This is purely a test - I think we will actually generate the resource if it doesn't exist going forward but I am trying to get an idea (when I build locally) on how many (if any) are missing

PR #291, opened 2020-11-02
Addition of Import docs for specific resource types

stack72:
We add this to the Description of the resource spec because we
want to ensure this is added to the schema AND also the sdk specific
doc comments

This means we get it in code files as well as being able to parse it
for the resource docs


komalali (commented):
@stack72 it looks like the downstream tests are not generating the import section correctly. It looks fine in the schema.json but not in the resource files themselves (the actual examples are missing).
[`b52bfe7..932a787`#diff-c58610af37](https://github.com/pulumi-bot/pulumi-random/compare/b52bfe7..932a787#diff-c58610af375cd0fb8a6156c5e60a034be94260073c7620776457186af0802b42R45-R50)

jaxxstorm (approved):

komalali (approved):
Just a couple of nitty comments, overall LGTM!

PR #294, opened 2020-11-09
Housekeeping regarding CI scripts

stack72:


komalali (changes requested):
Looks like there's something not quite working right. Seems to be looking for a `pulumi` directory in `pulumi-terraform-bridge`.

komalali (approved):

PR #295, opened 2020-11-12
Fix edge case in diffing.

komalali:
We got into this situation because the first item in the array is the one that is a `DELETE_REPLACE` gets overwritten when we cycle through the config diffs. I added a check that if the other diff (i.e. the state diff) is a `DELETE_REPLACE` then consider the current diff an `UPDATE_REPLACE` instead of `UPDATE`.

Fixes: https://github.com/pulumi/pulumi/issues/5637

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -52,14 +52,14 @@ func containsComputedValues(v resource.PropertyValue) bool {
 // the value.
 type propertyVisitor func(attributeKey, propertyPath string, value resource.PropertyValue) bool
 
-// visitPropertyDiff checks the given property for diffs and invokes the given callback if a diff is found.
+// visitPropertyValue checks the given property for diffs and invokes the given callback if a diff is found.
 //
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/diff.go
```diff
@@ -209,7 +209,7 @@ func makePropertyDiff(name, path string, v resource.PropertyValue, tfDiff shim.I
 					kind = pulumirpc.PropertyDiff_ADD
 				}
 			default:
-				if d.RequiresNew {
+				if d.RequiresNew || other.Kind == pulumirpc.PropertyDiff_DELETE_REPLACE {
```
pgavlin, line 212: I wonder if we should expand this to include the other `replace` cases as well.

komalali: I am having trouble thinking of test cases for the other `replace` cases but adding them doesn't break any existing tests. I'll defer to you if you think we should include those.

komalali:
@pgavlin and I talked through some of the issues here and we need to rework `Set` diffing. I'm gonna close this out and reopen once I have a fix.

PR #297, opened 2020-11-15
Add richer name transformation support

lukehoban:
Part of pulumi/pulumi-aws#967.


stack72 (approved):
Oh this is great - the ability to pass a custom function for the resource is going to be very interesting for these types of scenarios

PR #298, opened 2020-11-17
Bump versions of GHA that still reference add-path

stack72:


stack72:
Self merging this as no other builds will work without it - this is only mechanical within GHA bumping the versions - this will be automated in a future world :)

PR #299, opened 2020-11-18
Ensure that specific naming cases in the examples are picked up

stack72:
Fixes: #https://github.com/pulumi/docs/issues/4507

The GCP docs now use the format

```
## Example Usage - Secret Config Basic
```

Our parser was ignoring that as it only checked for

### <Named>

We have now broken that title up to ensure they are actually picked
up as specific child named examples


stack72:
I expect to see a *huge* diff in pulumi-gcp! As all of the examples will be brought back into play

mikhailshilkov (approved):

PR #300, opened 2020-11-24
Deprecate UsesIOClasses boolean.

komalali:
Related to: https://github.com/pulumi/pulumi/issues/5145

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -293,7 +293,7 @@ type JavaScriptInfo struct {
 type PythonInfo struct {
 	Requires      map[string]string // Pip install_requires information.
 	Overlay       *OverlayInfo      // optional overlay information for augmented code-generation.
-	UsesIOClasses bool              // Indicates whether the package generates input/output classes.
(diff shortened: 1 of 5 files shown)


pkg/tfbridge/info.go
```diff
@@ -293,7 +293,7 @@ type JavaScriptInfo struct {
 type PythonInfo struct {
 	Requires      map[string]string // Pip install_requires information.
 	Overlay       *OverlayInfo      // optional overlay information for augmented code-generation.
-	UsesIOClasses bool              // Indicates whether the package generates input/output classes.
+	UsesIOClasses bool              // Deprecated: No longer required, all providers use IO classes.
```
komalali, line 296: Should I just remove this instead of deprecating? It'll mean forced action to remove this from `resources.go` on the next provider update.

komalali: Actually, I think it makes sense to keep it deprecated rather than removing right away since that would break downstream tests.

stack72 (approved):

PR #301, opened 2020-11-30
Map the `google` TF provider to `google-beta`.

pgavlin:
This is necessary for the in-memory plugin info host to load the cached
info for the GCP provider.


stack72:
This is awesome! Thank you so much for fixing this up - it has sped up the GCP build by approx 10 mins

stack72 (approved):

PR #302, opened 2020-12-04
Bump pulumi/pulumi dependency to commit 26062043 needed for new multi-lang work

stack72:


stack72:
Self merging here - this is only to bump the version of the bridge

Issue #296, opened 2020-11-13
Diff function produces unexpected results when diffing Sets.


komalali:
[makeDetailedDiff](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/diff.go#L257-L288) produces unexpected results when there are multiple items in a set and an item is removed, added or updated.

There are a number of examples where this has manifested in panics:
* https://github.com/pulumi/pulumi-aws/issues/1207
* https://github.com/pulumi/pulumi-aws/issues/1236
* https://github.com/pulumi/pulumi/issues/5637
* https://github.com/pulumi/pulumi/issues/3953
* https://github.com/pulumi/pulumi-aws/issues/1086

Possibly more: https://github.com/issues?q=org%3Apulumi+Expected+diff+to+not+require+deletion+or+replacement+during

PR #304, opened 2020-12-07
Fix Set diffing to exclude items whose values are unchanged

komalali:
`tfDiff` includes each item of the set, even when that specific item hasn't changed (i.e. d.Old == d.New). So if we just check for and exclude those cases, we get the results we expect.

Fixes: #296

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Improve Set diffs to correctly identify replacements.
+  [#304](https://github.com/pulumi/pulumi-terraform-bridge/pull/304)
+
 - Deprecate UsesIOClasses bool.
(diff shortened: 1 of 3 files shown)


pgavlin:
> DiffKind is Update for a set item when there is actually no diff

Can you elaborate on this? Do we understand exactly the circumstances under which this occurs?

komalali:
@pgavlin Thanks for asking that question, it helped me figure out the last remaining piece of the puzzle. It turns out that tfDiff includes each item of the set, even when that specific item hasn't changed (i.e. d.Old == d.New). So if we just check for and exclude those cases, we get the results we expect.

komalali:
Actually, I think we don't need joe's fix at all. If we are just excluding the diffs where `d.Old == d.New` we don't ever end up in a situation where the paths get overwritten.

pgavlin (approved):
🎉

joeduffy:
Woohoo, so glad this ended up being even simpler!

PR #308, opened 2020-12-12
Preserve unknowns during provider preview.

pgavlin:
Just what it says on the tin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Ensure that nested unknowns are preserved during previews of `Create` and `Update` operations.
+  [#308](https://github.com/pulumi/pulumi-terraform-bridge/pull/308)
+
 - Fix the diff reader s.t. reading sets with partially-known elements does not panic.
   [#305](https://github.com/pulumi/pulumi-terraform-bridge/pull/305)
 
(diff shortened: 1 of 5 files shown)


pgavlin:
Fixes https://github.com/pulumi/pulumi/issues/5864.

I'm working on tests to verify that we don't regress this.

leezen (approved):

lukehoban:
> Just what it says on the tin.

Would love more details on what the issue was, what the design is, why these changes are the right delta to achieve the design, and what follow-ups (if any) we should consider in this area.

leezen:
Should be noted that this is looking to address https://github.com/pulumi/pulumi/issues/5864

pgavlin:
> > Just what it says on the tin.
> Would love more details on what the issue was, what the design is, why these changes are the right delta to achieve the design, and what follow-ups (if any) we should consider in this area.

Definitely--much of this got dropped when I squashed.

The issue was that we were not setting the `KeepUnknowns` flag in `MarshalOptions` when returning results from the provider's `Create` and `Update` operations. This caused the marshaler to omit nested unknowns. These unknowns were present in the property bags that were being marshaled, so the only delta necessary to fix this was to set this flag. I do not believe that we need much in the way of follow-up aside from more comprehensive testing in this repository for the various operations for which a provider is responsible.

PR #310, opened 2020-12-16
Propagate better errors from Check

### LINKED ISSUE #309: Bad error messages on `ConflictsWith`
We've seen many reports of bad error messages when `Check` on a resource fails due to a conflict.  

For example:

```
aws:autoscaling:Group (bastion-asg):
    error: aws:autoscaling/group:Group resource 'bastion-asg' has a problem: ConflictsWith
    error: aws:autoscaling/group:Group resource 'bastion-asg' has a problem: ConflictsWith
```

The error message does not make clear what property has a conflict (and with what other property).

Reports of issues with this error:
* https://github.com/pulumi/pulumi-aws/issues/1209
* https://github.com/pulumi/pulumi/issues/5586
* https://github.com/pulumi/pulumi-aws/issues/1184
* https://github.com/pulumi/pulumi-gcp/issues/273#issuecomment-744751985

lukehoban:
In the `sdk-v2` and `tfplugin5` shim implementations, ensure that we bubble up a more complete error message from the underlying Terraform provider.

Also add test coverage for Check errors.

Fixes #309.

### DIFF
--- a/pkg/tfbridge/provider_test.go
+++ b/pkg/tfbridge/provider_test.go
@@ -305,8 +305,38 @@ func testProviderPreview(t *testing.T, provider *Provider) {
 	}).DeepEquals(outs["nestedResources"]))
 }
 
-func testCheckFailures(t *testing.T, provider *Provider) []*pulumirpc.CheckFailure {
-	urn := resource.NewURN("stack", "project", "", "ExampleResource", "name")
+func TestProviderPreview(t *testing.T) {
+	provider := &Provider{
+		tf:     shimv1.NewProvider(testTFProvider),
+		config: shimv1.NewSchemaMap(testTFProvider.Schema),
+	}
+	provider.resources = map[tokens.Type]Resource{
+		"ExampleResource": {
+			TF:     shimv1.NewResource(testTFProvider.ResourcesMap["example_resource"]),
+			TFName: "example_resource",
+			Schema: &ResourceInfo{Tok: "ExampleResource"},
+		},
+	}
+	testProviderPreview(t, provider)
+}
+
+func TestProviderPreviewV2(t *testing.T) {
+	provider := &Provider{
+		tf:     shimv2.NewProvider(testTFProviderV2),
+		config: shimv2.NewSchemaMap(testTFProviderV2.Schema),
+	}
+	provider.resources = map[tokens.Type]Resource{
+		"ExampleResource": {
+			TF:     shimv2.NewResource(testTFProviderV2.ResourcesMap["example_resource"]),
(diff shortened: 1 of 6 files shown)


pgavlin (approved):

pkg/tfbridge/provider_test.go
```diff
@@ -331,5 +369,15 @@ func TestProviderPreviewV2(t *testing.T) {
...
+
 	testProviderPreview(t, provider)
+
+	failures := testCheckFailures(t, provider)
+	sort.SliceStable(failures, func(i, j int) bool { return failures[i].Reason < failures[j].Reason })
+	assert.Equal(t, "ConflictsWith: \"conflicting_property\": conflicts with conflicting_property2", failures[0].Reason)
+	assert.Equal(t, "", failures[0].Property)
+	assert.Equal(t, "ConflictsWith: \"conflicting_property2\": conflicts with conflicting_property", failures[1].Reason)
+	assert.Equal(t, "", failures[1].Property)
+	assert.Equal(t, "Required attribute is not set", failures[2].Reason)
```
lukehoban, line 381: It's unfortunate that the error here still does not include the path of the required attribute.  I believe Terraform relies on `AttributePath` in the diagnostic to communicate this - but not clear if/how we should be surfacing that ourselves into our errors.

pkg/tfbridge/provider_test.go
```diff
@@ -331,5 +369,15 @@ func TestProviderPreviewV2(t *testing.T) {
...
+	assert.Equal(t, "", failures[2].Property)
```
lukehoban, line 382: Slightly related to the above - it feels wrong that we don't populate `Property` on any of these failures - but it appears we've never done that for any errors coming from Check - so this is a separate issue.  It is not immediately clear to me whether we could reliably map an `AttributePath` into a `Property`.

pgavlin: I agree that it feels strange that we don't populate that information. If we can, we should, though I'm not entirely certain that anything would use it at the moment.

pkg/tfbridge/provider_test.go
pgavlin: > I believe Terraform relies on AttributePath in the diagnostic to communicate this 

Yes, I think that you're right. I do believe that we may be able to translate an `AttributePath` into a Pulumi property path--I think that all we need is the relevant TF schema and Pulumi annotations.

lukehoban: Opened https://github.com/pulumi/pulumi-terraform-bridge/issues/311 to track this as a follow-up.

PR #312, opened 2020-12-17
add additional licenses to tfbridge

jaxxstorm:


pkg/tfgen/generate_schema.go
```diff
@@ -261,9 +261,14 @@ func (g *schemaGenerator) genPackageSpec(pack *pkg) (pschema.PackageSpec, error)
 
 	downstreamLicense := g.info.GetTFProviderLicense()
 	licenseTypeURL := getLicenseTypeURL(downstreamLicense)
-	readme := fmt.Sprintf(
-		standardDocReadme, g.pkg, g.info.Name, g.info.GetGitHubOrg(), downstreamLicense, licenseTypeURL,
-		g.info.GetGitHubHost())
+
+	readme := ""
+	if downstreamLicense != tfbridge.UnlicensedLicenseType {
+		readme = fmt.Sprintf(
```
stack72, line 267: This is the only question I have about this PR but it makes sense that no-attribution and nolicense would go together

stack72 (approved):

pkg/tfgen/generate_schema.go
lukehoban: It is a little odd to couple "unlicensed" to "doesn't emit a README indicating it's a derived work of a Terraform provider".  These seem at least partly orthogonal. 

We'll likely need to revisit all of this anyway though - as these READMEs aren't really being used prominently anymore (we're using the attribution piece instead) - and this makes a minimal addition the the schema surface area - so happy to take this is as for right now.

lukehoban (approved):

PR #313, opened 2020-12-19
Update to the latest sdk-v2 fork.

pgavlin:


stack72:
This PR cannot be merged at this time due to the fact that terraform-provider-azure doesn't use v2 of the plugin SDK :/

pgavlin:
> This PR cannot be merged at this time due to the fact that terraform-provider-azure doesn't use v2 of the plugin SDK :/

That shouldn't prevent us from merging this... the shim is able to accommodate such providers. It looks like we might just need `terraform-provider-azure` to apply an replacement for its version of terraform-plugin-test.

stack72:
@pgavlin I've tried this - we can't do it - when I am back at my machine I will send the logs of what it complains about 

It's not the bridge being the problem here - that works, it's actually an issue with azure :/

stack72:
So the build gives us the following:

```
(cd provider && go build -a -o /Users/stack72/code/go/src/github.com/pulumi/pulumi-azure/bin/pulumi-tfgen-azure -ldflags "-X github.com/pulumi/pulumi-azure/provider/v3/pkg/version.Version=3.42.0+dirty" github.com/pulumi/pulumi-azure/provider/v3/cmd/pulumi-tfgen-azure)
# github.com/hashicorp/terraform-plugin-test
../../../../../pkg/mod/github.com/hashicorp/terraform-plugin-test@v1.3.0/helper.go:184:31: not enough arguments in call to zipDecompressor.Decompress
	have (string, string, bool)
	want (string, string, bool, os.FileMode)
^Cmake: *** [tfgen] Interrupt: 2
```

I've tried replacements for v1.3.0, v1.4.0 and still get the same error. I have also tried replacement using v1.4.1

```
(cd provider && go build -a -o /Users/stack72/code/go/src/github.com/pulumi/pulumi-azure/bin/pulumi-tfgen-azure -ldflags "-X github.com/pulumi/pulumi-azure/provider/v3/pkg/version.Version=3.42.0+dirty" github.com/pulumi/pulumi-azure/provider/v3/cmd/pulumi-tfgen-azure)
go: downloading github.com/hashicorp/terraform-plugin-test v1.4.1
# github.com/hashicorp/terraform-plugin-test
../../../../../pkg/mod/github.com/hashicorp/terraform-plugin-test@v1.4.1/config.go:46:31: not enough arguments in call to tfinstall.Find
	have ([]tfinstall.ExecPathFinder...)
	want (context.Context, ...tfinstall.ExecPathFinder)
../../../../../pkg/mod/github.com/hashicorp/terraform-plugin-test@v1.4.1/helper.go:189:31: not enough arguments in call to zipDecompressor.Decompress
	have (string, string, bool)
	want (string, string, bool, os.FileMode)
c^Cmake: *** [tfgen] Interrupt: 2
```

v1.4.2 / v1.4.3 is the same. 

terraform-provider-azurerm uses v1.16.1 of the terraform-plugin-sdk which uses v1.3.0 of the terraform-plugin-test pkg - it feels like we are a wee bit stuck here...

stack72:
FWIW, AzureAD has been updated today - there's a PR in our repo to merge to master to unblock that upgrade

stack72:
As 4 of our 5 providers work with this patch, the way forward is that we (temporarily) not build AzureRM and merge this...

thoughts @lukehoban ?

PR #316, opened 2021-01-11
Migrate the copied helper funcs for boolRef and tfProviderLicense to the bridge

stack72:
These funcs are copied and pasted all over providers so they are a good area to add to the bridge


stack72:
I expect random and aws to fail here - there is a pending PR that fixes these up https://github.com/pulumi/pulumi-terraform-bridge/pull/313

I need to work on getting this merged

mikhailshilkov (approved):

PR #317, opened 2021-01-15
Process ignoreChanges before asserting diff does not cause deletion or replacement

komalali:
Fixes a panic caused by ignoreChanges not being processed before the assertion leading to the panic.

The codegen test failures are unrelated to this PR.

Fixes: https://github.com/pulumi/pulumi-aws/issues/1276

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Fixes a panic caused by ignoreChanges not being processed before the assertion leading to the panic.
+  [#317](https://github.com/pulumi/pulumi-terraform-bridge/pull/317)
+
 - Add "Unlicensed" license type.
   [#312](https://github.com/pulumi/pulumi-terraform-bridge/pull/312)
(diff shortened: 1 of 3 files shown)


pgavlin:
The provider preview unit test provides a reasonable framework from which to start: https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/provider_test.go#L201-L306

Using a similar style, you can write a test that steps through the `Check`/`Diff`/`Update` sequence using inputs that repro the original issue. The schema for the test resource is [here](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/internal/testprovider/schema.go#L47-L80). It looks like you'll need to use the `set_property_value` property to test this, since it's the only property marked `ForceNew` (unless you add a new property, but that's more involved).

komalali:
@pgavlin Awesome! Confirmed that the test failed with `Expected diff to not require deletion or replacement during Update of...` prior to the change, and passes as expected with the change.

pgavlin (approved):
LGTM!

PR #319, opened 2021-01-19
Correctly handle replacements for sets and lists with nested elements

komalali:
There is a bug in our diffing function's handling of sets and lists with nested elements when the set or list itself has `ForceNew` but the nested element does not.

To correct this, we also check if the list/set count diff has `ForceNew` and treat `A`/`D` as `AR`/`DR` if so.

Fixes: https://github.com/pulumi/pulumi-aws/issues/1297

### DIFF
--- a/pkg/tfbridge/diff_test.go
+++ b/pkg/tfbridge/diff_test.go
@@ -219,6 +219,7 @@ func TestNestedAdd(t *testing.T) {
 			"outp": "bar",
 		},
 		map[string]DiffKind{
+			"prop":      A,
 			"prop.nest": A,
 		})
 }
@@ -237,6 +238,7 @@ func TestNestedAddReplace(t *testing.T) {
 			"outp": "bar",
 		},
 		map[string]DiffKind{
+			"prop":      AR,
 			"prop.nest": AR,
 		})
 }
@@ -345,6 +347,7 @@ func TestListAdd(t *testing.T) {
 			"outp": "bar",
 		},
 		map[string]DiffKind{
+			"prop":    A,
 			"prop[0]": A,
 		})
 }
@@ -363,6 +366,7 @@ func TestListAddReplace(t *testing.T) {
 			"outp": "bar",
 		},
 		map[string]DiffKind{
+			"prop":    AR,
 			"prop[0]": AR,
 		})
 }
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/diff_test.go
```diff
@@ -1474,3 +1496,191 @@ func TestCollectionsWithMultipleItems(t *testing.T) {
...
+			"prop": []interface{}{map[string]interface{}{"nest": "foo"}},
+		},
+		// state
+		map[string]interface{}{
+			"prop": nil,
+			"outp": "bar",
+		},
+		// expected
+		map[string]DiffKind{
+			"prop":         AR,
```
pgavlin, line 1556: Out of curiosity, have you checked to see how this renders on the CLI side?

komalali: Just tried it out. Looks fine to me 🤷 

Screen Shot 2021-01-22 at 8 46 53 PM

komalali: Here's another... this one falls in the `TestCollectionsWithMultipleItems.RemoveFirst` for a set.

Screen Shot 2021-01-25 at 9 18 07 AM

pgavlin: Awesome. I'll admit that it feels a little weird that we don't see simply `"prop[0].nest": AR,`, but I think that's okay given the rendering.

pgavlin (approved):
LGTM. Could you file a followup issue to track the potential of changing the plain add of the nested property to an add-replace (or documenting the reason the detailed diff is rendered the way it is)?

pkg/tfbridge/diff_test.go
komalali: Fixed this in the latest commit.

pgavlin (approved):
This does look better, but I think it might be cleaner to pass a bit through the property visitor that indicates whether or not we're inside an add/delete of an array element.

komalali:
> I think it might be cleaner to pass a bit through the property visitor that indicates whether or not we're inside an add/delete of an array element.

@pgavlin I'm not really clear on how exactly to do this, but please feel free to make a follow-up change with what you're suggesting.

PR #324, opened 2021-01-28
Move GetModuleMajorVersion to the bridge so it can be shared

stack72:
This function should only return a version string to be appended
to a go sdk path if it is v2 or later. v1 go modules do not specify
a path like github.com/org/my-repo/sdk/v1/go

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -15,6 +15,8 @@
 package tfbridge
 
 import (
+	"fmt"
+	"github.com/blang/semver"
 	pschema "github.com/pulumi/pulumi/pkg/v2/codegen/schema"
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/info.go
```diff
@@ -716,3 +718,16 @@ func (m *MarshallableProviderInfo) Unmarshal() *ProviderInfo {
 
 	return &info
 }
+
+// Calculates the major version of a go sdk
```
mikhailshilkov, line 722: Maybe add a comment that it ignores v1 and why

mikhailshilkov (approved):

pkg/tfbridge/info.go
stack72: Good call! Updated now

PR #326, opened 2021-02-10
Ensure PreConfigureCallback gets called

lukehoban:
It appears this was regressed by https://github.com/pulumi/pulumi-terraform-bridge/pull/242, which partially reverted https://github.com/pulumi/pulumi-terraform-bridge/pull/225 but left in some code that doesn't appear to be correct without the other reverted code (it's not immediately clear why it's correct even with that original code - but at least it seems intentional as part of the original change).

Also added tests to ensure that PreConfigureCallback is called.

Fixes pulumi/pulumi-aws#1316.


stack72 (approved):
ah :/ That's a great find! Thank you for fixing that

PR #332, opened 2021-03-08
Initial attempt to add attribute info for check errors

### LINKED ISSUE #311: Include property name in Check errors
Currently, the `sdk-v2` and `tfplugin5` shims do not include the `AttributePath` of a Terraform `Diagnostic` in the error message reported back from `Check` in Pulumi.  For some validation errors, it is likely this information is necessary for a helpful/actionable error.

This is a follow-up to #310 and #309.

viveklak:
Fixes #311

The PR currently gets additional context on check failures but still has some contextual gaps:
1. No insight into the user program's language so "pulumified" attributes may not directly match the user's program. Can't really use the provided input either since the problem may be due to required attributes not being specified.
2. Only supports v2 and tf5 shims. v1 doesn't seem to emit attribute information.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -560,7 +560,7 @@ var requiredFieldRegex = regexp.MustCompile("\"(.*?)\": required field is not se
 func (p *Provider) formatFailureReason(tokenType tokens.Type, res Resource, err error) string {
 	reason := err.Error()
 	var attributePath string
-	var d *diagnostics.Error
+	var d *diagnostics.ValidationError
 	if errors.As(err, &d) {
 		path := d.AttributePath
 		if len(path) > 0 {
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -17,6 +17,8 @@ package tfbridge
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/pulumi/pulumi-terraform-bridge/v2/pkg/tfshim/diagnostics"
 	"log"
 	"regexp"
 	"strings"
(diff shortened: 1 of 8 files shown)


pgavlin:
> The attribute path returned from the validation call has the following form: [cty.Path{cty.GetAttrStep{Name:"vpc_config"}, cty.IndexStep{Key:cty.NumberIntVal(0)}, cty.GetAttrStep{Name:"vpc_id"}}] even though vpc_config is not a list type.

What is the type of `vpc_config` in the TF schema? It's possible that this is a collection in the TF schema that we project into a single optional input.

pkg/tfbridge/provider.go
```diff
@@ -555,7 +557,18 @@ func (p *Provider) Configure(ctx context.Context,
...
+	var attributePath string
+	var d *diagnostics.Error
+	if errors.As(err, &d) {
+		path := d.AttributePath
+		if len(path) > 0 {
+			attrPath := pathToAttributePath(path, tokenType, res)
+			attributePath = strings.Join(attrPath, "")
+		}
+	}
+
```
pgavlin, line 571: Does this work for all of the various shims, or will this only work for the v2 shim?

viveklak:
> > The attribute path returned from the validation call has the following form: [cty.Path{cty.GetAttrStep{Name:"vpc_config"}, cty.IndexStep{Key:cty.NumberIntVal(0)}, cty.GetAttrStep{Name:"vpc_id"}}] even though vpc_config is not a list type.
> 
> What is the type of `vpc_config` in the TF schema? It's possible that this is a collection in the TF schema that we project into a single optional input.

Sigh - its a typelist: https://github.com/pulumi/terraform-provider-aws/blob/upstream-v3.31.0/aws/data_source_aws_eks_cluster.go#L100

pgavlin:
> Sigh - its a typelist: https://github.com/pulumi/terraform-provider-aws/blob/upstream-v3.31.0/aws/data_source_aws_eks_cluster.go#L100

I thought it might be :)

In that case, the traversal -> property path logic will need to take the `IsMaxItemsOne` details into account.

viveklak:
> > Sigh - its a typelist: https://github.com/pulumi/terraform-provider-aws/blob/upstream-v3.31.0/aws/data_source_aws_eks_cluster.go#L100
> 
> I thought it might be :)
> 
> In that case, the traversal -> property path logic will need to take the `IsMaxItemsOne` details into account.

Awesome thanks! I will update shortly.

pkg/tfbridge/provider.go
viveklak: > Does this work for all of the various shims, or will this only work for the v2 shim?

I believe it should but I focussed on the v2 shim for now. I will verify before marking PR ready for review.

viveklak:
> > Sigh - its a typelist: https://github.com/pulumi/terraform-provider-aws/blob/upstream-v3.31.0/aws/data_source_aws_eks_cluster.go#L100
> 
> I thought it might be :)
> 
> In that case, the traversal -> property path logic will need to take the `IsMaxItemsOne` details into account.

This seems to work, thanks. I will look to add v1 shim support next.

pkg/tfbridge/provider.go
viveklak: I was wrong. I added support for tf5 but it appears v1's terraform-plugin-sdk doesn't actually return attribute information. It does appear though that it provides some useful information in the returned error but short of parsing the error messages, don't think this approach extends to the v1 shim.

viveklak:
@pgavlin this is now ready for review. Please see some comments [here](https://github.com/pulumi/pulumi-terraform-bridge/pull/332#discussion_r591039450) regarding extending to other shims.

pgavlin (approved):
LGTM!

PR #336, opened 2021-03-18
Address potential panic in array refresh

jen20:
This commit addresses a potential panic when extracting inputs for arrays, which can present during refresh operations if the length of the "new" array is longer than the length of the "old" array:

```
panic: runtime error: slice bounds out of range [:2] with capacity 1

[backtrace elided] pkg/tfbridge/schema.go:1019
```

The solution here is to use the min(len(oldArray), len(newArray)) as the index point for the mutated old array instead of len(newArray) unconditionally.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -995,6 +995,13 @@ func convertTfStringToFloat(stringValue string) (interface{}, error) {
 	return floatVal, nil
 }
 
+func min(a int, b int) int {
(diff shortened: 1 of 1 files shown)


pgavlin (approved):
Would you mind adding a unit test? If not, I can follow up with one later.

jen20:
I can add some - it's not obvious which test path exercises refresh in this way though - any pointers?

PR #341, opened 2021-04-16
Upgrade to v3 module for use with Pulumi sdk v3

stack72:


stack72:
unfortunately, downstream providers are not feature-3.0 -> master yet so we won't get any codegen

mikhailshilkov (approved):

PR #347, opened 2021-05-13
Create provider output properties for all inputs

jen20:
This PR alters the generation of the provider schema from the Terraform Bridge definition to include all input properties as output properties, in order that the configuration of a provider instance can be accessed during Pulumi program execution.

This has been tested downstream with `pulumi-digitalocean` and `pulumi-aws` and permits access to the configured values.


pgavlin:
@stack72 are the downstream jobs here no longer generating diffs?

@jen20 could you post a snippet of the SDK diffs?

jen20:
https://gist.github.com/jen20/863c558e4c8ec88ed6e584e4916c9894 is the diff for `pulumi-aws` (though I removed the changes to `go.{mod,sum}` since they were mostly noise)

pgavlin (approved):

pgavlin:
Oh, could you bump the CHANGELOG with a blurb re: the new properties that users will see in the Go/.NET/TS SDKs?

PR #353, opened 2021-06-14
Allow missing properties in PCL codegen.

pgavlin:
Just what it says on the tin. These changes will help ensure that
examples are not lost if and when typechecking of resource input
properties is enabled.


pgavlin:
Depends on https://github.com/pulumi/pulumi/pull/7297.

lblackstone (approved):
Change LGTM, but didn't look into test failures

pgavlin:
> Change LGTM, but didn't look into test failures

Test failures should be build failures, b/c these changes use an option that doesn't exist yet :)

pgavlin:
Downstream tests are failing b/c the jobs do not run `go mod tidy`. Diffs look good locally.

PR #354, opened 2021-06-15
Fix a bug in list element projection.

pgavlin:
Ensure that tuples are projected as well as lists.


pgavlin:
Hm, I may have misinterpreted the results I was seeing. Closing this out for now.

PR #361, opened 2021-07-01
Add htmlspan to be rendered as a markdown text sequence

stack72:


pkg/tfgen/parse_markdown.go
```diff
@@ -427,6 +427,8 @@ func parseTextSeq(firstNode *bf.Node, useStarsForStrongAndEmph bool) (string, er
 				} else {
 					buffer.WriteString("_")
 				}
+			case bf.HTMLSpan:
```
t0yv0, line 430: I think it might swallow the contents of the span. In the case of the original issue:

```
panic: parseTextSeq found a tag it cannot yet render back to Markdown: [
  "[HTMLSpan]",
  "\u003cbr\u003e"
]
```

It will swallow whitespace, so we don't care.

However if someone had meaningful text inside the span we'd be swallowing it silently.

Perhaps print it out as a warning? At least more visible.

t0yv0 (approved):

PR #365, opened 2021-07-08
Add flags to skip docs and examples processing.

pgavlin:
These are intended to speed up inner-loop debugging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Add flags to skip docs and examples processing.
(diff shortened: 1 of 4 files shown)


stack72:
@pgavlin this is great - does this mean we would need to update the tf providers to pass the flag as part of tfgen? Can I enhance this and use Viper to bind an env var as well?

stack72 (approved):
Left a question and a question about enhancement but no blocker :)

pgavlin:
> does this mean we would need to update the tf providers to pass the flag as part of tfgen

nope—they’ll automatically pick this up with the next bridge update

PR #369, opened 2021-07-27
Stop emitting `mapCase` for Python properties

justinvp:
The Python SDK code generator does not use `mapCase` anymore, so there's no need to continue emitting it in the generated schema.

Follow-up from https://github.com/pulumi/pulumi/pull/7647


justinvp:
Ah, need to upgrade the `pulumi` dependency first to see intended diffs.

iwahbe:
Closing in favor of #879.

PR #373, opened 2021-08-05
Fixed incorrect early return from example generation if just one language fails

T-Vova:
It seems that this change was intended by commit 194de63e91e9ec6c0095b741e2e5c11a479e38ce, however, it wasn't fully working and would lead to languages being skipped if a prior one failed.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -29,6 +29,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/spf13/afero"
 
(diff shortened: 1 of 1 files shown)


pkg/tfgen/docs.go
```diff
@@ -1169,17 +1170,25 @@ func (g *Generator) convertHCL(hcl, path string) (string, string, error) {
 		err = convertHCL("go")
 	case Schema:
 		langs := []string{"typescript", "python", "csharp", "go"}
+		var anySucceeded bool = false
 		for _, lang := range langs {
 			if langErr := convertHCL(lang); langErr != nil {
-				return "", stderr.String(), langErr
+				err = multierror.Append(err, langErr)
```
pgavlin, line 1176: It would be nice if we could still capture stderr here. If you want to take care of that with a separate PR, please file an issue for followup.

pgavlin (approved):
LGTM—one comment re: not dropping stderr.

pkg/tfgen/docs.go
T-Vova: I am working on expanding coverage and diagnostics tracking, and will make sure that stderr is captured even if the language isn't able to convert.

PR #374, opened 2021-08-06
Created a system to let Tfgen track example coverage statistics, and to export the data in various JSON formats

T-Vova:
Early implementation of a system for measuring coverage during example generation


pkg/tfgen/main.go
```diff
@@ -116,16 +116,20 @@ func newTFGenCmd(pkg string, version string, prov tfbridge.ProviderInfo) *cobra.
 				root = afero.NewBasePathFs(afero.NewOsFs(), absOutDir)
 			}
 
+			// Creating an item to keep track of example coverage
+			var coverageTracker CoverageTracker = newCoverageTracker(prov.Name)
```
stack72, line 120: I don't think we should instantiate a coverageTracker if there is not the requirement to do so - i.e. an env var suggesting that you want an output

Thoughts?

pkg/tfgen/main.go
```diff
@@ -137,6 +141,12 @@ func newTFGenCmd(pkg string, version string, prov tfbridge.ProviderInfo) *cobra.
 				return err
 			}
 
+			// Exporting collected coverage data to a given directory if the COVERAGE_OUTPUT_DIR env is set
+			COD := os.Getenv("COVERAGE_OUTPUT_DIR")
```
stack72, line 145: we can actually do something like

```
cod, exists := os.LookupEnv("COVERAGE_OUTPUT_DIR")
if exists {
   coverageTracker.exportResults(cod)
}
```

then you don't need to check for an empty string

T-Vova: Fixed

pkg/tfgen/main.go
T-Vova: Changed coverage tracker to not be instantiated if environment variable isn't present.

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+	CT *CoverageTracker // Reference to the Coverage Tracker that wants to turn its data into a file
+}
+
+func newCoverageExportUtil(coverageTracker *CoverageTracker) CoverageExportUtil {
+	return CoverageExportUtil{coverageTracker}
+}
+
+// The entire export utility interface. Will attempt to export the Coverage Tracker's data into the
+// specified output directory, and will panic if an error is encountered along the way
+func (CE *CoverageExportUtil) tryExport(outputDirectory string) {
```
pgavlin, line 39: Variable names in Go are idiomatically `camelCase`.

```suggestion
func (ce *CoverageExportUtil) tryExport(outputDirectory string) {
```

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+// The export utility's main structure, where it stores the desired output directory
+// and a reference to the CoverageTracker that created it
+type CoverageExportUtil struct {
```
pgavlin, line 29: Does this type need to be exported? If we do not intend to consume it from other packages, I suggest we unexport it by changing its name to `coverageExportUtil`.

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+	CE.exportUploadableResults(outputDirectory, "summary.json")
+	CE.exportSummarizedResults(outputDirectory, "concise.json")
+}
+
+// Three different ways to export coverage data:
+// The first mode, using a large provider > example map
+func (CE *CoverageExportUtil) exportFullResults(outputDirectory string, fileName string) {
+
+	// The Coverage Tracker data structure remains identical, the only thing added in the file is the name of the provider
+	ProviderNameToExamplesMap := map[string]map[string]GeneralExampleInfo{CE.CT.ProviderName: CE.CT.EncounteredExamples}
```
pgavlin, line 49:
```suggestion
	providerNameToExamplesMap := map[string]map[string]GeneralExampleInfo{CE.CT.ProviderName: CE.CT.EncounteredExamples}
```

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+// The export utility's main structure, where it stores the desired output directory
+// and a reference to the CoverageTracker that created it
+type CoverageExportUtil struct {
+	CT *CoverageTracker // Reference to the Coverage Tracker that wants to turn its data into a file
```
pgavlin, line 30:
```suggestion
	Tracker *CoverageTracker // Reference to the Coverage Tracker that wants to turn its data into a file
```

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+		ProviderVersion string
+		ExampleName     string
+		FailedLanguages []LanguageConversionResult
+	}
+
+	jsonOutputLocation := createJsonOutputLocation(outputDirectory, fileName)
+
+	// All the examples in the map are iterated by key and marshalled into one large byte array
+	// separated by \n, making the end result look like a bunch of Json files that got concatenated
+	var result = []byte{}
```
pgavlin, line 71: You can simplify this to

```suggestion
	var result []byte
```

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+	for _, exampleInMap := range CE.CT.EncounteredExamples {
+		singleExample := SingleExampleResult{
+			ProviderName:    CE.CT.ProviderName,
+			ProviderVersion: CE.CT.ProviderVersion,
+			ExampleName:     exampleInMap.Name,
+			FailedLanguages: []LanguageConversionResult{},
+		}
+
+		for _, conversionResult := range exampleInMap.LanguagesConvertedTo {
+			if conversionResult.FailureSeverity > 0 {
```
pgavlin, line 81: Would we expect any failures to have a severity of 0?

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+				singleExample.FailedLanguages = append(singleExample.FailedLanguages, conversionResult)
+			}
+		}
+		marshalledExample, err := json.MarshalIndent(singleExample, "", "\t")
+		panicIfError(err, "Failed to MarshalIndent JSON file")
```
pgavlin, line 86: It's pretty non-idiomatic to panic on errors. Any reason that we're not checking these and returning an error on failure?

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+		result = append(append(result, marshalledExample...), uint8('\n'))
+	}
+	err2 := ioutil.WriteFile(jsonOutputLocation, result, 0600)
```
pgavlin, line 89: You can just reuse the existing error:

```suggestion
	err = ioutil.WriteFile(jsonOutputLocation, result, 0600)
```

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+
+	type CoverageStatistics struct {
+		NoErrors      NumPct
+		LowSevErrors  NumPct
+		HighSevErrors NumPct
+		Fatal         NumPct
+		Total         int
+	}
+
+	SummarizedResult := CoverageStatistics{NumPct{0, 0.0}, NumPct{0, 0.0}, NumPct{0, 0.0}, NumPct{0, 0.0}, 0}
```
pgavlin, line 111:
```suggestion
	summarizedResult := CoverageStatistics{NumPct{0, 0.0}, NumPct{0, 0.0}, NumPct{0, 0.0}, NumPct{0, 0.0}, 0}
```

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+			SummarizedResult.Total += 1
+			if conversionResult.FailureSeverity == 0 {
+				SummarizedResult.NoErrors.Number += 1
+			} else if conversionResult.FailureSeverity == 1 {
+				SummarizedResult.LowSevErrors.Number += 1
+			} else if conversionResult.FailureSeverity == 2 {
+				SummarizedResult.HighSevErrors.Number += 1
+			} else {
+				SummarizedResult.Fatal.Number += 1
+			}
```
pgavlin, line 125: This is well-suited for a `switch` statement:

```suggestion
			switch conversionResult.FailureSeverity {
			case 0:
				SummarizedResult.NoErrors.Number++
			case 1:
				SummarizedResult.LowSevErrors.Number++
			case 2:
				SummarizedResult.HighSevErrors.Number++
			default:
				SummarizedResult.Fatal.Number += 1
			}
```

pkg/tfgen/examples_coverage_exporter.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+
+	SummarizedResult.NoErrors.Pct = float32(SummarizedResult.NoErrors.Number) / float32(SummarizedResult.Total) * 100.0
+	SummarizedResult.LowSevErrors.Pct = float32(SummarizedResult.LowSevErrors.Number) / float32(SummarizedResult.Total) * 100.0
+	SummarizedResult.HighSevErrors.Pct = float32(SummarizedResult.HighSevErrors.Number) / float32(SummarizedResult.Total) * 100.0
+	SummarizedResult.Fatal.Pct = float32(SummarizedResult.Fatal.Number) / float32(SummarizedResult.Total) * 100.0
```
pgavlin, line 132: It's not really of consequence--we don't need the extra precision--but it's typical in Go to use `float64` rather than `float32` for most FP operations.

pkg/tfgen/examples_coverage_exporter.go
pgavlin: Also, it looks like there are `const`s for these numbers defined elsewhere, so please use those rather than the numbers in these cases.

pkg/tfgen/examples_coverage_tracker.go
```diff
@@ -0,0 +1,151 @@
...
+	Fatal  = 3
+)
+
+func newCoverageTracker(ProviderName string, ProviderVersion string) *CoverageTracker {
+	return &CoverageTracker{ProviderName, ProviderVersion, "", make(map[string]GeneralExampleInfo)}
+}
+
+//========================== Coverage Tracker Interface ===========================
+// At different stages, the code translator notifies the tracker of what is going on.
+// Notifications are treated as an ordered stream of events: foundExample must be first.
```
pgavlin, line 63: This is not a convention that we use. Please add this comment to the documentation on the `CoverageTracker` type, where it will be picked up by `go doc` et. al.

pkg/tfgen/examples_coverage_tracker.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	CT.insertLanguageConversionResult(LanguageConversionResult{
+		TargetLanguage:            targetLanguage,
+		FailureSeverity:           3,
+		FailureInfo:               panicInfo,
+		ExamplePossiblyDuplicated: false,
+	})
+}
+
+//=================================================================================
```
pgavlin, line 124:
```suggestion
```

pkg/tfgen/examples_coverage_tracker.go
```diff
@@ -0,0 +1,151 @@
...
+	} else {
+		fmt.Println("Error: attempted to log the result of a language conversion without first finding an example")
+		fmt.Println(conversionResult)
+		panic("")
+	}
+}
+
+// Exporting the coverage results
+func (CT *CoverageTracker) exportResults(outputDirectory string) {
+	var coverageExportUtil = newCoverageExportUtil(CT)
```
pgavlin, line 149:
```suggestion
	coverageExportUtil := newCoverageExportUtil(CT)
```

pkg/tfgen/main.go
```diff
@@ -116,16 +116,25 @@ func newTFGenCmd(pkg string, version string, prov tfbridge.ProviderInfo) *cobra.
 				root = afero.NewBasePathFs(afero.NewOsFs(), absOutDir)
 			}
 
+			// Creating an item to keep track of example coverage if the
+			// COVERAGE_OUTPUT_DIR env is set
+			var coverageTracker *CoverageTracker = nil
```
pgavlin, line 121:
```suggestion
			var coverageTracker *CoverageTracker
```

pgavlin (changes requested):

stack72:
Hi @T-Vova 

There's a panic on a unit test here - lets see if we can get this fixed up and we can ship this :)

PR #375, opened 2021-08-09
Correct data source naming in Python doc strings

jen20:
This commit corrects the name representation of a Terraform data source if used in a docstring for Python.

Before this change a data source named `aws_some_data_source` would have been converted to `aws.someDataSource` for Python, which does not follow the naming standards for Python. After this change the function would be correctly called `aws.some_data_source`.

### DIFF
--- a/pkg/tf2pulumi/gen/python/generator.go
+++ b/pkg/tf2pulumi/gen/python/generator.go
@@ -140,7 +140,7 @@ func (g *generator) GenerateResource(r *il.ResourceNode) error {
 	// data source call itself into the apply.
 	if r.IsDataSource {
 		// Qualified member names are snake cased for data sources.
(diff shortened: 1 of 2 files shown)


stack72:
As this is a fork, we can't send a branch upstream to test the diff but the diffs are as expected:

```
-        :param pulumi.Input[str] path_id: Path identifier of the product. This value is optional if the product has a default path, and required if the product has more than one path. To list the paths for a product, use `servicecatalog.getLaunchPaths`. When required, you must provide `path_id` or `path_name`, but not both.
+        :param pulumi.Input[str] path_id: Path identifier of the product. This value is optional if the product has a default path, and required if the product has more than one path. To list the paths for a product, use `servicecatalog.get_launch_paths`. When required, you must provide `path_id` or `path_name`, but not both.
```

stack72 (approved):

PR #376, opened 2021-08-17
tfbridge: handle unidirectional conflicts

benesch:
Some Terraform providers specify ConflictsWith in one direction but not
the other (e.g., pulumi/pulumi-snowflake#11). Teach applyDefaults to
handle this situation too to avoid spurious conflict warnings.

Arguably, specifying ConflictsWith in one direction but not the other is
a bug in the provider schema, but at the very least it's a mistake
that's easy to make. Since Terraform seems to handle it just fine, it
seems worth handling in Pulumi too.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -355,8 +355,18 @@ func (ctx *conversionContext) applyDefaults(result map[string]interface{}, olds,
 	if tfs != nil {
 		tfs.Range(func(name string, sch shim.Schema) bool {
 			if _, has := result[name]; has {
+				// `name` is present, so mark any names that are declared to
+				// conflict with `name` for exclusiion.
 				for _, conflictingName := range sch.ConflictsWith() {
 					conflictsWith[conflictingName] = struct{}{}
+				}
+			} else {
+				// `name` is not present, so mark it for exclusion if any fields
+				// that conflict with `name` are present.
+				for _, conflictingName := range sch.ConflictsWith() {
+					if _, has := result[conflictingName]; has {
+						conflictsWith[name] = struct{}{}
+					}
 				}
 			}
 			return true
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/schema.go
```diff
@@ -355,9 +355,19 @@ func (ctx *conversionContext) applyDefaults(result map[string]interface{}, olds,
...
+				// conflict with `name` for exclusiion.
 				for _, conflictingName := range sch.ConflictsWith() {
 					conflictsWith[conflictingName] = struct{}{}
 				}
+			} else {
+				// `name` is not present, so mark it for exclusion if any fields
+				// that conflict with `name` are present.
+				for _, conflictingName := range sch.ConflictsWith() {
+					if _, has := result[conflictingName]; has {
+						conflictsWith[name] = struct{}{}
```
pgavlin, line 368: _tiny_ nit:

```suggestion
						conflictsWith[name] = struct{}{}
						break
```

pkg/tfbridge/schema.go
```diff
@@ -355,9 +355,19 @@ func (ctx *conversionContext) applyDefaults(result map[string]interface{}, olds,
 	if tfs != nil {
 		tfs.Range(func(name string, sch shim.Schema) bool {
 			if _, has := result[name]; has {
+				// `name` is present, so mark any names that are declared to
+				// conflict with `name` for exclusiion.
```
pgavlin, line 359:
```suggestion
				// conflict with `name` for exclusion.
```

pgavlin (approved):
Looks great! Thanks for fixing this.

benesch:
Thanks very much!

stack72:
Hey @benesch 

I am going to get this rolled out to the Snowflake provider very shortly!

Paul

PR #379, opened 2021-08-25
Minor Improvements to Coverage Tracker

T-Vova:
Concise results are additionally sorted by alphabetic order if their quantity matches, and the summary now shows the original HCL code of the example if conversion wasn't successful.

### DIFF
--- a/pkg/tfgen/examples_coverage_tracker.go
+++ b/pkg/tfgen/examples_coverage_tracker.go
@@ -33,7 +33,7 @@ import (
 type CoverageTracker struct {
 	ProviderName        string                         // Name of the provider
 	ProviderVersion     string                         // Version of the provider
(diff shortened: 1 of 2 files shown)


pkg/tfgen/examples_coverage_tracker.go
```diff
@@ -153,14 +154,37 @@ func (ct *CoverageTracker) insertLanguageConversionResult(conversionResult Langu
...
+
+	// Helper method to check if results already have one of this diagnostic
+	contains := func(result []string, target string) bool {
+		for _, diag := range result {
+			if diag == target {
+				return true
+			}
+		}
+		return false
+	}
```
lukehoban, line 170: Kind of minor, but this could be a place where StringSet could be used?

lukehoban (approved):
LGTM

PR #380, opened 2021-08-28
Added Sha1 support for Golang

T-Vova:


pkg/tf2pulumi/convert/tf12.go
```diff
@@ -1902,6 +1902,13 @@ var tf12builtins = map[string]*model.Function{
 				ReturnType: elementType,
 			}, diagnostics
 		})),
+	"sha1": model.NewFunction(model.StaticFunctionSignature{
+		Parameters: []model.Parameter{{
+			Name: "input string",
```
lukehoban, line 1907: How is this name used?  In all other cases here - it is a valid identifier, but this new one includes a space. Should this case also be a valid identifier, like `input`?

T-Vova: Done.

pgavlin:
This allows the PCL binding to succeed, but it doesn't add support for emitting calls to `sha1` in Go, right?

pgavlin:
Ah, I see--looks like this PR is superseded by https://github.com/pulumi/pulumi/pull/7834

PR #388, opened 2021-09-13
Fixed "unknown property '0'" error

T-Vova:


pgavlin (approved):
I'll admit that I'm a bit confused about how this arises--i.e. an `IsMaxItemsOne` schema that accepts an object rather than a list--but given that it does, I think this change is good :+1:

PR #392, opened 2021-09-21
feat: Support omitting properties

BenjaminSchiborr:
Currently there is no way to not generate Pulumi bindings for a
`DataSource` or a `Resource` for which a property in the underlying
Terraform provider exists.

This commit adds a new `Omit` flag in the `SchemaInfo`. When `Omit` is
enabled, the property is dropped from the schema and no bindings will
be generated for it.


stack72:
Hey @BenjaminSchiborr 

we need to ensure that any omitted properties are not required properties or this can be problematic when trying to use the provider and would need a re-release

p.

PR #394, opened 2021-09-23
Upgrade to latest Pulumi commit to fix the Python version build

stack72:


komalali (approved):
:shipit:

komalali (changes requested):
@stack72 pull in this change too https://github.com/pulumi/pulumi/pull/8038/files

komalali (approved):

PR #401, opened 2021-10-26
refactor: return less meaningless error

hyde-zhang:
Don't think messages like: `expected non-nil error with nil state during Create of ...` should be returned up. 

There must be either response value or error not both empty. If the provider returns empty or unexpected response then we should either:

- Query resource state and if resource not actually created then retry or 
- Let the user do so by offering some error handling on resources like `new docker.RemoteImage()` or
- Fail but report meaningful error message like: `the underlying provider has returned empty response for resource creation` etc. instead of:

> docker:index:RemoteImage xxxx creating replacement [diff: ~pullTriggers]; error: expected non-nil error with nil state during Create of urn:pulumi:***::xxx::xxx:group$xxx:xxx$docker:index/remoteImage:RemoteImage::xxxx

The above error goes away with a simple re-run without any changes but it happens intermittently


mikhailshilkov:
Closing as stale...

PR #406, opened 2021-10-29
Fix #8306

t0yv0:
Context: 

https://github.com/pulumi/pulumi/issues/8306

This PR root-causes and makes the following pulumi-aws  compilation errors disappear:

```
error: failed to convert HCL for #/functions/aws:index/getAvailabilityZone:getAvailabilityZone to typescript: cannot bind expression: <anonymous>:0,61-62: Missing attribute separator; Expected a newline or comma to mark the beginning of the next attribute.
error: failed to convert HCL for #/functions/aws:index/getAvailabilityZone:getAvailabilityZone to python: cannot bind expression: <anonymous>:0,61-62: Missing attribute separator; Expected a newline or comma to mark the beginning of the next attribute.
error: failed to convert HCL for #/functions/aws:index/getAvailabilityZone:getAvailabilityZone to csharp: cannot bind expression: <anonymous>:0,61-62: Missing attribute separator; Expected a newline or comma to mark the beginning of the next attribute.
error: failed to convert HCL for #/functions/aws:index/getAvailabilityZone:getAvailabilityZone to go: cannot bind expression: <anonymous>:0,61-62: Missing attribute separator; Expected a newline or comma to mark the beginning of the next attribute.
``` 

The other part of 8306 is still remaining.


t0yv0:
The last two commits workaround the rest of the issue with loading plugins named ''.

t0yv0:
Something is busted in the new tests on Windows only, with poor visibility. Very unfortunate:

```
=== RUN   TestConvert
=== RUN   TestConvert/regress_cannot_bind_expression
=== CONT  TestConvert
    convert_test.go:94: fixHcl left intact: 
                          variable "region_number" {
                            default = {
                              us-east-1 = 1
                            }
                          }
        
                          resource "aws_vpc" "example" {
                            cidr_block = cidrsubnet("10.0.0.0/8", 4, 2)
                          }
--- FAIL: TestConvert (0.00s)
    --- FAIL: TestConvert/regress_cannot_bind_expression (0.00s)
panic: fatal: An assertion has failed [recovered]
	panic: fatal: An assertion has failed
```

t0yv0:
Deferring in https://github.com/pulumi/pulumi-terraform-bridge/issues/408

pkg/tf2pulumi/convert/model_helpers.go
```diff
@@ -0,0 +1,81 @@
...
+//        default = 1
+//     }
+//
+// The type is decoded downstream (as in
+// `github.com/pulumi/pulumi/pkg/codegen/pcl/binder.go`) in PCL
+// processing. Currently we cannot guarantee that all posible
+// `model.Type` values survive the string encoding and subsequent
+// decoding. Thereore this function mimicks downstream decoding to see
+// if the value turns around, and aborts processing if that is not the
+// case.
```
pgavlin, line 45: Just to clarify: if we remove the explicit type from this example, is the issue that the type of `default` is `const(1)` and the binder then infers the type of `foo` to be `const(1)`? If so, it seems like we ought to fix this in the PCL binder s.t. the type of `foo` is inferred to be the const type's underlying type.

t0yv0: @pgavlin can we discuss the surrounding code here?

In a sense, you are right, there is an issue with inferring Const types for a variable from a `default` value.  These can (and are) nested. In the failing example it's nested under `ObjectType`:

```
	constType := model.NewConstType(model.IntType, cty.NumberIntVal(1))
	checkError("constType", constType)

	objTypeWithConst := model.NewObjectType(map[string]model.Type{"foo": constType})
	checkError("objectTypeWithConstType", objTypeWithConst)
```

We could insert a transformation `func unConstify(t model.Type) model.Type` and use it in here, and it *may* fix the manifested issue. Would it be possible to backlog it and check the fix as is as it is on the critical path? I'd like to take the time on that to fully understand the implications and recur through the model.Type grammar.

Here are some of the things that are quite surprising to me that make me tread really cautiously:

1. Why does the system rely on turning model.Type into a string and then parsing it back? We know it does not work if model.Type tree includes a Const node. Will it work for all possible cases if we remove Const nodes? Can we test that? I do not yet trust that will work, therefore aggressively checking that it will work or backing off to Dynamic type seems warranted.

2. How does this even work, why does `model.BindExpressionText` parse a string formed by taking `model.Type` through a `Sprintf("%v")` into an expression with a type equivalent to the original type? I am looking at this doc string `// BindExpressionText parses and binds an HCL2 expression using the given scope.` and it seems extremely surprising. Is the type expression grammar embedded in the term expression grammar? I think there is a bit of a documentation gap here at the least, we can correct.

LMK what you think, I still lean to merging as is, backlogging type inference from default value to remove consts, then also backlogging testing and documenting model.Type turnaround through a string and removing the model_helpers code once we can trust that to work.

t0yv0: I'm doing a little exploration here, I think I recovered the grammar of model.Type so I can get some answers to my own questions shortly.

t0yv0: https://github.com/pulumi/pulumi-terraform-bridge/pull/410 - not only Const types do not turnaround; there's a lot more to it.... The current aggressive check should stay in I think for now.  Let me explore deconstifying the types.

t0yv0:
OK @pgavlin PTAL I'm adding some functions to generalize the ConstType. It's a bit hard to tease it out in a good unit test, but the good news I've confirmed is that the constType generalization alone solves the underlying concrete problem we've been having. The bad news in #410 is that something is still quite fishy so I suggest to leave the defensive code in for now until there is some clarity on how that's guaranteed to work in general.

pkg/tfgen/convert_test.go
```diff
@@ -0,0 +1,137 @@
...
+                    default = {
+                      us-east-1 = 1
+                    }
+                  }`)
+
+		expectedCode := `
+import * as pulumi from "@pulumi/pulumi";
+
+const config = new pulumi.Config();
+const regionNumber = config.get("regionNumber") || {
```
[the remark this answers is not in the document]
t0yv0: Filed https://github.com/pulumi/pulumi/issues/8348 - this is problematic but does not look like it's frequent

t0yv0:
Also note: the type generalizes as having known keys (`object { "us-east-1": ... }`) but we could instead try generalizing this to `map[string]...`.

pgavlin (approved):

PR #413, opened 2021-11-23
Add debugging information to panic output

stack72:
when the docs from the tfdocs tool can't be parsed, we paniced and
tfgen didn't tell us what resource paniced. We now output the resource
name and the section of the docs that can't be parsed before the
panic occurs


t0yv0 (approved):
This is good! You can also cram this message into the panic itself, like `panic(fmt.Errorf("error: Failure ... : %w", err))` to be sure - but if this works and you can read the message it should be good to go.

stack72:
> This is good! You can also cram this message into the panic itself, like `panic(fmt.Errorf("error: Failure ... : %w", err))` to be sure - but if this works and you can read the message it should be good to go.

I will update!

PR #415, opened 2021-12-02
Adding the new docs variants for section titles

stack72:


pkg/tfgen/docs.go
```diff
@@ -598,9 +598,9 @@ func (p *tfMarkdownParser) parseSection(section []string) error {
 		return nil
 	case "Example Usage":
 		sectionKind = sectionExampleUsage
-	case "Arguments Reference", "Argument Reference", "Nested Blocks", "Nested blocks":
+	case "Arguments Reference", "Argument Reference", "Argument reference", "Nested Blocks", "Nested blocks":
```
Frassle, line 601: Maybe header should just be lower/uppercased before being switched on?

Frassle (approved):

pkg/tfgen/docs.go
stack72: This is ALL being replaced! This is just a quick fix for now to unblock an issue with a provider

PR #435, opened 2022-01-21
feat: Support omitting properties

jen20:
Currently there is no way to not generate Pulumi bindings for a `DataSource` or a `Resource` for which a property in the underlying Terraform provider exists.

This commit adds a new `Omit` flag in the `SchemaInfo`. When `Omit` is enabled, the property is dropped from the schema and no bindings will be generated for it, provided the property is not required (which would render the provider unusable from the generated bindings).

Supercedes #392.


stack72 (approved):
Thanks for fixing #392

PR #437, opened 2022-01-26
External language hooks for pulumi-terraform-bridge

t0yv0:
These changes aim to support more languages than the standard set without taking a build-dependency in pulumi-terraform-bridge. For example I am able to do this:

```
pulumi-tfgen-random external:pascal --out sdk/pascal
```

And it will interop with `pulumi-tfgen-external-language-pascal` binary I can provide at the call site to have the right codegen backend. 

The programgen (docs, examples) is not yet supported for these.

I can proceed with experimental development without this being merged. Also marking WIP for now until I have evidence it's working end to end.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1116,8 +1116,12 @@ func (g *Generator) convertHCL(hcl, path string) (string, string, error) {
 		}
 
 		files, diags, err := convert.Convert(convert.Options{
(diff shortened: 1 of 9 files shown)


stack72:
@t0yv0 i think we can close this out now based on the new direction we’ve gone right?

t0yv0:
Totally.

PR #453, opened 2022-02-16
Elide empty inputs from Read

Frassle:
Currently when we issue a refresh to terraform it sets properties to a zero value if they were reported unset by the underlying provider. This results in a lot of properties with “”, false, or 0 values which should have just been unset. 

This commit uses the schema reported back to strip out these empty properties. This should result in much better Read results from the tfbridge providers.

### DIFF
--- a/internal/testprovider/schema.go
+++ b/internal/testprovider/schema.go
@@ -45,6 +45,7 @@ func ProviderV1() *schemav1.Provider {
 		ResourcesMap: map[string]*schemav1.Resource{
 			"example_resource": {
 				Schema: map[string]*schemav1.Schema{
+					"opt_bool":              {Type: schemav1.TypeBool, Optional: true, Default: true},
 					"nil_property_value":    {Type: schemav1.TypeMap, Optional: true},
 					"bool_property_value":   {Type: schemav1.TypeBool, Optional: true},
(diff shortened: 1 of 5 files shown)


jkodroff:
Assigning to @stack72 for review once he returns since I'm not familiar enough with tfbridge to give a meaningful review.

mikhailshilkov:
We can't wait for @stack72 if we want to meet the Import delivery on March 9... Hoping @pgavlin can help here.

stack72:
I'm happy with the code here - I'd really like to see if there's a short video of a smoke test in the following scenario:

## no provider CRUD functionality changes 
* create new s3 bucket with the latest released version of Pulumi-aws
* build a new version of the aws provider with this latest bridge
* "upgrade" to this new functionality and ensure that the behaviour still works as expected

Frassle:
@stack72 I can run that test. Will get back with an update in a bit.

pgavlin (approved):
LGTM.

@stack72  this should be scoped specifically to `Read` for import. `refresh` takes a different path in `extractInputsFromOutputs`, and this code does not run in other operations.

stack72:
@pgavlin "should be scoped" this is why I wanted to see the test - we really should have these tests before this PR was merged

PR #457, opened 2022-02-22
Add a language PCL that produces examples as PCL

iwahbe:
Add an additional language target: pcl. We use this to extract pcl from provider examples. The use case here is testing programgen against a larger corpus of examples.

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -767,7 +767,7 @@ func (g *Generator) Generate() error {
 		}
 		files = map[string][]byte{}
 		for path, code := range g.convertedCode {
-			path = strings.TrimPrefix(path, "#/") + ".pcl"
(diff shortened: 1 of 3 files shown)


pkg/tfgen/generate.go
```diff
@@ -79,11 +81,12 @@ const (
 	Python Language = "python"
 	CSharp Language = "dotnet"
 	Schema Language = "schema"
+	PCL    Language = "pulumi"
```
jkodroff, line 84: Why not `pcl`?

iwahbe: To be consistent with https://github.com/pulumi/pulumi-terraform-bridge/blob/82448a9d29c68e2831c64f5525ca00e5e00c7a29/pkg/tf2pulumi/convert/convert.go#L40

jkodroff (approved):

PR #464, opened 2022-03-01
Recurse into secrets when stripping defaults

Frassle:


pkg/tfbridge/schema.go
```diff
@@ -1103,14 +1103,27 @@ func getDefaultValue(tfs shim.Schema, ps *SchemaInfo) interface{} {
...
+		if v.IsNumber() {
+			if floatValue, ok := dv.(float64); ok {
+				return floatValue == v.V
+			} else if intValue, ok := dv.(int); ok {
+				return float64(intValue) == v.V
+			} else {
+				// Unexpected type, just return false
+				return false
+			}
+		}
```
pgavlin, line 1126: I think we could simplify this to just check for `int`.

```suggestion
		if i, ok := dv.(int); ok {
			dv = float64(i)
		}
```

pkg/tfbridge/schema.go
```diff
@@ -1103,14 +1103,27 @@ func getDefaultValue(tfs shim.Schema, ps *SchemaInfo) interface{} {
 	if dv, _ := tfs.DefaultValue(); dv != nil {
 		return dv
 	}
-	if ps != nil {
-		return ps.Default
-	}
+	// TODO: We should inspect SchemaInfo.Default for the default value as well
+	// if ps != nil {
+	// 	return ps.Default
+	// }
```
pgavlin, line 1109: Curious about this one. Why was this commented out?

Frassle: `SchemaInfo.Default` is a `*DefaultInfo`. Not the actual default value. I'm sure there's a way to turn a `DefaultInfo` into the value given the info on it:
```
// DefaultInfo lets fields get default values at runtime, before they are even passed to Terraform.
type DefaultInfo struct {
	// AutoNamed is true if this default represents an autogenerated name.
	AutoNamed bool
	// Config uses a configuration variable from this package as the default value.
	Config string
	// From applies a transformation from other resource properties.
	From func(res *PulumiResource) (interface{}, error)
	// Value injects a raw literal value as the default.
	Value interface{}
	// EnvVars to use for defaults. If none of these variables have values at runtime, the value of `Value` (if any)
	// will be used as the default.
	EnvVars []string
}
```

But that's a bigger change than just trying to fix secrets and integers. So comment out and TODO.

pgavlin: > `SchemaInfo.Default` is a `*DefaultInfo`. Not the actual default value. 

:facepalm:

pgavlin (approved):

PR #466, opened 2022-03-02
Change unparseable h3 from warning to debug.

### LINKED ISSUE #448: Ignored documentation sections should be a debug-level output
We deliberately ignore certain sections of upstream docs because they are not applicable to Pulumi, e.g. timeouts. The count of ignored sections currently prints as a warning. This should be a debug-level output because this is normal behavior that we don't need to let the user know about unless they're specifically debugging behavior in the bridge.

Update: Not entirely sure what messages I was referring to above when I wrote this, but we should change "unparseable h3" from a warn to a debug because it's not actionable and nothing may be wrong anyway - it may be a peculiarity of how we parse sections. Have yet to see a case where doc locations should be overridden per the current error message.

jkodroff:
Fixes #448 

@pgavlin Does this jibe with your understanding of "unparseable h3"? It seems consistent based on a few hours of debugging.


pgavlin (approved):
Yeah, that comment is helpful 👍

PR #468, opened 2022-03-04
Optionally fail tfgen for an extra field mapping

### LINKED ISSUE #452: Fail tfgen for superfluous "custom resource schema"
The "custom resource schema ... was not present" message means that an overlay is being applied to an upstream resource that does not exist. This should fail the build when `PULUMI_EXTRA_MAPPING_ERROR` (or whatever that env var is called) is truthy.

jkodroff:
Fixes #452 

This change expands PULUMI_EXTRA_MAPPING_ERROR to also cover
superfluous mappings on fields, not just resources/data sources. We've
also improved the error message to use the TF identifier and improved
working that more directly describes the issue, its effects, and how to
remedy.

Sample output when set:

```
error: failed to gather package metadata: problem gathering resources: 1 error occurred:
        * there is a custom mapping on resource 'google_cloudfunctions_function' for field 'foo', but the field was not found in the Terraform metadata and will be ignored. Remove the mapping.
```

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -1068,7 +1068,13 @@ func (g *Generator) gatherResource(rawname string,
 	// Ensure there weren't any custom fields that were unrecognized.
 	for key := range info.Fields {
 		if _, has := schema.Schema().GetOk(key); !has {
-			g.warn("custom resource schema %s.%s was not present in the Terraform metadata", name, key)
+			msg := fmt.Sprintf("there is a custom mapping on resource '%s' for field '%s', but the field was not found in the Terraform metadata and will be ignored. To fix, remove the mapping.", rawname, key)
+
+			if isTruthy(os.Getenv("PULUMI_EXTRA_MAPPING_ERROR")) {
+				return "", nil, fmt.Errorf(msg)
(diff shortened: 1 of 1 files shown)


guineveresaenger (approved):

pkg/tfgen/generate.go
```diff
@@ -1068,7 +1068,13 @@ func (g *Generator) gatherResource(rawname string,
 	// Ensure there weren't any custom fields that were unrecognized.
 	for key := range info.Fields {
 		if _, has := schema.Schema().GetOk(key); !has {
-			g.warn("custom resource schema %s.%s was not present in the Terraform metadata", name, key)
+			msg := fmt.Sprintf("there is a custom mapping on resource '%s' for field '%s', but the field was not found in the Terraform metadata and will be ignored. Remove the mapping.", rawname, key)
```
guineveresaenger, line 1071: Actually, can we say "please remove the mapping"? or "To fix, remove the mapping"?

guineveresaenger: the first time I read it, I thought it said "removing the mapping" and missed that it was an instruction

PR #469, opened 2022-03-04
Clarify warning for unexpected code snippets.

### LINKED ISSUE #449: Improve messaging for "unexpected code snippets"
We get the warning "unexpected code snippets" but do not provide any context as to the offending text. Sometimes these are harmless, e.g. when they indicate a `literal`. Sometimes they should be fixed because they contain HCL syntax, e.g. `some_resource.attribute`.

* We should indicate that we are passing the through the text and not truncating it.
* We should show the offending text if at all possible so the user can take action if it's HCL (by updating our fork).
* This should be an INFO level output because we can't easily tell between the harmless case (`SOME_LITERAL`) and the warning case that should be corrected (`aws_some_resource.attribute`).

jkodroff:
Fixes #449

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -216,7 +216,7 @@ func getMarkdownDetails(org string, provider string, resourcePrefix string, kind
 	return markdownBytes, markdownFileName, true
 }
 
-func (k DocKind) string() string {
+func (k DocKind) String() string {
 	switch k {
 	case DataSourceDocs:
 		return "data source"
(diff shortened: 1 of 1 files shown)


pkg/tfgen/docs.go
```diff
@@ -216,7 +216,7 @@ func getMarkdownDetails(org string, provider string, resourcePrefix string, kind
 	return markdownBytes, markdownFileName, true
 }
 
-func (k DocKind) string() string {
+func (k DocKind) String() string {
```
guineveresaenger, line 219: yisss

guineveresaenger (approved):

PR #480, opened 2022-03-11
Remove unused hclFailures var

jkodroff:


stack72 (approved):

pkg/tfgen/docs.go
```diff
@@ -1213,19 +1210,6 @@ func (g *Generator) convertHCL(hcl, path, exampleTitle string) (string, string,
 			return fmt.Errorf("failed to convert HCL for %s to %v: %w", path, languageName, err)
 		}
 		if diags.All.HasErrors() {
-			if stderr.Len() != 0 {
```
stack72, line 1216: We really don't need these lines?

jkodroff: We used to return `stderr.String()` in addition to the converted HCL and an error, and we would store the value in `hclFailures`, but nothing ever read from `hclFailures`. I believe removing `stderr.String()` as a return value means that writing anything to `stderr` is pointless.

Can you confirm my logic here?

jkodroff: Confirmed that this change does not change the output nor schema.json for the GCP provider.

jkodroff:
The Windows build is failing because of a network error. Given the small scope of this change, making a judgment call that it's safe to merge.

PR #486, opened 2022-03-16
Change HCL conversion for schema to use a map

jkodroff:
When converting HCL to generate the schema, use a map internally and
*then* convert to a string of Markdown so that we can inject examples in
a future commit.

Verified both the output from the bridge and the schema are identical for the GCP provider.


stack72:
The fact that we have no generated code changes is great so LGTM

stack72 (approved):

Issue #458, opened 2022-02-23
make the use of replaceOnChanges consistent in our providers


susanev:
the user experience team did the design work to add an indicator to the API docs in registry to indicate which resource properties cause replacement. the replaceOnChanges property can be used for this, but our providers do not consistently use it.

related issues:
https://github.com/pulumi/registry/issues/518
https://github.com/pulumi/pulumi/issues/9017

susanev:
closing this out, as this approach was not agreed upon.
see [internal slack convo](https://pulumi.slack.com/archives/C02VCJEBT2N/p1647886944621689)

Issue #337, opened 2021-04-01
Show in docs what properties will cause a resource to be replaced


pierskarsenbarg:
Would be useful if we can show in our docs if changing a property cause a resource to be replaced. Can use ForceNew property

PR #489, opened 2022-03-21
Ensure replaceOnChange is set on schema

stack72:
Fixes: #458
Fixes: #337

when a parameter will force a new resource, it is now marked in
the schema as ReplaceOnChange: true

You can see the generated code this changes by clicking on the name of the provider link

### DIFF
--- a/pkg/tfgen/generate_schema.go
+++ b/pkg/tfgen/generate_schema.go
@@ -425,6 +425,7 @@ func (g *schemaGenerator) genProperty(mod string, prop *variable, pyMapCase bool
 		DeprecationMessage: prop.deprecationMessage(),
 		Language:           language,
 		Secret:             secret,
+		ReplaceOnChanges:   prop.forceNew(),
 	}
 }
 
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -494,6 +494,21 @@ func (v *variable) deprecationMessage() string {
 	return ""
 }
 
+func (v *variable) forceNew() bool {
+	// Output properties don't forceNew so we can return false by default
+	if v.out {
+		return false
+	}
+
+	// if we have an explicit marked as ForceNew then let's return that as that overrides
+	// the TF schema
+	if v.info != nil && v.info.ForceNew != nil {
+		return *v.info.ForceNew
+	}
+
+	return v.schema != nil && v.schema.ForceNew()
+}
+
 // optional checks whether the given property is optional, either due to Terraform or an overlay.
 func (v *variable) optional() bool {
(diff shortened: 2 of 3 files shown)


pkg/tfgen/generate_schema.go
```diff
@@ -417,6 +417,11 @@ func (g *schemaGenerator) genProperty(mod string, prop *variable, pyMapCase bool
 		secret = *prop.info.Secret
 	}
 
+	var replaceOnChange bool
+	if !prop.optional() {
```
Frassle, line 421: Really! `optional` tells you this? That's really confusing naming.

stack72: this is a helper func that checks to see if there is a TF based ForceNew OR a Pulumi override for ForceNew... I agree that it's confusing but that's the func

lukehoban: :thinking: This doesn't seem right.  Why would a require property by definition be replace on changes?

stack72: d'oh! You are correct.... fixing now - thank you both

lukehoban: But also - why are we even trying to automatically emit relaceOnChanges?  That is meant to be a way for users to override the provider’s default behaviour, which will be picked up automatically as part of normal provider behaviour.  I’m pretty nervous about using it in this way - feels like a nontrivial fundamental change.

Frassle (approved):
Looks right to me.

Looking at the diffs I did notice a lot of the properties have "Changing this field forces a new resource to be created." explicitly in their description. Not sure where that text is coming from but maybe it could be cleaned up later to instead just generate from the replaceOnChanges flag.

stack72:
> Looks right to me.
> 
> Looking at the diffs I did notice a lot of the properties have "Changing this field forces a new resource to be created." explicitly in their description. Not sure where that text is coming from but maybe it could be cleaned up later to instead just generate from the replaceOnChanges flag.

That comes from the upstream TF provider - we just grab those docs by default - they are not calculated based on any individual properties

pkg/tfgen/generate_schema.go
Frassle: I think this started because @susanev wanted a way to query schemas to know what properties would cause a replace on change, rather than it just being text in "description" (https://github.com/pulumi/pulumi-terraform-bridge/issues/458). And I guess some wires have got crossed and confused about what relaceOnChanges in the schema means.

I would of thought that relaceOnChanges being an output of the provider would be a description like what Susan is after, but your comment suggests that it's an input to the provider to override it's behavior?

PR #491, opened 2022-03-28
Add the ability to define additional HCL examples

### LINKED ISSUE #364: Overriding TF Provider Resource Examples
We need a way of being able to inject Pulumi curated examples for a specific resource into the provider resources.go. This will give us the ability to create examples where no terraform examples exist or that examples do not play nicely with our codegen

Currently, our resources.go allow us to do the following:

```
			"azurerm_application_insights_web_test": {
				Tok: azureResource(azureAppInsights, "WebTest"),
				Docs: &tfbridge.DocInfo{
					Source: "application_insights_webtests.html.markdown",
				},
			},
```

This means we have some level of control over the documentation but we don't have a way of being able to pass in a set of examples. Our tfgen process currently requires these examples to be written in hcl so that they can be converted to all languages. 

-> Question for @infin8x here - do we want to continue this being the state of the world or do we need to allow these docs to be in another format that is predefined?

By allow this resource example override, we are then in a situation where we can accept pull requests to add new and more feature rich examples to our documentation

### LINKED ISSUE #434: Add the ability to inject examples
Some commonly used resources lack examples in all languages, e.g. [lambda.Function](https://www.pulumi.com/registry/packages/aws/api-docs/lambda/function/) for Go in the AWS Classic provider.

We want to add the ability to inject examples into the docs for both missing use cases and use cases which may not be documented in the upstream provider.

Be sure to add a suggested remedy to the error message `no resource plugin 'foo' found in the workspace or on your $PATH`. Prior to adding the ability to inject examples, there was no way to remedy this issue.

jkodroff:
Adds ExtraFunctionHclExamples and ExtraResourceHclExamples to
tfbridge.ProviderInfo to allow users to add supplemental HCL examples to
resources/functions. Examples will be inserted into resource/function
descriptions and translated like any original source content.

Sample usage:

```go
		ExtraResourceHclExamples: []tfbridge.HclExampler{
			tfbridge.LocalFileHclExample{
				ID:           "gcp:memcache/instance:Instance",
				RelativePath: "docs/test.hcl",
			},
		},
```

Fixes #434
Fixes #364

### DIFF
--- a/pkg/tfbridge/examples.go
+++ b/pkg/tfbridge/examples.go
@@ -1,0 +1,101 @@
+package tfbridge
+
+import (
+	"bytes"
+	"io/ioutil"
+	"path/filepath"
+	"text/template"
+)
+
+// HclExampler represents a supplemental HCL example for a given resource or function.
+type HclExampler interface {
+	// GetPulumiIdentifier returns the fully qualified path to the resource or function in the schema, e.g.
+	// "provider:module/getFoo:getFoo" (function), or
+	// "provider:module/bar:Bar" (resource)
+	GetPulumiIdentifier() string
+	// GetMarkdown returns the Markdown that comprises the entire example, including the header.
+	//
+	// Headers should be an H3 ("###") and the header content should not contain any prefix, e.g. "Foo with Bar" not,
+	// "Example Usage - Foo with Bar".
+	//
+	// Code should be surrounded with code fences with an indicator of the language on the opening fence, e.g. "```hcl".
+	GetMarkdown() (string, error)
+}
+
+// LocalFileHclExample represents a supplemental HCL example that is on a relative path within the Pulumi provider repo.
+type LocalFileHclExample struct {
+	// ID is the Pulumi identifier for the resource or function to which the example pertains, e.g.
+	// "provider:module/getSomething:getSomething" (function), or
+	// "provider:module/something:Something" (resource)
+	ID string
+	// Title is title of the example, e.g. "Basic Something", "Advanced Something with Something Else".
+	Title string
+	// RelativePath is the path to the file in the Pulumi repo relative to the repo root.
+	RelativePath string
+}
+
+func (e LocalFileHclExample) GetPulumiIdentifier() string {
+	return e.ID
+}
+
+func (e LocalFileHclExample) GetMarkdown() (string, error) {
+	absPath, err := filepath.Abs(e.RelativePath)
+	if err != nil {
+		return "", err
+	}
+
+	fileBytes, err := ioutil.ReadFile(absPath)
+	if err != nil {
+		return "", err
+	}
+
+	return renderTemplate(e.Title, string(fileBytes))
+}
+
+// InlineHclExample represents a literal HCL example and is primarily used for testing the Bridge.
+type InlineHclExample struct {
+	// ID is the Pulumi identifier for the resource or function to which the example pertains, e.g.
+	// "provider:module/getSomething:getSomething" (function), or
+	// "provider:module/something:Something" (resource)
+	ID string
+	// Title is title of the example, e.g. "Basic Something", "Advanced Something with Something Else".
+	Title string
+	// Contents is the HCL that comprises the example with no surrounding Markdown constructs (e.g. ```hcl`).
+	Contents string
+}
+
+func (e InlineHclExample) GetPulumiIdentifier() string {
+	return e.ID
+}
+
+func (e InlineHclExample) GetMarkdown() (string, error) {
+	return renderTemplate(e.Title, e.Contents)
+}
+
+func renderTemplate(title, contents string) (string, error) {
+	tmpl := `### {{ .Title }}
+
+{{ .CodeFences }}hcl
+{{ .Contents }}
+{{ .CodeFences }}
+`
+
+	outputTemplate, _ := template.New("dummy").Parse(tmpl)
+	data := struct {
+		CodeFences string
+		Title      string
+		Contents   string
+	}{
+		CodeFences: "```",
+		Title:      title,
+		Contents:   contents,
+	}
+
+	var buf = bytes.Buffer{}
+	err := outputTemplate.Execute(&buf, data)
+	if err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
+}
--- a/pkg/tfbridge/examples.go
+++ b/pkg/tfbridge/examples.go
@@ -9,10 +9,10 @@ import (
 
 // HclExampler represents a supplemental HCL example for a given resource or function.
 type HclExampler interface {
-	// GetPulumiIdentifier returns the fully qualified path to the resource or function in the schema, e.g.
+	// GetToken returns the fully qualified path to the resource or function in the schema, e.g.
 	// "provider:module/getFoo:getFoo" (function), or
 	// "provider:module/bar:Bar" (resource)
-	GetPulumiIdentifier() string
+	GetToken() string
 	// GetMarkdown returns the Markdown that comprises the entire example, including the header.
 	//
 	// Headers should be an H3 ("###") and the header content should not contain any prefix, e.g. "Foo with Bar" not,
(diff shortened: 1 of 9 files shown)


pkg/tfbridge/examples.go
```diff
@@ -0,0 +1,101 @@
...
+	//
+	// Code should be surrounded with code fences with an indicator of the language on the opening fence, e.g. "```hcl".
+	GetMarkdown() (string, error)
+}
+
+// LocalFileHclExample represents a supplemental HCL example that is on a relative path within the Pulumi provider repo.
+type LocalFileHclExample struct {
+	// ID is the Pulumi identifier for the resource or function to which the example pertains, e.g.
+	// "provider:module/getSomething:getSomething" (function), or
+	// "provider:module/something:Something" (resource)
```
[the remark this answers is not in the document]
guineveresaenger: that said, apparently the official name for this is "Type token" or "type name"

per [Slack comment](https://pulumi.slack.com/archives/C011EMDQ8JE/p1648569543546689?thread_ts=1648569270.180049&cid=C011EMDQ8JE)

pkg/tfbridge/examples_test.go
```diff
@@ -0,0 +1,33 @@
...
+	example := LocalFileHclExample{
+		Title:        "File Title",
+		RelativePath: "examples_test.hcl",
+	}
+
+	actual, err := example.GetMarkdown()
+
+	assert.Equal(t, expected, actual)
+	assert.Nil(t, err)
+}
```
guineveresaenger, line 33: 🎉 for this entire file

pkg/tfbridge/info.go
```diff
@@ -246,7 +252,7 @@ type ConfigInfo struct {
 // the raw string is still accepted as a possible input value.
 type Transformer func(resource.PropertyValue) (resource.PropertyValue, error)
 
-// DocInfo contains optional overrids for finding and mapping TD docs.
+// DocInfo contains optional overrides for finding and mapping TF docs.
```
guineveresaenger, line 255: bless you 🙇‍♀️

pkg/tfgen/docs_test.go
```diff
@@ -504,3 +505,36 @@ FooFactory fooFactory = new FooFactory();
 
 	assert.Equal(t, buf.String(), hclConversionsToString(input))
 }
+
+func TestGroupLines(t *testing.T) {
```
guineveresaenger, line 509: love this but unsure if it needs to be in this PR?

pkg/tfgen/generate_schema_test.go
```diff
@@ -0,0 +1,167 @@
...
+## Import
+
+import content
+`
+
+	assert.Equal(t, renderTemplate(expectedTmpl), appendExample(input, renderTemplate(markdownTmpl)))
+}
+
+/// renderTemplate allows us to easily use code fences with herestrings
+func renderTemplate(tmpl string) string {
```
guineveresaenger, line 155: this is just a helper function for the tests, right? could we name it renderTestTemplate?

pkg/tfgen/generate_schema.go
```diff
@@ -758,3 +759,92 @@ func (g *Generator) convertExamplesInSchema(spec pschema.PackageSpec) pschema.Pa
...
+	}
+
+	return err
+}
+
+func addExtraHclExamplesToFunctions(extraExamples []tfbridge.HclExampler, spec *pschema.PackageSpec) error {
+	var err error
+	for _, ex := range extraExamples {
+		token := ex.GetPulumiIdentifier()
+		fun, ok := spec.Functions[token]
```
guineveresaenger, line 791: can we say `func` here? Also trying and failing to avoid making obnoxious dad jokes about how we at Pulumi are making our own fun

pkg/tfgen/generate_schema.go
```diff
@@ -758,3 +759,92 @@ func (g *Generator) convertExamplesInSchema(spec pschema.PackageSpec) pschema.Pa
...
+				sections[i] = append(section, strings.Split(markdownToAppend, "\n")...)
+				break
+			}
+		}
+	} else {
+		// If not, we need to add the header and append before the first ## in the doc, or EOF, whichever comes first
+		markdownToAppend = fmt.Sprintf("%s\n\n%s", exampleUsageHeader, markdownToAppend)
+
+		// If there's no blank line after the content, we need to add it to ensure we have semantically valid Markdown:
+		if sections[0][len(sections[0])-1] != "" {
```
guineveresaenger, line 838: could we use https://pkg.go.dev/strings#HasSuffix here? maybe I don't entirely grok the glob of text we're wrestling though!

guineveresaenger (changes requested):
Most of these lean into the clarify/standardize kind of feedback.

I love that you added tests.

Where in our infrastructure are we planning to inject the examples? I realize this is laying the groundwork, but I'm curious!

pkg/tfgen/docs_test.go
jkodroff: I had to understand how it worked in order to do this PR. It's equivalent to adding documentation.

pkg/tfgen/generate_schema.go
jkodroff: No - `func` is a keyword in Go.

pkg/tfgen/generate_schema.go
jkodroff: No, because the code here is dealing with slices that represent lines, not a single string.

pkg/tfgen/generate_schema_test.go
jkodroff: It's in an `_test.go` file, so it's not exported, so I don't think renaming would clarify anything.

pkg/tfbridge/examples.go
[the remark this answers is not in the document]
jkodroff: Done.

pkg/tfbridge/examples.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	return renderTemplate(e.Title, string(fileBytes))
+}
+
+// InlineHclExample represents a literal HCL example and is primarily used for testing the Bridge.
+type InlineHclExample struct {
+	// ID is the Pulumi identifier for the resource or function to which the example pertains, e.g.
+	// "provider:module/getSomething:getSomething" (function), or
+	// "provider:module/something:Something" (resource)
```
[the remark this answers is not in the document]
jkodroff: Fixed.

pkg/tfbridge/examples.go
```diff
@@ -0,0 +1,101 @@
...
+func renderTemplate(title, contents string) (string, error) {
+	tmpl := `### {{ .Title }}
+
+{{ .CodeFences }}hcl
+{{ .Contents }}
+{{ .CodeFences }}
+`
+
+	outputTemplate, _ := template.New("dummy").Parse(tmpl)
+	data := struct {
```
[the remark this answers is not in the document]
jkodroff: No, because it's not used anywhere but inside the function.

pkg/tfbridge/examples.go
```diff
@@ -0,0 +1,101 @@
...
+		CodeFences string
+		Title      string
+		Contents   string
+	}{
+		CodeFences: "```",
```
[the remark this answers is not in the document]
jkodroff: I would say no because I don't think this suggested change would improve the readability and we're only referencing the string in a single place.

jkodroff:
@guineveresaenger We'd add these into commonly used providers where examples upstream are either non-existent or cannot be translated into any language.

pkg/tfgen/generate_schema.go
guineveresaenger: oh right lol 🤦‍♀️ I still don't want this to say `fun` though!

stack72:
FWIW, I am a big fan of progressive enhancement here - we won't get this 100% correct upfront and may need to change a few things :) This is a massive first step - thank you @jkodroff

PR #492, opened 2022-03-29
Perform metaschema validation during sdk generation

viveklak:
In https://github.com/pulumi/pulumi/pull/8543 we disabled metaschema validation in most typical user-facing code-paths. But we should try to validate at code-generation.


jkodroff (changes requested):
Gonna need some code that compiles before I can perform a meaningful review. ;)

PR #494, opened 2022-04-01
Add NoUpstreamDocs()

jkodroff:
@stack72 Feel free to reject if you think this is a bad idea, but it's a use case that's come up more than a few times in different providers, and so rather than repeating the same literal and adding the same comment, I figured an upstream function will make it clearer to the reader and also easier to manipulate since it's a single symbol to reference.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -261,6 +261,15 @@ type DocInfo struct {
 	IncludeAttributesFromArguments string // optionally include attributes from another raw resource's arguments.
 }
 
+// NoUpstreamDocs is used to indicate that a resource has no upstream docs. It returns a pointer to DocInfo which
+// contains a simple whitespace string for the literal contents of the documentation. This value will allow tfgen
+// to pass when PULUMI_MISSING_DOCS_ERROR=true while clearly indicating the reason for the docs override.
+func NoUpstreamDocs() *DocInfo {
+	return &DocInfo{
+		Markdown: []byte(" "),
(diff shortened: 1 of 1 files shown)


pkg/tfbridge/info.go
```diff
@@ -261,6 +261,15 @@ type DocInfo struct {
 	IncludeAttributesFromArguments string // optionally include attributes from another raw resource's arguments.
 }
 
+// NoUpstreamDocs is used to indicate that a resource has no upstream docs. It returns a pointer to DocInfo which
+// contains a simple whitespace string for the literal contents of the documentation. This value will allow tfgen
+// to pass when PULUMI_MISSING_DOCS_ERROR=true while clearly indicating the reason for the docs override.
+func NoUpstreamDocs() *DocInfo {
+	return &DocInfo{
+		Markdown: []byte(" "),
```
stack72, line 269: why do we add this rather than skipping the docs generation?

jkodroff: There's no way that I'm aware of to keep `PULUMI_MISSING_DOCS_ERROR=true` (which I think we both agree is a good default), while acknowledging that a resource which has *no docs at all* (as opposed to, say, combined docs with another resource, which is relatively common).

Now, if you like it better, we could add a `Skip bool` option to `DocInfo`, but it's not as clean a design as I would like: I'd prefer that `DocInfo` was an interface and we could have an explicit type for the use case of no upstream docs, but this was the closest I could get.

PR #500, opened 2022-04-05
Add ability to override import instructions in providers

### LINKED ISSUE #495: Missing import documentation
## Hello!

- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

A user reported [missing information on how to import a resource](https://github.com/pulumi/pulumi-aws/pull/1867)

I don't believe this is the only such missing information, but I wasn't sure whether we support `import` for all resources, and if we have the ability to inject docs quite yet, and where we would do so (in the upstream fork? the docs bridge?)

Opening up this issue to track.

guineveresaenger:
* Add GetImportDetails to ResourceOrDataSourceInfo interface
* Implement method; default to empty string for DataSources
* Add check condition to parseImports to find and exit early if
  overwritten instructions exist

Resolves #495

Once rolled out to providers this will enable us to fix issues such as
https://github.com/pulumi/pulumi-aws/pull/1867.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -141,6 +141,7 @@ type ResourceOrDataSourceInfo interface {
 	GetTok() tokens.Token              // a type token to override the default; "" uses the default.
 	GetFields() map[string]*SchemaInfo // a map of custom field names; if a type is missing, uses the default.
 	GetDocs() *DocInfo                 // overrides for finding and mapping TF docs.
+	GetImportDetails() string          // Get overwritten instruction for importing a resource
 }
 
 // ResourceInfo is a top-level type exported by a provider.  This structure can override the type to generate.  It can
@@ -155,11 +156,13 @@ type ResourceInfo struct {
 	Aliases             []AliasInfo            // aliases for this resources, if any.
 	DeprecationMessage  string                 // message to use in deprecation warning
 	CSharpName          string                 // .NET-specific name
+	ImportDetails       string                 // Docs overwrite for import instructions
 }
 
 func (info *ResourceInfo) GetTok() tokens.Token              { return tokens.Token(info.Tok) }
 func (info *ResourceInfo) GetFields() map[string]*SchemaInfo { return info.Fields }
 func (info *ResourceInfo) GetDocs() *DocInfo                 { return info.Docs }
+func (info *ResourceInfo) GetImportDetails() string          { return info.ImportDetails }
 
 // DataSourceInfo can be used to override a data source's standard name mangling and argument/return information.
 type DataSourceInfo struct {
@@ -172,6 +175,7 @@ type DataSourceInfo struct {
 func (info *DataSourceInfo) GetTok() tokens.Token              { return tokens.Token(info.Tok) }
 func (info *DataSourceInfo) GetFields() map[string]*SchemaInfo { return info.Fields }
 func (info *DataSourceInfo) GetDocs() *DocInfo                 { return info.Docs }
+func (info *DataSourceInfo) GetImportDetails() string          { return "" }
 
 // SchemaInfo contains optional name transformations to apply.
 type SchemaInfo struct {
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/info.go
```diff
@@ -155,11 +156,13 @@ type ResourceInfo struct {
 	Aliases             []AliasInfo            // aliases for this resources, if any.
 	DeprecationMessage  string                 // message to use in deprecation warning
 	CSharpName          string                 // .NET-specific name
+	ImportDetails       string                 // Docs overwrite for import instructions
```
jkodroff, line 159: This should be a proper Golang doc string. Could also change the other properties in the struct.

pkg/tfbridge/info.go
```diff
@@ -141,6 +141,7 @@ type ResourceOrDataSourceInfo interface {
 	GetTok() tokens.Token              // a type token to override the default; "" uses the default.
 	GetFields() map[string]*SchemaInfo // a map of custom field names; if a type is missing, uses the default.
 	GetDocs() *DocInfo                 // overrides for finding and mapping TF docs.
+	GetImportDetails() string          // Get overwritten instruction for importing a resource
```
jkodroff, line 144: Should be a proper doc string. Could change other members of interface.

pkg/tfbridge/info.go
```diff
@@ -172,6 +175,7 @@ type DataSourceInfo struct {
 func (info *DataSourceInfo) GetTok() tokens.Token              { return tokens.Token(info.Tok) }
 func (info *DataSourceInfo) GetFields() map[string]*SchemaInfo { return info.Fields }
 func (info *DataSourceInfo) GetDocs() *DocInfo                 { return info.Docs }
+func (info *DataSourceInfo) GetImportDetails() string          { return "" }
```
jkodroff, line 178: I don't like we are adding a member to an interface which is not applicable to one of its implementations (because data sources do not have Import sections), but I understand we also have to be pragmatic and ship.

Could leave a comment indicating the design decision here.

pkg/tfgen/docs.go
```diff
@@ -796,6 +796,14 @@ func (p *tfMarkdownParser) parseAttributesReferenceSection(subsection []string)
 }
 
 func (p *tfMarkdownParser) parseImports(subsection []string) {
+	// check for import overwrites
+	importDetails := p.info.GetImportDetails()
+	if importDetails != "" {
+		var details []string
+		details = append(details, importDetails)
```
jkodroff, line 803: Not sure why this is necessary. Could you explain with a comment or remove if unnecessary?

jkodroff (changes requested):

jkodroff:
@guineveresaenger Since it's not feasible to integration test this in an automated fashion, could you verify that you tested with a Tier 1 provider and that it produced the expected output in the generated schema?

pkg/tfgen/docs.go
stack72: This is currently a []string so we need to build that list of []string

pkg/tfbridge/info.go
guineveresaenger: Did you have a certain format in mind?

pkg/tfbridge/info.go
guineveresaenger: Agreed - I don't love this either

pkg/tfgen/docs.go
jkodroff: So why not:

```go
details := []string{
  importDetails,
}
```

guineveresaenger: Yeah, we really don't need this, as we're just writing to `entityDocs.Import` which is a plain string

pkg/tfbridge/info.go
```diff
@@ -259,8 +269,12 @@ type DocInfo struct {
 	IncludeAttributesFrom          string // optionally include attributes from another raw resource for docs.
 	IncludeArgumentsFrom           string // optionally include arguments from another raw resource for docs.
 	IncludeAttributesFromArguments string // optionally include attributes from another raw resource's arguments.
+	ImportDetails                  string // Overwrite for import instructions
```
jkodroff, line 272: go doc could go here as well as functions.

jkodroff (approved):

PR #502, opened 2022-04-06
Fix Read not returning Secrets

Frassle:
Fixs https://github.com/pulumi/pulumi-terraform-bridge/issues/501

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -951,7 +951,10 @@ func (p *Provider) Read(ctx context.Context, req *pulumirpc.ReadRequest) (*pulum
 			return nil, err
 		}
 
-		mprops, err := plugin.MarshalProperties(props, plugin.MarshalOptions{Label: label + ".state"})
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/provider.go
```diff
@@ -951,7 +951,10 @@ func (p *Provider) Read(ctx context.Context, req *pulumirpc.ReadRequest) (*pulum
 			return nil, err
 		}
 
-		mprops, err := plugin.MarshalProperties(props, plugin.MarshalOptions{Label: label + ".state"})
+		mprops, err := plugin.MarshalProperties(props, plugin.MarshalOptions{
+			Label:       label + ".state",
+			KeepSecrets: p.supportsSecrets,
```
Frassle, line 956: This is the fix. Chances are every other instance of MarshalOptions ought to have this added as well?

stack72 (approved):
Verified locally - removing the keepSecrets makes the tests fail for the nested schema

Issue #504, opened 2022-04-11
Ensure Superfluous id fields are omitted as part of codegen


stack72:
The Terraform Plugin SDK adds an "id" to the TF schema when `d.SetID` is used. When the TF schema already has an "id" parameter defined, it ignores the extra id. BUT when Pulumi calls tfgen, we get a second id added to our code that causes a compilation failure in python

```
      SyntaxError: duplicate argument 'id' in function definition
      error: an unhandled error occurred: Program exited with non-zero exit code: 1
```

Currently, we will run a fork of the provider, change the schema and use the fork but this is not scalable

jkodroff:
Some affected provider issues:
- https://github.com/pulumi/pulumi-tls/issues/120
- https://github.com/pulumi/pulumi-rke/pull/28

PR #507, opened 2022-04-12
Ensure top level id parameters are not included in Pulumi schema

stack72:
Fixes: #504

Sometimes in Terraform, a superfluous "id" parameter is included
in the Terraform schema for a resource. This is ignored by terraform
but was being generated in the Pulumi schema as an additional id parameter

### DIFF
--- a/pkg/tfgen/generate_schema.go
+++ b/pkg/tfgen/generate_schema.go
@@ -225,7 +225,7 @@ func (g *schemaGenerator) genPackageSpec(pack *pkg) (pschema.PackageSpec, error)
 	for _, mod := range pack.modules.values() {
 		// Generate nested types.
 		for _, t := range gatherSchemaNestedTypesForModule(mod) {
-			tok, ts := g.genObjectType(mod.name, t)
+			tok, ts := g.genObjectType(mod.name, t, false)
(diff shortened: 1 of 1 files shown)


pkg/tfgen/generate_schema.go
```diff
@@ -520,11 +530,11 @@ func (g *schemaGenerator) genDatasourceFunc(mod string, fun *resourceFunc) psche
 
 	// If there are argument and/or return types, emit them.
 	if fun.argst != nil {
-		_, t := g.genObjectType(mod, &schemaNestedType{typ: fun.argst, pyMapCase: true})
+		_, t := g.genObjectType(mod, &schemaNestedType{typ: fun.argst, pyMapCase: true}, false)
```
Frassle, line 533: Can datasources have "id"?  I thought these would be true.

Frassle (approved):

pkg/tfgen/generate_schema.go
stack72: I just checked - data sources can have an Id - but they can sometimes be optional for lookups. I'll change data sources to remove top level ids that are computed only

stack72: For data sources, I omitted "id" outputs only - not inputs as that is a legit param

PR #509, opened 2022-04-20
Adding ability to allow setting a custom DotNet namespace for the package

stack72:


pkg/tfbridge/info.go
```diff
@@ -339,6 +339,7 @@ type CSharpInfo struct {
 	PackageReferences map[string]string // NuGet package reference information.
 	Overlay           *OverlayInfo      // optional overlay information for augmented code-generation.
 	Namespaces        map[string]string // Known .NET namespaces with proper capitalization.
+	RootNamespace     string            // The root namespace if setting to something other than Pulumi in the package name
```
jkodroff, line 342: Do struct attributes have to be commented in the same was as funcs do in order to generate proper docs? I'm confused on this.

stack72: [this is only for internal documentation only - we should look at documenting our Public API better](https://pkg.go.dev/github.com/pulumi/pulumi-terraform-bridge/v3@v3.20.0/pkg/tfbridge#CSharpInfo) No sir :)

Issue #503, opened 2022-04-07
Refresh Panics When Provider Not Properly Provisioned


gil-snyk:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

When trying to refresh an instance using the Terraform AWS provider, if the provider is under-provisioned (i.e. the credentials with which the provider was instantiated are not permissioned for the API call that refresh uses) then the refresh function panics:

`invalid memory address or nil pointer dereference`

Stacktrace (with less relevant elements removed...)
```goroutine 37 [running]:
runtime/debug.Stack()
	/usr/local/go/src/runtime/debug/stack.go:24 +0x65
...
	/usr/local/go/src/runtime/panic.go:1038 +0x215
github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/tfplugin5.(*provider).Refresh(0xc0095e6100, {0x15e10c6, 0x16}, {0x1908ec0, 0xc006c39a70})
	/Users/gil/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.6.0/pkg/tfshim/tfplugin5/provider.go:448 +0x27d
...```

Based on the stacktrace, it seems that the call to `ReadResource` is not returning an error, but rather a nil value which causes a panic when trying to unmarshal the value - `msgpack.Unmarshal(resp.NewState.Msgpack, resource.ctyType)`

It seems to me that the proper behavior would be to surface the cloud provider's API response so the end user can see what the underlying error is.  

### Steps to reproduce
<!-- Provide a link to a live example or an unambiguous set of steps to reproduce this bug. Tell us what you expected to happen, and what happened instead. Please include code to reproduce, if relevant. -->
Some pseudo-code to help understand...
1. Provision a Lambda function in AWS
2. Create an IAM role that does not have `lambda:GetFunction` permission
3. Using pulumi-terraform-bridge, create a Terraform AWS provider plugin 
* `provider, err := tfplugin5.StartProvider(...)`
4. Using pulumi-terraform-bridge, configure the provider using the credentials of the underprovisioned IAM role
* `provider.Configure(provider.NewResourceConfig(...<region and credentials>...))`
5. Using pulumi-terraform-bridge, create an instance state for the lambda function using its ID (the name of the function)
* `resourceType := provider.ResourcesMap().GetOk("aws_lambda_function")`
[...]

PR #512, opened 2022-04-21
Check For nil State and Return Error on Refresh

gil-snyk:
When refreshing a resource, the provider may return a `nil` value for `NewState` and the error will be in the `Diagnostics` field.  This PR accounts for this possibility and bubbles up the provider error rather than having the terraform bridge panic.

Relates to issue https://github.com/pulumi/pulumi-terraform-bridge/issues/503


stack72:
Thanks for this @gil-snyk 

Sorry this took a while to merge - this is something we will ensure we are better on! 

Paul

Issue #490, opened 2022-03-28
Added `__defaults` key after refresh


pgavlin:
A user reports that after a `pulumi refresh`, AWS resources have diffs that indicate that a `__defaults` key was added:

```
        aws:ec2:SecurityGroup sg-default  [diff: +__defaults]
        kubernetes:rbac.authorization.k8s.io/v1:Role admin
        aws:ec2:SecurityGroupRule sg-default-internal-ingress  [diff: +__defaults]
        kubernetes:rbac.authorization.k8s.io/v1:RoleBinding admin-ias
        aws:ec2:SecurityGroupRule sg-default-internal-egress  [diff: +__defaults]
        aws:ec2:SecurityGroupRule sg-cluster-dns-egress-udp  [diff: +__defaults]
        aws:ec2:SecurityGroupRule sg-cluster-dns-egress-tcp  [diff: +__defaults]
        aws:ec2:SecurityGroupRule sg-cluster-apiserver-ingress  [diff: +__defaults]
        kubernetes:vpcresources.k8s.aws/v1beta1:SecurityGroupPolicy sg-policy
```

This should be innocuous, but it is certainly unexpected.

pgavlin:
The user is using version 4.38.1 of the AWS provider.

phillipedwards:
Below is a simple go aws repro.

Pulumi details:
CLI: 3.26.1
AWS: 5.1.0
OS: ubuntu

```go
package main

import (
	"github.com/pulumi/pulumi-aws/sdk/v5/go/aws/secretsmanager"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {

		secret, err := secretsmanager.NewSecret(ctx, "my-secret", &secretsmanager.SecretArgs{
			Name: pulumi.String("my-string"),
		})

		if err != nil {
			return err
		}

		_, err = secretsmanager.NewSecretVersion(ctx, "version", &secretsmanager.SecretVersionArgs{
			SecretId:     secret.ID(),
			SecretString: pulumi.String("super-secret"),
		})

		if err != nil {
			return err
		}

		return nil
	})
}

```

![image](https://user-images.githubusercontent.com/25461821/162280549-cb55b8bc-fc93-499b-88f6-f3884ba97a27.png)

pgavlin:
I tried to repro this with the following TS equivalent and the same version of the AWS provider, but could not:

[...]

PR #514, opened 2022-04-29
[tfbridge] Retain the defaults list if present.

pgavlin:
The input extraction code that runs during a refresh has a bug that
causes it to drop the default list entirely if no properties were set
using defaults. Unfortunately, the semantics of a property bag that does
not contain a default list are different from those of a property bag
that contains an empty default list. These changes ensure that the list
is not dropped if it exsits, even if it is empty.

Fixes part of #490.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -1051,8 +1051,9 @@ func extractInputs(oldInput, newState resource.PropertyValue, tfs shim.Schema, p
 
 		// If we have a list of inputs that were populated by defaults, filter out any properties that changed and add
 		// the result to the new inputs.
-		defaultNames := map[string]bool{}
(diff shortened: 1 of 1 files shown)


pgavlin:
For a little bit of elaboration on why this fixes part of #490:

The root cause of #490 was the dropping of the `__defaults` list by the provider during a refresh. This difference was only visible during a `pulumi up -r` because the display logic incorrectly diffs _input_ values during the refresh portion of such an update rather than _state_ values as it does during a `pulumi refresh`. The restored `__defaults` list was then rendered during the `up` portion of the update.

pgavlin:
(the other part of #490 that needs fixing is the pulumi/pulumi issue that caused the diff to render in the first place)

stack72 (approved):
Great fine - especially as this was related to the up -r only!

PR #519, opened 2022-05-04
Readd support for different random char sets for autonames

Frassle:
This now uses the crypto entropy source like NewUniqueHex.


Frassle:
I'm going to see about adding this function to Pulumi as we've got to make a new `NewUnique` function anyway.

PR #520, opened 2022-05-05
Update to v2.13.0 of the TF plugin SDK

pgavlin:


pgavlin:
The failures in the downstream tests are expected. Those tests will not pass until the bridge has been rolled out and those providers have the proper `replace` directives for `terraform-plugin-sdk`.

jkodroff (approved):

PR #525, opened 2022-05-23
Adding support for Java options in the bridge to allow configuring Java SDK basePath

stack72:


pkg/tfbridge/info.go
```diff
@@ -343,6 +344,10 @@ type CSharpInfo struct {
 	RootNamespace     string            // The root namespace if setting to something other than Pulumi in the package name
 }
 
+type JavaInfo struct {
+	BasePackage string // the Base package for the Java SDK
```
t0yv0, line 348: Does this need annotations like `json:"basePackage"` or it won't be serialized?

pkg/tfgen/generate_schema.go
```diff
@@ -338,6 +338,12 @@ func (g *schemaGenerator) genPackageSpec(pack *pkg) (pschema.PackageSpec, error)
 		})
 	}
 
+	if javai := g.info.Java; javai != nil {
+		spec.Language["java"] = rawMessage(map[string]interface{}{
```
t0yv0, line 342: So we have this type to model the contents of `.language.java`: https://github.com/pulumi/pulumi-java/blob/main/pkg/codegen/java/importer.go#L32 - it's a bit hidden but must be made more prominent. Is there a way to reuse the type or not worth it? Can do later.

t0yv0 (approved):

PR #527, opened 2022-05-25
[tfgen] Validate the generated schema.

pgavlin:
Validate the Pulumi Package schema during generation, and fail if
validation fails. This helps catch schema errors earlier in the
generation pipeline (e.g. prior to doing heavy lifting like example
conversion).


viveklak (approved):

jkodroff (approved):

stack72 (approved):

pgavlin:
Failures look unrelated (build failures downstream due to a Pulumi/Pulumi version mismatch)

PR #532, opened 2022-06-01
Add ability to control deleteBeforeReplace semantics

stack72:
In the situation where a resource cannot have the same name AND is
explicitly named (e.g. GPC InstanceTemplate or AWS Autoscaling Group)
the resource would default to deleteBeforeReplace. This wouldn't work
if the resource was attached to another resource

Pulumi wasn't taking into account the fact that a physical name field
was being updated and still defaulted to deleteBeforeReplace

We can now specific a list of physicalNameFields in the provider that
will override the deleteBeforeReplace semantics and it will force the
creation of a new resource (because the name is changing) before it
deletes the old one

[asciicast](https://asciinema.org/a/DPDWU14QgamK6z026BvIgLwx3)


stack72:
@Frassle if you are happy with this now, then please +1 it :)

Frassle (approved):

PR #535, opened 2022-06-02
Clean up Makefile

jkodroff:
* Remove the (basically) unused imported build/common.mk
* Remove unused targets and variables
* Rename targets for simplicity (test instead of test_all)
* Update GH Actions files for new Makefile targets


stack72 (approved):

jkodroff:
Per @stack72, ok to merge b/c AWS and Azure failure is due to OOM.

PR #537, opened 2022-06-02
Get the build to pass the linting step, add step to CI

jkodroff:


jkisk (approved):
Awesome! Glad to be using a linter that is currently maintained (revive)

PR #542, opened 2022-06-07
Re-apply minor refactoring

jkodroff:


jkodroff:
Locally verified no changes to schema for AWS (which was the indicator of the bug in the first place), nor GCP.

jkisk (approved):
Looks good, thanks for improving variable names so code is easier to follow.

PR #543, opened 2022-06-07
Refactor mergeDocs into 3 separate functions.

jkodroff:
Verified locally that AWS and GCP build cleanly. (AWS would be the only one affected because it's the only provider in the Pulumi GH org that uses the `Include*From*` options.


jkisk (approved):
Ahhh, better

PR #555, opened 2022-06-23
Refactor parseArgReferenceSection

### LINKED ISSUE #529: Nested properties that share the same name get overwritten
Example: https://www.pulumi.com/registry/packages/aws/api-docs/cloudfront/distribution/

Both `DistributionDefaultCacheBehaviorLambdaFunctionAssociation` and `DistributionDefaultCacheBehaviorFunctionAssociation` have `EventType` as a property.

`DistributionDefaultCacheBehaviorFunctionAssociation.EventType` should read:

> The specific event to trigger this function. Valid values: viewer-request or viewer-response

`DistributionDefaultCacheBehaviorLambdaFunctionAssociation.EventType` should read:

> The specific event to trigger this function. Valid values: viewer-request, origin-request, viewer-response, origin-response

Known issues caused by this bug (will require an update to the bridge and re-release to fix):

* https://github.com/pulumi/pulumi-aws/issues/1993
* https://github.com/pulumi/pulumi-aws/issues/1879
* https://github.com/pulumi/pulumi-aws/issues/1994
* https://github.com/pulumi/pulumi-gcp/issues/784
* https://github.com/pulumi/pulumi-aws/issues/1992
* https://github.com/pulumi/pulumi-gcp/issues/852
* https://github.com/pulumi/pulumi-aws/issues/2050
* https://github.com/pulumi/pulumi-aws/issues/1315
* https://github.com/pulumi/pulumi-aws/issues/2119
* https://github.com/pulumi/pulumi-aws/issues/2163
* https://github.com/pulumi/pulumi-aws/issues/2164
* https://github.com/pulumi/pulumi-aws/issues/1616
* https://github.com/pulumi/pulumi-aws/issues/1603
* https://github.com/pulumi/pulumi-aws/issues/2253

### LINKED ISSUE #553: Arguments duplicate second line of source content
For some arguments whose descriptions are multi-line in the source docs, we are repeating the second line (and possibly subsequent lines - haven't found a >2 line example yet) of the content.

Example:

In [the Pulumi docs](https://www.pulumi.com/registry/packages/aws/api-docs/appmesh/route/#httpretryevents_nodejs):

> List of HTTP retry events. Valid values: client-error (HTTP status code 409), gateway-error (HTTP status codes 502, 503, and 504), server-error (HTTP status codes 500, 501, 502, 503, 504, 505, 506, 507, 508, 510, and 511), stream-error (retry on refused stream). Valid values: client-error (HTTP status code 409), gateway-error (HTTP status codes 502, 503, and 504), server-error (HTTP status codes 500, 501, 502, 503, 504, 505, 506, 507, 508, 510, and 511), stream-error (retry on refused stream).

But in [the TF docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/appmesh_route#http_retry_events):

> List of HTTP retry events. Valid values: client-error (HTTP status code 409), gateway-error (HTTP status codes 502, 503, and 504), server-error (HTTP status codes 500, 501, 502, 503, 504, 505, 506, 507, 508, 510, and 511), stream-error (retry on refused stream).

jkodroff:
Change parseArgReferenceSection so that an argument is saved in exactly
place. This change will help make the code more readable, and will help
us address pulumi-terraform-bridge#529 in a future commit.

Fixes #553 

**Notes for reviewers**

The schema for AWS shows a diff. From what I can tell after a great deal of eyeballing, the diff comprises the following types of changes:

1. Changes in entity descriptions resulting from unrelated changes upstream in the bridge (possibly better parsing of HCL or similar). <-- **UPDATE:** These diffs are likely due to the upgrade to Pulumi 3.35.0, which has since been reverted.
2. Fixes to #553: While looking at the schema diff, I realized that we had a bug that this commit actually fixes, much to my surprise.
3. Incidents of #529: This code changes the behavior so that when 2 nested arguments share the same name, the first argument is now preferred over the second. So we still have wrong argument descriptions, but now `nested2.some_arg` has `nested1.some_arg`'s description instead of vice versa. In this way, things do not appear to be made any worse - just wrong in a slightly different way. The intent is that we'll fix #529 in a subsequent PR by keeping a list of arguments with colliding names and dropping all collisions since we do not uniquely identify arguments. Example: kmsKeyArn in dynamo DB tables ([upstream docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dynamodb_table#kms_key_arn))


pkg/tfgen/docs_test.go
```diff
@@ -197,7 +197,7 @@ func TestArgumentRegex(t *testing.T) {
 				"* `retention_policy` - (Required) A `retention_policy` block as documented below.",
 				"",
 				"---",
-				"* `retention_policy` supports the following:",
+				"`retention_policy` supports the following:",
```
jkodroff, line 200: Verified via cs.github.com that whatever docs this code was intending to test the parsing of no longer exists in this format. That is, I was unable to find a nested block that starts with a bullet point and ends with "contains the following:".

It appears to be attempting to emulate a resource in Azure, but I also checked AWS, etc.

Issue #442, opened 2022-02-02
Name deprecated properties in warnings


austinbutler:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

Creating some `NrqlAlertCondition`s I get the warning `verification warning: Argument is deprecated`, but it doesn't say _what_ argument is deprecated. Through trial-and-error it seems to be `nrql.evaluationOffset`.

### Steps to reproduce

1. Add `newrelic.NrqlAlertCondition` with `nrql: {evaluationOffset: 1}`
2. Run `pulumi up`
3. Notice warning

Expected: If there's a deprecated argument, tell me which one
Actual: I am told there's a deprecated argument, but I have to guess or do trial-and-error

jkodroff:
Did some legwork on this:

Here's where the error is coming from in terraform-plugin-sdk: https://github.com/hashicorp/terraform-plugin-sdk/blob/6de4ff033a23058dd0a383df46212c304ff16756/helper/schema/schema.go

And here's where we are error-izing it: https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfshim/diagnostics/error.go

I'm unsure of the level of effort because I would guess that `Diagnostic.AttributePath` would be the HCL attribute, which we'd then have to map to the Pulumi attribute, and I don't know how difficult that mapping is with the available context.

austinbutler:
This was with TypeScript and VS Code. It's possible that the IDE was surfacing that to me, and if I see an error like this again I'll definitely check there, but apparently I didn't notice it at the time.

austinbutler:
Hit this again today with https://github.com/newrelic/terraform-provider-newrelic/pull/1656. If I hover I do see some indication that the property is deprecated, but no actual warnings or errors from VS Code.

![image](https://user-images.githubusercontent.com/354741/174920054-ddd24df0-20e9-4268-ad42-239986aaf0ae.png)

TBH the bug in https://github.com/pulumi/pulumi-terraform-bridge/issues/77 is preferable to the current status quo. It at least gives me something to start with.

joeduffy:
This still seems to be an issue; I just received this:

```
Diagnostics:
  datadog:index:Dashboard (pulumi-metrics):
[...]

PR #562, opened 2022-07-05
Ensure we suggest the deprecated attribute in bridge deprecation warnings

stack72:
Fixes: #442

### DIFF
--- a/pkg/tfshim/sdk-v2/diagnostics.go
+++ b/pkg/tfshim/sdk-v2/diagnostics.go
@@ -14,7 +14,12 @@ func warningsAndErrors(diags diag.Diagnostics) ([]string, []error) {
 		case diag.Error:
 			errors = append(errors, fromV2Diag(d))
 		case diag.Warning:
-			warnings = append(warnings, d.Summary)
+			// the summary doesn't contain the parameter name for which the warning occurs to
+			details := d.Summary
+			if d.Detail != "" {
+				details = d.Detail
+			}
+			warnings = append(warnings, details)
 		}
 	}
 	return warnings, errors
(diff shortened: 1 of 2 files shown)


guineveresaenger (changes requested):
Since we have [some test coverage ](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfshim/tfplugin5/diagnostics_test.go), we should add a test case for diagnostic Details.

Otherwise LGTM.

guineveresaenger:
I'm confused - can we not add a `Detail` field to [this line]https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfshim/tfplugin5/diagnostics_test.go#L14 and test for that?

If that test file does not map to [pkg/tfshim/tfplugin5/diagnostics.go](https://github.com/pulumi/pulumi-terraform-bridge/pull/562/files#diff-f07ee4eef978964a5b3d3491ee005c8a3ec4cf4974bdc8df0156ba8cfa4aff00) then please disregard.

pkg/tfshim/sdk-v2/diagnostics.go
```diff
@@ -14,7 +14,11 @@ func warningsAndErrors(diags diag.Diagnostics) ([]string, []error) {
 		case diag.Error:
 			errors = append(errors, fromV2Diag(d))
 		case diag.Warning:
-			warnings = append(warnings, d.Summary)
+			details := d.Summary
```
jkodroff, line 17: Can we add a comment citing an example of what might be in `Summary` vs `Detail` and why we would would prefer the latter?

pkg/tfshim/tfplugin5/diagnostics.go
```diff
@@ -18,7 +18,11 @@ func unmarshalWarningsAndErrors(diags []*proto.Diagnostic) ([]string, []error) {
 		case proto.Diagnostic_ERROR:
 			errors = append(errors, fromTF5ProtoDiag(d))
 		case proto.Diagnostic_WARNING:
-			warnings = append(warnings, d.Summary)
+			details := d.Summary
```
jkodroff, line 21: Same as above.

jkodroff (changes requested):

pkg/tfshim/sdk-v2/diagnostics.go
stack72: So the fix kinda denotes that right? I mean, the summary doesn't include the parameter name, the detail does :)

jkodroff (approved):

PR #568, opened 2022-07-13
Implement GetPluginInfo and ResolvePlugin

iwahbe:
Fixes https://github.com/pulumi/pulumi/issues/10121


stack72 (approved):

pkg/tfgen/pluginHost.go
```diff
@@ -50,6 +69,17 @@ func (host *inmemoryProviderHost) Provider(pkg tokens.Package, version *semver.V
 	return host.Host.Provider(pkg, version)
 }
 
+func (host *inmemoryProviderHost) ResolvePlugin(kind workspace.PluginKind, name string, version *semver.Version) (*workspace.PluginInfo, error) {
```
jkodroff: Could we add a docstring here?

pkg/tfgen/pluginHost.go
```diff
@@ -36,6 +38,23 @@ func (p *inmemoryProvider) GetSchema(version int) ([]byte, error) {
 	return p.schema, nil
 }
 
+func (p *inmemoryProvider) GetPluginInfo() (workspace.PluginInfo, error) {
+	var version *semver.Version
+	if p.info.Version != "" {
+		v, err := semver.ParseTolerant(p.info.Version)
+		if err != nil {
+			return workspace.PluginInfo{}, fmt.Errorf("failed to parse pkg version: %w", err)
```
jkodroff, line 47: Can we include the provider name in the error? I'm guessing that the version may be empty string at least some of the time, like if if it's malformed HCL e.g. `resource somenonexistentthing_someresource`.

jkodroff (approved):
Don't want to stand in the way of getting this out the door, but added a few suggestions for readability.

PR #575, opened 2022-07-22
Upgrade hashicorp/terraform-plugin-sdk to v2.19.0

stack72:


jkodroff (approved):

stack72:
The downstream are expected to fail until the new override is in place - I will fix those when we merge this

PR #576, opened 2022-07-22
[sdk/v2] Fill out RawConfig.

### LINKED ISSUE #548: Ensure bridge is aware of GetRawConfig
Without this, we can get a panic where the provider is using that code - GetRawConfig returns a cty.Value

https://github.com/pulumi/pulumi-auth0/issues/145
https://github.com/pulumi/pulumi-aws/issues/2017

pgavlin:
These changes update the sdk/v2 shim to fill out the RawConfig field on
InstanceState and InstanceDiff. TF providers may inspect the value of
this field to determine the differences between an attribute that has
a default value because it is was not specified and an attribute that
has been explicitly set to its default value.

Fixes #548.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -208,4 +208,4 @@ require (
 	sourcegraph.com/sourcegraph/appdash v0.0.0-20211028080628-e2786a622600 // indirect
 )
 
-replace github.com/hashicorp/terraform-plugin-sdk/v2 => ../terraform-plugin-sdk
+replace github.com/hashicorp/terraform-plugin-sdk/v2 => github.com/pulumi/terraform-plugin-sdk/v2 v2.0.0-20220722214815-c13e6bdd0f51
--- a/go.mod
+++ b/go.mod
@@ -208,4 +208,4 @@ require (
 	sourcegraph.com/sourcegraph/appdash v0.0.0-20211028080628-e2786a622600 // indirect
 )
 
-replace github.com/hashicorp/terraform-plugin-sdk/v2 => github.com/pulumi/terraform-plugin-sdk/v2 v2.0.0-20220505215311-795430389fa7
+replace github.com/hashicorp/terraform-plugin-sdk/v2 => ../terraform-plugin-sdk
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -208,4 +208,4 @@ require (
 	sourcegraph.com/sourcegraph/appdash v0.0.0-20211028080628-e2786a622600 // indirect
 )
 
-replace github.com/hashicorp/terraform-plugin-sdk/v2 => github.com/pulumi/terraform-plugin-sdk/v2 v2.0.0-20220505215311-795430389fa7
+replace github.com/hashicorp/terraform-plugin-sdk/v2 => ../terraform-plugin-sdk
```
stack72, line 211: I think this replace has to change back right?

pkg/tfshim/sdk-v2/cty.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+// makeResourceRawConfig converts the decoded Go values in a terraform.ResourceConfig into a cty.Value that is
+// appropriate for Instance{Diff,State}.RawConfig.
+func makeResourceRawConfig(config *terraform.ResourceConfig, resource *schema.Resource) cty.Value {
+	original := schema.HCL2ValueFromConfigValue(config.Raw)
+	coerced, err := resource.CoreConfigSchema().CoerceValue(original)
```
stack72, line 14: Ok so this was what I was missing!!!

stack72 (approved):
1 small think about go.mod

pkg/tfshim/sdk-v2/cty.go
pgavlin: Yeah, it took my distressingly long to find that. I implemented the equivalent twice before noticing it.

go.mod
pgavlin: Yeah, just fixed that

pgavlin:
Downstream tests are failing due to the need for those modules to reference the updated terraform-plugin-sdk fork.

PR #589, opened 2022-08-11
Use NewUniqueName

Frassle:
Same as for the other provider updates, this is the new naming system for update plans. I've put the seed on PulumiResource  rather than as a parameter on the naming function because that signature is used in a lot of downstream repos.

This also extends AutoNameOptions to have a Charset option (as NewUniqueName supports that) which should allow a lot of the transform functions in downstream repos to be switch to just autonames instead.


stack72 (approved):
Thank you fro this @Frassle

PR #595, opened 2022-08-25
Update to HEAD of pulumi

Frassle:
We split `PluginInfo` into two type `PluginInfo` and `PluginSpec`, and info no longer has download url on it (as it's intended to represent a plugin from the ".pulumi/plugins" folder which don't recall where there were downloaded from.


stack72 (approved):
We need to come up with a better way of these downstream tests - depending on HEAD is not acceptable going forward

PR #598, opened 2022-08-30
Add the ability to inject specific language examples for a resource

stack72:
This will allow us to change the examples that come from the TF
resource. Currently, the functionality we have in place is to allow
us to enhance the examples. That doesn't help when we have incorrect
examples in our docs - we want to replace the TF based examples
completely

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -456,7 +456,7 @@ const (
 )
 
 func (p *tfMarkdownParser) parseSupplementaryExamples() (string, error) {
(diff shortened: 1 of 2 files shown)


pkg/tfgen/docs.go
```diff
@@ -455,6 +455,21 @@ const (
 	sectionImports             = 5
 )
 
+func (p *tfMarkdownParser) parseSupplementaryExamples() (string, error) {
+	examplesFileName := fmt.Sprintf("docs/%s/%s.md", p.kind, p.rawname)
```
danielrbradley, line 459: As discussed, consider making the files `*.examples.md` to be more future-proof.

danielrbradley (approved):

PR #605, opened 2022-09-02
Add ability to send logs to the Pulumi Enging from preConfigure callback

guineveresaenger:
*UPDATE* Adds separate functionality to `provider.Info` to avoid a major bridge version upgrade and ensure downstream compatibility, as per review below

~*BREAKING CHANGE* Alters the function signature expected by preConfigureCallback. Accepting this change will mean downstream changes in multiple bridged providers (complete list to follow). These changes will not affect users; only provider upgrades.~

This change is in preparation for being able to solve gracefully and auditably for the situation when we want to pass along configuration warnings during a `pulumi up`, such as in https://github.com/pulumi/pulumi-gcp/issues/890.

Here's an example output using these changes on a GCP stack:

Screen Shot 2022-09-02 at 2 00 12 PM


guineveresaenger:
note: downstream test failures are expected unfortunately

guineveresaenger:
List of providers with `preConfigureCallback`:

- [ ] pulumi-artifactory
- [ ] pulumi-aws
- [ ] pulumi-azure
- [ ] pulumi-azuread
- [ ] pulumi-consul
- [ ] pulumi-ec
- [ ] pulumi-gcp
- [ ] pulumi-linode
- [ ] pulumi-mysql
- [ ] pulumi-newrelic
- [ ] pulumi-postgresql
- [ ] pulumi-rancher2
- [ ] pulumi-rke
- [ ] pulumi-signalfx
- [ ] pulumi-tailscale
- [ ] pulumi-vault
- [ ] pulumi-venafi

stack72:
@guineveresaenger this is not something we can do I'm afraid without a new major version of the bridge.

What I do suggest we do is to add a new piece of functionality in the bridge, similar to the current func but that accepts the new params and then we can have the bridge check for that as well

stack72 (changes requested):
AS per the comment

stack72 (approved):
This is looking way better for us to use now @guineveresaenger - thank you so much for the hard work on doing this

PR #612, opened 2022-09-23
Mark sensitive properties in schema as secret

viveklak:


viveklak:
@stack72 I think this is ready to go. Could you take a look when you get a chance?

stack72 (approved):

PR #614, opened 2022-09-27
Do not render shell comments as h1 in Import docs.

### LINKED ISSUE #563: Comments in Import section are rendered as H1's
We seem to not be correctly fencing the Import section. When there's a comment in the import section, it renders as an H1:

image

Source doc: https://registry.terraform.io/providers/Snowflake-Labs/snowflake/latest/docs/resources/function#import
Doc in Pulumi Registry: https://www.pulumi.com/registry/packages/snowflake/api-docs/function/#import

jkodroff:
- Add tests for parseImports()

Fixes #563 (Snowflake)

I do not believe this fixes:

* https://github.com/pulumi/pulumi-terraform-bridge/issues/563
* https://github.com/pulumi/pulumi-aws/issues/2115

But I'm reasonably confident it does not make them worse.


pkg/tfgen/docs.go
```diff
@@ -899,10 +899,14 @@ func (p *tfMarkdownParser) parseImports(subsection []string) {
 			continue
 		}
 
+		// Remove the shell comment characters to avoid writing this line as a Markdown H1:
+		section = strings.TrimPrefix(section, "# ")
```
thomas11, line 903: I'm wondering if it isn't confusing to have comments rendered as normal text, since they often are written in a different style and with a different purpose. We could render them as comments instead, e.g., by wrapping them in backticks?

jkodroff: tfgen's code is fairly fragile and difficult to test. Practically speaking, you can eyeball the generate schemas of a provider or two, but you can't really view how the docs will look in the registry.

And because of the large number of providers and the unstructured nature of docs, I'm taking a super-duper incremental approach here: Just make the Snowflake issue better without (hopefully) making any of the other providers worse.

(I did consider your approach while working on this fix and decided it wasn't worth the risk of a potential regression.)

thomas11: fair enough!

thomas11 (approved):

PR #616, opened 2022-09-28
Switch to failing tfgen by default on failed mapping errors

viveklak:
This will now require adding the environment variables to specific providers like alicloud that need this to opt-out but we should be erring for the common case instead of the previous behavior.

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -954,9 +954,9 @@ func (g *Generator) gatherResources() (moduleMap, error) {
 	}
 	modules := make(moduleMap)
 
-	failBuildOnMissingMapError := isTruthy(os.Getenv("PULUMI_MISSING_MAPPING_ERROR")) ||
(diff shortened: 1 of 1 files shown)


lblackstone (approved):

viveklak:
GCP failure should be fixed here: https://github.com/pulumi/pulumi-gcp/pull/913

That is also proof of this working as expected.

stack72:
@viveklak i personally think this is a bad idea - there are a number of providers (like AliCloud) that will have issues here due to the upstream tf provider being broken

viveklak:
> @viveklak i personally think this is a bad idea - there are a number of providers (like AliCloud) that will have issues here due to the upstream tf provider being broken as you noted
> 
> Adding this as the default means that 3rd party users themselves may have issues and they don't know how to opt-out

I am working on a change in ci-mgmt which will opt-in the behavior for these providers that don't comply with this. I understand there may be a long tail here but we need to find them one-way or the other anyway.

PR #622, opened 2022-10-07
Fixes incorrect module tokens generated for nested types (issue 611)

### LINKED ISSUE #611: Inclusion of a `/` in a datasources module name produces incorrect tokens in `schema.json` for children properties
### What happened?

In `resources.go` when a datasource's module name includes a 2 parts separated by a `/` the resulting token, in `schema.json`, is correctly produced for the corresponding `get` function, however, the tokens for the function's children properties are incorrectly produced, if the properties are non-primitives.

For example, the datasource `aws_iam_policy_document` when given a module name of `x/iam` gets a corresponding token in `schema.json` of `aws:x/iam/getPolicyDocument:getPolicyDocument`. This is correct, however, if you examine the `statements` property, which is a non-primitive,  you'll notice the token is missing the `iam` portion of the module. Eg- `aws:x/getPolicyDocumentStatement:getPolicyDocumentStatement`.

### Steps to reproduce

1. `git clone git@github.com:phillipedwards/pulumi-aws.git`
2. `cd pulumi-aws`
3. `make tfgen`
4. Open `schema.json` and examine the get function `aws:x/iam/getPolicyDocument:getPolicyDocument` note: we've introduced the `x/` to detail the function is experimental.
5. `make build_python`
6. Open `sdk/python/pulumi_aws/x/iam/get_policy_document.py` and go to line 105. 
7. The `def statements...` function references a non-existent type of `_x.outputs.GetPolicyDocumentStatement`

### Expected Behavior

1. In `schema.json` the tokens for the non-primitive children of `getPolicyDocument` should have tokens that begin with `aws:x/iam/...`. Eg- `aws:x/iam/getPolicyDocumentStatement:getPolicyDocumentStatement`
2. The generated python sdk function `pulumi_aws.x.iam.get_policy_document.py` should be referencing `GetPolicyDocumentStatementResult` as the return type for `def statements...`

### Actual Behavior

1. The non-primitive children tokens incorrectly start with `aws:x/....`. See, `#/types/aws:x/getPolicyDocumentStatement:getPolicyDocumentStatement`.
2. The statements function signature is `def statements(self) -> Optional[Sequence['_x.outputs.GetPolicyDocumentStatement']]:`

### Output of `pulumi about`

_No response_

### Additional context

_No response_

### Contributing

t0yv0:
Fixes #611 

Removes a TODO from 2018.

Where possible, `string` replaced with `tokens.Module`, `tokens.ModuleName`, `tokens.Package`, etc. This helps a little with tracking where we have a fully qualified name vs the last token of a mutli-token name, as well as review any strange string manipulations.

Added a helper paths module with types to represent schema traversal locations where types are found. While not strictly necessary, it was a great debugging helper as printing it out in any function that accept is as a parameter can quickly localize the work-in-progress relative to schema.


pkg/tfgen/generate_schema.go
```diff
@@ -908,3 +944,47 @@ func appendExample(description, markdownToAppend string) string {
 	}
 	return strings.Join(reassembledLines, "\n")
 }
+
+func modulePlacementForType(pkg tokens.Package, path *paths.TypePath) tokens.Module {
```
viveklak, line 948: This feels like the heart of the change. Lets add some unit tests here since none of the downstream repos actually exhibit the format that would trigger the bug in #611?

pkg/tfgen/test_data/regress-611-schema.json
```diff
@@ -213,7 +213,7 @@
         "values"
       ]
     },
-    "aws:x/getPolicyDocumentStatement:getPolicyDocumentStatement": {
+    "aws:x/iam/getPolicyDocumentStatement:getPolicyDocumentStatement": {
```
t0yv0, line 216: @viveklak PTAL - I've rebased this on https://github.com/pulumi/pulumi-terraform-bridge/pull/624 so now we can see how the changes affect the schema generation. The only diff is visible here is the expected diff - "x" becomes "x/iam". The other types and functions in the schema add some confidence that the rest of the schema generation does not regress.

PR #624, opened 2022-10-12
Introduce a test case reproducing 611

t0yv0:
Introduces a test case reproducing 611, in the [golden test style](https://ro-che.info/articles/2017-12-04-golden-tests) reminiscent of what Pulumi uses to test SDK generation ([example](https://github.com/pulumi/pulumi/blob/a2844f7b4436d67f980f6f5c514b99fd27e38646/pkg/codegen/testing/test/sdk_driver.go#L406)).

How this works:

- expected PackageSpec is stored in indented JSON
- the test asserts that there are no changes vs the indented JSON from the expected file
- you can run `PULUMI_ACCEPT=1 go test` to rewrite the expected file to actual value, check it with `git diff`, and commit if the new output makes sense to serve as the subsequent expected value

The test provider is extracted from a subset of AWS from #611 

Extracting the test provider subset is what caused most trouble. The values from "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema" Provider type that serve as a logical input to our schema generation process are not serializable as they contain funcs. Therefore it is difficult to write code that automatically computes subset. I got by with a semi-manual procedure that ended relying on "github.com/hexops/valast" package and some custom code to get back a Go literal that can be coaxed into an example provider value.

### DIFF
--- a/internal/testing/schemas.go
+++ b/internal/testing/schemas.go
@@ -1,0 +1,50 @@
+package testing
+
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+)
+
+func AssertPackageSpecEquals(
+	t *testing.T,
+	expectedSchemaJsonFile string,
+	spec pschema.PackageSpec,
+) {
+
+	marshalSchema := func(s pschema.PackageSpec, w io.Writer) error {
+		enc := json.NewEncoder(w)
+		enc.SetIndent("", "  ")
+		return enc.Encode(s)
+	}
+
+	assertPackageSpecMatchesFile := func(spec pschema.PackageSpec, file string) {
+		expectedBytes, err := ioutil.ReadFile(file)
+		assert.NoError(t, err)
+		expected := string(expectedBytes)
+
+		buf := bytes.Buffer{}
+		err = marshalSchema(spec, &buf)
+		assert.NoError(t, err)
+		actual := string(buf.Bytes())
+
+		assert.Equal(t, expected, actual)
+	}
+
+	if os.Getenv("PULUMI_ACCEPT") != "" {
+		buf := bytes.Buffer{}
+		err := marshalSchema(spec, &buf)
+		assert.NoError(t, err)
+		err = ioutil.WriteFile(expectedSchemaJsonFile, buf.Bytes(), 0600)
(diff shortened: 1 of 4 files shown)


t0yv0:
Working through an import cycle, need to move code around.

pkg/tf2pulumi/internal/config/module/get.go
```diff
@@ -34,10 +33,11 @@ const (
 // can't be updated on its own.
 func GetCopy(dst, src string) error {
 	// Create the temporary directory to do the real Get to
-	tmpDir, err := ioutil.TempDir("", "tf")
+	tmpDirObj, err := os.CreateTemp("", "tf")
```
t0yv0, line 36: This one isn't right, need TempDir.

pkg/tf2pulumi/internal/config/module/module_test.go
```diff
@@ -12,15 +12,16 @@ import (
...
+		log.SetOutput(io.Discard)
 	}
 }
 
 const fixtureDir = "./testdata"
 
 func tempDir(t *testing.T) string {
 	t.Helper()
-	dir, err := ioutil.TempDir("", "tf")
+	dirObj, err := os.CreateTemp("", "tf")
```
t0yv0, line 23: TempDir here also.

pkg/tf2pulumi/internal/config/module/testing.go
```diff
@@ -10,7 +9,8 @@ import (
 // as a function that should be deferred to clean up resources.
 func TestTree(t *testing.T, path string) (*Tree, func()) {
 	// Create a temporary directory for module storage
-	dir, err := ioutil.TempDir("", "tf")
+	dirObj, err := os.CreateTemp("", "tf")
```
t0yv0, line 12: TempDir

internal/testing/schemas.go
```diff
@@ -0,0 +1,62 @@
...
+		return buf.String()
+	}
+
+	assertPackageSpecMatchesFile := func(actualSpec pschema.PackageSpec, file string) {
+		expectedSpec, err := readSchemaFromFile(file)
+		assert.NoError(t, err)
+		assert.Equal(t, schemaToString(expectedSpec), schemaToString(actualSpec))
+	}
+
+	if os.Getenv("PULUMI_ACCEPT") != "" {
```
viveklak, line 53: Nice! Should probably add some documentation around this to update the reference schema.

viveklak (approved):
Couple of minor comments. I will follow up with fixing them. Thanks for adding the testing framework!

PR #629, opened 2022-10-27
Auto update pulumi/pulumi weekly

viveklak:
I have no idea how to test this.. I copied this from ci-mgmt and made some changes. Will try it out and see if it works after merge.

### DIFF
--- a/.github/workflows/weekly-pulumi-update.yml
+++ b/.github/workflows/weekly-pulumi-update.yml
@@ -1,0 +1,66 @@
+name: weekly-pulumi-update
+on:
+  schedule:
+  - cron: 35 12 * * 4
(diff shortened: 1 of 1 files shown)


t0yv0:
There's a way to make workloads trigger-able from GitHub UI. Like this one: https://github.com/pulumi/examples/blob/master/.github/workflows/performance_metrics_cron.yml#L3

It has to be on main branch first. The way I tested before:

1. merge a PR making an empty triggerable workflow on main branch
2. stand up a PR with the meat of the changes for the logic
3. trigger it manually *from the branch* to test the PR
4. when happy merge the PR

This assumes the logic is reasonably non-destructive and has safe side-effects for testing.

PR #633, opened 2022-11-07
Removes TODOs around the distinction between Config and PlannedState

t0yv0:
The meat of the change is removing TODOs around distinct Config vs PlannedState. Planning the state is factored out into `provider.plan` method. Parameters of Create, Diff, Update are renamed to be more suggestive.

There is a wrinkle that Pulumi CLI does not pass the planned state between Diff and say Update method, so this ends up calling it twice: Pulumi sequence Check, Diff, Update will call plan in Diff and in Update. That is okay. What this buys is the the protocol continues to be stateless.

As part of the change, had to rejigger some tests. Computed values generate unexpected diffs as in "x" changing to "<unknown>".

UseStateForUnknown plan modifier is an idiomatic way to suppress those (found this in the random resource). 

Also, specifically in the test resource written here, there are computed values that copy optional values; they are added to give more test cases to the serde testing. But TF gets confused and generates diffs showing that, say, "optionalInputBoolCopy" is "<unknown>" even when "optionalInputBool" is nil. Added a custom PlanModifier to recognize this case.


t0yv0:
On top of that: add support for lists; lists indicated a further problem with plan/diff implementation necessitating the rewrite of PlannedNewState.

Issue #477, opened 2022-03-10
tfgen.convert.Convert panics instead of returning an error


jkodroff:
The code does the meat of converting HCL panics when it encounters an error, which complicates the code for downstream callers, like `tfbridge.convertHCL()`:

https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tf2pulumi/convert/tf12.go#L161-L167

If we panic for good reason, we need to document the reasoning in a comment. Otherwise, we should return an error in order to allow downstream callers to handle issues parsing HCL as they see fit.

stack72:
@jkodroff I *think* you have changed this behaviour now right?

jkodroff:
@stack72 No - I have not touched any code in the `tf2pulumi` package. What I did was wrap the call in `tfgen` so that _that_ particular call to `convert.Convert()` catches the panic.

PR #634, opened 2022-11-09
Include stacktrace in error message

RobbieMcKinstry:
In TFGen, HCL conversion is allowed to panic. Panics are recovered and converted into errors, which are then displayed to the user.

Currently, these panics don't provide a way to inspect the stacktrace, making it very hard to debug P1s. This PR captures the stacktrace, and logs traces when `debug` is enabled.

# Related

* https://github.com/pulumi/pulumi-terraform-bridge/issues/477

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -17,6 +17,7 @@ package tfgen
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"github.com/hashicorp/go-multierror"
 	"golang.org/x/text/cases"
@@ -1209,6 +1210,11 @@ func (err *ConversionError) Error() string {
 	return err.wrappedErr.Error()
 }
 
+// Unwrap provides error as returned by the conversion panic.
+func (err *ConversionError) Unwrap() error {
+	return err.wrappedErr
+}
+
 // Statically enforce that ConversionError implements the Error interface.
 var _ error = &ConversionError{}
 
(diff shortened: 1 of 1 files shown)


pkg/tfgen/docs.go
```diff
@@ -1174,6 +1175,43 @@ func (g *Generator) convertExamples(docs, name string, stripSubsectionsWithError
...
+	}
+}
+
+// StackTrace returns the stacktrace of the error.
+func (err *ConversionError) StackTrace() string {
+	return err.trace
+}
+
+// Return the err-representation of this struct.
+func (err *ConversionError) Error() string {
```
viveklak, line 1208: Nit - should we add `Unwrap` as well to support unwrapping and `errors.As` and `errors.Is`?

pkg/tfgen/docs.go
```diff
@@ -1226,6 +1264,9 @@ func (g *Generator) convertHCLToString(hcl, path, languageName string) (string,
 	if err != nil {
 		// Because this condition is presumably the result of a panic that we wrap as an error, we do not need to add
 		// anything to g.coverageTracker - that's covered in the panic recovery above.
+		if convErr, ok := err.(*ConversionError); ok {
```
viveklak, line 1267: Switch to `errors.As` instead?

viveklak (approved):
Minor nits. LGTM otherwise.

pkg/tfgen/docs.go
RobbieMcKinstry: TIL about `Unwrap`!

PR #636, opened 2022-11-10
Connect schema generation

t0yv0:
Implements schema generation by reusing tfgen.

To interface with it, builds up a minimal schema-only shim provider (`pf.Provider -> shim.Provider`). Runtime methods panic.

Removes hand-written and byte-checked-in schemas from the tests.

Implements tfgen.Main.

To pass test on the autogenerated schemas, works around pluralization mangling. There is more due diligence TODO on renaming.

Adds make targets to rebuild testproviders "the tfgen way" by building their schemas first and then embedding them.


pkg/tfpfbridge/naming.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func toPropertyKey(name string, typ tftypes.Type) resource.PropertyKey {
+	if willPluralize(name, typ) {
+		return resource.PropertyKey(inflector.Pluralize(name))
+	}
+	return resource.PropertyKey(name)
+}
+
+func willPluralize(name string, typ tftypes.Type) bool {
```
viveklak, line 38: Nit: how about `func pluralize` instead which returns `(string, bool)` where the second return type suggests if name was pluralized instead?

viveklak (approved):
This looks amazing, great to see the dogfooding we can do here already!

PR #639, opened 2022-11-14
Start of the PCL test framework for Eject

Frassle:


pkg/tf2pulumi/convert/eject_test.go
```diff
@@ -0,0 +1,149 @@
...
+				name string
+				path string
+			}{
+				name: info.Name(),
+				path: filepath.Join(testDir, info.Name()),
+			})
+		}
+	}
+
+	loader := &testLoader{path: filepath.Join(testDir, "schema")}
```
Zaid-Ajaj, line 102: Since we skipped a directory called `schemas` that include the test schemas, should this also be `schemas`?

Frassle: hah yes! Adding a test that actually needed a schema was going to be the next thing I added and I would of stumbled on this then.

Zaid-Ajaj (approved):

PR #642, opened 2022-11-14
Support nested attributes

t0yv0:
Just enough support added for a NestedList example to work end to end, inspired by [this](https://github.com/fly-apps/terraform-provider-fly/blob/28438713f2bdf08dbd0aa2fae9d74baaca9845f1/internal/provider/machine_resource.go#L179) schema example:

- schema generation for object types
- convert support for object types
- new integration test

The difficulty here is that this does need a small changes in the production tfgen to generate schemas for these object types. Previously tfgen never had to deal with this. To minimize the change-set, the code here employs a trick with a pseudoResource struct to push through the object type metadata into tfgen. I should lift this out of the tfpf-stubs into a separate PR against master so we can get that in and reason about bw-compat safety on the change.


t0yv0:
We'll need to find a way to retroactively review the code, merging for now.

PR #643, opened 2022-11-15
Support TypeObject

t0yv0:
With Terraform Plugin Framework arises a need for nested Object types. With this minor modification tfgen can start supporting these and generating Pulumi Schema for these types. It is a bit quirky to overload Elem() to return a pseudo-resource for TypeObject, but it keeps the change minimal without digging further into errors in MashalSchema. Open to feedback on that, if we want a cleaner encoding I can dig deeper and try to work it out.

The change should be backwards-compatible since TypeObject entries are not yet generated by any of the production providers.

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -334,6 +334,8 @@ func makePropertyType(objectName string, sch shim.Schema, info *tfbridge.SchemaI
 		t.kind = kindMap
 	case shim.TypeSet:
 		t.kind = kindSet
+	case shim.TypeObject:
+		t.kind = kindObject
 	}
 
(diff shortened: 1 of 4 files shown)


t0yv0:
Cleaned up to V2 that introduces a special ObjectTypeSchema interface to make the distinction between object types and resources a little cleaner (thanks @guineveresaenger). Added extra docstrings around Elem() and where that comes from, a little digging uncovered the Elem() returning Resource case comes from blocks in Terraform.

t0yv0:
Second thoughts here - looks like this extension is not needed, and returning Resource out of Elem() passes the test and achieves the desired effect. We don't need this change. This is good. I've put the doc-string on Elem() into a separate PR since that should be useful down the line.

PR #646, opened 2022-11-15
Single nested support

t0yv0:
Integration test for tfsdk.SingleNestedAttributes and a few minor fixes.

### DIFF
--- a/pkg/tfpfbridge/schemashim/attr_schema.go
+++ b/pkg/tfpfbridge/schemashim/attr_schema.go
@@ -92,8 +92,10 @@ func (s *attrSchema) Elem() interface{} {
 		lT := t.(tftypes.List)
 		var schema shim.Schema = &typeSchema{lT.ElementType}
 		return schema
+	case t.Is(tftypes.Object{}):
+		return &objectPseudoResource{t.(tftypes.Object)}
 	default:
-		panic(fmt.Errorf("TODO: unhanded elem case: %v", t))
+		panic(fmt.Errorf("TODO: unhandled elem case: %v", t))
 	}
 }
 
--- a/pkg/tfpfbridge/schemashim/attr_schema.go
+++ b/pkg/tfpfbridge/schemashim/attr_schema.go
@@ -72,8 +72,7 @@ func (s *attrSchema) ForceNew() bool {
 
 func (*attrSchema) StateFunc() shim.SchemaStateFunc { panic("TODO") }
 
-// Needs to return a shim.Schema, a shim.Resource, or nil. IN the case of attrSchema it is always a
-// shim.Schema specifically a &typeSchema, or else nil if the type has no element.
+// Needs to return a shim.Schema, a shim.Resource, or nil.
 func (s *attrSchema) Elem() interface{} {
 	ctx := context.TODO()
 	t := s.attr.FrameworkType().TerraformType(ctx)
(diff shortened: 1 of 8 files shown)


pkg/tfpfbridge/schemashim/attr_schema.go
```diff
@@ -92,8 +92,10 @@ func (s *attrSchema) Elem() interface{} {
 		lT := t.(tftypes.List)
 		var schema shim.Schema = &typeSchema{lT.ElementType}
 		return schema
+	case t.Is(tftypes.Object{}):
```
t0yv0, line 95: This needed to be filled in for SingleNestedAttributes to work.

pkg/tfpfbridge/tests/internal/cmd/pulumi-resource-testbridge/schema.json
```diff
@@ -81,6 +81,21 @@
...
+            "properties": {
+                "description": {
+                    "type": "string"
+                },
+                "quantity": {
+                    "type": "number"
+                }
+            },
+            "type": "object",
+            "required": [
```
t0yv0, line 95: The optional/required distinction is lost due to an underlying issue in TFPF mapping types to TF types. Perhaps we could work around this and recover the info that is still present in the TPFF-only attr types.

pkg/tfpfbridge/tests/nested_attr_test.go
```diff
@@ -32,5 +37,20 @@ func TestNestedType(t *testing.T) {
...
+	ctx := context.TODO()
+	nestedOptional := tfsdk.SingleNestedAttributes(map[string]tfsdk.Attribute{
+		"nestedOptional": {
+			Optional: true,
+			Type:     types.StringType,
+		},
+	})
+	assert.True(t, nestedOptional.GetAttributes()["nestedOptional"].IsOptional())
+	// Projecting to tftypes.Object seems to forget the optional-ity of nestedOptional.
+	assert.Contains(t, nestedOptional.Type().TerraformType(ctx).(tftypes.Object).OptionalAttributes, "nestedOptional")
```
t0yv0, line 55: Here's the issue that surprised me, TerraformType() projection lost the optionality of the attributes.

pkg/tfpfbridge/schemashim/attr_schema.go
viveklak: Add this as documentation to the code?

pkg/tfpfbridge/schemashim/schema_decorators.go
```diff
@@ -0,0 +1,75 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schemashim
+
+import (
+	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+)
+
+type schemaDecorator struct {
```
viveklak, line 21: Nit - thoughts on embedding the shim.Schema interface anonymously - in which case you wouldn't need to implement all the other methods?
```
type schemaWithOptional struct {
    shim.Schema // embedded interface
    optional func(shim.Schema) bool
}

func (o *schemaWithOptional) Optional() bool {
	if o.optional != nil {
		return o.optional(o.Schema)
	}
	return o.Schema.Optional()
}
```

pkg/tfpfbridge/tests/internal/cmd/pulumi-resource-testbridge/schema.json
viveklak: This would be problematic for updating random and other providers no?

viveklak (approved):

viveklak:
Some minor nits, but otherwise looks good!

pkg/tfpfbridge/tests/internal/cmd/pulumi-resource-testbridge/schema.json
t0yv0: It didn't show in the random schema diff, but it will bite at some point, I'll file an issue to fix it up.

PR #647, opened 2022-11-16
Harden ID handling

t0yv0:


pkg/tfpfbridge/provider_resources.go
```diff
@@ -27,15 +27,19 @@ import (
...
+type resourceHandle struct {
+	makeResource          func() pfresource.Resource
+	terraformResourceName string
+	schema                tfsdk.Schema
+	pulumiResourceInfo    *info.ResourceInfo // optional
+	idExtractor           idExtractor
+}
 
+func (p *Provider) resources(ctx context.Context) (res pfutils.Resources, err error) {
 	p.resourcesOnce.Do(func() {
```
guineveresaenger, line 39: Not the topic of this PR but what is `resourcesOnce`?

pkg/tfpfbridge/provider_resources.go
```diff
@@ -27,15 +27,19 @@ import (
 	"github.com/pulumi/pulumi-terraform-bridge/pkg/tfpfbridge/pfutils"
 )
 
-func (p *Provider) resources(ctx context.Context) (res pfutils.Resources, err error) {
+type resourceHandle struct {
+	makeResource          func() pfresource.Resource
+	terraformResourceName string
+	schema                tfsdk.Schema
+	pulumiResourceInfo    *info.ResourceInfo // optional
+	idExtractor           idExtractor
```
guineveresaenger, line 35: I'm curious why this is using functions as struct fields instead of creating methods on the `resourceHandle` type? is it because each resource handle needs to have its own unique extractor, based on the resource type?

guineveresaenger (approved):
Thank you for all the comments here! It's really nice to have context. 👍

pkg/tfpfbridge/provider_resources.go
t0yv0: it's making sure resource scanning happens only once lazily on first ask. Should probably move to a constructor instead somewhere.

pkg/tfpfbridge/provider_resources.go
t0yv0: Ah these are some bad habits from my FP past. It's true that each idExtractor is slightly different but the only difference really is knowing the resource type for prettier errors. Moving this to a method and its state into the resourceHandle struct makes sense.

PR #648, opened 2022-11-17
Test invoke using new mapping interfaces

Frassle:


pkg/tf2pulumi/convert/testdata/simple_invoke/pcl/main.pp
```diff
@@ -0,0 +1,7 @@
+aDataSource = invoke("simpledata::source", {
```
Frassle, line 1: All of this work just to get this one little test out.

Zaid-Ajaj (approved):
LGTM. Just one question, how did we generate `schemas/simpledata.json` (Pulumi schema) from `mappings/simpledata.json` (what I assume to be terraform schema)?

iwahbe (approved):

Frassle:
> how did we generate...

I wrote them both by hand.

PR #649, opened 2022-11-18
Tabulate renaming decisions

t0yv0:
The function to convert TF names to Pulumi names may not be invertible. But both directions of conversion are needed (tf->pu and pu->tf) in serde that happens at runtime in the bridged providers. Also tf->pu direction is needed in the tf2pulumi etc tools, and could be made foolproof via tables.

What if we tabulated the renaming decisions and stored it in metadata, perhaps as an extension to PackageSpec? Exploring this in this draft PR, though the approach may not be final.

To build tables we need a way to uniquely address properties. 

I think for this purpose what is sufficient for addressing properties in Pulumi is (token|config|provider, name) where token is a resource, function, or named type token, that is without making a distinction between state/input/output properties. Provider resource has a token. Config doesn't have a token but can be a special key.

But the implementation reuses an internal TypePath that is more refined (carries extra information).  TypePath identifies locations of paths during the translation, making the distinction between args and outputs for example; a subset of TypePath can be used to identify all places where a language SDK would generate a struct, even if these are not addressable in. 

The PR may be cramming a bit too much into TypePath. It's been useful for me to trace it through the execution as it's a good debugging aide but it may be trying to do too much. 

It is also a bit clumsy to correlate object type tokens to a notion of TF type locations that map to them, and it is actually a one-to-many relation which makes things a little more interesting still.


iwahbe (commented):
Can you clarify what consumes this table? It sounds like the table is only consumed by the provider that embeds it. 

I'm a little concerned that this will prevent pulumi from round-tripping the schema. I'm not sure that we do, but it seems like a capability we will decide to want.

Frassle:
> What if we tabulated the renaming decisions and stored it in metadata, perhaps as an extension to PackageSpec? Exploring this in this draft PR, though the approach may not be final.

This should not be part of schema, it's only going to be used by two things. 
One the provider itself, which is what defined the naming logic anyway so should be able to keep track of the renames however it likes (if a full rename table is the best way then fine do that)
Two the program conversion system, which would pull this data from the [ProviderInfo struct](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/info.go#L47), so the rename table should probably be thrown into there.

That second part is what would interact with the conversion mapper changes I'm making in https://github.com/pulumi/pulumi/pull/11378, and is a good prompt that we probably need to version the data format there for plugins even though to the engine its just bytes.

t0yv0:
I don't mind not mixing this data into the schema, that is just a technical convenience to achieve the effect of precomputing the data and baking it into the provider for runtime access:

```
pulumi-tfgen-foo --> renames.json --[embed] --> pulumi-resource-foo
```

Can totally use another route to get that to happen.

Now extending ProviderInfo does not make sense to me at the moment, ProviderInfo seems to be authored by hand or some semi-automated process to configure tfgen. So ProviderInfo is an input to tfgen, while Renames are an output of tfgen. Renames are f(input, tfgen.logic). 

> program conversion system, which would pull this data from the [ProviderInfo struct](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/info.go#L47), so the rename table should probably be thrown into there.

The part I want to understand here is versioning/releases. Currently when we release a new version of pulumi-provider-aws say, we get a new schema built and as well as a new provider binary, and we track bw-compatibility there. I'd like to think TF renamings are published and versioned as part of this process. 

If this is how it's versioned, any tool like tf2pulumi should be able to grab any version of pulumi-provider-aws and respect that version's renaming decisions, even if those are changing between versions. 

So perhaps Program conversion system should instead query the provider for this? Or how does tf2pulumi read ProviderInfo, does it query the provider for it?

pkg/tfgen/renames.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen/internal/paths"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+)
+
+// Tabulates information about custom inflected or pluralied names. Pulumi bridged providers and Terraform conversion
+// tooling may need to consult these tables to accurately translate between Pulumi and Terraform names.
+type Renames struct {
+	// RenamedResources[t] stores the Terraform name for a resource identified by the Pulumi token t when this name
+	// is not trivially recoverable from the token itself.
```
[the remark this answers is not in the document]
t0yv0: Ah yes, I was trying to find it and it's' both duplicated in a few places and little complicated. Search for this:

```go
// resourceName translates a Terraform name into its Pulumi name equivalent, plus a module name.
func resourceName(provider string, rawname string,
	info *tfbridge.ResourceInfo, isProvider bool) (string, tokens.ModuleName) {
	if isProvider {
		return "Provider", indexMod
	}
	if info == nil || info.Tok == "" {
		// default transformations.
		name := withoutPackageName(provider, rawname)                  // strip off the pkg prefix.
		camel := tfbridge.TerraformToPulumiName(name, nil, nil, false) // camelCase the module name.
		pascal := tfbridge.TerraformToPulumiName(name, nil, nil, true) // PascalCase the resource name.

		modName := tokens.ModuleName(camel)
		return pascal, modName
	}
	// otherwise, a custom transformation exists; use it.
	return info.Tok.Name().String(), info.Tok.Module().Name()
}
```

IN the end I decided to just say ok, whatever, let's just record every resource name instead of mire-optimizing this.

Frassle:
> Or how does tf2pulumi read ProviderInfo, does it query the provider for it?

Yes I'm adding a new GetMapping method to the provider interface for providers to implement so they can send data to tf2pulumi.

t0yv0:
That works! Keep me looped in I'd like to implement this method for the TFPF-bridged provider (also for classical bridge maybe), and I'd love to agree on the schema of the data it sends.

viveklak:
> I don't mind not mixing this data into the schema, that is just a technical convenience to achieve the effect of precomputing the data and baking it into the provider for runtime access:
> 
> ```
> pulumi-tfgen-foo --> renames.json --[embed] --> pulumi-resource-foo
> ```

Just FYI - this is something we do pretty heavily in the native providers with the use of metadata.json, e.g.: https://raw.githubusercontent.com/pulumi/pulumi-azure-native/master/provider/cmd/pulumi-resource-azure-native/metadata.json. e.g. https://github.com/pulumi/pulumi-azure-native/blob/master/provider/pkg/resources/resources.go#L48 `SdkName` is used to rename fields. This is used both in the provider and arm2pulumi to perform the translation. 
 
Currently these metadata files are somewhat artisanal in nature and don't have a lot of consistency. We have the same model in aws-native as well but it looks slightly different.

I have previously thought about standardizing this (optional?) metadata and exposing this through another RPC similar to GetSchema etc. Is this a line of thinking worth entertaining?

Frassle:
> I have previously thought about standardizing this (optional?) metadata and exposing this through another RPC similar to GetSchema etc. Is this a line of thinking worth entertaining?

I think metadata that's just needed for how the provider works should be baked into the provider itself. For data that's needed from the provider to drive things like tf2pulumi/arm2pulumi I am adding a new RPC method ([GetMapping](https://github.com/pulumi/pulumi/pull/11378))

t0yv0:
@viveklak to your comment, do you have a schema for that JSON file somewhere? Or "artisanal in nature" means it's per provider? 

Cross-linking https://github.com/pulumi/pulumi-terraform-bridge/pull/654/files here also. It sounds like I missed a structure called MarshallableProviderInfo  that is currently consumed by tf2pulumi ; Fraser's PR is threading it through the engine to dis-intermediate tf2pulumi plugin loading through the engine. 

It's a big structure. I think this structure may be redundant with a schema extension here, I think it encodes token/name decisions. I'll double check. Sounds like TFPF should hit the same structure which I missed. 

If the information is available from MarshallableProviderInfo and we continue supporting it - it might make this PR redundant.

t0yv0:
Rewrote description, removed edits to PackageSpec to address comments. Still seems like a good idea to me.

Frassle:
Not clear how this would be used, also doesn't look like its threaded into the provider info structure to feed back to tf2pulumi which surely _also_ needs to know about these renames to be able to convert correctly?

t0yv0:
I would like to make use of this on the t0yv0/tfpf-stubs branch right away, with a possible refactor of the current bridge later on. The [new features](https://developer.hashicorp.com/terraform/plugin/which-sdk#do-you-need-more-features-than-sdkv2-provides) include "structured types like objects". I cannot quite work out how existing code supports these. 

So for now rewriting how current tfbridge and tf2pulumi inflect names to use this table is out of scope of this PR. Presumably tf2pulumi is naming things right as implemented at the moment. But such a refactor would be directionally aligned with this PR. There's two options:

1. continue with current approach to inflecting names; TerraformToPulumiName + MarshallableProviderInfo; drop this PR; adopt this approach also in the PluginFramework version.

Here is a demonstration on how this works for resource properties:

https://github.com/pulumi/pulumi-terraform-bridge/blob/b6bc4c4d8721895699b1b1d376e31b21a38a8b87/pkg/tfpfbridge/tests/schema_test.go#L37-#L37

It's not too bad for resources, but how does it work for named object types (which also may have inflected properties presumably)? 

I found something like [this](https://github.com/pulumi/pulumi-terraform-bridge/blob/b6bc4c4d8721895699b1b1d376e31b21a38a8b87/pkg/tfbridge/provider.go#L628-#L628) but it does not seem to be used at the right place? 

2. move away from TerraformToPulumiName to tabulated renames eventually. Use it in PF branch first. Later refactor the bridge to use it. Expose the tables to tf2pulumi - extending MarshallablePluginInfo or the GetSchema() response all sounds reasonable, I'm not as familiar with the details yet.


2 is not immediately beneficial but I think much more fool-proof long term if we can afford storing these tables.

pkg/tfgen/test_data/regress-611-renames.json
```diff
@@ -0,0 +1,331 @@
+{
```
viveklak, line 1: I guess we will have this file now available in the repo and linked into the provider binary? Just a note that we will need to be mindful to support rolling this out across the providers as part of a rollout for tf bridge. Might involve some ci-mgmt changes etc.

viveklak (approved):

pkg/tfgen/main.go
```diff
@@ -33,14 +33,38 @@ import (
...
+		err = g.Generate()
+		if err != nil {
+			return err
+		}
+
+		return err
+	})
+}
+
+// Like Main but allows to customize the generation logic past the parsing of cmd-line arguments.
```
t0yv0, line 54: @guineveresaenger and @viveklak  I had questions about this in a few places, answering here. This annoying looking logic supports a divergent behavior between legacy bridged providers and TFPF providers around generating the renames.json file. 

So that:

- legacy bridged providers continue using Main that does not generate renames.json
- whereas TFPF bridged providers use a MainWithCustomGenerate that goes ahead and generates the renames.json

SO the good news is that accepting this PR does not generate any downstream work in the bridged provider boilerplates, but the bad news perhaps is that this does not look particularly elegant or clear; if you have a better idea of where to make the option of generating renames configurable LMK I can clean this up in a follow-up PR.

PR #653, opened 2022-11-28
Nested optionals

t0yv0:
Fixes pulumi/home#2432

When working out the unexpected Float->Int schema change in Random, a repeated theme emerged that github.com/hashicorp/terraform-plugin-framework/attr has more information than "github.com/hashicorp/terraform-plugin-go/tftypes":

- it distinguishes int/float vs number
- it recalls whether object attrs are optional for not

This PR is working toward preserving attr.* values during schema translation so it can use all available information. As a nice effect of this change, the optional-ity information is now preserved through nested object types so it fixes 2432.

Unfortunately the attr.* API is a bit cumbersome to consume for this purpose as it quickly drops into internal packages. This PR uses reflect to compensate.


pkg/tfpfbridge/schemashim/attr.go
```diff
@@ -28,3 +47,59 @@ type attr interface {
...
+	if attrMap.IsNil() {
+		return attrs
+	}
+
+	// Sometimes need to call a.GetAttributes().GetAttributes() to get to a map.
+	if attrMap.Type().Kind() != reflect.Map {
+		attrMap = callGetAttributes(attrMap)
+	}
+
+	if attrMap.Type().Kind() != reflect.Map {
```
guineveresaenger, line 66: so if it's not a map after two calls, we panic? Do we catch this panic somewhere?
what is the condition under which we cannot translate this particular attribute? does it mean we're having bad input?
 Does this panic cause the entire provider to fail to build?

pkg/tfpfbridge/schemashim/attr.go
```diff
@@ -28,3 +47,59 @@ type attr interface {
...
+	return attrs
+}
+
+func callGetAttributes(v reflect.Value) reflect.Value {
+	m := v.MethodByName("GetAttributes")
+	t := v.Type().String()
+	if !m.IsValid() {
+		panic(fmt.Sprintf("Expected a value of type %s to implement GetAttributes method", t))
+	}
+	attrCallResult := m.Call(nil)
```
guineveresaenger, line 89: whoa, very neat

pkg/tfpfbridge/schemashim/attr.go
```diff
@@ -15,10 +15,29 @@
...
 
 import (
+	"fmt"
+	"reflect"
+
 	pfattr "github.com/hashicorp/terraform-plugin-framework/attr"
 )
 
 type attr interface {
+	attrLike
```
guineveresaenger, line 25: Would it not be possible to combine `attr` and `attrLike`? Perhaps I'm missing some dependency implications?

guineveresaenger (approved):
Do we know how much extra time using reflection will cause?
And thank you for the clear PR description!

pkg/tfpfbridge/schemashim/attr.go
t0yv0: Yes it would. I think I'd like that actually - this is essentially an assert that should not happen. Schema generation should not have these issues, if it does, it means some assumptions here are broken and we need to fix it. Fortunately this code does not run at runtime in the bridged provider at all so blast radius is only pertaining to builds.

PR #654, opened 2022-11-29
Second phase of convert mapper

Frassle:


Frassle:
Blocked on https://github.com/pulumi/pulumi/issues/11452

pkg/tfgen/pluginHost.go
```diff
@@ -53,6 +54,15 @@ func (p *inmemoryProvider) GetSchema(version int) ([]byte, error) {
 	return p.schema, nil
 }
 
+func (p *inmemoryProvider) GetMapping(key string) ([]byte, string, error) {
+	info := tfbridge.MarshalProviderInfo(&p.info)
```
t0yv0, line 58: Whoah. Is all of this needed🙏  

This is a large-ish export, I need to get a bit more of a handle on all the things here to be comfortable, as presumably this now also becomes an obligation for PF-built providers to hit the same contract.

Frassle: This is not new, this is the data structure that tf2pulumi _already_ expects for bridged providers to export for conversions. It's just currently it access it via running the plugin with the "-get-provider-info" flag (see https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/main.go#L55-L60)

This is just moving that interface to be a well known engine interface so that the engine can do the plugin lookup and loading, rather than tf2pulumi having to do it.

t0yv0: Ah, gotcha so tf2pulumi consumes it. This is my miss then, I need to target supporting this from tfpf branch as well!

iwahbe (approved):
I'm not familiar with the underlying `p.info`, but this looks like the expected continuation of the original PR.

PR #656, opened 2022-11-30
Schema support for recognizing Int64 type

t0yv0:
On top of t0yv0/nested-optionals. 

This makes schema generation recognize int vs number. 

It turns out runtime support for integers may not be necessary. Both on PropertyValue side and on tftypes side the only numeric type is float64 and in case of TF it's big.Float, there's no dedicated int transport types.


viveklak (approved):
Nice!

PR #657, opened 2022-11-30
Fix race on ID allocation causing flaky tests

t0yv0:
Adding a mutex and not relying on os.Stat to sync has fixed `make test.integration` target from being flaky. What was happening is id "2" was allocated to two resources simultaneously, confusing Read from pseudo-cloud-state.

### DIFF
--- a/pkg/tfpfbridge/tests/internal/testprovider/testbridge_resource_testres.go
+++ b/pkg/tfpfbridge/tests/internal/testprovider/testbridge_resource_testres.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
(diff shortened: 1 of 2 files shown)


pkg/tfpfbridge/tests/internal/testprovider/testbridge_resource_testres.go
```diff
@@ -503,19 +506,36 @@ func (e *testres) Delete(ctx context.Context, req resource.DeleteRequest, resp *
...
+	if err := mu.Lock(); err != nil {
+		return "", err
+	}
+	defer func() {
+		if err := mu.Unlock(); err != nil {
+			panic(err)
 		}
+	}()
+
+	cF := filepath.Join(statedir, "testres.counter")
```
guineveresaenger, line 519: is `cF` the location of the resource file?

guineveresaenger (approved):

pkg/tfpfbridge/tests/internal/testprovider/testbridge_resource_testres.go
t0yv0: It's a special counterFile that stores the integer that we use as a resource ID. We increment it and store to disk. This has to work across processes. A mini database of sorts.

PR #661, opened 2022-12-01
Refine path model and path assumptions

t0yv0:
The paths module is used internally to track where we are in the schema during conversion. This refactors it a bit to simplify the model and introduces an interesting correction that helps clarify the tfgen process.

When allocating a named object type, tfgen actually merges structurally identical entries into a single new named object type. Therefore, several distinct paths in the schema may correspond to a new object type token. This is now explicit via PathSet. 

It is interesting whether it is possible for the merging to happen across modules, and which module to place the generated type into if that's the case. I believe it's not possible, but now there is a panic/assert checking for that case.

### DIFF
--- a/pkg/tfgen/generate_schema.go
+++ b/pkg/tfgen/generate_schema.go
@@ -53,7 +53,9 @@ type schemaNestedType struct {
 	requiredInputs  codegen.StringSet
 	requiredOutputs codegen.StringSet
 	pyMapCase       bool
-	typePath        *paths.TypePath
+
+	// The same *propertyType may be found at multiple typePaths and reused. Non-empty.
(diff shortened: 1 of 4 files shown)


pkg/tfgen/generate_schema.go
```diff
@@ -135,12 +131,16 @@ func (nt *schemaNestedTypes) declareType(path *paths.TypePath,
 		requiredOutputs = required
 	}
 
+	// Merging makes sure that structurally identical types are shared and not generated more than once.
```
t0yv0, line 134: This is the interesting bit in the tfgen algorithm! Adding a comment to highlight it.

viveklak (approved):

PR #666, opened 2022-12-05
Dedicated convert package

t0yv0:
- break down convert.go into more files in a dedicated package
- thread through Pulumi metadata so that object converters are aware of Pulumi type token they're working with
- basic abstractions for token-indexed renames eventually intended to work with https://github.com/pulumi/pulumi-terraform-bridge/pull/649


t0yv0:
Vivek had doubts reviewing this but I still think  I need this and use it as the foundation for the rest.. We need to convert Pulumi values to TF values. The current approach here is a direct type-aware mapping between tftypes.Value and resource.PropertyValue. It's not the only way but I quite like how explicit it is.

jazzyfresh (approved):

PR #667, opened 2022-12-05
Basic secret transcoding

t0yv0:
Quick support for dealing with Secret values on the wire. Pulumi passes these values for properties marked as secret in the schema. This change is necessary to scale up to passing Random provider tests.

### DIFF
--- a/pkg/tfpfbridge/internal/convert/secret.go
+++ b/pkg/tfpfbridge/internal/convert/secret.go
@@ -1,0 +1,72 @@
+// Copyright 2016-2022, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


guineveresaenger (approved):

pkg/tfpfbridge/internal/convert/secret.go
```diff
@@ -0,0 +1,70 @@
...
+		return tftypes.Value{}, fmt.Errorf("PropertyValue should be secret but is not")
+	}
+	v := p.SecretValue().Element
+	return enc.elementEncoder.FromPropertyValue(v)
+}
+
+func (dec *secretDecoder) ToPropertyValue(v tftypes.Value) (resource.PropertyValue, error) {
+	if !v.IsKnown() || v.IsNull() {
+		return dec.elementDecoder.ToPropertyValue(v)
+	}
```
viveklak, line 64: This special treatment isn't obvious given the use of `dec.elementDecoder.ToPropertyValue(v)` later as well...

viveklak (approved):

PR #675, opened 2022-12-10
Ensure forward build compatibility for resource provider

AaronFriel:
See: https://github.com/pulumi/pulumi/issues/11616


pkg/tfgen/pluginHost.go
```diff
@@ -32,6 +33,8 @@ import (
 var _ = (plugin.Provider)((*inmemoryProvider)(nil))
 
 type inmemoryProvider struct {
+	pulumirpc.UnimplementedResourceProviderServer
```
Frassle, line 36: Could clean up all the panic unimplemented methods in this file

AaronFriel: I thought about it but I wondered if it would change the behavior in any way. The unimplemented server returns a grpc error rather than panic, yeah?

Does seem unlikely anything depends on these methods panicking.

Frassle: Yeh I think it was just an easier way to write the method, its used in a part of code that doesn't call any of this stuff anyway.

AaronFriel: Ah, the embedding here was incorrect anyway. I've added a PR to pu/pu to add an embeddable `plugin.UnimplementedProvider`: https://github.com/pulumi/pulumi/pull/11618

Frassle (approved):

pkg/tfgen/pluginHost.go
AaronFriel: @Frassle updated this targeting latest Pulumi, embedded `plugin.UnimplementedProvider` and deleted the panicking methods.

PR #677, opened 2022-12-10
Experimental support to use the Terraform parser

Frassle:
This is the start of a new tf2pulumi converter that uses the Terraform parser and object model to parse and understand the Terraform source code.

I've added this as an option when using PULUMI_EXPERIMENTAL, but have extended the eject tests to cover both this new converter and the existing converter.

The new converter works by invoking terraforms config parser and then converting their object model directly into hcl blocks that will result in a valid PCL program (It might be better to convert directly into a PCL object model and then write the program out from that, and maybe that's doable with the hcl2.model package, but it wasn't clear to me how).

We get access to terraforms parser by using a fork of terraform which moves "internal" to "pkg" allowing us to link and use things like the config parser in this repo.


dixler (approved):

t0yv0:
One caveat here is we're imposing build dependency to [pulumi/terraform](https://github.com/pulumi/terraform) fork on all providers. Hoping this should be OK. The prior art in the bridge was depending on code from that repo that got vendored in as opposed to linking in a modifier fork; but we still depend on code from TF CLI alread. 


CC @pulumi/providers  @mikhailshilkov  @lukehoban  just in case.

From earlier, the stats are compelling and I'd like to err on the side of shipping:

```
Baseline     2328
Provider:     google-beta
Success rate: 95.16% (13144/13812)

Converted 93.96% of csharp examples (2163/2302)
Converted 93.92% of go examples (2162/2302)
Converted 97.87% of java examples (2253/2302)
Converted 93.96% of python examples (2163/2302)
Converted 93.96% of typescript examples (2163/2302)
Converted 97.31% of yaml examples (2240/2302)

Experimental 2328
Provider:     google-beta
Success rate: 99.69% (13464/13506)

Converted 99.78% of csharp examples (2246/2251)
Converted 99.73% of go examples (2245/2251)
Converted 99.60% of java examples (2242/2251)
Converted 99.78% of python examples (2246/2251)
Converted 99.78% of typescript examples (2246/2251)
Converted 99.47% of yaml examples (2239/2251)
```

t0yv0 (approved):

Frassle:
> LMK what are your plans here, are you merging this, or still working on this, or done and need me to resolve conflicts and merge thanks.

It's pretty stable, it could be merged. I'm only still working on it because it makes a little more sense than opening PRs against a PR for every little change I do.

Shouldn't be any conflicts against master currently, I've been keeping this branch up-to-date.

t0yv0:
I'll check tomorrow for any further comments on the dependency above, and if no comments I'd like this merged so I can release a bridge update with it inside and start rolling it out. Thanks!

t0yv0 (changes requested):
Blocking till we decide on the dependency question above.

Frassle:
Superceded by https://github.com/pulumi/pulumi-converter-terraform/pull/75

PR #689, opened 2022-12-14
Fix complex_resource/main.tf

Frassle:


Frassle:
Amusingly the current converter somehow trundled through this invalid Terraform, but the new converter flagged up an error.

PR #690, opened 2022-12-14
Fix issues around confused nested block encoding

t0yv0:
Adding test cases we recently started hitting panics in code that computes renames (panics coming from paths.go) and introduced a workaround. This PR reverts the workaround and clarifies the code, fixing the underlying issue in paths computation. Although renames and paths are not yet used in prod code, I'd like to keep using them for Plugin Framework support.

I took the liberty to move around some code in makePropetyType to clarify the case analysis as I was getting quite confused about the cases there. The only functional change should be the computation of typePath fixes for the recursion. 

TestRegress611 and Test_GenerateTestDataSchemas give some good coverage here.


pkg/tfgen/internal/paths/paths.go
```diff
@@ -285,8 +285,7 @@ func (p *ElementPath) UniqueKey() string {
 }
 
 func (p *ElementPath) String() string {
-	// TODO This seems wrong but it gets my tests passing!?
-	return p.parent.String()
+	return fmt.Sprintf("%s.$", p.parent.String())
```
t0yv0, line 288: This is the workaround revert.

t0yv0:
Interesting downstream failures! The asserts are too strict hm... It sounds like in the case of random provider 

```
panic: at resource[key="random_id",token="random:index/randomId:RandomId"].outputs.keepers: sch.Elem() == nil should imply sch.Type() having a scalar type, but got 6
```

with 6 = TypeMap. So it's a Map with "nil" elementType, like unknown element type. Fascinating.

pkg/tfgen/generate.go
```diff
@@ -322,50 +322,72 @@ func (g *Generator) makePropertyType(typePath paths.TypePath,
...
-	case shim.TypeSet:
-		t.kind = kindSet
+		return t
 	}
 
-	// We should carry across any of the deprecation messages, to Pulumi, as per Terraform schema
-	if sch.Deprecated() != "" && elemInfo != nil {
-		elemInfo.DeprecationMessage = sch.Deprecated()
+	// Handle single-nested blocks next.
+	if blockType, ok := sch.Elem().(shim.Resource); ok && sch.Elem() == shim.TypeMap {
```
t0yv0, line 347: Blip, this should say sch.Type() == shim.TypeMap, AND the test didn't caught this, so we need a test to cover single-nested blocks.

pkg/tfgen/generate.go
```diff
@@ -322,50 +322,72 @@ func (g *Generator) makePropertyType(typePath paths.TypePath,
...
+	// IsMaxItemOne lists and sets are flattened, transforming List[T] to T. Detect if this is the case.
+	flatten := false
+	switch sch.Type() {
+	case shim.TypeList, shim.TypeSet:
+		if tfbridge.IsMaxItemsOne(sch, info) {
+			flatten = true
+		}
+	}
+
+	// The remaining cases are collections, List[T], Set[T] or Map[T], and recusion needs NewElementPath except for
```
t0yv0, line 360: s/recusion/recursion/

pkg/tfgen/generate_test.go
```diff
@@ -132,3 +136,112 @@ func Test_GenerateTestDataSchemas(t *testing.T) {
 		bridgetesting.AssertEqualsJSONFile(t, schemaPath, schema)
 	}
 }
+
+// Encoding of Elem() and Type() is tricky when recognizing types, cover all the cases here documented in the
+// shim.Schema.Elem() docstring.
+func Test_makePropertyType(t *testing.T) {
```
t0yv0, line 142: @guineveresaenger thanks for insisting on proper unit tests cases here, here we go. Covering the interesting bits.

PR #713, opened 2023-01-03
make/build: Don't install test provider globally

abhinav:
`make build` currently runs `go install` with `cmd/...`,
which ends up installing the test provider globally.
This is unnecessary since this provider is needed for tests only.

To resolve this, move the test provider into internal/testing.
We now build it manually and pass its path on to tests.

Another option that was considered was putting bin/ onto PATH, 
but unfortunately, it's extremely difficult to add a new entry to PATH
from the Makefile in a cross-platform way.


AaronFriel (changes requested):
Hmm. I was torn on this because I didn't know if this is a good precedent for other repos. The nice thing about that is we can rely on $GOBIN being "developer loop output", and I'm at least happy to periodically `rm ~/go/bin/pulumi*`.

My dev workflow and I expect some of our cross-repo tests rely on a single global install path. For my dev workflow, I use a Go workspace and a tool like `watchexec` to build binaries and run programs involving N repos.

But a few things stood out to me:

 * "make build" shouldn't _install_ anything
 * This repo is only a library, so there shouldn't be anything to build
 * The only item built, in `./cmd`, is used exclusively for tests in this repo

I think to avoid any precedent setting, let me know what's your opinion of these changes:

First moving this bin out of `cmd` because it's not a real provider we want people building:

```sh
mv cmd/pulumi-terraform-bridge-test-provider internal/testing/
```

And then modify the makefile like so:

```Makefile
build::
  # kept for build scripts/CI, as this is a library
  go mod tidy
  go vet ./pkg/...
  go vet ./internal/...

test::
  # explicitly build the test provider as part of tests and add the path
  go build -o bin ./internal/testing/pulumi-terraform-bridge-test-provider
  PATH="${PWD}/bin:${PATH}" go test -v -count=1 -cover -timeout 2h -parallel ${TESTPARALLELISM} ./...
```

abhinav:
Sure, that sounds reasonable!

abhinav:
@AaronFriel Taking a closer look to how this is used, another option is to eliminate the extraneous binary completely.
We can put that code inside a function inside the only test file that uses this, create a copy of the test executable with the name "pulumi-terraform-bridge-test-provider" and dispatch on `os.Args[0]` in TestMain.

Basically,

```go
func testProvider() {
  plugin.Serve(...)
}

func TestMain(m *testing.M) {
  if filepath.Base(os.Args[0]) == "pulumi-terraform-bridge-test-provider" {
    testProvider()
    os.Exit(0)
  }
  os.Exit(m.Run())
}
```

So for the tests that need this, they'll create a copy of the test executable to a temporary directory with that name, and everything else remains the same.

If this sounds reasonable, I can follow-up with that instead.

AaronFriel:
Saw your edit, that makes sense. I think we'd want a separate binary as that's part of our plugin model, and `TestMain` is maybe a step too magical/surprising for people onboarding to this repository.

AaronFriel (approved):
LGTM

On other PRs I've seen that the GCP downstream tests might be failing/timing out and that's being separately triaged. If that's the only hold out I'm OK with merging.

AaronFriel:
@abhinav 

Hmm, I think in other places we've assumed the shell is bash/sh-like, even setting it in CI: 

```yaml
defaults:
  run:
    shell: bash
```

Think that's the path of least resistance here?

https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/3832453619/jobs/6522774305#step:8:3

abhinav:
> @abhinav
> 
> Hmm, I think in other places we've assumed the shell is bash/sh-like, even setting it in CI:
> 
> ```yaml
> defaults:
>   run:
>     shell: bash
> ```
> 
> Think that's the path of least resistance here?
> 
> https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/3832453619/jobs/6522774305#step:8:3

Yeah, I was just about to respond about that. 
If we're assuming bash/sh elsewhere, then yeah, that sounds like the simplest route.
I've made that change for all workflows that run `make test` in this repo.

Longer-term, I think it would be better to have something more platform agnostic on the test side.
Maybe running `go build` on the provider code from inside the test is an option.

abhinav:
Yeah, that bash thing didn't work.
I'll just make it run `go build` as part of the test.

AaronFriel (approved):

Issue #120, opened 2020-02-26
Improve GoDocs 


lukehoban:
Our GoDocs have a few issues:
- [ ] Lots of `nolint: lll`
- [ ] Markdown formatting not rendered looks a little awkward
- [ ] Sub-packages don't have any package level comments at all
- [ ] Most types/functions have no doc comments at all

See for example https://godoc.org/github.com/pulumi/pulumi-aws/sdk/go/aws.

lukehoban:
The first two are fixed in the latest codegen - which has been fixed up in e.g. https://pkg.go.dev/github.com/pulumi/pulumi-aws/sdk@v1.31.0/go/aws?tab=doc.

The other two are less critical - will push out to post GA.

PR #717, opened 2023-01-06
golangci-lint: Enable nolintlint

abhinav:
The nolintlint linter included with golangci-lint
finds bad usages of the `//nolint` directive
such as `// nolint:foo` (leading space).
Such incorrect usage ends up in the package's godocs (see #120).

This change enables the linter and fixes all occurrences
discovered by golangci-lint.

Depends on #716

Note that #716 is the base for this PR.
When that merges, the base branch for this will be updated.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -17,6 +17,7 @@ linters:
     - megacheck
     - misspell
     - nakedret
+    - nolintlint
     - revive
     - unconvert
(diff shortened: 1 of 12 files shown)


RobbieMcKinstry (approved):
LGTM!

AaronFriel (approved):

PR #718, opened 2023-01-10
Fail build on missing data sources, like for resources

thomas11:
This PR replicates #616 but for data sources. Without this, it's deceptively easy to assume that when there are no errors, all mappings are complete, but it would only be true for resources.


t0yv0 (approved):

t0yv0:
All things equal stricter seems better at this point. Thank you.

PR #725, opened 2023-01-18
Support Plugin Framework v1.1.1

t0yv0:
This PR does a quick upgrade of the Plugin Framework support to v1.1.1. It is just enough to pass the previous test suite. Still would be good to read the [changelog](https://github.com/hashicorp/terraform-plugin-framework/blob/main/CHANGELOG.md#111-january-13-2022) to see if there's anything new the bridge is missing altogether, possibly CustomTypes support.

- [x] fill in the stubs in pfutils
- [x] rewrite plan modifiers used in test
- [x] test code does not compile
- [x] no diff on `make build.testproviders`
- [x] make test
- [x] make test.integration


pkg/tfpfbridge/tests/internal/cmd/pulumi-resource-tls/schema.json
```diff
@@ -74,15 +74,7 @@
...
-            "type": "object",
-            "required": [
-                "fromEnv"
-            ],
-            "language": {
-                "nodejs": {
-                    "requiredInputs": []
-                }
-            }
+            "type": "object"
```
t0yv0, line 77: Looks like fromEnv stopped being recognized as required, need to double check why.

t0yv0: OK this is now simply marked `Optional: true` in the upstream version of the tls provider and has nothing to do with the framework changes being made.

t0yv0:
Friendly ping @jazzyfresh would it be possible to review?

jazzyfresh (approved):
LGTM

Thank you for your patience, it took me a bit of time to read through each commit and understand

t0yv0:
> Thank you for your patience, it took me a bit of time to read through each commit and understand

NP at all, please take your time. Quality over speed here. I'm trying to itemize change into smaller chunks going forward so it's less to digest.

PR #734, opened 2023-01-24
Remove python from tfbridge

iwahbe:
We don't seem to use the custom python generation anymore. We should have moved all our python specific output into https://github.com/pulumi/pulumi/pkg/codegen/python.


t0yv0 (approved):
Thank you.

PR #735, opened 2023-01-24
Remove nodejs from tfbridge

iwahbe:
We don't seem to use the custom nodejs generation anymore. We should have moved all our nodejs specific output into [github.com/pulumi/pulumi/pkg/codegen/nodejs](https://github.com/pulumi/pulumi/tree/master/pkg/codegen/nodejs).


t0yv0 (approved):
Nice, thank you.

Issue #729, opened 2023-01-20
[Plugin Framework] PropertyValue should be secret but is not


t0yv0:
### What happened?

When testing Pulumi TLS on top of the Plugin Framework bridge, one of the tests failed with:

```
Diagnostics:
[239](https://github.com/pulumi/pulumi-tls/actions/runs/3969410176/jobs/6803960643#step:21:241)
    tls:index:PrivateKey (my-private-key):
[240](https://github.com/pulumi/pulumi-tls/actions/runs/3969410176/jobs/6803960643#step:21:242)
      error: objectEncoder failed on property "private_key_pem_pkcs8": PropertyValue should be secret but is not
[241](https://github.com/pulumi/pulumi-tls/actions/runs/3969410176/jobs/6803960643#step:21:243)
  
[242](https://github.com/pulumi/pulumi-tls/actions/runs/3969410176/jobs/6803960643#step:21:244)
    pulumi:pulumi:Stack (private-key-tls-dotnet-p-it-fv-az438-2-dotnet-cfad51c7):
[243](https://github.com/pulumi/pulumi-tls/actions/runs/3969410176/jobs/6803960643#step:21:245)
      error: update failed
[244](https://github.com/pulumi/pulumi-tls/actions/runs/3969410176/jobs/6803960643#step:21:246)
```

### Steps to reproduce

See https://github.com/pulumi/pulumi-tls/actions/runs/3969410176/jobs/6803960643#step:21:240

### Expected Behavior

Either something is wrong with the bridge, or something is wrong with the user program after the provider got tighter with marking properties as secrets. In the latter case ideally we should have an error message that directs the user how to fix the program.

### Actual Behavior

Confusing error message.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
Root caused as a race in Go SDK. https://github.com/pulumi/pulumi/issues/11971

Can work around by loosening the condition in the TFPF bridge to automatically wrap values as secrets if they are not wrapped.

PR #736, opened 2023-01-24
Relax secret encoder checks

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/729

Instead of erroring, automatically coerce secrets to secret form if they are not wrapped already.


iwahbe (approved):
https://github.com/pulumi/pulumi/issues/11971 scares me, but this seems like a reasonable patch.

PR #740, opened 2023-01-25
Add interface for inmemory doc edits

iwahbe:


iwahbe:
Something like this eventually went in as:
https://github.com/pulumi/pulumi-terraform-bridge/blob/81c1dfbd579ac17c3f5bfebb4cbbfb95ddde5753/pkg/tfbridge/info/info.go#L119-L120

PR #741, opened 2023-01-25
Plugin framework API cleanup 

t0yv0:
This PR reduces the exposed API surface and documents the public API a little better.

- no logical changes
- deduplicate asking the user for Version - use ProviderInfo.Version
- custom Info classes removed in favor of unifying with the tfbridge.ProviderInfo
- schemashim, pfutils moved to internal
- tfpfbridge and tfpfbridge/tfgen are the only two packages exposed
- tfpfbridge now has package name matching folder name as you'd expect
- these now have package comments
- helper ProviderMetadata type so the users exploring in IDE get some pointers to where it's coming from
- renames.json moved to bridge-metadata.json to future-proof doing some changes over this opaque blob
- all non-essential members from tfpfbridge are hidden as private


pkg/tfpfbridge/info.go
```diff
@@ -0,0 +1,33 @@
...
+
+package tfpfbridge
+
+import (
+	pfprovider "github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+)
+
+// Configures Pulumi provider metadata and bridging options.
+type ProviderInfo struct {
```
t0yv0, line 23: This is the new ProviderInfo struct. We don't yet support tall the knobs from tfbridge.ProviderInfo but it seems reasonable to embed. We share tfgen internals with the v3 tfbridge.

pkg/tfpfbridge/metadata.go
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+package tfpfbridge
+
+// Defines bridged provider metadata that is pre-computed at build time with tfgen (tfgen
+// ("github.com/pulumi/pulumi-terraform-bridge/pkg/tfpfbridge/tfgen") and typically made available to the provider
+// binary at runtime with [embed].
+//
+// [embed]: https://pkg.go.dev/embed
+type ProviderMetadata struct {
```
t0yv0, line 22: This is the new helper Metadata type.

pkg/tfpfbridge/tests/internal/cmd/pulumi-tfgen-tls/main.go
```diff
@@ -20,5 +20,5 @@ import (
 )
 
 func main() {
-	tfgen.Main("tls", "4.0.4", testprovider.TlsProvider())
+	tfgen.Main("tls", testprovider.TlsProvider())
```
t0yv0, line 23: Example of new-style tfgen.

pkg/tfpfbridge/tests/internal/cmd/pulumi-resource-tls/main.go
```diff
@@ -24,15 +25,10 @@ import (
...
 
 func main() {
-	tfbridge.Main(
-		"tls",
-		"4.0.4",
-		testprovider.TlsProvider(),
-		schema,
-		renames,
-	)
+	meta := tfbridge.ProviderMetadata{PackageSchema: schema, BridgeMetadata: bridgeMetadata}
```
t0yv0, line 32: Example of new-style provider main.

AaronFriel (approved):
Many renames. Appreciate the callouts in comments and reduction in public surface area.

PR #742, opened 2023-01-26
fix: Allow aiven provider to compile without fork

AaronFriel:
Attempting to use the upstream provider directly fails due to the assert removed here.

I believe this change is safe to make, despite affecting all providers, as the assert causes tfgen to panic.

Background:

The Aiven provider has data sources such as:

> - `pg_partman_bgw__dot__interval` (String)
> - `pg_partman_bgw__dot__role` (String)
> - `pg_stat_statements__dot__track` (String)
(Copied from their changelog.)

The upstream provider uses a pair of routines to map `.` to `__dot__` and vice versa, and the patch in our fork of Aiven modifies that mapping to avoid double underscores:

https://github.com/pulumi/terraform-provider-aiven/commit/296d9e0d

By allowing the bridge to generate the name, the upstream provider can be used without requiring the fork. The casing behavior is preserved, and `__dot__` is cased exactly as `_dot_` was before.


pkg/tfbridge/names.go
```diff
@@ -102,8 +101,6 @@ func TerraformToPulumiName(name string, sch shim.Schema, ps *SchemaInfo, upper b
 	casingActivated := false // tolerate leading underscores
 	for i, c := range name {
 		if c == '_' && casingActivated {
-			// skip underscores and make sure the next one is capitalized.
-			contract.Assertf(!nextCap, "Unexpected duplicate underscore: %v", name)
```
iwahbe, line 106: Nit: Might be nice to have a comment explaining that we will ignore duplicate `_`.

iwahbe (approved):
LGTM

pkg/tfbridge/names.go
AaronFriel: Added

Issue #747, opened 2023-01-26
[Plugin Framework] Support Private State


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

[Private State Management](https://developer.hashicorp.com/terraform/plugin/framework/resources/private-state) feature of Terraform providers should carry over to Pulumi-bridged providers.

PR #760, opened 2023-01-27
Ensure that Update preserves schema version

t0yv0:


pf/tfbridge/provider_update.go
```diff
@@ -96,10 +96,15 @@ func (p *provider) Update(
 	}
 
 	// TODO handle resp.Private
```
jazzyfresh, line 98: should we file a ticket for this

jazzyfresh (approved):

pf/tfbridge/provider_update.go
t0yv0: This is https://github.com/pulumi/pulumi-terraform-bridge/issues/747 but you are correct that it should be cross-linked to the TODO. I'll do a sweep to cross-link TODOs as a separate PR.

PR #761, opened 2023-01-28
Ensure that Read preserves schema version

t0yv0:
Make sure Read():

- performs state upgrades (this seems to be part of sdk-v2 behavior, we probably have to)
- notes the state version in what it returns to the Pulumi CLI

The test was extracted from a `pulumi refresh`. We might need to test "get not refresh" scenario based on some if-then-else code I've seen in the v3/tfbridge.  Not sure yet which high-level Pulumi command it corresponds to.

I've also not had time make sure the input pieces are implemented, need to ticket that.

### DIFF
--- a/pf/tests/provider_read_test.go
+++ b/pf/tests/provider_read_test.go
@@ -28,9 +28,32 @@ func TestReadFromRefresh(t *testing.T) {
 	// Specifically testing for:
 	//
 	// - __meta writing out the schema version
+	// - implicit upgrade from version 1 to version 3 is performed ("numeric": true) appears
 	// - inputs being populated
+
 	server := newProviderServer(t, testprovider.RandomProvider())
-	testCase := `
+
+	testCase := `[
+	{
+	  "method": "/pulumirpc.ResourceProvider/Configure",
+	  "request": {
+	    "args": {
+	      "version": "4.8.0"
+	    },
+	    "acceptSecrets": true,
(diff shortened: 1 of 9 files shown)


pf/tfbridge/provider_read.go
```diff
@@ -80,25 +79,25 @@ func (p *provider) Read(
...
-	readID, err := rh.idExtractor.extractID(readResourceStateValue)
+	readStateMap, err := readState.ToPropertyMap(&rh)
 	if err != nil {
 		return plugin.ReadResult{}, 0, err
 	}
 
 	return plugin.ReadResult{
-		ID: resource.ID(readID),
+		ID: readID,
 		// TODO support populating inputs, see extractInputsFromOutputs in the prod bridge.
```
jazzyfresh, line 99: should we file a ticket for this

pf/tests/provider_read_test.go
```diff
@@ -0,0 +1,153 @@
...
+		"value": "Ps7XGKxa"
+	      },
+	      "special": true,
+	      "upper": true
+	    },
+            "inputs": {}
+	  }
+	}]`
+
+	// TODO populate inputs
```
jazzyfresh, line 129: ticket

jazzyfresh (approved):
Let's make sure we have tickets for all the action items you mentioned, even if it's design or more research required

PR #762, opened 2023-01-28
Token mapping quality of life

iwahbe:
Having used the token mapping library for a week, this PR introduces a few quality of life improvements.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -104,11 +104,16 @@ type ProviderInfo struct {
 // NOTE: Experimental; We are still iterating on the design of this function, and it is
 // subject to change without warning.
 func (info *ProviderInfo) ComputeDefaults(r ResourceStrategy, d DatasourceStrategy) error {
+	var errs multierror.Error
 	err := info.ComputeDefaultResources(r)
 	if err != nil {
-		return err
+		errs.Errors = append(errs.Errors, fmt.Errorf("resources:\n%w", err))
 	}
-	return info.ComputeDefaultDataSources(d)
+	err = info.ComputeDefaultDataSources(d)
+	if err != nil {
+		errs.Errors = append(errs.Errors, fmt.Errorf("datasources:\n%w", err))
+	}
+	return errs.ErrorOrNil()
 }
 
 // Apply strategy to generate missing resources.
--- a/pkg/tfbridge/token.go
+++ b/pkg/tfbridge/token.go
@@ -57,7 +57,8 @@ type MakeToken func(module, name string) (string, error)
 // Convert a Terraform token to a Pulumi token with the standard mapping.
 //
 // The mapping is
-//   (pkg, module, name) => pkg:module/lowerFirst(name):name
+//
+//	(pkg, module, name) => pkg:module/lowerFirst(name):name
 //
 // NOTE: Experimental; We are still iterating on the design of this function, and it is
 // subject to change without warning.
--- a/pkg/tfbridge/token.go
+++ b/pkg/tfbridge/token.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"unicode"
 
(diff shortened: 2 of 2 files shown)


pkg/tfbridge/info.go
```diff
@@ -104,11 +104,16 @@ type ProviderInfo struct {
...
 	if err != nil {
-		return err
+		errs.Errors = append(errs.Errors, fmt.Errorf("resources:\n%w", err))
+	}
+	err = info.ComputeDefaultDataSources(d)
+	if err != nil {
+		errs.Errors = append(errs.Errors, fmt.Errorf("datasources:\n%w", err))
 	}
-	return info.ComputeDefaultDataSources(d)
+	return errs.ErrorOrNil()
```
t0yv0, line 116: 👍

pkg/tfbridge/token.go
```diff
@@ -53,6 +54,20 @@ type Strategy[T ResourceInfo | DataSourceInfo] func(tfToken string, tfTokens []s
 // subject to change without warning.
 type MakeToken func(module, name string) (string, error)
 
+// Convert a Terraform token to a Pulumi token with the standard mapping.
```
t0yv0, line 57: 👍

t0yv0 (approved):
LGTM

pkg/tfbridge/token.go
```diff
@@ -53,6 +54,20 @@ type Strategy[T ResourceInfo | DataSourceInfo] func(tfToken string, tfTokens []s
...
+//
+// The mapping is
+//   (pkg, module, name) => pkg:module/lowerFirst(name):name
+//
+// NOTE: Experimental; We are still iterating on the design of this function, and it is
+// subject to change without warning.
+func MakeStandardToken(pkgName string) MakeToken {
```
t0yv0, line 64: Is this different from `MakeResource` and `MakeDataSource`?

iwahbe: It's more honest that this is all string manipulation, but `MakeStandardToken(pkg)(mod, name)` is equivalent to `string(MakeResource(pkg, mod, name))` which is itself equivalent to `string(MakeDataSource(pkg, mod, name))`.

t0yv0: I disagree with strings being more honest - these are not free-form strings, and having a type marker gives me a place to "jump to definition" and find the documentation and tests of what the permissible values are. There is at least one bug I fixed in tf-bridge by systematically pushing the token types in place of string types and finding a semantic mismatch where the codebase crossed the wires on "strongly typed" logic.'

Also hmm MakeStandardToken does look equivalent to MakeResource, MakeDataSource but having all of them in th module doesn't make sense to me. Should we mark some of them as deprecated?

iwahbe: I didn't say it's better to treat them as free form strings, I said it's more honest to admit that we are already doing so. `MakeResource` and `MakeDataSource` accept strings. They do some basic string manipulation and then call `MakeMember`, which does raw string manipulation with a cast to a `tokens` type. At no point do they call any token manipulation functions.

This is what `MakeResource` looks like with all `tfbridge` functions inlined:

```go
func MakeResource(pkg string, mod string, res string) tokens.Type {
	fn := string(unicode.ToLower(rune(res[0]))) + res[1:]
        stk0 := mod+"/"+fn
	return tokens.Type(tokens.ModuleMember(pkg + ":"+ stk0 + ":" + res))
}
```

All calls to `tokens` are just casts, there is no additional validation. Rewritten to include validation from `tokens`, `MakeResource` looks like this:

```go
func MakeResource(pkg string, mod string, res string) tokens.Type {
	fullMod := mod + "/" + string(unicode.ToLower(rune(res[0]))) + res[1:]
	pkgT := tokens.NewPackageToken(tokens.PackageName(pkg))
	modT := tokens.NewModuleToken(pkgT, tokens.ModuleName(fullMod))
	return tokens.NewTypeToken(modT, tokens.TypeName(res))
}
```

PR #764, opened 2023-01-30
WIP: Automapping api ideas

t0yv0:


pkg/tfbridge/info_defaults.go
```diff
@@ -0,0 +1,180 @@
...
+	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+
+	multierror "github.com/hashicorp/go-multierror"
+)
+
+// SetDefaults populates mappings such as the Resources mapping by computing sensible defaults in code. This is
+// especially useful for larger providers that have thousands of entries in Resources. Manually specified mappings
+// always take precedence: SetDefaults will never override an entry in the Resources map, only add new entries.
+func (info *ProviderInfo) SetDefaults(defaults ProviderInfoDefaults) error {
```
t0yv0, line 32: ComputeDefaults was a good name, just picking something here so it can exist side by side for discussion.

pkg/tfbridge/info_defaults.go
```diff
@@ -0,0 +1,180 @@
...
+			info.DataSources = map[string]*DataSourceInfo{}
+		}
+		if err := setDefaults(info.P.DataSourcesMap(), info.DataSources, defaults.DataSources); err != nil {
+			errs.Errors = append(errs.Errors, fmt.Errorf("datasources:\n%w", err))
+		}
+	}
+	return errs.ErrorOrNil()
+}
+
+type ProviderInfoDefaults struct {
```
t0yv0, line 54: This is a struct on purpose so we can easily extend it with more things in the future.

pkg/tfbridge/info_defaults.go
```diff
@@ -0,0 +1,180 @@
...
+	Resources   func(resourceToken string) (*ResourceInfo, error)
+	DataSources func(dataSourceToken string) (*DataSourceInfo, error)
+}
+
+func (d ProviderInfoDefaults) Unmappable(substring string) ProviderInfoDefaults {
```
t0yv0, line 59: Just copying this from the previous design, curious did his turn out useful ? 

ALT is user explicitly doing:

```
Resources: func(resourceToken string) (*ResourceInfo, error) {
    if strings.Contans(resourceToken, "foo") {
        return nil, fmt.Errorf("Refusing to map...")
    }
    ...
}
```

pkg/tfbridge/info_defaults.go
```diff
@@ -0,0 +1,180 @@
...
+	Name   string
+}
+
+type TFTokenParser func(tfToken string) (*TFToken, error)
+
+func SingleModule(prefix, defaultModule string) TFTokenParser {
+	return KnownModules(prefix, defaultModule, nil)
+}
+
+func KnownModules(prefix, defaultModule string, modules []string) TFTokenParser {
```
t0yv0, line 90: So it looks to me that this is the essence of SingleModule/KnownModule strategies in the base branch. It tries to guess "terraform module" from the token, then feeds the result into standard token manglers. Perhaps this is useful then as its own mini-abstraction?

I'm thinking that on the base branch it's pre-packaged into types that carry ResourceInfo, when it's really only ever dealing with tokens, it feels a bit more natural to unpack it to only work on tokens, but YMMV it's not a big deal; the user can always compensate with a few more lambdas.

pkg/tfbridge/info_defaults.go
```diff
@@ -0,0 +1,180 @@
...
+	mod := camelCase(token.Module)
+	res := upperCamelCase(token.Name)
+	return MakeResource(pkg, mod, res)
+}
+
+func SuggestDataSourceToken(pkg string, token *TFToken) tokens.ModuleMember {
+	mod := camelCase(token.Module)
+	fn := upperCamelCase(token.Name)
+	submod := string(unicode.ToLower(rune(fn[0]))) + fn[1:]
+	return MakeMember(pkg, mod+"/"+submod, "get"+fn)
```
t0yv0, line 149: So this bit I found on base branch I was surprised by. Is there a story for the "get" prefix? Is there a reason that MakeDataSource doesn't do the "get" prefix? Is this how it's conventionally done?

pkg/tfbridge/info_defaults_test.go
```diff
@@ -0,0 +1,145 @@
...
+				"foo_very_special_source": {},
+			},
+		},
+	}
+
+	err := info.SetDefaults(tfbridge.SuggestTokens("foo", tfbridge.SingleModule("foo_", "index")))
+	assert.NoError(t, err)
+
+	expectedResources := map[string]*tfbridge.ResourceInfo{
+		"foo_fizz_buzz":       {Tok: "foo:index/fizzBuzz:FizzBuzz"},
```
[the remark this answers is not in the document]
iwahbe: Modern Pulumi preferred `pkg:index:Resource` as it allows a cleaner separation between token and module. For backwards compatibility reasons, we need to support `pkg:index/resource:Resource`. The second scheme is used in almost all TF providers by convention.

pkg/tfbridge/info_defaults.go
iwahbe: I wanted it to work on `ResourceInfo` because I'm planning to follow up with support for stateful remappings (which benefit from alias support).

pkg/tfbridge/info_defaults.go
```diff
@@ -0,0 +1,180 @@
...
+				t, substring)
+		}
+		return f(t)
+	}
+}
+
+type TFToken struct {
+	Module string
+	Name   string
+}
```
iwahbe, line 82: I like the idea of having an intermediate representation here.

pkg/tfbridge/info_defaults.go
t0yv0: Ah interesting. So you will need strategy combinators for that probably. I'd like to see that before taking a look at the naming again!

t0yv0:
Closing for now. This was useful as an exercise to clarify some issues for me.

PR #766, opened 2023-01-30
Make explicit the tuple type for strategies

iwahbe:
Even though this is technically a breaking change, I only expect the `datadog` provider to require a fix.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -104,7 +104,7 @@ type ProviderInfo struct {
 //
 // NOTE: Experimental; We are still iterating on the design of this function, and it is
 // subject to change without warning.
-type ComputeDefaultInfo struct {
+type DefaultStrategy struct {
 	Resource   ResourceStrategy
 	DataSource DataSourceStrategy
 }
@@ -113,7 +113,7 @@ type ComputeDefaultInfo struct {
 //
 // NOTE: Experimental; We are still iterating on the design of this function, and it is
 // subject to change without warning.
-func (info *ProviderInfo) ComputeDefaults(opts ComputeDefaultInfo) error {
+func (info *ProviderInfo) ComputeDefaults(opts DefaultStrategy) error {
 	var errs multierror.Error
 	err := info.computeDefaultResources(opts.Resource)
 	if err != nil {
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -99,39 +99,47 @@ type ProviderInfo struct {
 	PreConfigureCallbackWithLogger PreConfigureCallbackWithLogger
 }
 
+// Describe the mapping from resource and datasource tokens to Pulumi resources and
+// datasources.
+//
+// NOTE: Experimental; We are still iterating on the design of this function, and it is
+// subject to change without warning.
+type ComputeDefaultInfo struct {
+	Resource   ResourceStrategy
+	DataSource DataSourceStrategy
+}
+
 // Add mapped resources and datasources according to the given strategies.
 //
 // NOTE: Experimental; We are still iterating on the design of this function, and it is
 // subject to change without warning.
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/info.go
```diff
@@ -99,39 +99,47 @@ type ProviderInfo struct {
...
 	}
 	return errs.ErrorOrNil()
 }
 
-// Apply strategy to generate missing resources.
-//
-// NOTE: Experimental; We are still iterating on the design of this function, and it is
-// subject to change without warning.
-func (info *ProviderInfo) ComputeDefaultResources(strategy ResourceStrategy) error {
+func (info *ProviderInfo) computeDefaultResources(strategy ResourceStrategy) error {
```
iwahbe, line 129: Now that we have a struct for passing strategies, we never have a reason to expose these helpers. We may as well remove them during this phase.

pkg/tfbridge/token.go
```diff
@@ -34,13 +34,13 @@ type ResourceStrategy = Strategy[ResourceInfo]
...
 // subject to change without warning.
-type DatasourceStrategy = Strategy[DataSourceInfo]
+type DataSourceStrategy = Strategy[DataSourceInfo]
 
 // A generic remapping strategy.
 //
 // NOTE: Experimental; We are still iterating on the design of this type, and it is
 // subject to change without warning.
-type Strategy[T ResourceInfo | DataSourceInfo] func(tfToken string, tfTokens []string) (*T, error)
+type Strategy[T ResourceInfo | DataSourceInfo] func(tfToken string) (*T, error)
```
iwahbe, line 43: No strategy ever used the "all tokens" argument, so I'm remove it. Again, while this is technically breaking, no providers will be impacted.

pkg/tfbridge/token.go
```diff
@@ -121,40 +121,55 @@ func tokensKnownModules[T ResourceInfo | DataSourceInfo](
...
+				tk, err := finalize(mod, "get"+tk)
+				if err != nil {
+					return nil, err
+				}
+				return &DataSourceInfo{Tok: tokens.ModuleMember(tk)}, nil
+			}),
+	}
+}
+
+func (ts ComputeDefaultInfo) Unmappable(substring, reason string) ComputeDefaultInfo {
```
iwahbe, line 150: In practice, marking something as unmappable without explaining why leads to confusion later down the line. I'm augmenting this API to allow the user to explain why something is unmappable.

pkg/tfbridge/info.go
t0yv0: Yeah! I think either use a struct or remove ComputeDefaults completely and only provide the individual helpers, keep surface area a bit smaller! I think the struct is appropriate as there are the multierror helpers baked in.

pkg/tfbridge/token.go
t0yv0: SGTM.

t0yv0 (approved):

pkg/tfbridge/info.go
```diff
@@ -99,39 +99,47 @@ type ProviderInfo struct {
 	PreConfigureCallbackWithLogger PreConfigureCallbackWithLogger
 }
 
+// Describe the mapping from resource and datasource tokens to Pulumi resources and
+// datasources.
+//
+// NOTE: Experimental; We are still iterating on the design of this function, and it is
+// subject to change without warning.
+type ComputeDefaultInfo struct {
```
t0yv0, line 107: Nit: I don't like the *Info name. It puts it as a synonym of ResourceInfo in my head which is not as helpful. Perhaps. DefaultsStrategy?

iwahbe: Changed in https://github.com/pulumi/pulumi-terraform-bridge/pull/766/commits/8145571811581e9e848bcaccbba2527b956079a0

PR #767, opened 2023-01-30
Use tokens.* functions for creating tokens

iwahbe:
This adds validation to bridged tokens. Invalid tokens now panic, instead of being ignored.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -888,11 +888,17 @@ func GetModuleMajorVersion(version string) string {
 }
 
 // MakeMember manufactures a type token for the package and the given module and type.
+//
+// Deprecated: Use MakeResource or call into the `tokens` module in
+// "github.com/pulumi/pulumi/sdk/v3/go/common/tokens" directly.
 func MakeMember(pkg string, mod string, mem string) tokens.ModuleMember {
 	return tokens.ModuleMember(pkg + ":" + mod + ":" + mem)
 }
 
 // MakeType manufactures a type token for the package and the given module and type.
+//
+// Deprecated: Use MakeResource or call into the `tokens` module in
+// "github.com/pulumi/pulumi/sdk/v3/go/common/tokens" directly.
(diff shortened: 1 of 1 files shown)


pkg/tfbridge/info.go
```diff
@@ -888,29 +888,64 @@ func GetModuleMajorVersion(version string) string {
 }
 
 // MakeMember manufactures a type token for the package and the given module and type.
+//
+// Depreciated: Use MakeResource or call into the `tokens` module in
+// "github.com/pulumi/pulumi/sdk/v3/go/common/tokens" directly.
```
t0yv0, line 893: It think the word you're looking for is spelled "Deprecated".

pkg/tfbridge/info.go
```diff
@@ -888,29 +888,64 @@ func GetModuleMajorVersion(version string) string {
...
+//
+// Invalid inputs panic.
+func makeModule(pkg, mod, member string) tokens.Module {
+	mod += "/" + string(unicode.ToLower(rune(member[0]))) + member[1:]
+	if !tokens.IsQName(pkg) {
+		panic(fmt.Errorf("invalid pkg name: '%s'", pkg))
+	}
+	pkgT := tokens.NewPackageToken(tokens.PackageName(pkg))
+	if !tokens.IsQName(mod) {
+		panic(fmt.Errorf("invalid module name: '%s'", mod))
```
t0yv0, line 933: It's a bit ruthless to panic on invalid inputs but we can't change the signature to `(T, err)` in a non-breaking way so that's probably an improvement overall, justified move.

iwahbe: Non-compliant tokens will fail during schema checks later down the line. I agree it's ruthless, but at least it won't break any existing providers. It also matches the way the `tokens` module is written (panicking on invalid inputs).

t0yv0: Yes. LGTM - good move for now.

PR #768, opened 2023-01-30
Fixes #720: ignore invalid defaults

### LINKED ISSUE #720: Error on creating azure.storage.ManagementPolicy
### What happened?

Trying to deploy a `azure.storage.ManagementPolicy` resource. If not all *optional* parameters for actions are set explicitly, an error is thrown stating that `-1` is not in the range of acceptable values.

image

### Steps to reproduce

1. Create Storage Account
2. Create Storage Account Container `azure.storage.Container`
3. Create Storage Account Management Policy as described in the current documentation https://www.pulumi.com/registry/packages/azure/api-docs/storage/managementpolicy/ and not set all possible action values.

```typescript
  const exampleManagementPolicy = new azure.storage.ManagementPolicy("managementPolicyRobotFramework", {
    storageAccountId: args.storageAccount.id,
    rules: [
      {
        name: "robot-framework",
        enabled: true,
        filters: {
          prefixMatches: [
            containerName,
          ],
          blobTypes: [
            "blockBlob",
          ],
        },
        actions: {
          baseBlob: {
            deleteAfterDaysSinceModificationGreaterThan: 30,
            // deleteAfterDaysSinceCreationGreaterThan: 0,
            // deleteAfterDaysSinceLastAccessTimeGreaterThan: 0,
            // tierToArchiveAfterDaysSinceCreationGreaterThan: 0,
            // tierToArchiveAfterDaysSinceLastAccessTimeGreaterThan: 0,
            // tierToArchiveAfterDaysSinceLastTierChangeGreaterThan: 0,
            // tierToArchiveAfterDaysSinceModificationGreaterThan: 0,
            // tierToCoolAfterDaysSinceCreationGreaterThan: 0,
            // tierToCoolAfterDaysSinceLastAccessTimeGreaterThan: 0,
            // tierToCoolAfterDaysSinceModificationGreaterThan: 0,
          },
          snapshot: {
            deleteAfterDaysSinceCreationGreaterThan: 30,
            // changeTierToArchiveAfterDaysSinceCreation: 0,
            // changeTierToCoolAfterDaysSinceCreation: 0,
            // tierToArchiveAfterDaysSinceLastTierChangeGreaterThan: 0,
          },
          version: {
            deleteAfterDaysSinceCreation: 30,
            // changeTierToArchiveAfterDaysSinceCreation: 0,
            // changeTierToCoolAfterDaysSinceCreation: 0,
            // tierToArchiveAfterDaysSinceLastTierChangeGreaterThan: 0,
          }
        }
      }
    ]
  })
```

### Expected Behavior

Storage Account Management Policy  got created by Pulumi

### Actual Behavior

Error thrown while performing `pulumi preview`

image

### Output of `pulumi about`

`pulumi about` with sensitive information removed.

```
CLI          
Version      3.38.0
Go Version   go1.19
Go Compiler  gc

Plugins
NAME     VERSION
azure    5.27.0
azuread  5.32.0
nodejs   unknown
random   4.8.2

Host     
OS       debian
Version  11.4
Arch     x86_64

This project is written in nodejs: executable='/usr/local/bin/node' version='v16.16.0'

Dependencies:
NAME             VERSION
@types/node      12.20.55
md5-file         5.0.0
@pulumi/azure    5.27.0
@pulumi/azuread  5.32.0
@pulumi/pulumi   3.38.0
@pulumi/random   4.8.2
```

### Additional context

_No response_

### Contributing

t0yv0:
Fixes #720 

When an upstream provider defines a schema with a non-nil Default value and a non-nil ValidateFunc check such that the DefaultValue does not pass the validations according to ValidateFunc, the bridge will now ignore the DefaultValue and proceed as if it has been unset or nil. This works around issues in creating azure.storage.ManagementPolicy.

### DIFF
--- a/pkg/tfshim/sdk-v2/defaults.go
+++ b/pkg/tfshim/sdk-v2/defaults.go
@@ -30,14 +30,12 @@ func withPatchedDefaults(s *schema.Schema) *schema.Schema {
 }
 
 func hasInvalidDefault(s *schema.Schema) bool {
-	if s.Default == nil && s.ValidateFunc == nil {
+	if s.Default == nil || s.ValidateFunc == nil {
 		return false
 	}
 	_, errors := s.ValidateFunc(s.Default, "field")
 	for _, err := range errors {
-		glog.V(9).Infof("ignoring a Default value %v that does not validate with ValidateFunc: %v",
-			s.Default,
-			err)
+		glog.V(9).Infof("ignoring an invalid (per ValidateFunc) Default %v: %v", s.Default, err)
 	}
 	return len(errors) > 0
 }
--- a/pkg/tfshim/sdk-v2/defaults.go
+++ b/pkg/tfshim/sdk-v2/defaults.go
@@ -1,0 +1,45 @@
+// Copyright 2016-2023, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sdkv2
+
+import (
+	"github.com/golang/glog"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func withPatchedDefaults(s *schema.Schema) *schema.Schema {
+	if hasInvalidDefault(s) {
+		var schema schema.Schema = *s
+		schema.Default = nil
+		return &schema
+	}
+	return s
+}
+
+func hasInvalidDefault(s *schema.Schema) bool {
+	if s.Default != nil && s.ValidateFunc != nil {
+		_, errors := s.ValidateFunc(s.Default, "field")
+		if len(errors) > 0 {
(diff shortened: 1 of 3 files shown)


pkg/tfshim/sdk-v2/defaults.go
```diff
@@ -0,0 +1,45 @@
...
+			for _, err := range errors {
+				glog.V(9).Infof("ignoring a Default value %v that does not validate with ValidateFunc: %v",
+					s.Default,
+					err)
+			}
+			return true
+		}
+	}
+	return false
+}
```
iwahbe, line 45: Playing code golf, I think this is easier to read:
```suggestion
func hasInvalidDefault(s *schema.Schema) bool {
	if s.Default == nil && s.ValidateFunc == nil {
		return false
	}
	_, errors := s.ValidateFunc(s.Default, "field")
	for _, err := range errors {
		glog.V(9).Infof("ignoring a Default value %v that does not validate with ValidateFunc: %v",
			s.Default,
			err)
	}
	return len(errors) > 0
}
```

iwahbe (approved):
I'm not a domain expert here, but it seems like we are implicitly deciding to drop the default instead of the validation.

I see three responses when `validate(default) is false`:
1. Drop the default
2. Drop the validation
3. Don't valid the default

I think dropping the default is a valid choice, but I would love an explanation on why we went with option (1) instead of (2) or (3).

t0yv0:
The code that runs the validator is outside of the control of tfbridge afaik, it happens inside some provider method that tfbridge ends up calling. So we can't not run the validator, at least not easily. Likewise, if we drop the validator here this will only affect the tfbridge view of the underlying provider, not the actual provider, so the validator still will run and cause the error messages to appear in the user's example.

PR #769, opened 2023-01-31
Respect IgnoredMapping when computing defaults

iwahbe:
We shouldn't provide default mapping for resources listed in `IgnoreMappings`.


jazzyfresh (approved):

aq17 (approved):
LGTM

PR #771, opened 2023-01-31
Split builtins test by experimental/non-experimental

Frassle:
This adds support for `#if EXPERIMENTAL ... #else ... #endif` blocks in the eject tests and updates the builtin tests to make use of that. This allows us to start adding bits to tests that only work in the new experimental converter without needing to split into completely separate tests that only run on experimental runs.

An alternative design is to do something like we've done for "partial" tests, and name some of the test folders "experimental_xyz" and only run those in experimental mode. If we went with that we'd want to split the builtins test (and some others) into two different tests.


pkg/tf2pulumi/convert/testdata/builtins/pcl/main.pp
```diff
@@ -63,4 +66,5 @@
 }
 output funcSplit2 {
   value = split(",", "")
-}
+}#endif
```
AaronFriel, line 69: #endif introduced on its own?

pkg/tf2pulumi/convert/eject_test.go
```diff
@@ -74,6 +78,50 @@ func isTruthy(s string) bool {
 	return s == "1" || strings.EqualFold(s, "true")
 }
 
+func formatTestFile(src io.Reader, dst io.Writer, isExperimental bool) error {
+	scanner := bufio.NewScanner(src)
+	inIf := false
+	inElse := false
+	for scanner.Scan() {
+		line := scanner.Text()
```
iwahbe, line 86: It looks like calls to `scanner.Scan` will give back lines. Let's make sure we aren't effected by whitespace within the line here. 
```suggestion
		line := strings.TrimSpace(scanner.Text())
```

pkg/tf2pulumi/convert/eject_test.go
```diff
@@ -74,6 +78,50 @@ func isTruthy(s string) bool {
 	return s == "1" || strings.EqualFold(s, "true")
 }
 
+func formatTestFile(src io.Reader, dst io.Writer, isExperimental bool) error {
```
iwahbe, line 81: Can we change the name of the function to `expandPragmas` or something similar. I would not call a pre-processor "formatting". I also think it needs a comment explaining what its doing, and the intended grammar.

pkg/tf2pulumi/convert/eject_test.go
```diff
@@ -74,6 +78,50 @@ func isTruthy(s string) bool {
...
+			(inElse && !isExperimental) ||
+			(!inIf && !inElse)
+
+		if doWrite {
+			_, err := dst.Write([]byte(line + "\n"))
+			if err != nil {
+				return err
+			}
+		}
+	}
```
iwahbe, line 121: Let's forbid as much as we can now, since we don't have any backwards compatibility concerns. 
```suggestion
	}
	if inIf {
		return fmt.Errorf("Unclosed if statement")
	}	
	if inElse {
		return fmt.Errorf("Unclosed else statement")
	}
```

scripts/generate_builtins.py
```diff
@@ -40,110 +42,114 @@ def trimext(file: str) -> str:
     ],
 }
 
-# There's a number of functions we _don't_ support yet, so we exclude emitting these to the file
-unsupported = set([
+# There's a number of functions we only support in the new experimental converter
+experimental = set([
```
iwahbe, line 46: Is there any way to tell what is and is not supported, or is it just an exhaustive list? Can this comment point to a place in our code to check if the list is correct?


P.S. Python has set literals:
```suggestion
experimental = {
```

pkg/tf2pulumi/convert/eject_test.go
```diff
@@ -74,6 +78,50 @@ func isTruthy(s string) bool {
...
+	return scanner.Err()
+}
+
```
iwahbe, line 124: Just like the comment block on `TestGeneratePackage`, I would like an explanation of what running `PULUMI_EXPERIMENTAL=true go test ./...` is supposed to do.

iwahbe (commented):
In theory, I think it's fine to have a pre-processor for our tests to split on experimental. We need to make sure it's well documented.

scripts/generate_builtins.py
Frassle: > Is there any way to tell what is and is not supported, or is it just an exhaustive list? 

In this PR nothing. There's a list of handled functions in the new converter, so this could point to that.

AaronFriel:
@Frassle how are we are testing both regular and experimental flows on CI here?

scripts/generate_builtins.py
iwahbe: I meant "Is there a list anywhere in our code base". Pointing to the list of handled functions would be awesome!

Frassle:
> how are we are testing both regular and experimental flows on CI here?

Currently we're not, because this doesn't yet add the new converter to do anything in experimental mode. It just setting up some test framework changes which we're large enough I wanted them considered alone. 

In the PR to add the converter I'll make CI run tests in normal and experimental mode.

scripts/generate_builtins.py
Frassle: > the list of handled functions would be awesome

I mean that's what this script generates :) It grabs every function from TF's docs and unless it's marked "unsupported" here writes it to the builtins.tf test file which the converter then checks can convert to pcl.

pkg/tf2pulumi/convert/testdata/builtins/pcl/main.pp
Frassle: Ha it was writing the pragmas out to the HCL file, which mostly got dropped because we don't parse comments fully but this one somehow survived the translation process 😆 
Fixed it so the converted HCL doesn't have the pragma lines left in it.

t0yv0:
I'm not opposed to maintaining this - the two-way if-then-else can be tolerated in the test files. The other alternatives are heavier weight like spilt test suites or even Git feature branches. The heavier weight would make more sense if there was more than N=2 alternatives to support.

PR #772, opened 2023-01-31
Added FilteringProvider to tfshim/util

t0yv0:
Added FilteringProvider to tfshim/util to assist debugging issues in large providers.


aq17 (approved):
LGTM

Issue #778, opened 2023-02-01
 Pulumi name "base32StringSeed" incorrectly translates back to Terraform as "base32_string_seed"


t0yv0:
### What happened?

There is an issue with translating property names that contain numbers, these property names do not turn around between Pulumi and Terraform. Attempting to use resource with such property names will result in run-time error.

Since v3.39.1 this is detected at build time as a warning. For example, running pulumi-aws gives:

```
warning: Naming error in Terraform Resource "aws_iam_virtual_mfa_device" attribute "base_32_string_seed": Pulumi name "base32StringSeed" incorrectly translates back to Terraform as "base32_string_seed"
warning: Naming error in Terraform Resource "aws_dms_s3_endpoint" attribute "rfc_4180": Pulumi name "rfc4180" incorrectly translates back to Terraform as "rfc4180"
```

It appears that this issue can be solved in pulumi-terraform-bridge itself.

### Expected Behavior

Properties such as base32_string_seed work as expected.

### Steps to reproduce

Run `make tfgen` on `pulumi-aws` against the v3.39.1 version of the bridge.

### Output of `pulumi about`

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
CC @iwahbe following up from yesterday, this wrinkle is still an issue I think.

t0yv0:
@danielrbradley there were a few of these warnings surfaced for AWS that are now fixed with @iwahbe latest change! We'll release a new tfbridge version tomorrow, AWS is now clean from these naming warnings.

PR #780, opened 2023-02-01
[WIP] Invert name tables

t0yv0:
An experiment in making PulumiToTerraformName trivially invertible by tabulating TerraformToPulumiName.

As written this fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/778

But it has some downsides, opening for discussion.

### DIFF
--- a/pkg/tfbridge/names.go
+++ b/pkg/tfbridge/names.go
@@ -28,6 +28,15 @@ import (
 // PulumiToTerraformName performs a standard transformation on the given name string, from Pulumi's PascalCasing or
 // camelCasing, to Terraform's underscore_casing.
 func PulumiToTerraformName(name string, tfs shim.SchemaMap, ps map[string]*SchemaInfo) string {
+
+	if tfs != nil {
+		// Inefficient but improves precision of inverting TerraformToPulumiName.
+		t := pulumiToTerraformNameTable(tfs, ps)
+		if result, ok := t[name]; ok {
+			return result
+		}
+	}
+
 	var result string
 	for i, c := range name {
 		if c >= 'A' && c <= 'Z' {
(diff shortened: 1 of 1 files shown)


pkg/tfbridge/names.go
```diff
@@ -28,6 +28,15 @@ import (
 // PulumiToTerraformName performs a standard transformation on the given name string, from Pulumi's PascalCasing or
 // camelCasing, to Terraform's underscore_casing.
 func PulumiToTerraformName(name string, tfs shim.SchemaMap, ps map[string]*SchemaInfo) string {
+
+	if tfs != nil {
```
t0yv0, line 32: One question I have is do we always have a complete set of TF names passed in `tfs shim.SchemaMap`, or are there situations where we legitimately are trying to "make up a name" on the fly that's not part of the schema, or else we're just forgetting to pass the complete set of TF names?

If we can make this stricter and we are certain we have complete name information available, then the fallback name computation should probably be removed, and this function deprecated into a version that returns an `(string, error)` and errors out if the TF name is not found in the table.

pkg/tfbridge/names.go
```diff
@@ -66,6 +75,17 @@ func PulumiToTerraformName(name string, tfs shim.SchemaMap, ps map[string]*Schem
 	return result
 }
 
+// A strict inverse of TerraformToPulumiName(upper=false) that tabulates it based on the SchemaMap.
+func pulumiToTerraformNameTable(schemaMap shim.SchemaMap, ps map[string]*SchemaInfo) map[string]string {
```
t0yv0, line 79: Calling this as part of PulumiToTerraformName when that loops over all properties is quadratic in properties, which maybe is bad if we have 1000-property situations. There is a way to mitigate this inside the tfbridge though with a bit of refactoring. Let me link another experimental PR shortly.

t0yv0:
Closing in favor of https://github.com/pulumi/pulumi-terraform-bridge/pull/782

PR #781, opened 2023-02-01
[WIP] Refactor getInfoFromPulumiName to iterators

t0yv0:
This PR relates to the efficiency concerns in https://github.com/pulumi/pulumi-terraform-bridge/pull/780 inverting the name tables. As it turns out, inside the bridge codebase itself, getInfoFromPulumiName is *always* called in loops, most often looping over a PropertyMap, and in one case looping over a subset of two PropertyMap instances over a `[]string` of fields. 

And getInfoFromPulumiName is the only caller of PulumiToTerraformName.

While a bit heavy-handed, this PR replaces these loops into uses of iterators. Now the iterators can be further adjusted to pre-compute information before iterating. That is, getInfoFromPulumiName can precompute inverted name tables at each loop.

Looking at this with fresh eyes, we probably don't want iterators but simply factor getInfoFromPulumiName into two calls, one to be called outside the loop and one inside.

Another alternative is memoizing PulumiToTerraformName through some static sync.Map or similar.

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -134,7 +134,9 @@ func visitPropertyValue(name, path string, v resource.PropertyValue, tfs shim.Sc
 		}
 
 		rawElementNames := rawNames || useRawNames(tfs)
-		for k, e := range v.ObjectValue() {
+		iter := newPropertyMapIterator(v.ObjectValue(), tfflds, psflds, rawElementNames)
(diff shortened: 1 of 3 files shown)


t0yv0:
We agreed to not bother with O(N^2) perf here for the moment as we couldn't find any property contents in the Pulumi corpus with N> 300.

PR #782, opened 2023-02-01
Enforce Bijection between `PulumiToTerraformName` and `TerraformToPulumiNameV2`

### LINKED ISSUE #778: Pulumi name "base32StringSeed" incorrectly translates back to Terraform as "base32_string_seed"

iwahbe:
Fixes #778

This PR adds `TerraformToPulumiNameV2`: A version of `TerraformToPulumiName` that captures the same amount of information as `PulumiToTerraformName`. 

This allows us ensure `TerraformToPulumiNameV2(PulumiToTerraformName(tfName)) == tfName` whenever possible[^1].

[^1]: The user can override this information. If there are manual overrides that conflict, we can't ensure the bijection.

### DIFF
--- a/pkg/tfbridge/names.go
+++ b/pkg/tfbridge/names.go
@@ -61,7 +61,7 @@ func PulumiToTerraformName(name string, tfs shim.SchemaMap, ps map[string]*Schem
 	return result
 }
 
-func checkTfMaxItems(tfs shim.Schema, maxItemsOne bool) bool {
+func isTfPlural(tfs shim.Schema) bool {
 	if tfs == nil {
 		return false
 	}
@@ -70,7 +70,7 @@ func checkTfMaxItems(tfs shim.Schema, maxItemsOne bool) bool {
 		return false
 	}
 
-	return (tfs.MaxItems() == 1) == maxItemsOne
+	return tfs.MaxItems() != 1
 }
 
 func isPulumiMaxItemsOne(ps *SchemaInfo) bool {
@@ -114,7 +114,7 @@ func terraformToPulumiName(name string, sch shim.SchemaMap, ps map[string]*Schem
 	}
 
 	// Pluralize names that will become array-shaped Pulumi values
-	if sch != nil && !isPulumiMaxItemsOne(psInfo) && checkTfMaxItems(sch.Get(name), false) {
+	if sch != nil && !isPulumiMaxItemsOne(psInfo) && isTfPlural(sch.Get(name)) {
 		candidate := inflector.Pluralize(name)
 
 		var conflict bool
--- a/pkg/tfbridge/names.go
+++ b/pkg/tfbridge/names.go
@@ -85,9 +85,21 @@ func TerraformToPulumiNameV2(name string, sch shim.SchemaMap, ps map[string]*Sch
 // string, from Terraform's underscore_casing to Pulumi's PascalCasing (if upper is true)
 // or camelCasing (if upper is false).
 //
-// Deprecated: Convert to TerraformToPulumiNameV2
+// Deprecated: Convert to TerraformToPulumiNameV2, see comment for conversion instructions.
 //
-// TODO: Playbook for transition
+// TerraformToPulumiNameV2 includes enough information for a bijective mapping between
+// Terraform attributes and Pulumi properties. If the full naming context cannot be
+// acquired, you can construct a partial naming context:
+//
+//	TerraformToPulumiNameV2(name,
+//		schema.SchemaMap(map[string]shim.Schema{name: sch}),
+//		map[string]*SchemaInfo{name: ps})
+//
+// If the previous (non-invertable) camel casing is necessary, it must be implemented
+// manually:
+//
+//	name := TerraformToPulumiNameV2(key, sch, ps)
+//	name = strings(uniode.ToUpper(rune(name[0]))) + name[1:]
 func TerraformToPulumiName(name string, sch shim.Schema, ps *SchemaInfo, upper bool) string {
 	return terraformToPulumiName(name,
 		schema.SchemaMap(map[string]shim.Schema{name: sch}),
--- a/pkg/tfbridge/names.go
+++ b/pkg/tfbridge/names.go
@@ -32,8 +32,6 @@ func PulumiToTerraformName(name string, tfs shim.SchemaMap, ps map[string]*Schem
 	var result string
 	// First, check if any tf name points to this one.
 	if tfs != nil {
-		// NOTE: TerraformToPulumiNameV2 is O(n) with n = # of properties. That makes
-		// PulumiToTerraformName O(n^2).
 		tfs.Range(func(key string, value shim.Schema) bool {
 			v := TerraformToPulumiNameV2(key, tfs, ps)
 			if v == name {
@@ -116,19 +114,7 @@ func terraformToPulumiName(name string, sch shim.SchemaMap, ps map[string]*Schem
 	// Pluralize names that will become array-shaped Pulumi values
 	if sch != nil && !isPulumiMaxItemsOne(psInfo) && isTfPlural(sch.Get(name)) {
 		candidate := inflector.Pluralize(name)
-
-		var conflict bool
-		sch.Range(func(key string, value shim.Schema) bool {
-			if key == name {
-				return true
-			}
-			if key == candidate {
-				conflict = true
-			}
-			return !conflict
-		})
-
-		if !conflict {
+		if _, conflict := sch.GetOk(candidate); !conflict {
 			name = candidate
 		}
 	}
--- a/pkg/tfbridge/names.go
+++ b/pkg/tfbridge/names.go
@@ -29,14 +29,23 @@ import (
 // PulumiToTerraformName performs a standard transformation on the given name string, from Pulumi's PascalCasing or
 // camelCasing, to Terraform's underscore_casing.
 func PulumiToTerraformName(name string, tfs shim.SchemaMap, ps map[string]*SchemaInfo) string {
-	// First, check if a .Name override applies
-	for k, v := range ps {
-		if v.Name == name {
-			return k
+	var result string
+	// First, check if any tf name points to this one.
+	if tfs != nil {
+		// NOTE: TerraformToPulumiNameV2 is O(n) with n = # of properties. That makes
+		// PulumiToTerraformName O(n^2).
+		tfs.Range(func(key string, value shim.Schema) bool {
+			v := TerraformToPulumiNameV2(key, tfs, ps)
+			if v == name {
+				result = key
+			}
+			return result == ""
+		})
+		if result != "" {
+			return result
 		}
 	}
 
-	var result string
 	for i, c := range name {
 		if c >= 'A' && c <= 'Z' {
 			// if upper case, add an underscore (if it's not #1), and then the lower case version.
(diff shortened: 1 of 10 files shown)


pkg/tfbridge/names.go
```diff
@@ -82,25 +77,60 @@ func isPulumiMaxItemsOne(ps *SchemaInfo) bool {
...
-			//			contract.Assertf(
-			//				inflector.Pluralize(name) == name || inflector.Singularize(inflector.Pluralize(name)) == name,
-			//				"expected to be able to safely pluralize name: %s (%s, %s)", name, inflector.Pluralize(name),
-			//				inflector.Singularize(inflector.Pluralize(name)))
-
-			name = pluralized
+	if sch != nil && !isPulumiMaxItemsOne(psInfo) && checkTfMaxItems(sch.Get(name), false) {
+		candidate := inflector.Pluralize(name)
+
+		var conflict bool
```
t0yv0, line 120: Aha so this refuses to pluralize on conflict. Good!

pkg/tfbridge/names.go
```diff
@@ -82,25 +77,60 @@ func isPulumiMaxItemsOne(ps *SchemaInfo) bool {
...
 	var result string
 	var nextCap bool
 	var prev rune
 
+	var psInfo *SchemaInfo
+	if ps != nil {
+		psInfo = ps[name]
+	}
+
+	if psInfo != nil {
```
t0yv0, line 110: This now respects psInfo.Name. Good!

pkg/tfbridge/names.go
```diff
@@ -23,12 +23,29 @@ import (
...
 
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
 )
 
 // PulumiToTerraformName performs a standard transformation on the given name string, from Pulumi's PascalCasing or
 // camelCasing, to Terraform's underscore_casing.
 func PulumiToTerraformName(name string, tfs shim.SchemaMap, ps map[string]*SchemaInfo) string {
 	var result string
+	// First, check if any tf name points to this one.
```
t0yv0, line 33: Good, tables FTW.

pkg/tfbridge/names_test.go
```diff
@@ -210,6 +200,41 @@ func TestBijectiveNameConversion(t *testing.T) {
...
+					Name: "plural",
+				},
+			},
+			expected: map[string]string{
+				"singular": "singular_property",
+				"plural":   "plural_property",
+			},
+		},
+		{ // Check number mapping
+			schema: map[string]*schemav2.Schema{
```
t0yv0, line 227: Nice.

pkg/tfgen/generate.go
```diff
@@ -1483,18 +1486,19 @@ func input(sch shim.Schema, info *tfbridge.SchemaInfo) bool {
 //
 //	would need to understand how to unmarshal names in a language-idiomatic way (and specifically reverse the
 //	name transformation process).  This isn't impossible, but certainly complicates matters.
-func propertyName(key string, sch shim.Schema, custom *tfbridge.SchemaInfo) string {
+func propertyName(key string, sch shim.SchemaMap, custom map[string]*tfbridge.SchemaInfo) string {
 	// Use the name override, if one exists, or use the standard name mangling otherwise.
 	if custom != nil {
-		if custom.Name != "" {
-			return custom.Name
+		v := custom[key]
```
t0yv0, line 1492: Oof. Is this blocks till needed? The TerraformToPulumiNameV2 bit takes care of SchemaInfo.Name now, no? Or is the worry about the `// BUGBUG` effects ?

pkg/tfgen/generate.go
```diff
@@ -1546,9 +1560,12 @@ func resourceName(provider string, rawname string,
...
 	if info == nil || info.Tok == "" {
 		// default transformations.
-		name := withoutPackageName(provider, rawname)                  // strip off the pkg prefix.
-		camel := tfbridge.TerraformToPulumiName(name, nil, nil, false) // camelCase the module name.
-		pascal := tfbridge.TerraformToPulumiName(name, nil, nil, true) // PascalCase the resource name.
+		name := withoutPackageName(provider, rawname)              // strip off the pkg prefix.
+		camel := tfbridge.TerraformToPulumiNameV2(name, nil, nil)  // camelCase the module name.
+		pascal := tfbridge.TerraformToPulumiNameV2(name, nil, nil) // PascalCase the resource name.
+		if pascal != "" {
+			pascal = string(unicode.ToUpper(rune(pascal[0]))) + pascal[1:]
```
t0yv0, line 1567: What's this change doing here? This feels unexpected?

t0yv0:
There's remaining code that has not migrated to the new version. This seems dangerous to me? Or are we sure all those use cases are benign?

```
t0yv0@Antons-MacBook-Pro> git grep TerraformToPulumiName | grep -v V2 | grep -v test                                                                                                          ~/code/pulumi-terraform-bridge
pkg/tfbridge/names.go:// TerraformToPulumiName performs a standard transformation on the given name
pkg/tfbridge/names.go:func TerraformToPulumiName(name string, sch shim.Schema, ps *SchemaInfo, upper bool) string {
pkg/tfbridge/provider.go:       return TerraformToPulumiName(name, nil, nil, false),
pkg/tfbridge/provider.go:               TerraformToPulumiName(name, nil, nil, true)
pkg/tfbridge/provider.go:                       name := TerraformToPulumiName(key, meta, nil, false)
pkg/tfbridge/provider.go:                       name := TerraformToPulumiName(parts[1], schema, info, false)
pkg/tfbridge/provider.go:                       name := TerraformToPulumiName(selector.Name, schema, info, true)
pkg/tfbridge/schema.go:                 name = TerraformToPulumiName(key, getSchema(tfs, key), ps[key], false)
```

t0yv0:
Thinking about the API a little bit.

Current API:

```
func TerraformToPulumiNameV2(name string, sch shim.SchemaMap, ps map[string]*SchemaInfo) string
func PulumiToTerraformName(name string, tfs shim.SchemaMap, ps map[string]*SchemaInfo) string
```

Suggested API:

```
type NameContext interface {
   TerraformToPulumiName(name string) string
   PulumiToTerraformName(name string) string
}
func NewNameContext(shim.SchemaMap, map[string]*SchemaInfo) NameContext
```

Having NameContext spelled out can help us refacoring this further to lift NewNameContext out of loops to avoid
quadratics. For starters we can just search/replace to `NewNameContext(schemaMap, infos).TerraformToPulumiName(n)`.

I am also wondering if this should really be this? Some callsites don't need invented names that don't exist
in the tables.

```
type NameContext interface {
   TerraformToPulumiName(name string) (string, error)
   PulumiToTerraformName(name string) (string, error)
}
```

Or perhaps:

```
type NameContext interface {
   TerraformToPulumiName(name string) string
   PulumiToTerraformName(name string) string
   KnownTerraformToPulumiName(name string) (string, error)
   KnownPulumiToTerraformName(name string) (string, error)
}
```

t0yv0:
This fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/778 I think?

pkg/tfbridge/names.go
```diff
@@ -82,25 +77,60 @@ func isPulumiMaxItemsOne(ps *SchemaInfo) bool {
...
-	if !isPulumiMaxItemsOne(ps) && checkTfMaxItems(sch, false) {
-		pluralized := inflector.Pluralize(name)
-		if inflector.Singularize(pluralized) == name {
-			//			contract.Assertf(
-			//				inflector.Pluralize(name) == name || inflector.Singularize(inflector.Pluralize(name)) == name,
-			//				"expected to be able to safely pluralize name: %s (%s, %s)", name, inflector.Pluralize(name),
-			//				inflector.Singularize(inflector.Pluralize(name)))
-
-			name = pluralized
+	if sch != nil && !isPulumiMaxItemsOne(psInfo) && checkTfMaxItems(sch.Get(name), false) {
```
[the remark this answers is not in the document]
iwahbe: Good catch. It now reads:
```go
sch != nil && !isPulumiMaxItemsOne(psInfo) && isTfPlural(sch.Get(name))
```

pkg/tfgen/generate.go
iwahbe: That block can be yanked. The `// BUGBUG` only effects the key, which is not used in `.Name` override lookup.

pkg/tfgen/generate.go
iwahbe: Previously, the code looked like this:
https://github.com/pulumi/pulumi-terraform-bridge/blob/59a7281edaa87aa3e7c3d8544521d9d4232c05e3/pkg/tfgen/generate.go#L1550-L1551

Since `tfbrige.TerraformToPulumiName` doesn't support a (non-reversible) PascalCase option anymore, it needed to be manually reimplemented.

iwahbe:
@t0yv0 It looks like the number of resource properties caps out at 70 (aws), with other T1 providers less then that[^1]. The performance of `PulumiToTerraformName` and `TerraformToPulumiName` are both `O(n)`. Can you point me to the hot loops where you predict this will be a problem?

I think that the API you are suggesting will give performance wins, but I'm not sure they will be noticeable. I don't think we need an API that supports known and unknown variants. Given the same schema and info maps, all operations will be reversible.

[^1]: `jq '.resources[].properties | length' provider/cmd/pulumi-resource-aws/schema.json  | sort -n | tail -10`

pkg/tfgen/generate.go
t0yv0: Got it. That makes sense.

t0yv0:
Regardless of whether we actually optimize it or not here I have ever so slight preference for something like NameContext (though hard to find a name for it..) as it can be more obvious to document a constraint on these two functions together that they need to be invertible. Also it incidentally solves the TerraformToPulumiNameV2 wrinkle. With TerraformToPulumiNameV2 we don't have a way to ever get rid of "V2" even if we remove the original version. Would be nice to shoot for something that becomes clean down the line. Any other ideas to avoid the wrinkle?

The rest is non-binding.

#### perf

N=375 on object types.

```
$ jq '.types[].properties | length' provider/cmd/pulumi-resource-aws/schema.json  | sort -n | tail -5                                                                            
48
48
88
375
375
```

https://github.com/pulumi/pulumi-terraform-bridge/pull/781/ edits all the locations that are looping over PulumiToTerraformName.

Just a matter of time till someone attempts a provider with 1K props, but we can choose to leave this for later.

#### errors

Definitely for later. It feels that at least in some contexts PulumiToTerraformName should not be inventing Terraform names that don't exist on the fly. Catching this can be a chance to present a better error. But admittedly I'm not sure how to repro that situation or what the current error presentation is. In the PF version of the bridge if the SDK sends properties TF does not understand, it currently simply ignores them.

t0yv0 (approved):

t0yv0:
Approving as it moves us forward significantly and fixes an important issue though I'm still not a fan of V2 function naming :)

iwahbe:
> Approving as it moves us forward significantly and fixes an important issue though I'm still not a fan of V2 function naming :)

I think a clean set of exports is a 4.0 problem. We can fix the performance problem if it comes up by implementing a context cached but not exposing it.

PR #783, opened 2023-02-02
Cleanup go sdk modules

iwahbe:
This brings Go up to parity with the rest of our SDKs vis-a-vis cleanup.


t0yv0 (approved):
Comments from slack: gogen doesn't yet take extraFiles option, this code compensates. LGTM.

PR #784, opened 2023-02-02
Add Aliasing strategy

iwahbe:
This PR adds opt-in automatic aliasing to the bridge. 

This allows us to implement deterministic but unstable naming, such as https://github.com/pulumi/pulumi-terraform-bridge/pull/806.

At a technical level, we save a look aside table of all names using [ProviderMetadata](https://github.com/pulumi/pulumi-terraform-bridge/pull/873), and then check for mismatches when running `resources.go`. Mismatches are fixed up via already existing hard alias functions:

https://github.com/pulumi/pulumi-terraform-bridge/blob/e30e417750326d395a8e1e6c5a89f99e11ea83ef/pkg/tfbridge/provider.go#L1294-L1295

https://github.com/pulumi/pulumi-terraform-bridge/blob/e30e417750326d395a8e1e6c5a89f99e11ea83ef/pkg/tfbridge/provider.go#L1337-L1338

Fixes https://github.com/pulumi/home/issues/2592


pkg/tfbridge/token.go
```diff
@@ -189,3 +191,169 @@ func (err UnmappableError) Error() string {
...
+}
+
+type alias[T ~string] struct {
+	Name      T    `json:"name"`      // The previous token.
+	InCodegen bool `json:"inCodegen"` // If the alias is a fully generated resource, or just a schema alias.
+}
+
+type aliasHistory struct {
+	Resources   map[string]*tokenHistory[tokens.Type]         `json:"resources"`
+	DataSources map[string]*tokenHistory[tokens.ModuleMember] `json:"datasources"`
```
t0yv0, line 207: Interesting, so this data is not stored in schema? Is it stored in MarshallableProviderInfo? Perhaps a dedicated struct is better anyway.

iwahbe: The alias history isn't really part of the ProviderInfo or the schema. It is a historical artifact of the mapping between the two.

t0yv0: Got it, the coin dropped. Yes it needs an accumulating history.

t0yv0:
The functionality makes a lot of sense to me and it moves us forward a lot! Automatically doing aliasing sounds awesome.

Could we consider simplifying the end-user API here somehow? Something like this?

```
providerInfo.AutoAlias("data/alias-history.yml") 
```

This change does strategy combinators, but squinting at it it seems to me the combination is purely sequential?  So we take ProviderInfo, and modify it to auto-map tokens, then modify it to auto-alias? If that's all that's going on then sticking with sequential mutable API to express this pipeline would seem to me least confusing and more in line with what we have there already, e.g.:

```
providerInfo.ComputeDefaults(..)
providerInfo.AutoAlias("data/alias-history.yml") 
```

It seems to me FinishAlias is also a bit strange. But perhaps it is needed?

Think of the changes we make here in the bridge as something we need to document and explain to a wide audience. It's worth IMO to iterate a bit on the naming.

iwahbe:
> The functionality makes a lot of sense to me and it moves us forward a lot! Automatically doing aliasing sounds awesome.
> 
> Could we consider simplifying the end-user API here somehow? Something like this?
> 
> ```
> providerInfo.AutoAlias("data/alias-history.yml") 
> ```

Unfortunately, it's not that easy. The binary needs to be able to access the history during both `make tfgen` time, and during normal runtime. This means we need to embed history with `go:embed`.  Since `go:embed` takes a static path in the comment, our API can't deal purely with runtime data like `"data/alias-history.yml"`. This is part of why the current API deals with data blobs (`[]byte`), and requires the user to handle pathing. 

I think it's also helpful to be able to control what is and isn't aliased. Right now, `tfbridge.Aliasing` only aliases what it wraps, not all resources. This makes opt-in and opt-out controls automatic.

> This change does strategy combinators, but squinting at it it seems to me the combination is purely sequential? So we take ProviderInfo, and modify it to auto-map tokens, then modify it to auto-alias? If that's all that's going on then sticking with sequential mutable API to express this pipeline would seem to me least confusing and more in line with what we have there already, e.g.:
> 
> ```
> providerInfo.ComputeDefaults(..)
> providerInfo.AutoAlias("data/alias-history.yml") 
> ```
> 
> Like in F# it would be like this (and yes order matters, which is a bit of a gotcha):
> 
> ```fsharp
> providerInfo
> |> computeDefaults
> |> autoAlias
> ```

Does this desugar into `computeDefaults(providerInfo); autoAlias(providerInfo)`? I don't know F#.

> Instead it seems we're headed toward something like the following, with a bit of a Haskell bend to it.
> 
> ```
> providerInfo.ComputedDefaults(strat1 . strat2 . strat3) 
> ```

This is exactly what we are doing.

> It seems to me FinishAlias is also a bit strange. But perhaps it is needed?

`FinishAlias` is needed because we are modeling aliases as a default strategy. It provides the side channel necessary for communicating back the updated history blob for the user to save.

> Think of the changes we make here in the bridge as something we need to document and explain to a wide audience. It's worth IMO to iterate a bit on the naming.

t0yv0:
I'm quite alarmed at the runtime dependency bit on the alias history, and didn't notice it in the review, I should have looked at the given thing. Let me think about this for a bit..

The fact resources.go runs both get-time and run-time and is expected to give the same result is concerning to me long term. Perhaps we could stage it. There's MarshallableProviderInfo. The staged design would run resources.go to compute MarshallableProviderInfo at gen-time, embed that, and at runtime it'd only be dealing with ProviderInfo recovered from MarshallableProviderInfo. Then alias-history would very visibly not be a runtime dependency.

Also, less importantly, stylistically I'd err toward some "provider-metadata.json" that's opaque and we evolve over time rather than a proliferation of new embed directives, that's the path taken in the pf/ bridge, seems could help here a bit too - at least make it easier to interpret the diffs downstream.. Oh okay, bridge is changing the internal provider-metadata blob, nothing to see here.

iwahbe:
> I'm quite alarmed at the runtime dependency bit on the alias history, and didn't notice it in the review, I should have looked at the noted PR that applies this change. It makes sense.. Let me think about this for a bit..
> 
> The fact resources.go runs both get-time and run-time and is expected to give the same result is concerning to me long term. Perhaps we could stage it. There's MarshallableProviderInfo. The staged design would run resources.go to compute MarshallableProviderInfo at gen-time, embed that, and at runtime it'd only be dealing with ProviderInfo recovered from MarshallableProviderInfo. Then alias-history would very visibly not be a runtime dependency.

I agree this is not ideal, but it's orthogonal to this PR as long as alias application is deterministic (it is). If we convert to a model where we serialize `MarshallableProviderInfo`, then we won't need to make any changes to `Aliasing` or its underlying technology.

> Also, less importantly, stylistically I'd err toward some "provider-metadata.json" that's opaque and we evolve over time rather than a proliferation of new embed directives, that's the path taken in the pf/ bridge, seems could help here a bit too - at least make it easier to interpret the diffs downstream.. Oh okay, bridge is changing the internal provider-metadata blob, nothing to see here.

I agree. I'll introduce a `ProviderMetadata` type so it's clearer that we only want a single embed, not 1 for every feature used.

t0yv0:
Thank you! I need to take a raincheck on reviewing this on time.

For prior art I link you to https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pf/tfgen/main.go#L66 - I don't have a type yet but the generated stuff I called `bridge-metadata.json` and the intention is for it to evolve together with the bridge releases without any other code peeking in. In the PF case it carries renames for now but we might want to change that later now that we trust the name bijection.

iwahbe:
> Thank you! I need to take a raincheck on reviewing this on time.

No problem. Maybe when your not on triage rotation 😄.

> For prior art I link you to https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pf/tfgen/main.go#L66 - I don't have a type yet but the generated stuff I called `bridge-metadata.json` and the intention is for it to evolve together with the bridge releases without any other code peeking in. In the PF case it carries renames for now but we might want to change that later now that we trust the name bijection.

I saw that. I was hoping this type could be used by both `pkg` and `pf`, hence the shenanigans with `internal/metadata` to allow `pf` to read and write fields from `metadata.Provider`.

t0yv0:
I'd like that, that makes sense! Back to this later..

internal/metadata/metadata.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return &Data{m}, nil
+}
+
+func (d *Data) Marshal() []byte {
+	if d == nil {
+		d = &Data{m: make(map[string]json.RawMessage)}
+	}
+	bytes, err := json.MarshalIndent(d.m, "", "    ")
+	contract.AssertNoErrorf(err, "internal: failed to marshal metadata")
```
AaronFriel, line 44: My only nit on this PR is the use of `contract`. 

I think we had been moving away from panics in non-test code, in light of how panics hid errors lurking in docsgen, which used `recover()`.

AaronFriel (changes requested):
I'd like to review the use of contract assertions, there seem to be some which could be returned as errors instead.

internal/metadata/metadata.go
iwahbe: I'll add a note explaining why the `contract` assertion is there, but I think it should stay. In my mind, I separate error handling (something has gone wrong, let's deal with it) from assertions (this should never happen, and if it does I don't want to deal with it. Let the compiler know). The second type shouldn't be returned as errors, but I agree that there needs to be a better explanation in the code.

Here, the explanation is
```
`d.m` is a `map[string]json.RawMessage`. `json.MarshalIndent` errors only when it is 
asked to serialize an unmarshalable type (complex, function or channel) or a cyclic data
structure. Because `string` and `json.RawMessage` are trivially marshallable and cannot
contain cycles, all values of `d.m` can be marshaled without error. 
```

pkg/tfgen/generate.go
```diff
@@ -55,6 +56,10 @@ const (
 	maxWidth      = 120 // the ideal maximum width of the generated file.
 )
 
+// Additional files to put next schema.json when it is generated.
+var additionalSchemaFilesHook = map[string][]byte{}
```
t0yv0, line 60: Eh, static state is a red flag. What would it take to get this to naturally flow through the generator invocation?

pkg/tfgen/generate.go
```diff
@@ -1786,3 +1797,18 @@ func ignoreMappingError(s []string, str string) bool {
 	}
 	return false
 }
+
+// Add a file to be emitted next to the schema.json only when the schema is itself
+// emitted.
+//
+// `panic`s if called with a `path` that conflicts with an existing file, including
+// `schema.json`.
+func AdditionalSchemaFile(path string, bytes []byte) {
```
t0yv0, line 1806: I've used https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pf/tfgen/main.go#L40-#L40 for this purpose before which does not feel quite right either.

internal/metadata/metadata.go
```diff
@@ -0,0 +1,97 @@
...
+	//
+	// See https://pkg.go.dev/encoding/json#Marshal for details.
+	contract.AssertNoErrorf(err, "internal: failed to marshal metadata")
+	return bytes
+}
+
+// Set a piece of metadata to a value.
+//
+// Set errors only if value fails to serialize.
+func Set(d *Data, key string, value any) error {
```
t0yv0, line 58: Would this be more natural as `func (*Data) Set(key, value)`?

pkg/metadata/metadata.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Re-export metadata in an opaque format.
+package metadata
+
+import (
+	"github.com/pulumi/pulumi-terraform-bridge/v3/internal/metadata"
+)
+
+// A store persisted between `tfgen` and a running provider.
+type Provider *metadata.Data
```
t0yv0, line 23: Could we name this ProviderMetadata? A bit more verbose but much less confusing. This sounds like it's a Provider that provides services about metadata.

t0yv0:
Suggest imperative API here https://github.com/pulumi/pulumi-terraform-bridge/pull/858

t0yv0:
The piece that bugs me the most, but I don't have a solution to, is the lack of phase separation.  We almost need to consider some breaking changes or detecting if the ProviderInfo computation is running in tfgen context or in provider startup context perhaps using some tricks?

Typical provider does this:

```
//go:embed schema-embed.json
var pulumiSchema []byte

func main() {
	tfbridge.Main("tls", version.Version, tls.Provider(), pulumiSchema)
}
```

That's making `tls.Provider()` run *again* after it was already run by tfgen.

I'm not sure that full alias history needs to be runtime dependency (dep of provider) as opposed to build-time dependency in the tfgen world only, but it almost has to because of this coupling..

D'you have an example of one provider that users the API from this PR as is for me to ponder how you solved the knot?

internal/metadata/metadata.go
iwahbe: I'm trying to create a separation between consumers of `Data` (the tfbridge package) and holders/passers of metadata (end providers). To name and pass around metadata, you need to import `github.com/pulumi/pulumi-terraform-bridge/v3/pkg/metadata`. To manipulate metadata, you need to import `github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata`. 

If we make `Set` a method (as opposed to a function), it will be usable from `github.com/pulumi/pulumi-terraform-bridge/v3/pkg/metadata.Data` and not require the `unstable` import. 

Note: I used to have `Set` in `internal/metadata`, but `pf` can't use that, so I changed `internal` to `unstable`.

iwahbe: https://github.com/pulumi/pulumi-terraform-bridge/blob/a7f49a7d5354a98b924aa3ffc62dafcdcdbd2850/unstable/doc.go#L15-L20

pkg/metadata/metadata.go
iwahbe: Sure. Because `metadata.ProviderMetadata` is overly verbose, and because all functions in `metadata` also need to be renamed to match, I folded `metadata` into `tfbridge`. The new name is `tfbridge.NewProviderMetadata(...)`.

iwahbe:
> D'you have an example of one provider that users the API from this PR as is for me to ponder how you solved the knot?

@t0yv0 The example PR is https://github.com/pulumi/pulumi-alicloud/pull/345. Basically, the caller is in charge of embedding and persisting the metadata bundle.

t0yv0:
Can we see the diff on top of the PR with the metadata store that would help a lot! :) 

This probably LGTM since you've kindly adopted the imperative API suggestions, just the PR description is a bit out of date now as it still talks about wrapping strategies, let's make sure the final commit msg is accurate.

Issue #779, opened 2023-02-01
Validate object type property name assignment in tfgen


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

The work in https://github.com/pulumi/pulumi-terraform-bridge/pull/775 ideally should extend not only to top-level properties, but also properties of nested object types. These types also use the naming functions such as this one:

[TerraformToPulumiName](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/names.go#L87-#L87)

The code related to Renames may be helpful in finding and correlating all nested object types.

PR #786, opened 2023-02-03
Better namecheck

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/779

Name checks are now extended to cover object types, I think it covers everything now. Very reassuringly the name-checks pass on pulumi-aws as https://github.com/pulumi/pulumi-terraform-bridge/pull/782 took care of the missing bits. This double checks that the naming functions are indeed invertible for the entire provider after applying user-supplied Name overrides and pluralization.

- [x] check TF and Pulumi property name uniqueness for all objects
- [x] check object types (not just resources and functions)
- [x] cross-check that rename tables correspond to the renaming functions used at runtime
- [x] more actionable messages 
- [x] correlate SchemaInfo on object types

Example error presentation (note that's no longer an error anymore, it was a spurious warning due to a bug in namecheck, but it's here to illustrate the format):

```
warning: TerraformToPulumiNameV2("expiration") unexpectedly returns "expiration", expecting "expirations" from schema generation.
  Please report this as a bug to pulumi/pulumi-terraform-bridge.
  Location:
    Type: "aws:s3/BucketV2LifecycleRule:BucketV2LifecycleRule"
    TypePath: resource[key="aws_s3_bucket",token="aws:s3/bucketV2:BucketV2"].inputs.lifecycle_rule[pulumi:"lifecycleRules"].$
```

### DIFF
--- a/pkg/tfgen/namecheck.go
+++ b/pkg/tfgen/namecheck.go
@@ -17,150 +17,355 @@ package tfgen
 import (
 	"fmt"
 	"os"
-	"sort"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
 
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+
+	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen/internal/paths"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 )
 
-type nameErrors []nameError
-
-func (ne nameErrors) Sort() {
-	sort.Slice(ne, func(i, j int) bool {
-		if ne[i].context < ne[j].context {
-			return true
-		}
-		if ne[i].tfKey < ne[j].tfKey {
-			return true
-		}
-		if ne[i].tfAttr < ne[j].tfAttr {
-			return true
-		}
-		if ne[i].message < ne[j].message {
-			return true
-		}
-		return false
-	})
+type nameCheckPropSet struct {
+	location  string
+	props     []paths.PropertyName
+	schemaMap shim.SchemaMap
+	infos     map[string]*tfbridge.SchemaInfo
 }
 
-func (ne nameErrors) Report(sink diag.Sink) {
-	if len(ne) == 0 {
-		return
-	}
-
+// Checks a generated schema to detect property name collisions and verify automatic naming.
+//
+// Bridged providers need to perform property name translation between Pulumi and Terraform names. Historically there
+// were two classes of issues with naming, both leading to confusing runtime errors:
+//
+//   - problems from customized ProviderInfo overrides, such as custom SchemaInfo.Name that picks a Pulumi property name
+//     that conflicts with another property
+//
+//   - automatic naming bugs in tfbridge code such as https://github.com/pulumi/pulumi-terraform-bridge/issues/778
+//
+// Best-effort attempt is made here to detect both classes of issues early (at schema generation aka provider build
+// time), decide which one it is.
+func nameCheck(
+	prov tfbridge.ProviderInfo,
+	spec pschema.PackageSpec,
+	renamesBuilder *renamesBuilder,
+	sink diag.Sink,
+) error {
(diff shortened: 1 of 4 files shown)


pkg/tfgen/namecheck.go
```diff
@@ -17,150 +17,341 @@ package tfgen
...
+	for _, p := range props.props {
+		index[p.Name] = append(index[p.Name], p.Key)
+	}
+	for k, v := range index {
+		if len(v) <= 1 {
+			continue
+		}
+		m := "Pulumi property %q maps to multiple Terraform properties %v.\n" +
+			"  This will cause runtime errors in the provider.\n" +
+			"  Please set SchemaInfo.Name to rename the Terraform properties.\n" +
```
t0yv0, line 174: @danielrbradley as we discussed the warnings should include some steps to remediation; if this ever pops up again, this is what it would say - does this make sense? Or can I change the wording further. Full paths to SchemaInfo can be involved, e.g. ProviderInfo.Resources[t].Fields.Elem....Name

pkg/tfgen/namecheck.go
```diff
@@ -17,150 +17,341 @@ package tfgen
...
 
-type nameErrors []nameError
+type nameCheckPropSet struct {
+	location  string
+	props     []paths.PropertyName
+	schemaMap shim.SchemaMap
+	infos     map[string]*tfbridge.SchemaInfo
+}
+
+func nameCheck(
```
iwahbe, line 40: Can we add an explanation of what this function is checking for?

pkg/tfgen/namecheck.go
```diff
@@ -17,150 +17,341 @@ package tfgen
...
+}
+
+func (x *scalarType) SchemaInfos() (map[string]*tfbridge.SchemaInfo, error) {
+	return nil, fmt.Errorf("Fields are undefined on an scalarType")
+}
+
+var _ typ = (*scalarType)(nil)
 
-		return true
+func parseSchema(s shim.Schema, info *tfbridge.SchemaInfo) typ {
```
iwahbe, line 283: `parseSchema` is not a clear name for what this function is doing. It looks like it is converting a schema + info into a `typ`. Would it make sense to call it `newTypFromSchemaAndInfo`?

pkg/tfgen/renames.go
```diff
@@ -102,40 +102,116 @@ func (r *renamesBuilder) registerNamedObjectType(paths paths.TypePathSet, token
...
+
+func (*renamesBuilder) isConfig(propertyParentPath paths.TypePath) bool {
+	_, ok := propertyParentPath.(*paths.ConfigPath)
+	return ok
+}
+
+func (r *renamesBuilder) propetyParentToken(propertyParentPath paths.TypePath) (tokens.Token, error) {
+	switch parent := propertyParentPath.(type) {
+	case *paths.ConfigPath:
+		return "", fmt.Errorf("ConfigPath should have been detected with isConfig")
```
iwahbe, line 123: I don't have context, but the message implies that this should never happen, regardless of user input. Should we just `panic` here?

iwahbe (approved):
This is an excellent improvement to our validation!

I have a couple requests for code clarity. After that, LGTM!

pkg/tfgen/namecheck.go
t0yv0: This is some 1am code. `tip` is not a clear name either. It's all internal helpers to make sense out of pretty crazy type encoding that is tfbridge.SchemaInfo so that the code here can step down and lookup matching SchemaInfo by TypePath.

t0yv0: Let me think about this a bit, how to name this.

pkg/tfgen/renames.go
t0yv0: In theory yes, in practice I'm leery of panics since I had to work through quite a few of those last night. I think the trouble is that I don't fully trust the TypePath steppers yet. In particular the path that has an ElementPath fragment in it - that is going from `List[String]` into `String` must be aware of Pulumi flattening but isn't aware yet, so the code has hacks to compensate for this..

I think I'll leave it as not having any panics now (and clean up existing ones) but we can tighten up once we have more confidence from a large-scale rollout.

iwahbe: My concern is that we won't get confidence from a large scale roll out because we have so many warnings, which makes our warnings easy to miss.

t0yv0: Yes, ok this one is a panic. I think what you're implicitly asking for also is raising the warnings to errors here, we can come back to this in a few weeks.

PR #788, opened 2023-02-03
[WIP] pulumi aws 2352

t0yv0:


t0yv0:
I'd like to check that in together with the parent PR, so that https://github.com/pulumi/pulumi-terraform-bridge/pull/844 can be rebased on top and demonstrate that it can make the test pass by removing the skip.

AaronFriel (approved):

PR #789, opened 2023-02-04
Accommodate conflict remaps

iwahbe:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/790

We were keeping fields singular when they would cause a conflict, even if the conflict was manually remapped. This caused a breaking change in the azurerm_frontdoor resource. 

The problem was discovered via https://github.com/pulumi/pulumi-azure/actions/runs/4078132838/jobs/7048043503.


t0yv0 (approved):

t0yv0:
Linter: 

```
Error: pkg/tfbridge/names.go:144:77: `emited` is a misspelling of `emitted` (misspell)
[10](https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/4088862669/jobs/7050957866#step:7:11)
		// The non-bijectivity will be caught at tfgen time and a warning will be emited.
```

PR #805, opened 2023-02-10
Fix schema cache poisoning for colliding provider versions

### LINKED ISSUE #801: Bridge versions 3.39.x occasionally missing examples during generation
Reproduced in azure-classic in this PR: https://github.com/pulumi/pulumi-azure/pull/1205

- Downgrading to 3.38.1 resolves the issue.
- Using both 3.39.0 and 3.39.3 introduces unreliable generation.

Having performed a clean build locally, CI will fail with dirty working tree due to generation behaviour being inconsistent.

t0yv0:
Fixes #801 

Looks like this bug was around for a bit but it only manifests as dropped examples in provider builds when 2+ providers accidentally have the same version.

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -772,16 +772,13 @@ func NewGenerator(opts GeneratorOptions) (*Generator, error) {
 	}
 
 	return &Generator{
-		pkg:          pkg,
-		version:      version,
-		language:     lang,
-		info:         info,
-		root:         root,
-		providerShim: providerShim,
-		pluginHost: &cachingProviderHost{
-			Host:  host,
-			cache: map[string]plugin.Provider{},
-		},
+		pkg:              pkg,
+		version:          version,
+		language:         lang,
+		info:             info,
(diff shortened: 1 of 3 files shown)


lukehoban:
Can we add test coverage?

Notably - had https://github.com/pulumi/pulumi-terraform-bridge/commit/c12dcdbc79f2a86c4459ed4ea46c220d00ca64f6 had tests included they nearly certainly would have caught this before we shipped it 2 years ago.

t0yv0:
I'm wondering if this wrapper even needed? Does platform have in-mem caching facilities in the provider host we can use here? Is the underlying host doing caching already.

t0yv0:
Couldn't figure out quickly if caching is redundant, but tests are easy to add so tests are added.

Frassle (approved):
> Couldn't figure out quickly if caching is redundant, but tests are easy to add so tests are added.

I had a quick look at this as well, it doesn't look like platform exposes anything for this so I think this fix is the right one for now.

PR #806, opened 2023-02-10
Add Inferred Token Module Strategy

iwahbe:
Adds a default strategy for heuristic module discovery.

This strategy performs global analysis on the complete set of resources and datasources in the underlying provider to determine which modules should be created.

This requires https://github.com/pulumi/pulumi-terraform-bridge/pull/784 for backwards compatibility.


aq17 (approved):
This LGTM, I may want to test it out on something such as https://github.com/pulumi/pulumi-newrelic/pull/370 first

pkg/tfbridge/x/token_test.go
```diff
@@ -169,6 +169,112 @@ func TestIgnored(t *testing.T) {
...
+			name: "detect a simple module",
+			resourceMapping: map[string]string{
+				"pkg_hello_world":   "hello:World",
+				"pkg_hello_pulumi":  "hello:Pulumi",
+				"pkg_hello":         "hello:Hello",
+				"pkg_goodbye_folks": "index:GoodbyeFolks",
+				"pkg_hi":            "index:Hi",
+			},
+			opts: &InferredModulesOpts{
+				MinimumModuleSize: 2,
```
t0yv0, line 198: A comment here highlighting what MinimumModuleSize is - I'm guessing this means that at least two entries are needed so "hello" got recognized as a module as it has 3 entries.

pkg/tfbridge/x/token_test.go
```diff
@@ -169,6 +169,112 @@ func TestIgnored(t *testing.T) {
...
+				"pkg_mod_sub_r2": "modSub:R2",
+				"pkg_mod_sub_r3": "modSub:R3",
+				"pkg_mod_sub_r4": "modSub:R4",
+				"pkg_mod_not_r1": "mod:NotR1",
+				"pkg_mod_not_r2": "mod:NotR2",
+			},
+			opts: &InferredModulesOpts{
+				TfPkgPrefix:          "pkg_",
+				MinimumModuleSize:    3,
+				MimimumSubmoduleSize: 4,
```
t0yv0, line 218: How is MimimumSubmoduleSize: 4 in play in this example?

pkg/tfbridge/x/token_test.go
```diff
@@ -169,6 +169,112 @@ func TestIgnored(t *testing.T) {
 	}, info.Resources)
 }
 
+func TestTokensInferredModules(t *testing.T) {
+	tests := []struct {
+		name            string
+		resourceMapping map[string]string
+		opts            *InferredModulesOpts
+	}{
+		{
```
t0yv0, line 178: Suggest adding example here based on our conversation from OCI with a bit of long-form explanation motivating the tree algorithm tricks.

```

			"oci_apm_apm_domain": {Tok: tfbridge.MakeResource(mainPkg, apmMod, "ApmDomain")},

			"oci_apm_config_config": {Tok: tfbridge.MakeResource(mainPkg, apmConfigMod, "Config")},

			"oci_apm_synthetics_monitor":                 {Tok: tfbridge.MakeResource(mainPkg, apmSyntheticsMod, "Config")},
			"oci_apm_synthetics_script":                  {Tok: tfbridge.MakeResource(mainPkg, apmSyntheticsMod, "Script")},
			"oci_apm_synthetics_dedicated_vantage_point": {Tok: tfbridge.MakeResource(mainPkg, apmSyntheticsMod, "DedicatedVantagePoint")},

```

The algo looking at "oci_apm_apm_domain" needs to figure out if the module is oci_apm or oci_apm_apm. It's not allowed to know the list of all known submodules. So the way it does inference is it notices that there's only one "oci_apm_apm" prefix, therefore it's not a module; if the user added "oci_apm_apm_thingy" resource, then there'd be two entries and it might decide oci_apm_apm is now a module.

t0yv0:
Thanks @iwahbe for explaining the motivation. 

My chief concern is that the module assignment produced by this constraint based system can be non-trivial to understand and unstable under extension (adding new resources might move around old resources). This concern is mitigated by the automatic aliasing machinery that will make movement of resources fully transparent and non-breaking to the user, I didn't realize we have this machinery - that's brilliant. SO the cost of unpredictable module re-assignments is just bloat, not breaking changes.

We also discussed that the user can recover a degree of control by injecting special cases to manually assign resources to modules and having the strategy respect that choice. This is great.

One might imagine this can be made to compose in some way also with "known modules", that is, the user might be able to list what is or what isn't a module and have those constraints respected before doing further inference.

Finally this is just one of the strategies and the user may opt out and use KnownModules or some such, which I'd personally prefer for smaller providers to get started with.

So it sounds like this is a tool that has it's place 🚢

t0yv0 (approved):

pkg/tfbridge/x/token_test.go
iwahbe: I have annotated that test. Here are is the new object:
```go
		{
			name: "nested modules",
			resourceMapping: map[string]string{
				"pkg_mod_r1":     "mod:R1",
				"pkg_mod_r2":     "mod:R2",
				"pkg_mod_r3":     "mod:R3",
				"pkg_mod_r4":     "mod:R4",
				"pkg_mod_sub_r1": "modSub:R1",
				"pkg_mod_sub_r2": "modSub:R2",
				"pkg_mod_sub_r3": "modSub:R3",
				"pkg_mod_sub_r4": "modSub:R4",
				"pkg_mod_not_r1": "mod:NotR1",
				"pkg_mod_not_r2": "mod:NotR2",
			},
			opts: &InferredModulesOpts{
				TfPkgPrefix: "pkg_",
				// We set the minimum module size to 4. This ensures that
				// `pkg_mod` is picked up as a module.
				MinimumModuleSize: 4,
				// We set the MimimumSubmoduleSize to 3, ensuring that
				// `pkg_mod_sub_*` is is given its own `modSub` module (4
				// elements), while `pkg_mod_not_*` is put in the `mod`
				// module, since `pkg_mod_not` only has 2 elements.
				MimimumSubmoduleSize: 3,
			},
		}
```

iwahbe:
Downstream tests are known-broken due to a failure to pin gradle to 7.x.x. Bridge only tests are green, so I'm merging.

PR #808, opened 2023-02-14
PF: Support Tuple Type

### LINKED ISSUE #730: [Plugin Framework] Support Tuple type
## Hello!

## Issue details

Terraform Plugin Framework allows providers to specify types of properties as a Tuple (known length, known element types). Bridged Pulumi providers should support working with such properties.

iwahbe:
Fixes #730

This PR currently supports value conversion between TF Tuples and Pulumi Objects. Schema generation is not yet supported.


t0yv0:
Good job on finding the convert bit and adding tests!

The other important bit is support of schema generation for tuple types. This will also force to think about which representation would work..


An approach I would take here is taking a look at this test:

- https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pf/tests/testdata/basicprogram/Pulumi.yaml#L5
- run it as part of `make test.integration`
- the Go wrapper is https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pf/tests/integration/program_test.go#L32

This test has been carrying a lot of the end-to-end tests on the serializers for various types so feels good to extend with one more case. TO extend the test:

- add a property to https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pf/tests/internal/testprovider/testbridge_resource_testres.go of a tuple type (or if this gets nasty, add another resource just to test tuples)

- use this resource in basicprogram/Pulumi.yaml

- code some asserts in  basicprogram/Pulumi.yaml that it turns around


🙇

pf/internal/convert/encoding.go
```diff
@@ -268,15 +269,30 @@ func (e *encoding) deriveEncoder(typeSpec *pschema.TypeSpec, t tftypes.Type) (En
...
 	case "array":
-		lt, ok := t.(tftypes.List)
-		if !ok {
-			return nil, fmt.Errorf("expected a List, got %s", t.String())
-		}
-		elementEncoder, err := e.deriveEncoder(typeSpec.Items, lt.ElementType)
-		if err != nil {
-			return nil, err
+		switch t := t.(type) {
+		case tftypes.List:
```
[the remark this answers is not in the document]
iwahbe: I shied away from objects, since that would require naming, and `(P1_T1, P1_T2) != (P2_T1, P2_T2)` even if `P1_T1 = P2_T1 = string` and `P1_T2 = P2_T2 = int`, which will be confusing (either at runtime or compile time, depending on the language). I'll try an object iteration on this, since I agree it preserves type information while lists do not, but I'm concerned python users will complain that `pkg.mod.StringIntTuple("string", 0)` is absurdly verbose compared to `"string", 0`.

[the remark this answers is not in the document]
t0yv0: Well Python has native tuples, so perhaps a feature request we could make upstream to the Pulumi schema to support tuple types natively and map them to Python tuples, or perhaps more precariously to magic-recognize encoded tuples in the Python encoding so that Python tuples are supported as tuples.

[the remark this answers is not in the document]
t0yv0: The request would need some motivating examples though.. So far I've not seen this feature widely used, but we need more providers. From what we know today it sounds like preserving type info in the projection is worth it and worth some awkwardness, with the awkwardness not affecting too many users because it seems like a rare feature.

iwahbe:
I now convert tuples into objects. There is an end to end test using `basicprogram/Pulumi.yaml` that shows how a tuple is consumed.

pf/tfbridge/provider_plan.go
```diff
@@ -68,5 +70,12 @@ func (p *provider) plan(
 		return nil, err
 	}
 
-	return planResp, nil
+	var diags multierror.Error
```
t0yv0, line 73: Won't block on it but this looks useful.

Does it need to be special cased or it's just a way to handle things? there is this typical stanza I try to convert everything into:

```
if err := p.processDiagnostics(planResp.Diagnostics); err != nil {
   return err
}
```

If that works perhaps your handlers could go into improving processDiagnostics implementation? Multierror sounds useful.

We also need to hook up logging but it's a different ticket.

pf/tests/testdata/basicprogram/Pulumi.yaml
```diff
@@ -86,6 +86,14 @@ resources:
             - port: 22
               handlers:
                 - sshd
+  tupleElementAttrRes:
```
t0yv0, line 89: Brilliant.

pf/tests/internal/testprovider/tuple_type.go
```diff
@@ -0,0 +1,179 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+
+	"github.com/pulumi/pulumi-terraform-bridge/pf/internal/pfutils"
+)
+
+type TupleType struct {
```
t0yv0, line 29: So if I understand this right this code is needed to even use tuples in PF via CustomType hooks.. So users are expected to write this up. Very interesting.

pf/tests/internal/cmd/pulumi-resource-testbridge/schema.json
```diff
@@ -231,6 +256,13 @@
                 "statedir": {
                     "type": "string",
                     "description": "Dir to store pseudo-cloud state in.\n"
+                },
+                "tuplesOptionals": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/types/testbridge:index/TestresTuplesOptional:TestresTuplesOptional"
```
t0yv0, line 263: Ah every tuple will generate a named type! That's not ideal.. I think we also have type merging in tfgen. Two structurally identical types will possibly collapse to the same type name so maybe that's good. The name itself is weird though because ideally it'd be a system type. Regardless given our conversations on lack of evidence of prevalence we can shelve this till we find actual use cases.

pf/internal/schemashim/type_schema.go
```diff
@@ -66,6 +66,9 @@ func (s *typeSchema) Elem() interface{} {
 	case types.MapType:
 		contract.Assert(s.nested == nil || len(s.nested) == 0)
 		return newTypeSchema(tt.ElemType, nil)
+	case pfattr.TypeWithElementTypes:
```
t0yv0, line 69: This might match more than tuples, it might match set.; perhaps lift the case types.SetType into a separate match.

pf/internal/schemashim/object_pseudoresource.go
```diff
@@ -119,3 +125,117 @@ func (*objectPseudoResource) Set(key string, value shim.Schema) {
...
+	return newTypeSchema(types[i], r.attrs), true
+}
+
+func (r *tuplePseudoResource) Len() int {
+	return len(r.tuple.ElementTypes())
+}
+
+func (r *tuplePseudoResource) Range(each func(key string, value shim.Schema) bool) {
+	for i, v := range r.tuple.ElementTypes() {
+		k := fmt.Sprintf("t%d", i)
```
t0yv0, line 192: That looks exactly like objectPseudoResource with `t1,t2,t3` no? Or is tuplePseudoResource has slightly different behavior? Curious, ok either way.

pf/internal/schemashim/object_pseudoresource.go
```diff
@@ -28,13 +31,16 @@ import (
 // which assumes schema.Elem() would return either a Resource or a Schema. This struct packages the object field names
 // an types schema through a pseudo-Resource.
 type objectPseudoResource struct {
+	schemaOnly
```
t0yv0, line 34: Neat trick!

pf/internal/schemashim/convert_type.go
```diff
@@ -15,36 +15,43 @@
...
 
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 )
 
 func convertType(typ pfattr.Type) (shim.ValueType, error) {
+	// typ.TerraformType is 1:1 to the TF wire format, and is thus closed to outsider
+	// implementation. This allows us to maintain an exhaustive list of possible types and
+	// values.
+	tftype := typ.TerraformType(context.Background())
+	is := func(typ tftypes.Type) bool { return typ.Is(tftype) }
```
t0yv0, line 33: Is is better than Equal?

pf/internal/schemashim/attr_schema.go
```diff
@@ -91,6 +91,10 @@ func (s *attrSchema) Elem() interface{} {
 		var res shim.Resource = newObjectPseudoResource(tt, s.attr.Nested())
 		return res
 	}
+	if tt, ok := t.(pfattr.TypeWithElementTypes); ok {
```
t0yv0, line 94: Also here worried this will catch more than tuples..

pf/internal/convert/tuple.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2016-2022, Pulumi Corporation.
```
t0yv0, line 1: It's 2023

t0yv0 (approved):
Looking good, had a few questions and nits.

pf/tests/internal/testprovider/tuple_type.go
iwahbe: I think tuples are less supported than possible right now. Defining our own tuple lets us test our ability to handle the tuple wire format and test support custom types at the same time.

pf/internal/schemashim/type_schema.go
iwahbe: It shouldn't match `set`. According to the documentation, set should implement `pfattr.TypeWithElementType` (sets, lists and map), but not `pfattr.TypeWithElementTypes` (tuple). Note the plural difference.

pf/internal/schemashim/object_pseudoresource.go
iwahbe: It's very similar, but it doesn't involve an `types.ObjectType`. They could probably be unified more cohesively into a `pseudoObjectType[T]` or a type that just holds a map without understanding its origin.

pf/internal/schemashim/type_schema.go
t0yv0: Aha!

pf/internal/schemashim/object_pseudoresource.go
t0yv0: Ah so there is a difference. It's the right call then to have a separate struct. Thanks!

pf/internal/schemashim/convert_type.go
iwahbe: On the types we are using it for, I believe it is equivalent. Conceptually, `Is` is a shallow comparison, while `Equal` is a deep comparison. Since we are using it on inherently shallow types, there is no difference. I can switch back to `Equal`.

pf/internal/schemashim/attr_schema.go
iwahbe: It shouldn't. See the [documentation](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/attr#TypeWithElementTypes) for `pfattr.TypeWithElementTypes`.

It says
> TypeWithElementTypes extends the Type interface to include information about the types of each element. **Element types are part of the definition of a tuple type**.

(emphasis mine)

iwahbe:
Bridge tests are passing. The downstream tests are failing due to gradle upgrading to 8.0.0, which is a known problem unrelated to this PR.

PR #811, opened 2023-02-14
Introduce a package for experimental APIs

t0yv0:
Per team conversation, a better home for experimental APIs that are subject to renaming.


aq17 (approved):
😄

t0yv0:
CC @iwahbe moving these APIs into `x/` per @AaronFriel solid suggestion - it sounds like this is somewhat idiomatic Go to have experimental APIs in `x`. AS the apis mature we can continue shipping `x` for a bit to make it smooth for consumers to migrate, at least our own consumers in repos that ecosystem controls.

PR #813, opened 2023-02-14
Move tests also

t0yv0:
The last PR had the omission of not moving the unit tests.


iwahbe (approved):
Lets wait for the tests to go green before merging.

Issue #751, opened 2023-01-26
[Plugin Framework] Support ignoreChanges correctly in the Diff method


t0yv0:
Pulumi allows users specifying [ignoreChanges](https://www.pulumi.com/docs/intro/concepts/resources/options/ignorechanges/) option when creating resources, and the bridge needs to support this correctly for providers bridged from the Terraform Plugin Framework. Specifically, Pulumi CLI will pass the user's chosen options via the [Diff method](https://github.com/pulumi/pulumi/blob/master/proto/pulumi/provider.proto#L55) of the Provider Protocol. It is currently the provider's responsibility to take this option into account, and failing to do so correctly has a large impact on the behavior of the Pulumi program, as the Diff method is responsible for deciding whether the resource needs to be Updated in-place or replaced.

This is further complicated by the ignoreChanges having a complex grammar for nested properties and wildcards. Perhaps  [ParsePropertyPath](https://github.com/pulumi/pulumi/blob/fc27707511f646c111a338e7424c63f61386de53/sdk/go/common/resource/properties_path.go#L47-#L47) functionality can be reused to help correctly match ignoreChanges patterns to diffs.

PR #816, opened 2023-02-15
Implement ignoreChanges for PluginFramework

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/751

### DIFF
--- a/pf/tfbridge/ignore_changes.go
+++ b/pf/tfbridge/ignore_changes.go
@@ -108,6 +108,24 @@ type schemaStepper interface {
 	PropertyKey(convert.TerraformPropertyName) resource.PropertyKey
 }
 
+type terminalStepper struct {
+	inner schemaStepper
+}
+
+func (s *terminalStepper) Property(resource.PropertyKey) schemaStepper {
+	return nil
+}
+
+func (s *terminalStepper) Element() schemaStepper {
+	return nil
+}
+
+func (s *terminalStepper) PropertyKey(n convert.TerraformPropertyName) resource.PropertyKey {
+	return s.inner.PropertyKey(n)
+}
+
+var _ schemaStepper = (*terminalStepper)(nil)
+
 // Matching names at an intermediary type such as List[T].
 type typeSchemaStepper struct {
 	renames convert.PropertyNames
@@ -188,6 +206,23 @@ func typeStepper(renames convert.PropertyNames, s *schema.PackageSpec, t *schema
 				schema:  s,
 				token:   tokens.Token(tok),
 				renames: renames,
+			}
+		}
+
+		if strings.HasPrefix(t.Ref, "#/resources/") {
+			tok := strings.TrimPrefix(t.Ref, "#/resources/")
+			// dangling refs not supported
+			if _, ok := s.Resources[tok]; !ok {
+				return nil
+			}
+			// #/resources/ refs do not need ignoreChanges to recur into the referenced resource properties,
+			// hence the terminalStepper.
+			return &terminalStepper{
+				&namedEntitySchemaStepper{
+					schema:  s,
+					token:   tokens.Token(tok),
+					renames: renames,
+				},
 			}
 		}
 
--- a/pf/tfbridge/ignore_changes.go
+++ b/pf/tfbridge/ignore_changes.go
@@ -247,11 +247,22 @@ var _ schemaStepper = (*namedEntitySchemaStepper)(nil)
 type matchingContext struct {
 	schemaStepper    schemaStepper
 	remainingPattern resource.PropertyPath
+	terminated       bool
 	matchFailed      bool
 }
 
+func (mc *matchingContext) terminate() *matchingContext {
+	return &matchingContext{
+		terminated:  true,
+		matchFailed: mc.matchFailed,
+	}
+}
+
 func (*matchingContext) fail() *matchingContext {
-	return &matchingContext{matchFailed: true}
+	return &matchingContext{
+		terminated:  true,
+		matchFailed: true,
+	}
 }
 
 func (*matchingContext) pulumiNameMatches(pattern interface{}, puName resource.PropertyKey) bool {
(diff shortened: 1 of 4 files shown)


t0yv0:
This wouldn't quite work for maxitem=1 flattening as the paths will be off by 1. I'll have to defer that. But as is should be plenty useful and unblocks the upgrade workaround for tls.

pf/tfbridge/ignore_changes.go
```diff
@@ -0,0 +1,301 @@
...
+	return false
+}
+
+type schemaStepper interface {
+	Property(resource.PropertyKey) schemaStepper
+	Element() schemaStepper
+	PropertyKey(convert.TerraformPropertyName) resource.PropertyKey
+}
+
+type typeSchemaStepper struct {
```
iwahbe, line 102: Can I get a doc comment explaining what this is? I read it as "The type of a schema stepper", but that seems wrong.

pf/tfbridge/ignore_changes.go
```diff
@@ -0,0 +1,301 @@
...
+	return r.renames.PropertyKey(r.token, n, nil)
+}
+
+func typeStepper(renames convert.PropertyNames, s *schema.PackageSpec, t *schema.TypeSpec) schemaStepper {
+	// properties may refer to named object types via refs
+	if t.Ref != "" {
+		// understand local type refs
+		if strings.HasPrefix(t.Ref, "#/types/") {
+			tok := strings.TrimPrefix(t.Ref, "#/types/")
+			// dangling refs not supported
```
iwahbe, line 173: Dangling regs convert into token types, which are primitives. I think we can say that dangling refs don't have fields. I'm not sure if we should say that they aren't supported, since I think they are...

pf/tfbridge/ignore_changes.go
```diff
@@ -0,0 +1,301 @@
...
+	return nil
+}
+
+func (r *namedEntitySchemaStepper) PropertyKey(n convert.TerraformPropertyName) resource.PropertyKey {
+	return r.renames.PropertyKey(r.token, n, nil)
+}
+
+func typeStepper(renames convert.PropertyNames, s *schema.PackageSpec, t *schema.TypeSpec) schemaStepper {
+	// properties may refer to named object types via refs
+	if t.Ref != "" {
```
iwahbe, line 169: `t.Ref` can refer to `"#/resources/"` as well. I think we can handle that easily in this first draft.

pf/tfbridge/ignore_changes.go
```diff
@@ -0,0 +1,301 @@
...
+			schema: schema,
+			token:  token,
+			// Another complication is that Pulumi renames the properties, this uses the Renames framework
+			// to track the name tables.
+			renames: renames,
+		},
+	}, nil
+}
+
+func (ic *ignoreChanges) IsIgnored(ap *tftypes.AttributePath) bool {
```
iwahbe, line 67: This is `O(n)` and only used in a loop. Is it worth optimizing here?

pf/tfbridge/ignore_changes.go
```diff
@@ -0,0 +1,301 @@
...
+		return typeStepper(p.renames, p.schema, p.t.AdditionalProperties)
+	}
+	if p.t.Items != nil {
+		return typeStepper(p.renames, p.schema, p.t.Items)
+	}
+	return nil
+}
+
+func (*typeSchemaStepper) Property(k resource.PropertyKey) schemaStepper {
+	// there are no named properties here, those are supported by namedEntitySchemaStepper
```
iwahbe, line 121: I thought you could have a map with an ignored property. Is that wrong?

iwahbe (commented):
I think this is good, but it would really benefit from some comments explaining what the various `schemaStepper`s are stepping on.

pf/tfbridge/ignore_changes.go
t0yv0: I can try, but these types should have been anonymous.. This is all machinery to support matching algo reusing tftypes.WalkAttributePath...

pf/tfbridge/ignore_changes.go
t0yv0: Hmm that's interesting. Do you have an example where we use that? I can search some schemas.

pf/tfbridge/ignore_changes.go
t0yv0: Not sure what change you'd like to see here. Optimizing path matching from current `O(len(rawIgnoreChanges)) * O(len(pathsInDiff))` would need some semi-advanced code and I think's not worth it, or are you suggesting something else?

pf/tfbridge/ignore_changes.go
t0yv0: Sure and it works https://github.com/pulumi/pulumi-terraform-bridge/blob/509f13f49641c069d7a7c9259fada738909e3002/pf/tfbridge/ignore_changes_test.go#L128-#L128

TF framework makes a distinction between tftypes.ElementKeyString and tftypes.AttributeName in paths. It actually helps us here because then we know not to apply name mangling to tftypes.ElementKeyString. Property func is for situations where name mangling is required (attirs).

pf/tfbridge/ignore_changes.go
iwahbe: Nope, I just wanted to make sure we knew its `O(n^2)`. I think it's fine not optimizing until we see a slow down.

pf/tfbridge/ignore_changes.go
iwahbe: We use it in `awsx` a lot. I'm not sure if this ever appears in a TF bridged provider. We could just provided a custom error message telling the user to let us know if they see it. After seeing that TF bridged providers never use this functionality, it doesn't seem like a high priority.

pf/tfbridge/ignore_changes.go
iwahbe: You can reference types across packages, but it has a different syntax. It doesn't start  with `#/`. See https://www.pulumi.com/docs/guides/pulumi-packages/schema/#type (`$ref`) for details.

pf/tfbridge/ignore_changes.go
t0yv0: Yeah perfect I think just a solid comment could work here for now.

iwahbe (approved):

t0yv0:
@iwahbe worth another look at the addendum 🙏 
 
I've noticed a few things:

- didn't handle prefix matches right; if p is in ignoreChanges, then likely `p.x.y.z` paths should be ignored wholesale (prefix match)

- resources refs didn't work right, this is now fixed; thanks for pointing that out

- I don't have a clean solution without a bit of refactoring for maxitems=1 situation but I think I found a workaround that can be in there for now

iwahbe (approved):
This looks good as far as I can tell. LGTM!

PR #826, opened 2023-02-16
Clean up TODOs

t0yv0:
Implements simple TODOs, removes TODOs from things that permanently should panic, cross-links important misses to the issue tracker.

### DIFF
--- a/pf/Makefile
+++ b/pf/Makefile
@@ -67,6 +67,9 @@ tidy::
 	cd tests/internal/tlsshim && go mod tidy
 	cd tests/testdatagen/genrandom && go mod tidy
 
-.PHONY: lint
+.PHONY: lint todos
 lint:
 	golangci-lint run
+
+todos::
+	git grep TODO | grep -v terraform-bridge | grep -v context.TODO
--- a/pf/Makefile
+++ b/pf/Makefile
@@ -70,6 +70,3 @@ tidy::
 .PHONY: lint todos
 lint:
 	golangci-lint run
-
-todos::
-	git grep TODO | grep -v terraform-bridge | grep -v context.TODO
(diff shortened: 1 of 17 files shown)


pf/internal/schemashim/block_schema.go
```diff
@@ -95,12 +95,38 @@ func (s *blockSchema) Deprecated() string { return s.block.GetDeprecationMessage
...
-func (*blockSchema) ConflictsWith() []string                            { panic("TODO") }
-func (*blockSchema) Default() interface{}                               { panic("TODO") }
-func (*blockSchema) DefaultFunc() shim.SchemaDefaultFunc                { panic("TODO") }
-func (*blockSchema) DefaultValue() (interface{}, error)                 { panic("TODO") }
-func (*blockSchema) ExactlyOneOf() []string                             { panic("TODO") }
-func (*blockSchema) SetElement(config interface{}) (interface{}, error) { panic("TODO") }
-func (*blockSchema) SetHash(v interface{}) int                          { panic("TODO") }
-func (*blockSchema) StateFunc() shim.SchemaStateFunc                    { panic("TODO") }
-func (*blockSchema) UnknownValue() interface{}                          { panic("TODO") }
+func (*blockSchema) ConflictsWith() []string {
```
iwahbe, line 98: It's probably worth embedding a `schemaGenerationOnly` type in `blockSchema`, `objectPseudoResource` and `typeSchema`. That way we don't need to repeat this message for each type.

pf/Makefile
```diff
@@ -67,6 +67,9 @@ tidy::
 	cd tests/internal/tlsshim && go mod tidy
 	cd tests/testdatagen/genrandom && go mod tidy
 
-.PHONY: lint
+.PHONY: lint todos
 lint:
 	golangci-lint run
+
+todos::
+	git grep TODO | grep -v terraform-bridge | grep -v context.TODO
```
iwahbe, line 75: I don't think we need a permanent target for this. We aren't going to use it, are we?

iwahbe (approved):
LGTM

pf/internal/schemashim/block_schema.go
t0yv0: It's in the "schemashim" package but feel free to rename. I'll just merge for now.

PR #843, opened 2023-02-28
Add workflow to update providers on release

aq17:
Fixes https://github.com/pulumi/home/issues/2589
Automatically trigger [update-bridge-all-providers](https://github.com/pulumi/ci-mgmt/blob/master/.github/workflows/update-bridge-all-providers.yml) in `pulumi/ci-mgmt` on release

### DIFF
--- a/.github/workflows/update-providers-ci-mgmt.yml
+++ b/.github/workflows/update-providers-ci-mgmt.yml
@@ -1,11 +1,19 @@
 name: Update Providers with new bridge version via ci-mgmt
 on:
   workflow_dispatch:
+  release:
+    types: [published]
+env:
+  GH_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 jobs:
-  update-providers:
-    name: update-providers
+  upgrade_provider:
+    name: upgrade-provider
     runs-on: ubuntu-latest
     steps:
-      - name: Check out source code
-        uses: actions/checkout@master
-    # TBD
+    - name: Call upgrade provider action
+      uses: pulumi/pulumi-upgrade-provider-action@v0.0.4
+      with:
+        slack-channel: provider-upgrade-status
+        slack-webhook: ${{ secrets.SLACK_WEBHOOK_URL }}
+        kind: "bridge"


.github/workflows/update-providers-ci-mgmt.yml
```diff
@@ -0,0 +1,38 @@
+name: Update Providers with new bridge version via ci-mgmt
```
t0yv0, line 1: Perhaps this workflow can live in ci-mgmt repo?

t0yv0: Ah never mind, I see this dispatches from release to update-bridge-all-providers.yml...

I think it would be good to first test out if update-bridge-all-providers.yml works, and narrow it down to ecosystem-only repos.

.github/workflows/update-providers-ci-mgmt.yml
```diff
@@ -1,4 +1,50 @@
...
+      - uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.PULUMI_BOT_TOKEN }}
+          script: |
+            await github.rest.actions.createWorkflowDispatch({
+              owner: 'pulumi',
+              repo: 'ci-mgmt',
+              workflow_id: 'update-bridge-ecosystem-providers.yml',
+              ref: 'master',
+              inputs: {"bridge_version": "${{ needs.get-bridge-latest-version.outputs.bridge-latest }}", "sdk_version": "${{ needs.get-sdk-latest-version.outputs.sdk-latest }}", "automerge": true}
```
aq17, line 48: @t0yv0 do you see anything wrong with this request? Failing with https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/4326824183/jobs/7554718807 but the three parameters look like they match with https://github.com/pulumi/ci-mgmt/blob/master/.github/workflows/update-bridge-ecosystem-providers.yml#L11-L27

AaronFriel: It looks like `automerge` might need to be a string, despite the workflow dispatch config. https://github.com/cli/cli/issues/5246

.github/workflows/update-providers-ci-mgmt.yml
```diff
@@ -1,4 +1,50 @@
 name: Update Providers with new bridge version via ci-mgmt
 on:
+  release:
+    types: [published]
```
t0yv0, line 4: Hmm. I publish releases for pf/* tags separately from normal release. Will this confuse it?

t0yv0 (approved):
I'm ok to start trying this out and fixing incrementally. If it works at all this is good value. I worry it might confuse tags but we can observe it and see what happens. 

@aq17 perhaps try to run manually before merging and report how that went?

.github/workflows/update-providers-ci-mgmt.yml
```diff
@@ -1,4 +1,50 @@
...
+          echo "::set-output name=version::$version"
+    outputs:
+      bridge-latest: ${{ steps.bridge-version.outputs.version }}
+  get-sdk-latest-version:
+    runs-on: ubuntu-latest
+    steps:
+      - id: get-sdk-latest
+        uses: pozetroninc/github-action-get-latest-release@d1dafdb6e338bdab109e6afce581a01858680dfb
+        with:
+          repository: pulumi/pulumi
```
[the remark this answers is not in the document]
aq17: The target workflow requires an SDK version, so figure it's best to use the latest for now

.github/workflows/update-providers-ci-mgmt.yml
aq17: Good catch, yes it's getting confused.

aq17:
@t0yv0 coming back to this, much easier now that we have our upgrade provider action. We should be able to do something similar in `pu/pu` to target `pu/sdk` and `pu/pu` updates on release also

PR #844, opened 2023-02-28
Add PlanState DiffStrategy to sdkv2 Diff under a feature flag

### LINKED ISSUE #834: Inconsistent previews due to mis-application of set hashing
### What happened?

Continuing to investigate https://github.com/pulumi/pulumi-terraform-bridge/issues/791 issue, as the root cause may be causing inconsistent previews across resources we do not know about.

### Expected Behavior

Consistent update/no-change plans.

### Steps to reproduce

See #791.

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

t0yv0:
Fixes #834 

This introduces a change to how sdkv2 Diff method operates, making it more compatible with how Terraform CLI does change planning. Given that many production providers rely on sdkv2, the change is off-by-default under a feature flag, and is to be rolled out gradually.


pkg/tfshim/sdk-v2/cty_test.go
```diff
@@ -415,20 +417,20 @@ func TestMakeResourceRawConfig(t *testing.T) {
 			},
 			expected: cty.ObjectVal(map[string]cty.Value{
 				"allowed_logout_urls": cty.NullVal(cty.List(cty.String)),
-				"change_password": cty.ListValEmpty(cty.Object(map[string]cty.Type{
+				"change_password": cty.NullVal(cty.List(cty.Object(map[string]cty.Type{
```
t0yv0, line 420: The fix to cty.go replaced ListValEmpty with NullVal in some of the inputs to the SimpleDiff method, which sounds benign but I wasn't 100% sure so just in case I subjected the fix to flagging also.

pkg/tfshim/sdk-v2/diff_strategy.go
```diff
@@ -0,0 +1,74 @@
...
+package sdkv2
+
+import (
+	"fmt"
+	"os"
+)
+
+// Configures how the provider performs Diff. Since this is a sensitive method that can result in unexpected breaking
+// changes, using a configurable DiffStrategy as a feature flag assists gradual rollout.
+type DiffStrategy int
```
t0yv0, line 24: This is the feature flag.

pkg/tfshim/sdk-v2/diff_strategy.go
```diff
@@ -0,0 +1,74 @@
...
+	case "PlanState":
+		return PlanState, nil
+	case "TryPlanState":
+		return TryPlanState, nil
+	default:
+		return 0, fmt.Errorf("Unknown DiffStrategy: %q", text)
+	}
+}
+
+const diffStrategyEnvVar = "PULUMI_DIFF_STRATEGY"
```
t0yv0, line 65: This is the env var to override.

pkg/tfshim/sdk-v2/provider_options.go
```diff
@@ -0,0 +1,51 @@
...
+
+package sdkv2
+
+type providerOptions struct {
+	diffStrategy DiffStrategy
+}
+
+type providerOption func(providerOptions) (providerOptions, error)
+
+func WithDiffStrategy(s DiffStrategy) providerOption { //nolint:revive
```
t0yv0, line 23: This is the option to pass at compile time to opt into the default DiffStrategy for a provider.

pkg/tfshim/sdk-v2/diff_strategy.go
```diff
@@ -0,0 +1,74 @@
...
+
+func ParseDiffStrategyFromEnv() (DiffStrategy, bool, error) {
+	s := os.Getenv(diffStrategyEnvVar)
+	if s == "" {
+		return 0, false, nil
```
iwahbe, line 70:
```suggestion
		return ClassicDiff, false, nil
```

pkg/tfshim/sdk-v2/internal/tf/generate.go
```diff
@@ -0,0 +1,170 @@
+// Copyright 2016-2023, Pulumi Corporation.
```
iwahbe, line 1: I would prefer to transfer this over to linker tricks if possible, but we can do that at a later date.

pkg/tfshim/sdk-v2/proposed_new.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	queue := newQueue(schema.BlockTypes)
+
+	destinations := map[interface{}]map[string]*configschema.NestedBlock{}
+	for _, b := range schema.BlockTypes {
+		destinations[b] = block.BlockTypes
+	}
+
+	for !queue.empty() {
```
iwahbe, line 125: It would be really helpful to have a comment describing what this algorithm is doing.

iwahbe (approved):
A couple nits, but LGTM!

pkg/tfshim/sdk-v2/internal/tf/generate.go
t0yv0: The question I have about linker tricks, is there a way to do this that does not infect downstream consumers of the lib to have to apply he same "go mod replace" in the consuming Go module? I was assuming that's not going to work, so we'd be infecting the downstream with more "go mod replace", but I can run a quick experiment to verify.

pkg/tfshim/sdk-v2/proposed_new.go
t0yv0: Ah yes this one is quite an embarrassment. I have `a->b` function that' recursive but since type a is an internal type from a foreign go module I can't seem to be able to write out `func (x a) b` because I can't reference `a`! This does not prevent the program from manipulating values of type `a` and in fact storing them in a queue :) The rest are workarounds. Let me try a few more direct workarounds using generics, if not I'll just add a comment.

t0yv0: Added a comment.

pkg/tfshim/sdk-v2/internal/tf/generate.go
t0yv0: Yes I'm pretty sure that linker tricks would impose `go mod replace` hell on our library consumers which is pretty unfortunate.

iwahbe: There are different kinds of linker tricks. If we use one like our shamed providers, then we impose downstream costs. If we do some of the ([`go:linkname`](https://github.com/pulumi/pulumi/blob/4dc2943ae42f153872641140c48349b6f4681b87/sdk/go/pulumi/provider_linked.go#L48)) module dodging that pulumi/pulumi does, it doesn't have downstream impacts.

t0yv0: Hm interesting, that wasn't part of my toolkit thanks!

t0yv0:
Filed https://github.com/pulumi/pulumi-terraform-bridge/issues/865

PR #849, opened 2023-03-02
Share Replay API in a public go module

t0yv0:
I'm finding the need to use Replay-style testing outside of Plugin Framework, for example in https://github.com/pulumi/pulumi-terraform-bridge/pull/788 to regress a pre-Plugin Framework issue. This is impossible since it's currently hiding as an internal package inside the `pf` Go module. This PR lifts it to a new go module that's now public, so that it can be consumed by both `pf` Go module and `github.com/pulumi/pulumi-terraform-bridge/v3` Go module.

API is cleaned up a bit and documented.

`module github.com/pulumi/pulumi-terraform-bridge/testing` is the new Go module. It's a separate one as it's intended just for test dependencies.

The `x` package is indicating the API is experimental before we have some more alignment where to put it.

JSON match functionality is added to be able to use wildcards "*" in the tests to ignore properties irrelevant for the test. It's a frequent need.


AaronFriel (approved):
This looks great, and I'm sorry I missed the original PR adding replay tests. 

This looks a lot like snapshot testing I've seen in the frontend space. Could we use something like [hexops/autogold](https://github.com/hexops/autogold) to generate inline snapshots? 

I like inline snapshot testing utils like autogold because the diff is usually near the test, and when I intentionally change behavior I can use a tool to regenerate the snapshot.

t0yv0:
https://github.com/hexops/autogold looks lovely and well thought out! Can be nice to think about, also perhaps in how we do codegen tests! I've also enjoyed using hexops/valast before. Great lib.

A fair bit of Replay/ReplaySequence tests end up being human edited after being extracted from the program, and the original program is often too complex to include, so we can't re-gen them, but ReplayFile is definitely amenable to re-generation of "golden" values.

t0yv0:
Also FWIW - a lot of pf coverage is done through these Replay-style tests, this is kind of where it got incubated.

PR #860, opened 2023-03-03
Introduce PreStateUpgradeHook for Plugin Framework

t0yv0:
PreStateUpgradeHook seems useful in a few recent exceptional situations:

- pulumi-random provider versions to an early version of PF that contaminated users's state do not let us ship an easy upgrade that just works; writing some logic in PreStateUpgradeHook could be used to auto-fix it

- pulumi-tls provider updates to FP are breaking because upstream did not code up state migrations properly, but using this hook can let us compensate at Pulumi level: https://github.com/pulumi/pulumi-tls/pull/190

Open to suggestions on alternative mechanism and/or API design.

### DIFF
--- a/pf/tfbridge/resource_state.go
+++ b/pf/tfbridge/resource_state.go
@@ -156,7 +156,12 @@ func updateTFSchemaVersion(m resource.PropertyMap, version int64) (resource.Prop
 	} else {
 		meta = map[string]interface{}{}
 	}
-	meta["schema_version"] = version
+	if version != 0 {
+		meta["schema_version"] = fmt.Sprintf("%d", version)
+	}
+	if len(meta) == 0 {
+		return m, nil
+	}
 	updatedMeta, err := json.Marshal(meta)
 	if err != nil {
 		return nil, err
--- a/pf/tfbridge/resource_state.go
+++ b/pf/tfbridge/resource_state.go
@@ -47,7 +47,7 @@ func (u *upgradedResourceState) ToPropertyMap(rh *resourceHandle) (resource.Prop
 		return nil, err
 	}
 
-	return addTFSchemaVersion(propMap, u.state.TFSchemaVersion)
+	return updateTFSchemaVersion(propMap, u.state.TFSchemaVersion)
 }
 
 func (u *upgradedResourceState) ExtractID(rh *resourceHandle) (resource.ID, error) {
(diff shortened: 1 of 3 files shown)


t0yv0:
Umm, this will require a ci-mgmt change to actually run the unit tests. I can move the code to 'examples' to avoid having to do that change now.

pkg/tfbridge/info.go
```diff
@@ -180,6 +180,9 @@ type ResourceInfo struct {
 	DeprecationMessage  string      // message to use in deprecation warning
 	CSharpName          string      // .NET-specific name
 
+	// Optional hook to run before upgrading the state. This is currently only supported for Plugin-Framework based
+	// providers.
+	PreStateUpgradeHook PreStateUpgradeHook
```
iwahbe, line 185: Do we plan to support `PreStateUpgradeHook` on all bridged providers in the future? If so, can you comment with the tracking issue. If not, could we put this in `pf/tfbridge.ProviderInfo`?
https://github.com/pulumi/pulumi-terraform-bridge/blob/92f231195177017c423321aa4f0c17d524f17e1d/pf/tfbridge/info.go#L22-L33

pf/tfbridge/resource_state.go
```diff
@@ -139,3 +157,24 @@ func parseTFSchemaVersion(m resource.PropertyMap) (int64, error) {
...
+	if metaProperty, hasMeta := m[metaKey]; hasMeta && metaProperty.IsString() {
+		if err := json.Unmarshal([]byte(metaProperty.StringValue()), &meta); err != nil {
+			err = fmt.Errorf("expected %q special property to be a JSON-marshalled string: %w",
+				metaKey, err)
+			return nil, err
+		}
+	} else {
+		meta = map[string]interface{}{}
+	}
+	meta["schema_version"] = version
```
iwahbe, line 172: If we didn't have `updateTFSchemaVersion` before, what did we use `"schema_version"` for in the past? Is this the TF schema version, or a construct internal to the bridge?

iwahbe (approved):
I'm not 100% sure how `"schema_version"` is used, but the mechanics look clean.

pkg/tfbridge/info.go
t0yv0: Pondering this, a bit of a design question.. I think it should be implementable and occasionally useful for pre-PF providers as well. I've added an issue:

https://github.com/pulumi/pulumi-terraform-bridge/issues/864

PR #863, opened 2023-03-06
`namecheck` resilient to unmapped resources/functions

iwahbe:
This allows pulumi-docker/v4 to build. Previously, it crashed with

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x200ca6d]

goroutine 1 [running]:
github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen.nameCheck({{0x2c91018, 0xc000948d80}, {0x2667432, 0x6}, {0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, ...}, ...)
        $HOME/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.41.0/pkg/tfgen/namecheck.go:93 +0x48d
github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen.(*Generator).Generate(0xc0007498c0)
        $HOME/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.41.0/pkg/tfgen/generate.go:890 +0x6d8
github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen.Main.func1({{0x2667432, 0x6}, {0x2c581a0, 0x1f}, {0x7ffc70d971fd, 0x6}, {{0x2c91018, 0xc000948d80}, {0x2667432, 0x6}, ...}, ...})
```

### DIFF
--- a/pkg/tfgen/namecheck.go
+++ b/pkg/tfgen/namecheck.go
@@ -84,26 +84,44 @@ func nameCheck(
 		all = append(all, ps)
 	}
 
+	// It is not required that all resources/functions are mapped in `prov`. We need
+	// to be resilient to missing items.
+	schemaGetter := func(resourceMap func() shim.ResourceMap) func(string) shim.SchemaMap {
+		m := resourceMap()
+		return func(key string) shim.SchemaMap {
+			if m == nil {
+				return nil
+			}
+			r, ok := m.GetOk(key)
+			if !ok || r == nil {
+				return nil
+			}
+			return r.Schema()
+		}
+	}
+
+	resourceSchema := schemaGetter(prov.P.ResourcesMap)
 	for tok := range spec.Resources {
 		key := renames.Resources[tokens.Type(tok)]
 		loc := fmt.Sprintf("Resource: %q\n    Token: %q", tok, key)
 		ps := nameCheckPropSet{
 			location:  loc,
 			props:     props[tokens.Token(tok)],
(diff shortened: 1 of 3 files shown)


pkg/tfgen/namecheck.go
```diff
@@ -84,26 +84,44 @@ func nameCheck(
 		all = append(all, ps)
 	}
 
+	// It is not required that all resources/functions are mapped in `prov`. We need
```
t0yv0, line 87: Are we sure his is factually correct? What happens with unmapped resources, don't they result in an error from tfgen and/or have no tokens allocated, or are not present in the schema? Directionally it makes sense though that we should not require all resources to be mapped.

t0yv0 (approved):

t0yv0:
A quickie test could be handy here. LGTM.

pkg/tfgen/namecheck.go
iwahbe: Unmapped errors result in an error from tfgen... by default. There are settings that turn off unmapped resource errors, so the rest of the process needs to be resilient to missing resources/functions.

t0yv0: Ah that's right. Well that makes sense!

PR #867, opened 2023-03-06
Pull in terraform for the new converter

Frassle:
We get access to terraforms parser by doing a git read-tree against https://github.com/hashicorp/terraform.git and writing it to a subfolder in tf2pulumi, followed by some automatic fix ups. We already had an old copy of some internal terraform code in pkg/tf2pulumi/internal, I've moved all that code down one level so it's at pkg/tf2pulumi/internal/tf12 (It was taken from version 0.12.0).


Makefile
```diff
@@ -45,3 +44,26 @@ tidy::
...
+	git remote remove terraform || true
+	git remote add -f -t main --no-tags terraform https://github.com/hashicorp/terraform.git
+	git rm -rf ./pkg/tf2pulumi/internal/tf
+	git read-tree --prefix=pkg/tf2pulumi/internal/tf/ -u terraform/main:internal
+	# Remove the parts of Terraform we don't need
+	find ./pkg/tf2pulumi/internal/tf -mindepth 1 -maxdepth 1 -type d -regex ".*/tf/\($(TFPKGS)\)" -exec rm -rf {} \;
+	# No need for us to keep any of Terraforms tests
+	find ./pkg/tf2pulumi/internal/tf -type f -regex ".*_test.go" -delete
+	# Update the .go file import paths
+	find ./pkg/tf2pulumi/internal/tf -type f -name "*.go" -exec sed -i -e 's|github.com/hashicorp/terraform/internal|github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tf2pulumi/internal/tf|g' {} \;
```
abhinav, line 60: Note that sed behaves differently on macOS vs Linux (GNU sed vs BSD sed).
Specifically, on macOS, `-i` _requires_ an argument and if one isn't specified, it'll treat `-e` as  the backup file suffix, which will generate "foo-e" files for every "foo" file, which will get checked in and committed below.

We could try to make the sed cross-platform, or we could use perl which behaves the same everywhere.
(Also, this can also use the `+` instead of `;` regardless of sed or perl.)

```suggestion
	find ./pkg/tf2pulumi/internal/tf -type f -name "*.go" -exec perl -p -i -e 's|github.com/hashicorp/terraform/internal|github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tf2pulumi/internal/tf|g' {} '+'
```

Makefile
```diff
@@ -45,3 +44,26 @@ tidy::
...
+
+# Update the subtree in pkg/tf2pulumi/internal/tf with the latest code from terraform master
+TFPKGS=backend\|command\|provider-terraform\|provider-simple\|provider-simple-v6\|provisioner-local-exec\|cloud\|builtin\|grpcwrap\|initwd\|earlyconfig\|moduledeps\|plugin\|tfplugin5\|plugin6\|tfplugin6\|communicator\|providercache\|e2e\|legacy\|moduletest\|helper\|repl\|terminal
+pull_terraform::
+	git remote remove terraform || true
+	git remote add -f -t main --no-tags terraform https://github.com/hashicorp/terraform.git
+	git rm -rf ./pkg/tf2pulumi/internal/tf
+	git read-tree --prefix=pkg/tf2pulumi/internal/tf/ -u terraform/main:internal
+	# Remove the parts of Terraform we don't need
+	find ./pkg/tf2pulumi/internal/tf -mindepth 1 -maxdepth 1 -type d -regex ".*/tf/\($(TFPKGS)\)" -exec rm -rf {} \;
```
abhinav, line 56: optional: `-exec` supports `+` instead of `;` to pass in as many paths as it can in one invocation.
So instead of one rm -rf per matched path, this could rm -rf them all in one invocation.
```suggestion
	find ./pkg/tf2pulumi/internal/tf -mindepth 1 -maxdepth 1 -type d -regex ".*/tf/\($(TFPKGS)\)" -exec rm -rf {} '+'
```

Makefile
```diff
@@ -45,3 +44,26 @@ tidy::
...
+	# No need for us to keep any of Terraforms tests
+	find ./pkg/tf2pulumi/internal/tf -type f -regex ".*_test.go" -delete
+	# Update the .go file import paths
+	find ./pkg/tf2pulumi/internal/tf -type f -name "*.go" -exec sed -i -e 's|github.com/hashicorp/terraform/internal|github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tf2pulumi/internal/tf|g' {} \;
+	# Workaround the duplicate registration of rpcFriendlyDiag in tf/tfdiags/rpc_friendly.go and same type in the public plugin repo
+	sed -i -e 's|rpcFriendlyDiag|internalRpcFriendlyDiag|g' ./pkg/tf2pulumi/internal/tf/tfdiags/rpc_friendly.go
```
abhinav, line 62: If you end up using perl:
```suggestion
	perl -p -i -e 's|rpcFriendlyDiag|internalRpcFriendlyDiag|g' ./pkg/tf2pulumi/internal/tf/tfdiags/rpc_friendly.go
```

pkg/tfgen/docs.go
```diff
@@ -1331,11 +1331,20 @@ func (g *Generator) convertHCLToString(hcl, path, languageName string) (string,
 		// we write an example to a temp file internally in order to pass to convert.Convert().
 		//
 		// fileName starts with a "/" which is not present in the resulting error, so we need to skip the first rune.
-		errMsg := strings.ReplaceAll(diags.All.Error(), fileName[1:], "")
+		errMsg := ""
+		first := true
+		for _, diag := range diags.All {
+			if !first {
+				errMsg += ", and "
+			}
```
abhinav, line 1339: Minor bug here: first is never reassigned so this will never add the ", and ".
Can omit it entirely with indexes.

```suggestion
		for idx, diag := range diags.All {
			if idx > 0 {
				errMsg += ", and "
			}
```

pkg/tfgen/docs.go
```diff
@@ -1331,11 +1331,20 @@ func (g *Generator) convertHCLToString(hcl, path, languageName string) (string,
...
+			errMsg += diag.Error()
+		}
+
+		errMsg = strings.ReplaceAll(errMsg, fileName[1:], "")
```
abhinav, line 1343: The comment above about removing the temporary filename from the error message should probably be moved here.

pkg/tfgen/docs.go
```diff
@@ -1331,11 +1331,20 @@ func (g *Generator) convertHCLToString(hcl, path, languageName string) (string,
...
 
 		g.warn("failed to convert HCL for %s to %v: %v", path, languageName, errMsg)
 		g.coverageTracker.languageConversionFailure(languageName, diags.All)
-		return "", fmt.Errorf(errMsg)
+		return "", fmt.Errorf("%s", errMsg)
```
abhinav, line 1347: minro nit: unecessary fmt.Errorf here.
This can be `errors.New(errMsg)`.

t0yv0:
I just had to do something similar for another terraform sub-tree and ended up packaging it via `go generate`.

Frassle:
Looking at a new approach for this

Issue #797, opened 2023-02-07
[Plugin Framework] Ensure CI runs tests


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Quick omission, the tests in `./pf` are not yet checked in CI.

PR #870, opened 2023-03-06
Integrate PF testing into main test targets

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/797


iwahbe (approved):
I appreciate keeping the logic in `Makefile` instead of GH actions.

PR #871, opened 2023-03-06
Improve lint, Go 1.20.x

t0yv0:
- Update go to 1.20.x
- Update setup-go to v3
- Update golangci-lint to a version compatible with 1.20.x
- Setup golangci-lint via official action
- `make lint` runs golangci-lint automatically on every module it finds in source (git ls-files go.mod search)
-  code is fixed up to pass the stricter lint checks


aq17 (approved):
LGTM

PR #872, opened 2023-03-07
Fix unknown value encoding in Plugin Framework

t0yv0:
There is a regression in pulumi-random https://github.com/pulumi/pulumi-random/issues/279 motivating this change.

Before the change, Go SDK emits Unknown Number value sentinel that confuses the Node SDK in the user's program.

After the change, only generic Unknown sentinel values are emitted and then the Node program works as expected.

### DIFF
--- a/pf/internal/convert/bool.go
+++ b/pf/internal/convert/bool.go
@@ -49,8 +49,7 @@ func (*boolEncoder) FromPropertyValue(p resource.PropertyValue) (tftypes.Value, 
 
 func (*boolDecoder) ToPropertyValue(v tftypes.Value) (resource.PropertyValue, error) {
 	if !v.IsKnown() {
-		c := resource.Computed{Element: resource.NewBoolProperty(false)}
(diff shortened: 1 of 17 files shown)


pf/internal/convert/bool.go
```diff
@@ -49,8 +49,7 @@ func (*boolEncoder) FromPropertyValue(p resource.PropertyValue) (tftypes.Value,
 
 func (*boolDecoder) ToPropertyValue(v tftypes.Value) (resource.PropertyValue, error) {
 	if !v.IsKnown() {
-		c := resource.Computed{Element: resource.NewBoolProperty(false)}
-		unknown := resource.NewComputedProperty(c)
+		unknown := resource.NewOutputProperty(resource.Output{Known: false})
```
iwahbe, line 52: This could be cleaned up with a constant in the convert module:

```go
var unknownProperty = resource.NewOutputProperty(resource.Output{Known: false})
```

iwahbe (approved):
LGTM

PR #873, opened 2023-03-07
Hook ProviderMetadata into ProviderInfo

iwahbe:
Adds a untyped KV store managed by the `tfbridge.ProviderInfo` life cycle.

The store is read-write when the schema is being generated, and is persisted to disk with `schema.json`. During normal provider operation (`pulumi-resource-${PKG}`), the store if not persisted (making it effectively read-only).

Because the user must opt-in to using `ProviderMetadata` (by embedding the file using `go:embed`), the store is only available when `ProviderInfo.MetadataInfo` is non-nil.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -98,9 +98,7 @@ type ProviderInfo struct {
 
 	// Information for the embedded metadata file.
 	//
-	// If provided, the user should call NewMetadataInfo(path, bytes) where path is
-	// the path (relative to schema.json) of the embedded file, and bytes is the
-	// compile time embedded file.
+	// See NewProviderMetadata for in-place construction of a *MetadataInfo.
 	MetadataInfo *MetadataInfo
 
 	UpstreamRepoPath string // An optional path that overrides upstream location during docs lookup
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -98,8 +98,9 @@ type ProviderInfo struct {
 
 	// Information for the embedded metadata file.
 	//
-	// If provided, the user should embed the file at `MetadataInfo.Path`, and store
-	// the resulting embedded bytes in `MetadataInfo.Bytes`.
+	// If provided, the user should call NewMetadataInfo(path, bytes) where path is
+	// the path (relative to schema.json) of the embedded file, and bytes is the
+	// compile time embedded file.
 	MetadataInfo *MetadataInfo
 
 	UpstreamRepoPath string // An optional path that overrides upstream location during docs lookup
@@ -121,7 +122,7 @@ func (info ProviderInfo) GetResourcePrefix() string {
 
 func (info ProviderInfo) GetMetadata() ProviderMetadata {
 	info.MetadataInfo.assertValid()
-	return info.MetadataInfo.data
+	return info.MetadataInfo.Data
 }
 
 func (info ProviderInfo) GetGitHubOrg() string {
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -96,6 +96,12 @@ type ProviderInfo struct {
 	PreConfigureCallback           PreConfigureCallback // a provider-specific callback to invoke prior to TF Configure
 	PreConfigureCallbackWithLogger PreConfigureCallbackWithLogger
 
+	// Information for the embedded metadata file.
+	//
+	// If provided, the user should embed the file at `MetadataInfo.Path`, and store
+	// the resulting embedded bytes in `MetadataInfo.Bytes`.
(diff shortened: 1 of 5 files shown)


unstable/doc.go
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// A package that exposes an unstable interface across go module boundaries. This is
+// necessary to get around granularity limitations of the `internal` path that go modules
+// provides.
+//
+// It is not recommended to rely on exports from this package. We do not provide stability
+// guarantees for this package.
+package unstable
```
t0yv0, line 21: Should this just go to x or unstable is a meaningful distinction?

pkg/tfbridge/metadata.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+)
+
+// A store persisted between `tfgen` and a running provider.
+type ProviderMetadata *metadata.Data
+
+type MetadataInfo struct {
+	Path string
```
t0yv0, line 26: Maybe give a default value to path, like "bridge-metadata.json" used in TFPF was something I came up with.

unstable/metadata/metadata.go
```diff
@@ -0,0 +1,97 @@
...
+	//
+	// See https://pkg.go.dev/encoding/json#Marshal for details.
+	contract.AssertNoErrorf(err, "internal: failed to marshal metadata")
+	return bytes
+}
+
+// Set a piece of metadata to a value.
+//
+// Set errors only if value fails to serialize.
+func Set(d *Data, key string, value any) error {
```
t0yv0, line 58: Consider `func (*Data) Set(..)`?

unstable/metadata/metadata.go
```diff
@@ -0,0 +1,97 @@
...
+	data, err := json.Marshal(value)
+	if err != nil {
+		return err
+	}
+	d.m[key] = data
+	return nil
+}
+
+func Get[T any](d *Data, key string) (T, bool, error) {
```
t0yv0, line 67: Ah, `func (*Data) Get()` does not work because generics? Funny.

unstable/metadata/metadata.go
t0yv0: Ah I see why you might want to make this symmetric to Get.

unstable/metadata/metadata.go
```diff
@@ -0,0 +1,97 @@
...
+	data, ok := d.m[key]
+	var t T
+	if !ok {
+		return t, false, nil
+	}
+	err := json.Unmarshal(data, &t)
+	return t, true, err
+}
+
+func Clone(data *Data) *Data {
```
t0yv0, line 77: Nit: This still would read better as `func (*Data) Clone()`

t0yv0:
> The store is read-write when the schema is being generated, and is persisted to disk with schema.json. During normal provider operation (pulumi-resource-${PKG}), the store if not persisted (making it effectively read-only).

This finally clicked! Sorry I didn't get this yesterday. Consider including this in the comment, this is important.

t0yv0 (approved):
A few nits but this looks *great*! This neatly solves the problem with passing opaque information from schema generation time to provider runtime through a stable interface we can evolve over time.

unstable/doc.go
iwahbe: There is a distinction, though I'm not sure it's meaningful. I would like the functionality in `x` to mature and be stabilized. Ideally, we will depreciate `x` for the `tfbridge` and we become confident in these APIs. `unstable` is an implementation detail of the bridge, and should never be depended upon out of repo. We don't want to stabilize it.

pkg/tfbridge/info.go
```diff
@@ -96,6 +96,13 @@ type ProviderInfo struct {
 	PreConfigureCallback           PreConfigureCallback // a provider-specific callback to invoke prior to TF Configure
 	PreConfigureCallbackWithLogger PreConfigureCallbackWithLogger
 
+	// Information for the embedded metadata file.
+	//
+	// If provided, the user should call NewMetadataInfo(path, bytes) where path is
```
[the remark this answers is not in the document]
iwahbe: The zero value works great (which is why the type is `*MetadataInfo`). I'll fix the language.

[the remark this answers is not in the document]
t0yv0: Ahh I see. I meant something like MetadataInfo{bytes: ..} forgetting the path, which would default to something. I see you have some validation in New so that's fine too.

PR #875, opened 2023-03-08
remove double-slashes to fix examples docs generation

jazzyfresh:
for https://github.com/pulumi/docs/issues/8672

pulumi-docker reached an edge case in docsgen that uses a template with invalid syntax (double-slashes, changed to single-slash should be the way to go according to docs team)

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1090,7 +1090,7 @@ func (g *Generator) convertExamples(docs, name string, stripSubsectionsWithError
 		strings.Contains(docs, "```csharp") {
 		// we have explicitly rewritten these examples and need to just return them directly rather than trying
 		// to reconvert them. But we need to surround them in the examples shortcode for rendering on the registry
-		return fmt.Sprintf("{{%% examples %%}}\n%s\n{{%% //examples %%}}", docs)
+		return fmt.Sprintf("{{%% examples %%}}\n%s\n{{%% /examples %%}}", docs)
 	}
 
 	output := &bytes.Buffer{}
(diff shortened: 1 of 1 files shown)


guineveresaenger (commented):
Line 1088 should also include `java`, no?

pkg/tfgen/docs.go
```diff
@@ -1090,7 +1090,7 @@ func (g *Generator) convertExamples(docs, name string, stripSubsectionsWithError
 		strings.Contains(docs, "```csharp") {
 		// we have explicitly rewritten these examples and need to just return them directly rather than trying
 		// to reconvert them. But we need to surround them in the examples shortcode for rendering on the registry
-		return fmt.Sprintf("{{%% examples %%}}\n%s\n{{%% //examples %%}}", docs)
+		return fmt.Sprintf("{{%% examples %%}}\n%s\n{{%% /examples %%}}", docs)
```
AaronFriel, line 1093: Is this simpler?

```suggestion
		return docs
```

guineveresaenger: I don't think so - I think it's possible that this is supplying short codes to other overwritten examples maybe?

guineveresaenger (approved):

jazzyfresh:
Fix was resolved https://github.com/pulumi/pulumi-docker/pull/524

iwahbe (approved):
LGTM!

t0yv0:
Can we have a test or an explanation why a test isn't possible/practical and how it's been tested thanks.

t0yv0:
Oh I see it's a P0. In that case let's backlog the test and get it in.

t0yv0 (approved):

jazzyfresh:
@t0yv0 we need to keep this branch alive temporarily

t0yv0:
Apologies! Got it.

jazzyfresh:
we can delete this branch since the schema file is already generated

PR #879, opened 2023-03-09
Remove mapCase from python generation

iwahbe:
This is an updated version of https://github.com/pulumi/pulumi-terraform-bridge/pull/369.


aq17 (approved):
LGTM!

Issue #732, opened 2023-01-20
[Plugin Framework] Verify defaults application


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Providers that are built on top of Terraform Plugin Framework and then bridged to Pulumi have notable differences around applying default values to properties.  The task is to add test coverage and understand various behaviors, possibly adding bug fixes or enhancements.

Prior to Plugin Framework, pulumi-terraform-bridge used an [involved algorithm](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/schema.go#L111) for managing default value application. It used a special `__defaults` pseudo-property to remember which values are set by default and handle "non-deterministic defaults" scenarios.

With Plugin Framework, the Framework likely itself currently performs default value application as part of PlanResourceChange (needs verified in test). Currently no attempt is made to populate or honor `__defaults` or do any additional default value application at Pulumi level. This might need revising.

Finally there is an additional layer of features when bridged Pulumi providers specify default value overrides, for example by populating [DefaultInfo](https://github.com/pulumi/pulumi-terraform-bridge/blob/6ffb2faf051e5faaf2234a0d0b926cba19c339ec/pkg/tfbridge/info.go#L322-#L322). This is currently not supported at all on top of Plugin Framework and might complicate upgrade scenarios for users wishing to move the bridged provider from the older bridge to the Plugin Framework.

t0yv0:
Deciding to not do anything further for default values here, logging rationale.

At a basic level defaults seem to be working on Plugin Framework. For example, this program uses RandomPet that has as default value for "separator" as "_", and it behaves similarly pre-PF and post-PF. Since this is a force-replace property, adding or removing the separator setting replaces the resource, and the default value of "-" is applied in both cases. 

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as r from "@pulumi/random";

const m = new r.RandomPet("monkey" /* , {separator: "_"} */);

export const separator = m.separator;
export const length = m.length;
```

This example illustrates how PF defines default values, which is through plan modifiers:

https://github.com/hashicorp/terraform-provider-random/blob/6c1912d96acae7fb7964c54e9b4452eb2f565722/internal/provider/resource_pet.go#L74

[...]

PR #882, opened 2023-03-09
Implement Read variant via ImportResourceState

### LINKED ISSUE #795: [Plugin Framework] Populate ReadResult.Inputs
## Hello!

## Issue details

When implementing Read, the Pulumi protocol makes a distinction between input and output properties, and the provider is supposed to implement Inputs population on the result. 

There is this function in the pre-PF bridge that does the work:

[extractInputsFromOutputs](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/schema.go#L1423-
#L1423)

It looks fairly involved so there maybe a bit of research/recovering test cases to understand precisely why this is happening.

t0yv0:
Fixes #795 

The motivation is to support Random resource import scenario:

```
pulumi import random:index/randomPassword:RandomPassword newPassword supersecret
```

With these changes I am able to do the above and get a program back. The program does successfully import the secret password:

```
new_password = random.RandomPassword("newPassword",
    length=11,
    lower=True,
    number=True,
    numeric=True,
    special=True,
    upper=True,
    opts=pulumi.ResourceOptions(protect=True))
```

But has a few extra parameters like lower, number, numeric, special, etc that are unnecessary, and only appear because of the difficulties we have with detecting default values, see #732.

The PR has to fix #795 because otherwise the generated program is this, which doesn't work:

```
new_password = random.RandomPassword("newPassword", length=0)
```

### DIFF
--- a/pf/internal/schemashim/attr_schema.go
+++ b/pf/internal/schemashim/attr_schema.go
@@ -75,7 +75,9 @@ func (s *attrSchema) ForceNew() bool {
 }
 
 func (*attrSchema) StateFunc() shim.SchemaStateFunc {
-	panic("StateFunc() should not be caleld during schema generation")
+	// StateFunc() should not be called by tfgen, but it currently may be called by ExtractInputsFromOutputs, so
+	// returning nil is better than a panic.
+	return nil
 }
 
 // Needs to return a shim.Schema, a shim.Resource, or nil.
--- a/pf/internal/schemashim/attr_schema.go
+++ b/pf/internal/schemashim/attr_schema.go
@@ -55,7 +55,9 @@ func (*attrSchema) DefaultFunc() shim.SchemaDefaultFunc {
 	panic("DefaultFunc() should not be called during schema generation")
 }
 func (*attrSchema) DefaultValue() (interface{}, error) {
-	panic("DefaultValue() should not be called during schema generation")
+	// DefaultValue() should not be called by tfgen, but it currently may be called by ExtractInputsFromOutputs, so
+	// returning nil is better than a panic.
+	return nil, nil
 }
 
 func (s *attrSchema) Description() string {
--- a/pf/internal/server/server.go
+++ b/pf/internal/server/server.go
@@ -1,0 +1,92 @@
+// Copyright 2016-2023, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"context"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
+	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
+)
+
+func NewProviderServer(provider plugin.Provider) pulumirpc.ResourceProviderServer {
+	return &providerServer{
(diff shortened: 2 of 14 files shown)


pkg/tfbridge/schema.go
```diff
@@ -1437,7 +1437,7 @@ func extractSchemaInputs(state resource.PropertyValue, tfs shim.Schema, ps *Sche
 	}
 }
 
-func extractInputsFromOutputs(oldInputs, outs resource.PropertyMap,
+func ExtractInputsFromOutputs(oldInputs, outs resource.PropertyMap,
```
t0yv0, line 1440: The implementation here dates back to #330 which appears incomplete with respect to end-user test cases and nailing down why the input extraction algorithm has the quirks it has. Because of this reimplementing for Plugin Framework seems like a substantial lift. So here I'm opting to re-export and reuse the existing implementation, and continue with it until we accrete more test cases or bug reports to refine.

It unfortunately creates some other inconsistencies, for example in how PF vs pre-PF bridge handle property naming, but so far appears OK in practice.

pf/tests/provider_read_test.go
```diff
@@ -122,32 +122,61 @@ func TestReadFromRefresh(t *testing.T) {
...
 	testutils.ReplaySequence(t, server, testCase)
 }
+
+func TestImportRandomPassword(t *testing.T) {
+	server := newProviderServer(t, testprovider.RandomProvider())
+	testCase := `
+	{
+	  "method": "/pulumirpc.ResourceProvider/Read",
+	  "request": {
+	    "id": "supersecret",
```
AaronFriel, line 149: Can we implement this without putting the password in the ID field?

pf/internal/server/server.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil, err
+	}
+
+	rpcInputs, err := plugin.MarshalProperties(result.Inputs, p.marshalOptions("newInputs"))
+	if err != nil {
+		return nil, err
+	}
+
+	return &pulumirpc.ReadResponse{
+		Id:         string(result.ID),
```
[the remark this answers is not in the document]
AaronFriel: Ah, I see - you've already looked at this.

pf/tests/provider_read_test.go
```diff
@@ -122,32 +122,61 @@ func TestReadFromRefresh(t *testing.T) {
...
+	    "urn": "urn:pulumi:v2::re::random:index/randomPassword:RandomPassword::newPassword",
+	    "properties": {}
+	  },
+	  "response": {
+	    "id": "none",
```
AaronFriel, line 154: Does this need to be a random value to have a unique URN?

t0yv0: I don't think so it doesn't participate in URN that's computed by `random:index/randomPassword:RandomPassword newPassword` parameters. Perhaps another q @Frassle can doublecheck with us. The "none" here is returned straight from Plugin Framework provider.

iwahbe (approved):
I have a question, but this looks good in general.

pf/internal/schemashim/attr_schema.go
```diff
@@ -55,7 +55,9 @@ func (*attrSchema) DefaultFunc() shim.SchemaDefaultFunc {
 	panic("DefaultFunc() should not be called during schema generation")
 }
 func (*attrSchema) DefaultValue() (interface{}, error) {
-	panic("DefaultValue() should not be called during schema generation")
+	// DefaultValue() should not be called by tfgen, but it currently may be called by ExtractInputsFromOutputs, so
+	// returning nil is better than a panic.
```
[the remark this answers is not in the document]
t0yv0: It's not OK but I don't think we can do anything about it, because we can't detect what's a default value, see #732 

Perhaps there's some workarounds we can think of to detect?

One specific downside of this is that import spits out all the extraneous fields like 'lower=True' that would have better been omitted:

```
new_password = random.RandomPassword("newPassword",
    length=11,
    lower=True,
    number=True,
    numeric=True,
    special=True,
    upper=True,
    opts=pulumi.ResourceOptions(protect=True))
```

pf/internal/server/server.go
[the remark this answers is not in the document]
t0yv0: Upstreamed to pulumi/pulumi; will fix up later here.

Issue #812, opened 2023-02-14
[Plugin Framework] Issues with output strings


t0yv0:
### What happened?

```
error: Preview failed: objectEncoder failed on property "url": Expected a string, got: {&{{[http://vm-4liajfa3.staging.azure.tetrate.com/}}}](http://vm-4liajfa3.staging.azure.tetrate.com/%7D%7D%7D)
```

```
pulumi.output(args.endpoint).apply(
                    (endpoint) =>
                        new checkmate.HttpHealth(
                            `${this.pulumiName}-check-endpoint`,
                            {
                                timeout: 1000 * 60 * 10, // 10 min, to allow for DNS to propagate (2 x TTL)
                                interval: 1000 * 2, // 2s
                                consecutiveSuccesses: 100,
                                url: endpoint,
                            },
                            { dependsOn: this.resources },
                        ).id,
                )
```

```
endpoint: pulumi.interpolate`vm-${vmPrefix}.${standaloneCluster.dnsDomain}
```

Courtesy of Rob Martens.

### Expected Behavior

Program succeeds

### Steps to reproduce

N/A

### Output of `pulumi about`

tf framework version v1.1.1, and pf bridge version [github.com/pulumi/pulumi-terraform-bridge/pkg/tfpfbridge](http://github.com/pulumi/pulumi-terraform-bridge/pkg/tfpfbridge) v0.0.0-20230120143314-6ffb2faf051e

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
Reproduces on pf/v0.2.1 also

jordanbinskin:
Can reproduce when trying to interpolate more than one value into the name string using `RandomString` from `@pulumi/random` (v4.11.1)

t0yv0:
@tmeckel you were kindly looking to reproduce this, is there any success? This would really help.

tmeckel:
@t0yv0 Sorry for the delay. I applied the change using `pulumi up -r`.

**Preview**

![image](https://user-images.githubusercontent.com/14177833/223848469-d706a7f8-b59a-4b01-86c8-4a408212aa3d.png)

**Output**

![image](https://user-images.githubusercontent.com/14177833/223848711-621ef0fd-6772-4535-9c1d-fc6df10541a8.png)

Issue is that everywhere in the output you'll find the illegal default value (-1). 

[...]

PR #883, opened 2023-03-10
Fix handling first-class secrets in Plugin Framework

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/812

### DIFF
--- a/pf/internal/convert/convert.go
+++ b/pf/internal/convert/convert.go
@@ -24,6 +24,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 
+	"github.com/pulumi/pulumi-terraform-bridge/pf/internal/propertyvalue"
 	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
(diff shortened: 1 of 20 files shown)


iwahbe (approved):
Looks good. LGTM!

pf/tfbridge/provider_create.go
```diff
@@ -34,6 +35,11 @@ func (p *provider) Create(
 ) (resource.ID, resource.PropertyMap, resource.Status, error) {
 	ctx := context.TODO()
 
+	// There is impedance mismatch for first-class secret values. These may be coming out of Pulumi programs but
```
t0yv0, line 38: Oops this bit is not needed, let me revert.

pf/internal/convert/secret.go
```diff
@@ -20,49 +20,20 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 )
 
-type secretEncoder struct {
-	elementEncoder Encoder
-	tfType         tftypes.Type
-}
-
+// Secrets only have a Decoder, not an Encoder because they are discarded when reading a resource.PropertyValue before
```
t0yv0, line 23: Drop encoder here, since it now removes secrets prior to encoding.

pf/internal/convert/convert.go
```diff
@@ -90,19 +91,19 @@ func NewConfigPropertyNames(pn PropertyNames) LocalPropertyNames {
 }
 
 type Encoder interface {
-	FromPropertyValue(resource.PropertyValue) (tftypes.Value, error)
+	fromPropertyValue(resource.PropertyValue) (tftypes.Value, error)
```
t0yv0, line 94: These are made internal since they're not secret-aware; to make sure the secret handling does not leak outside this module.

PR #887, opened 2023-03-13
Introduce InstanceStateStrategy CtyInstanceState flag to fix pulumi-aws#1423

t0yv0:
Fixes pulumi/pulumi-aws#1423

It appears that `func (s v2InstanceState) Object` is using customized github.com/pulumi/terraform-diff-reader/sdk-v2 code that is behind upstream TF source edits, and is causing pulumi-aws to misbehave. In particular when it reads and applies a diff like the following:

```
    rule.1860330423.name ResourceAttrDiff{New: "josh-test-aws-rules"}
    rule.1860330423.override_action.# ResourceAttrDiff{New: "1"}
    rule.1860330423.override_action.0.count.# ResourceAttrDiff{New: "1"}
    rule.1860330423.override_action.0.none.# ResourceAttrDiff{New: "0"}
```

it incorrectly decides to populate `count` with a `nil` value instead of the expected value of an empty object `{}`. This cause confusion. Switching to more higher-level API from TF SDKv2 resolves the problem. 

However, using this API seems to break some other tests in the repo and causes several observable differences. For example, Pulumi starts returning explicit `"foo": null` fields for objects (distinct from maps) when it previously elided them. It also causes some set element ordering issues.

Given that it is not a clean replacement, opting to be conservative and feature-flagging the change in a way that it can be activated for a single resource only.

I've tested  pulumi/pulumi-aws#1423 on top of this change plus a one-line configuration in `resources.go` and verified that the problem is resoled, moreover the resource responds correctly to further edits. Configuration line:

```go
	shimv2.SetInstanceStateStrategy(prov.P.ResourcesMap().Get("aws_wafv2_web_acl"), sdkv2.CtyInstanceState)
```

Previous attempt to fix 1423 by editing terraform-diff-reader/sdk-v2 did not work end to end and is reverted here.


t0yv0:
From tracing SetHash

```
diffToShim rule.283902798.override.0.count.# => ..
diffToShim rule.283902798.priority => ..
diffToShim rule.1219255360.name => ..
diffToShim rule.283902798.override.# => ..
diffToShim rule.1219255360.override.0.none.# => ..
diffToShim rule.1219255360.priority => ..
diffToShim rule.283902798.override.0.none.# => ..
diffToShim rule.283902798.name => ..
diffToShim rule.1219255360.override.# => ..
diffToShim rule.1219255360.override.0.count.# => ..
SetHash(map[name:josh-test-aws-rules override:[<nil>] priority:1]) => 487022137
SetHash(map[name:josh-test-aws-rules override:[map[count:[<nil>] none:[]]] priority:1]) => 283902798
SetHash(map[name:josh-test-aws-rules override:[<nil>] priority:1]) => 487022137
SetHash(map[name:josh-test-aws-rules override:[map[count:[<nil>] none:[]]] priority:1]) => 283902798
SetHash(map[name:josh-test-aws-rules override:[<nil>] priority:1]) => 487022137
```

t0yv0:
If I replace Set Hash function with a constant value, we get the desired diff None, so I think this is a bug about set element hashing, especially influenced by the override_action parameter in the schema, that affects the hash. This parameter is a little weird. A list of objects of emptySchema. I'm guessing there's a discrepancy in how TF vs Pulumi computes set hashes for these.

thomas11 (approved):
A test is a great start - any reason it's a draft PR?

t0yv0:
I thought maybe we can find a quick fix! But it's not so easy :)

t0yv0:
Debugging this a little further. @danielrbradley if you're interested.

This till reproduces. I think the bug is actually in Create method. We send this to create

```
          "overrideAction": {
            "__defaults": [],
            "count": {
              "__defaults": []
            }
          },
```

But it sends this back:

```
          "overrideAction": {
            "count": null,
            "none": null
          },
```

That's different. That's as if count property wasn't set.

My program is

```typescript

const args: aws.wafv2.WebAclArgs = {
    scope: "REGIONAL",
    rules: [
        {
            name: "rule-1",
            priority: 1,
            overrideAction: {
                count: {},
            },
            statement: {
                managedRuleGroupStatement: {
                    name: "AWSManagedRulesCommonRuleSet",
                    vendorName: "AWS",
                }
            },
            visibilityConfig: {
                cloudwatchMetricsEnabled: false,
                metricName: "friendly-rule-metric-name",
                sampledRequestsEnabled: false,
            },
        }
    ],
    defaultAction: {
        allow: {},
    },
    visibilityConfig: {
        cloudwatchMetricsEnabled: false,
        metricName: "friendly-rule-metric-name",
        sampledRequestsEnabled: false,
    },
};

const acl = new aws.wafv2.WebAcl("aclw", args);

export const arn = acl.arn;

```

pkg/tfshim/sdk-v2/cty.go
```diff
@@ -28,9 +29,9 @@ func makeResourceRawConfig(
 	config *terraform.ResourceConfig,
 	resource *schema.Resource,
 ) cty.Value {
```
t0yv0, line 31: Having to commit to the new method here, because "classic" parses sets incorrectly as tuples, causing type confusion in cty.Value representation.

pkg/tfshim/sdk-v2/instance_state.go
```diff
@@ -36,68 +37,41 @@ func (s v2InstanceState) ID() string {
 }
 
 func (s v2InstanceState) Object(sch shim.SchemaMap) (map[string]interface{}, error) {
```
t0yv0, line 39: This method wan't doing quite the right thing. The empty object value ended up parsed as `nil`. This caused problems because a `List[EmptyObj]` value under MaxItems=1 flattening got translated to `nil` which is indistinguishable from a `nil` list.

t0yv0: The new implementation leans into TF primitives. I wonder how to ramp up testing to know we can make this replacement.

t0yv0: The new implementation trips up some tests because it now generates `"foo": nil` entries in maps instead of eliding them.

t0yv0:
Opting to flag it so we can deploy the fix for the motivating issue, without taking the risk wholesale.

t0yv0:
https://gist.github.com/t0yv0/3c25627a43e5326ccfd176f7e481f02c temp investigation of TF expected behavior in a unit test.

pkg/tfshim/sdk-v2/instance_state.go
```diff
@@ -1,23 +1,46 @@
...
-	tf   *terraform.InstanceState
-	diff *terraform.InstanceDiff
+	resource *schema.Resource
+	tf       *terraform.InstanceState
+	diff     *terraform.InstanceDiff
 }
 
-func NewInstanceState(s *terraform.InstanceState) shim.InstanceState {
-	return v2InstanceState{s, nil}
+func NewInstanceState(resource *schema.Resource, s *terraform.InstanceState) shim.InstanceState {
```
iwahbe, line 38: I'm assuming we are not concerned about changing this technically public interface. I just wanted to call out that we are.

pkg/tfshim/sdk-v2/instance_state_read_strategy.go
```diff
@@ -0,0 +1,107 @@
...
+
+func SetInstanceStateStrategy(resource shim.Resource, strategy InstanceStateStrategy) {
+	v2Res, ok := resource.(v2Resource)
+	contract.Assertf(ok, "expecting resource to be shimmed with sdkv2 shim, got %v", reflect.TypeOf(resource))
+	instanceStateStrategyByResource.Store(v2Res, strategy)
+}
+
+func GetInstanceStateStrategy(resource shim.Resource) InstanceStateStrategy {
+	if resource == nil {
+		return ClassicInstanceState
```
iwahbe, line 50: I'm confused as to the difference between `ClassicInstanceState` and `defaultInstanceStateStrategy`. If `resource == nil`, shouldn't we return `defaultInstanceStateStrategy`.

pkg/tfshim/sdk-v2/instance_state_read_strategy.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	v2Res, ok := resource.(v2Resource)
+	contract.Assertf(ok, "expecting resource to be shimmed with sdkv2 shim, got %v", reflect.TypeOf(resource))
+	v, ok := instanceStateStrategyByResource.Load(v2Res)
+	if !ok {
+		return defaultInstanceStateStrategy
+	}
+	r, ok := v.(InstanceStateStrategy)
```
iwahbe, line 58: Can this cast ever fail?

pkg/tfshim/sdk-v2/instance_state_read_strategy.go
```diff
@@ -0,0 +1,107 @@
...
+
+// Feature flag to configures how the provider reads instance state.
+type InstanceStateStrategy int
+
+const (
+	// The legacy strategy.
+	ClassicInstanceState InstanceStateStrategy = 0
+
+	// The strategy based on transcoding to cty.Value.
+	CtyInstanceState InstanceStateStrategy = 1
```
iwahbe, line 73: Why not `iota`?

pkg/tfshim/sdk-v2/instance_state_read_strategy.go
```diff
@@ -0,0 +1,107 @@
...
+
+const instanteStateStrategyEnvVar = "PULUMI_INSTANCE_STATE_STRATEGY"
+
+func ParseInstanceStateStrategyFromEnv() (InstanceStateStrategy, bool, error) {
+	s := os.Getenv(instanteStateStrategyEnvVar)
+	if s == "" {
+		return ClassicInstanceState, false, nil
+	}
+	p, err := ParseInstanceStateStrategy(s)
+	return p, err == nil, err
```
iwahbe, line 106: It feels weird to have all 3 values returned here. Why is `(InstanceStateStrategy, error)` not enough?

pkg/tfshim/sdk-v2/instance_state.go
t0yv0: Fixed.

pkg/tfshim/sdk-v2/instance_state_read_strategy.go
t0yv0: Not really, overly defensive coding.

t0yv0:
@iwahbe addressed the comments, ready for another look.

pkg/tests/regress_aws_1423_test.go
```diff
@@ -107,6 +107,8 @@ func TestRegressAws1423(t *testing.T) {
 		[]byte{}, /* pulumiSchema */
 	)
 
+	shimv2.SetInstanceStateStrategy(p.ResourcesMap().Get("aws_wafv2_web_acl"), shimv2.CtyInstanceState)
```
AaronFriel, line 110: Do we need to regenerate the test case, or should we not expect a diff below this line?

t0yv0: The test case does discriminate the functionality. It fails the test once the changes to undo the previous non-fix made to pulumi-terraform-diff-reader are made. It's incomplete in a sense that it couldn't discriminate the pulumi-terraform-diff-reader non-fix from a real fix.

AaronFriel: I see

AaronFriel (approved):

Issue #877, opened 2023-03-08
write test for edge case examples generation


jazzyfresh:
test this PR that was merged last night to fix the docs https://github.com/pulumi/pulumi-terraform-bridge/pull/875

jazzyfresh:
Not sure how to test docs generation in the schema cc @t0yv0 @guineveresaenger for prior art

t0yv0:
Fixed by https://github.com/pulumi/pulumi-terraform-bridge/pull/888

PR #888, opened 2023-03-13
Add a  test for example generation

t0yv0:
For #877 

Example rendering seems pretty coupled with the rest of tfgen. It doesn't have a testable exported API. Perhaps it should. In the meanwhile there's a way to test example rendering through an entire schema generation test.

### DIFF
--- a/pkg/tfgen/docs_test.go
+++ b/pkg/tfgen/docs_test.go
@@ -17,13 +17,20 @@ package tfgen
 
 import (
 	"bytes"
+	"io"
 	"strings"
 	"testing"
 	"text/template"
 
-	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
-	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen/internal/testprovider"
 )
 
 type testcase struct {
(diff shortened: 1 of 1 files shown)


pkg/tfgen/docs_test.go
```diff
@@ -846,6 +853,51 @@ func TestFindRepoPath(t *testing.T) {
...
+func TestExampleGeneration(t *testing.T) {
+	info := testprovider.ProviderMiniRandom()
+
+	markdown := []byte(`
+## Examples
+
+There is some more code in here.
+
+~~~java
+throw new Exception("!");
```
t0yv0, line 865: This is our new code that should go through example rendering.

pkg/tfgen/docs_test.go
```diff
@@ -846,6 +853,51 @@ func TestFindRepoPath(t *testing.T) {
...
+	}
+
+	inmem := afero.NewMemMapFs()
+
+	g, err := NewGenerator(GeneratorOptions{
+		Package:      info.Name,
+		Version:      info.Version,
+		Language:     Schema,
+		ProviderInfo: info,
+		Root:         inmem,
```
t0yv0, line 882: Using afero will write "pseudo files" in-memory instead of spilling to disk. What we expect is schema.json file is produced.

pkg/tfgen/docs_test.go
```diff
@@ -846,6 +853,51 @@ func TestFindRepoPath(t *testing.T) {
...
+	err = g.Generate()
+	assert.NoError(t, err)
+
+	f, err := inmem.Open("schema.json")
+	assert.NoError(t, err)
+
+	schemaBytes, err := io.ReadAll(f)
+	assert.NoError(t, err)
+
+	assert.NotContains(t, string(schemaBytes), "{{% //examples %}}")
```
t0yv0, line 898: Deep in the schema.json file we now have the rendered doc for the random_integer resource to assert against.

jazzyfresh (approved):

jazzyfresh:
@t0yv0 I approved but i was never able to figure out the failing lint step here https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/4410350502/jobs/7764819576?pr=888

also that actions page sometimes becomes unresponsive 😬

t0yv0:
> but i was never able to figure out the failing lint step 

Oh don't do that just let me know it's broken, I'll take care of it.

jazzyfresh:
It seemed like something weird with the action runner. I re-ran the test suite.

PR #890, opened 2023-03-14
Automatic Alias Removal on Major Version Bump

### LINKED ISSUE #889: Automatic Alias Removal on Major Version Bump
## Hello!

### Problem

Aliases are used when a Pulumi provider moves resources/functions. This allows the underlying resource/function to have a new and preferred name. The bridge makes "hard" aliases, generating both the new and old name as its own resource/function in the schema. Right now, aliases are effectively permanent, since there is no process to remove them. 

### Solution
When a provider is built, it understands what major version it was built with. We can "teach" `AutoAliasing` to drop "hard" aliases on major version updates, the only time when we are allowed to remove code. "soft" token aliases should remain, to help users migrate from old to new aliases. By keeping the process automatic, we remove the need for manual supervision. 

### Implementation Details
For each resource/function, `AutoAliasing` persists the set of previous Pulumi tokens in a [tokenHistory](https://github.com/pulumi/pulumi-terraform-bridge/blob/660f36d0519ab8f4a88424c7dd573f8fb40db5a6/pkg/tfbridge/x/tokens.go#L462-L465): the current token and a set of [alias](https://github.com/pulumi/pulumi-terraform-bridge/blob/660f36d0519ab8f4a88424c7dd573f8fb40db5a6/pkg/tfbridge/x/tokens.go#L467-L470)s. We will augment `alias` with the provider version used when it was created. When `make tfgen` is run with a new provider version, aliases from previous versions will be dropped from generated code (maintaining "soft" aliases).

### Affected area/feature

https://github.com/pulumi/pulumi-terraform-bridge/blob/660f36d0519ab8f4a88424c7dd573f8fb40db5a6/pkg/tfbridge/x/tokens.go#L477

aq17:
Fixes #889 

When a major provider upgrade is performed, old aliases can be removed since breaking changes are allowed.

### DIFF
--- a/pkg/tfbridge/x/tokens.go
+++ b/pkg/tfbridge/x/tokens.go
@@ -507,7 +507,6 @@ func AutoAliasing(providerInfo *b.ProviderInfo, artifact b.ProviderMetadata) err
 				}
 			}
 		}
-		providerInfo.Version = fmt.Sprintf("%d.0.0", currentVersion)
 	}
 
 	for tfToken, computed := range providerInfo.Resources {
--- a/pkg/tfbridge/x/tokens.go
+++ b/pkg/tfbridge/x/tokens.go
@@ -484,13 +484,30 @@ func AutoAliasing(providerInfo *b.ProviderInfo, artifact b.ProviderMetadata) err
 		return err
 	}
 
-	currentVersion := -1
+	var currentVersion int
+	// If version is missing, we assume the most recent major version in history
 	if providerInfo.Version != "" {
 		v, err := semver.NewVersion(providerInfo.Version)
 		if err != nil {
 			return err
 		}
 		currentVersion = int(v.Major())
+	} else {
+		for _, r := range hist.Resources {
+			for _, p := range r.Past {
+				if p.MajorVersion > currentVersion {
+					currentVersion = p.MajorVersion
+				}
+			}
+		}
+		for _, d := range hist.DataSources {
+			for _, p := range d.Past {
+				if p.MajorVersion > currentVersion {
+					currentVersion = p.MajorVersion
+				}
+			}
+		}
+		providerInfo.Version = fmt.Sprintf("%d.0.0", currentVersion)
 	}
 
 	for tfToken, computed := range providerInfo.Resources {
@@ -538,14 +555,6 @@ func aliasResource(
 	version int,
 ) {
 	prev, hasPrev := hist[tfToken]
-	// If version is -1, we assume the most recent major version
-	if version == -1 {
-		for _, p := range prev.Past {
-			if p.MajorVersion > version {
-				version = p.MajorVersion
-			}
-		}
-	}
 	if !hasPrev {
 		// It's not in the history, so it must be new. Stick it in the history for
 		// next time.
@@ -612,13 +621,6 @@ func aliasDataSource(
 	version int,
 ) {
 	prev, hasPrev := hist[tfToken]
-	if version == -1 {
-		for _, p := range prev.Past {
-			if p.MajorVersion > version {
-				version = p.MajorVersion
-			}
-		}
-	}
 	if !hasPrev {
 		// It's not in the history, so it must be new. Stick it in the history for
 		// next time.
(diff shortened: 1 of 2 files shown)


iwahbe (commented):
Looking good!

I know this is marked as a draft, so If you were already planning to do this, just ignore it:
- We need to apply this to data sources as well.
- There are a bunch of tests for `AutoAlias` already. We will need to expand test coverage to include this feature.

pkg/tfbridge/x/tokens.go
```diff
@@ -565,11 +574,12 @@ func aliasOrRenameResource(p *b.ProviderInfo, tfToken string, hist *tokenHistory
 		hist.Past = append(hist.Past, alias[tokens.Type]{
 			Name:      hist.Current,
 			InCodegen: true,
+			Version:   currentVersion,
 		})
 	}
 	for _, a := range hist.Past {
 		legacy := a.Name
-		if a.InCodegen {
+		if a.InCodegen && (a.Version == currentVersion || a.Version == 0) {
```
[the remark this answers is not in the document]
aq17: Got it, I wasn't sure about that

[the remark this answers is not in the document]
iwahbe: As a though experiment, if we were concerned about backwards compatibility:

We still don't want to keep names from version 0 forever. One solution would be to record the version we will remove a name at, instead of the version when the name was added. Since `0` is never a valid version to remove a name, we can safely update `0` to one past the current major version. We would remain backwards compatible without a long term compromise in functionality.

pkg/tfbridge/x/tokens.go
```diff
@@ -563,13 +576,17 @@ func aliasOrRenameResource(p *b.ProviderInfo, tfToken string, hist *tokenHistory
...
+			InCodegen:    true,
+			MajorVersion: currentVersion,
 		})
 	}
 	for _, a := range hist.Past {
 		legacy := a.Name
-		if a.InCodegen {
+		// A version set as "-1" means the provider did not have a version set
+		// so we refrain from removing aliases
+		fmt.Println(legacy, a.MajorVersion, currentVersion)
```
iwahbe, line 588: We never remove old aliases from the history, which means that aliases multiple major versions back might exist. I think we can conservatively assume that the most recent major version referenced in the history is the most recent major version. This way unsetting the version won't introduce multiple generations of aliases into a build.

I appreciate the comment explaining what's going on!

iwahbe (commented):
If we're going to try to accommodate a version less provider, we should have a test that we are doing this correctly.

pkg/tfbridge/x/tokens.go
aq17: This makes more sense. In our tests previously, we didn't have a version set on the mock providers since it isn't a required attribute, so I wasn't sure if we needed to check for that.
I think it'd be better to default to the most recent version referenced if one doesn't exist.

iwahbe (commented):
It looks like tests are panicking. The fix I mentioned in my comment will fix that. Then we can merge.

pkg/tfbridge/x/tokens.go
```diff
@@ -482,12 +484,30 @@ func AutoAliasing(providerInfo *b.ProviderInfo, artifact b.ProviderMetadata) err
...
+	var currentVersion int
+	// If version is missing, we assume the most recent major version in history
+	if providerInfo.Version != "" {
+		v, err := semver.NewVersion(providerInfo.Version)
+		if err != nil {
+			return err
+		}
+		currentVersion = int(v.Major())
+	} else {
+		for _, r := range hist.Resources {
```
iwahbe, line 496: We also keep track of function versions

pkg/tfbridge/x/tokens.go
```diff
@@ -484,13 +484,30 @@ func AutoAliasing(providerInfo *b.ProviderInfo, artifact b.ProviderMetadata) err
...
+			}
+		}
+		for _, d := range hist.DataSources {
+			for _, p := range d.Past {
+				if p.MajorVersion > currentVersion {
+					currentVersion = p.MajorVersion
+				}
+			}
+		}
+		providerInfo.Version = fmt.Sprintf("%d.0.0", currentVersion)
```
aq17, line 510: Not sure if we want to default to this, or leave empty.

iwahbe: I would say leave it empty. We can't guarantee that the version we have computed is correct. Potentially false information is much more confusing then no information here.

iwahbe (approved):
Let's delete 
```L510: 		providerInfo.Version = fmt.Sprintf("%d.0.0", currentVersion)```
Beyond that, LGTM!

Issue #896, opened 2023-02-02
Docsgen/Programgen: Improper Capitalization


julien-skillcorner:
There is currently a bug in python doc generation where the inner _ is not escaped, see: https://github.com/pulumi/pulumi-aws/pull/2347#discussion_r1095157137

For an example, check out this error: There is a typo in Python example using lambda attribute instead of lambda_

[themes/default/content/registry/packages/aws/api-docs/lambda/getfunction/_index.md](https://www.pulumi.com/registry/packages/aws/api-docs/lambda/getfunction/)

Frassle:
Testing the raw python schemagen and programgen I can't reproduce these issues so it looks like this is specific to the fixups tfgen does to schemas per-language.

PR #897, opened 2023-03-16
Don't pass open/close quotes to PyName

Frassle:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/896

https://github.com/pulumi/pulumi-terraform-bridge/pull/734 removed `PyName` from tfbridge and replaced uses of it with a call to `PyName` in "github.com/pulumi/pulumi/pkg/v3/codegen/python".
Unfortunately the behavior of the two functions was slightly different for characters that were not valid identifier characters. That coupled with the fact that `open` and `close` characters were being passed into PyName in `fixupPropertyReferences` meant that when this changed we started outputing docs with text like `"This data source is different from the _directconnect_get_locations_ data source"` rather than ``"This data source is different from the `directconnect_get_locations` data source"``.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1690,7 +1690,7 @@ func fixupPropertyReferences(language Language, pkg tokens.Package, info tfbridg
 				return open + modname + getname.String() + close
 			case Python:
 				// Use `ec2.get_ami` format
-				return python.PyName(open + modname + getname.String() + close)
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
Thanks for tracking this down! LGTM!

There should really be tests for these, but I don't see any and it looks like adding a comprehensive test set for this is non-trivial. I'll open an issue for it.

PR #900, opened 2023-03-21
Try setup-go@v4

t0yv0:
Fixes breakage of golangci-lint, optimizes builds a bit with more aggressive caching.


aq17 (approved):

iwahbe (approved):
Does `skip-cache` skip using the cache, or caching? Or will `setup-go@v4` do all that for us?

Lets try it

t0yv0:
setup-go@v4 does all the caching. The setup-lint action was stalling for 8 min as it did something dubious with the cache. This works better for now.

PR #902, opened 2023-03-21
Update golangci-lint-action to v4

Frassle:
v3 doesn't seem to be quite working anymore (see https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/4480726203/jobs/7876429264?pr=677)


t0yv0:
Also fixed by https://github.com/pulumi/pulumi-terraform-bridge/pull/900

Frassle:
Ah fair enough :) even better

PR #903, opened 2023-03-21
Fix pulumi dependency update

t0yv0:
Due to an issue with an auto-upgrade script, dependencies in top-level go.mod did not yet get updated. And once they did, there was a lot more lint work to do to make the linter pass.

### DIFF
--- a/scripts/build.go
+++ b/scripts/build.go
@@ -98,7 +98,7 @@ func execCommandOrLogFatal(cwd, name string, arg ...string) {
 // Finds directories containing go.mod files in the repository.
 func findGoModuleRoots() (result []string) {
 	var buf bytes.Buffer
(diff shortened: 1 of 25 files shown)


scripts/build.go
```diff
@@ -98,7 +98,7 @@ func execCommandOrLogFatal(cwd, name string, arg ...string) {
 // Finds directories containing go.mod files in the repository.
 func findGoModuleRoots() (result []string) {
 	var buf bytes.Buffer
-	cmd := exec.Command("git", "ls-files", "-z", "**/go.mod")
+	cmd := exec.Command("git", "ls-files", "-z", "**go.mod")
```
t0yv0, line 101: This was the bug.. Hrm... `**/go.mod` does not match `./go.mod` in the cwd.

aq17 (approved):

PR #912, opened 2023-03-23
WIP :: added formatPath function to include the attribute in question in an error

phillipedwards:


t0yv0:
Thanks @phillipedwards this is a super impactful usability issue, and your PR fixes it - I finally started looking into this. I'm suggesting a slight variation in https://github.com/pulumi/pulumi-terraform-bridge/pull/1153 - basically I've discovered some existing machinery that translates TF property paths to Pulumi property paths and was able to reuse it for the purpose of formatting the validation errors for provider config as well; that'd be slightly more accurate for explicit providers as it will be in terms of Pulumi names.

PR #919, opened 2023-03-28
Support set type

### LINKED ISSUE #731: [Plugin Framework] Support Set type
## Hello!

## Issue details

Terraform Plugin Framework allows plugins to define properties as having a Set type. Bridged Pulumi providers should support such properties.

iwahbe:
Fixes #731


pf/internal/convert/set.go
```diff
@@ -0,0 +1,95 @@
...
+
+	if propertyValueIsUnkonwn(p) {
+		return tftypes.NewValue(setTy, tftypes.UnknownValue), nil
+	}
+	if p.IsNull() {
+		return tftypes.NewValue(setTy, nil), nil
+	}
+	if !p.IsArray() {
+		return tftypes.NewValue(setTy, nil),
+			fmt.Errorf("Expected an Array PropertyValue")
```
t0yv0, line 58: Mark the message saying we're in a set encoder, might help.

pf/internal/convert/encoding.go
```diff
@@ -267,15 +267,22 @@ func (e *encoding) deriveEncoder(typeSpec *pschema.TypeSpec, t tftypes.Type) (En
...
+				return nil, err
+			}
+			return newListEncoder(t.ElementType, elementEncoder)
+		case tftypes.Set:
+			elementEncoder, err := e.deriveEncoder(typeSpec.Items, t.ElementType)
+			if err != nil {
+				return nil, err
+			}
+			return newSetEncoder(t.ElementType, elementEncoder)
+		default:
```
t0yv0, line 283: fmt.Erorrf should now say List or Set.

pf/internal/schemashim/type_schema.go
```diff
@@ -68,11 +68,14 @@ func (s *typeSchema) Elem() interface{} {
 		contract.Assertf(s.nested == nil || len(s.nested) == 0,
 			"s.t==MapType should not have any s.nested attrs")
 		return newTypeSchema(tt.ElemType, nil)
+	case types.SetType:
+		contract.Assertf(s.nested == nil || len(s.nested) == 0,
+			"s.t==SetType should not have any s.nested attrs")
```
t0yv0, line 73: I think that's right, but I'm forgetting this a bit already.

pf/tests/testdata/basicprogram/Pulumi.yaml
```diff
@@ -94,6 +94,22 @@ resources:
...
+      setOptionals:
+        - "hi"
+        - "bye"
+  setElementRevAttrRes:
+    type: testbridge:index:Testres
+    properties:
+      statedir: state
+      requiredInputString: "x"
+      setOptionals:
+        - "bye"
```
t0yv0, line 111: Hm it preserves order?

pf/tfbridge/ignore_changes.go
```diff
@@ -265,35 +265,30 @@ func (*matchingContext) fail() *matchingContext {
...
-	if p, ok := pattern.(string); ok {
-		if p == "*" {
-			return true
-		}
-		if p == string(puName) {
-			return true
-		}
-	}
-	return false
+func (mc *matchingContext) pulumiNameMatches(pattern interface{}, puName resource.PropertyKey) bool {
```
t0yv0, line 268: Maybe drop ignore_changes from this PR and instead let's go for fixing https://github.com/pulumi/pulumi-terraform-bridge/issues/920 using Pulumi-level approach you outlined, that's going to be oblivious to sets anyway.

pf/internal/convert/convert_test.go
```diff
@@ -121,6 +120,34 @@ func TestConvertTurnaround(t *testing.T) {
...
+		}
+	}
+
+	cases = append(cases,
+		setCase("empty", tftypes.String, resource.NewArrayProperty([]resource.PropertyValue{})),
+		setCase("string-alpha", tftypes.String, resource.NewArrayProperty([]resource.PropertyValue{
+			resource.NewStringProperty("hello"),
+			resource.NewStringProperty("world"),
+		}),
+			"hello",
```
t0yv0, line 141: I'd be really curious to have these tested:

- what happens if Pulumi passes duplicate items?
- what if some of the items inside are Null, or Unknown, how does that fare? duplicative?
- this seems to indicate order preservation turnaround (interesting), does this property relax to something like order-preserving deduplication?
- TF can't pass duplicate items back to Pulumi right?

t0yv0:
I'd try one set with MaxItems=1 also at integration level. It might just work but we might need to remember how we did that for lists. It's worth checking out as it actually shows up a fair bit. 

Perhaps also try insertion and deletion of new elements via throwing some of set-powered resources in TestUpdateProgram? This uses ProgramTest steps. 

Other than that LGTM, let's try this out in real providers, excited!

pf/internal/convert/convert_test.go
iwahbe: > I'd be really curious to have these tested:
> 
> * what happens if Pulumi passes duplicate items?

The TFPF framework will error. 

> * what if some of the items inside are Null, or Unknown, how does that fare? duplicative?


> * this seems to indicate order preservation turnaround (interesting), does this property relax to something like order-preserving deduplication?
> * TF can't pass duplicate items back to Pulumi right?

t0yv0: Interesting. Well if it's an error that's a bit unfortunate, we might need to backlog a bug with an example of how this error presents to the user. I'm guessing it's going to start with something not user friendly.

pf/tests/testdata/basicprogram/Pulumi.yaml
iwahbe: It seems like it. Sets are implemented as `[]tftypes.Value`, so we seem to get array-like properties. This would all be very different if sets were implemented as `map[ValueHash]Value`.

t0yv0: Huh that's good to know. Interesting. I think in TF making data as a set mostly affects how they go through diffing operations, sounds like the repr isn't helpful to suss out how this works.

pf/tests/testdata/updateprogram-3/Pulumi.yaml
```diff
@@ -0,0 +1,31 @@
...
+  testRequiredStringAfterUpdate__actual: ${r.requiredInputStringCopy}
+  testRequiredStringAfterUpdate__expect: |-
+    input1-updated
+
+  testOptionalStringAfterUpdate__actual: ${r.optionalInputStringCopy}
+  testOptionalStringAfterUpdate__expect: |-
+    input2-updated
+
+  testSetOptionals__actual: ${set.setOptionals}
+  testSetOptionals__expect: [ a, b, c, b ]
```
t0yv0, line 31: This is interesting so it preserved the duplicate entry `b` through the turnaround?

pf/internal/pfutils/attr.go
```diff
@@ -133,7 +133,7 @@ func LookupTerraformPath(schema Schema, path *tftypes.AttributePath) (LookupResu
 	if parent := path.WithoutLastStep(); parent != nil {
 		parentRes, parentOk, parentErr := tryLookupAttrOrBlock(schema, parent)
 		if parentErr == nil && parentOk {
-			if parentRes.IsAttr && parentRes.Attr.HasNestedObject() {
+			if parentRes.IsAttr {
```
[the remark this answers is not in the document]
t0yv0: This should be unnecessary once you rebase on master.

PR #925, opened 2023-03-29
Apply ignore to updates

### LINKED ISSUE #920: [Plugin Framework] IgnoreChanges only applies to Diff
### What happened?

I noticed that `ignoreChanges` only applies for diffs. Once an `update` is desired, `ignoreChanges` no longer takes effect.

### Expected Behavior

We should see the following

Initial:
```ts
import * as random from "@pulumi/random";

const a = new random.RandomInteger(
    "ri",
    {min: 1, max: 10},
    {ignoreChanges: ["min"]}
)
```
With state `{min: 1, max: 10}`

---

Transition 1:
```ts
import * as random from "@pulumi/random";

const a = new random.RandomInteger(
    "ri",
    {min: 2, max: 10},
    {ignoreChanges: ["min"]}
)
```
With state: `{min: 1, max: 10}` (no change)

---

Transition 2:
```ts
import * as random from "@pulumi/random";

const a = new random.RandomInteger(
    "ri",
    {min: 2, max: 20},
    {ignoreChanges: ["min"]}
)
```
With state: `{min: 1, max: 20}`

### Steps to reproduce

What we actually get is a final state (see Expected Behavior) of `{min: 2, max: 20}`. This is wrong. 

### Output of `pulumi about`

Bridge version v3.42.1 (pf version [v0.6.0](https://github.com/pulumi/pulumi-terraform-bridge/releases/tag/pf%2Fv0.6.0))

### Additional context

_No response_

### Contributing

iwahbe:
Fixes #920


iwahbe:
@t0yv0 right now, this change is failing `TestTlsResourceUpgrade`. The new version of ignoreChanges is behaving correctly for that test case though.

The documentation for ignoreChanges says:
> Ignored properties will still be used from the program when there is no previous value in the state, most importantly when creating the resource.

The ignored properties are not present in `olds` ([here](https://github.com/pulumi/pulumi-terraform-bridge/blob/228b97f2c51a58bf3077019d9c1c5e7690f66749/pf/tests/provider_diff_test.go#L214-L239)), so they are correctly ignored.

I know that this test exists to ensure there is a safe migration path from tls v3 to v4, so I'm hesitant to change the output so it passes (even though that is the correct behavior in the abstract). What would you recommend.

t0yv0:
Hmm.

I think my intuition is that the test is valid and should continue passing, and if we did not have the documentation site then I would assume ignoreChanges lets user change a plan with changes into a plan without changes; a special case of that is when a plan introduces a property that wasn't there before.

We can also consult prior art specifically doIgnoreChanges implementation in the production version of the bridge:

https://github.com/pulumi/pulumi-terraform-bridge/blob/d92cf2d9336d7632ab31823dfb561bed653e27e4/pkg/tfbridge/diff.go#L263

Clearly this implementation is looking at both olds and news to build ignoreKeySet and then knocking it off the diff structure. So if we change the semantic if PF version this might cause unwanted changes in behavior to users.

I was hesitant to reuse the doIgnoreChanges implementation because visitPropertyValue calls MakeTerraformInput input for the set case which looked highly surprising but we can revisit that code.

pf/tfbridge/provider_diff.go
```diff
@@ -46,6 +46,11 @@ func (p *provider) DiffWithContext(
 ) (plugin.DiffResult, error) {
 	ctx = p.initLogging(ctx, p.logSink, urn)
 
+	checkedInputs, err := applyIgnoreChanges(priorStateMap, checkedInputs, ignoreChanges)
```
t0yv0, line 49: This is very different than doIgnoreChanges in the pre-PF bridge that looks at news and olds post Diff, post planning the resource change (which also applies defaults and such erm).. I think we can construct test cases where this will behave a bit differently from the current bridged provider behavior, and I think the pressure will be to match existing behavior as we're trying to make as little disruptive changes as possible 😓

t0yv0: Could not construct a test case here that would be relevant. One possibility if state upgrade is migrating the format (or type) of a property P that's being ignored. In that case copying from state to news will bypass the migration on the property and news will have a bad format, which will bork TF processing.  Sounds sufficiently niche to wait on that until such time as we need to deal with it.

pf/tfbridge/ignore_changes_test.go
```diff
@@ -17,205 +17,132 @@ package tfbridge
...
-type testRenames struct{}
+			prev, _ := path.Get(resource.NewObjectProperty(old))
+			value, _ := path.Get(resource.NewObjectProperty(new))
 
-func (*testRenames) PropertyKey(_ tokens.Token,
-	property convert.TerraformPropertyName, _ tftypes.Type) resource.PropertyKey {
-	return resource.PropertyKey(tfbridge.TerraformToPulumiNameV2(property, nil, nil))
-}
+			if c.expectChange {
+				assert.NotEqual(t, prev, value)
```
t0yv0, line 141: Is assert.Equal same as prev.DeepEqual(value)?

pf/tfbridge/ignore_changes.go
```diff
@@ -15,363 +15,153 @@
 package tfbridge
 
 import (
-	"strings"
+	"fmt"
 
-	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/hashicorp/go-multierror"
```
t0yv0, line 20: Should we be using "errors" now?

t0yv0:
The correctness of '*' broke my brain a bit but if it's passing the test cases that aren't substantially modified there is some reassurance here. 

The tests seem a bit weak spec for '*' especially since they just check values at path.Get. 

I'm curious if paths with '*' will copy entries into news even if not in news?

```
func TestIgnoreChangesCopiesEntries(t *testing.T) {
	olds := resource.PropertyMap{"k": resource.NewObjectProperty(resource.PropertyMap{
		"a": resource.NewStringProperty("A"),
		"b": resource.NewStringProperty("B"),
	})}
	news := resource.PropertyMap{"k": resource.NewObjectProperty(resource.PropertyMap{
		"a": resource.NewStringProperty("A"),
	})}
	news2, err := applyIgnoreChanges(olds, news, []string{"k[*]"})
	assert.NoError(t, err)
	assert.Equal(t, olds, news2)
}

```

t0yv0:
And definitely want this one to pass as we discussed. Thank you! 

```go
func TestIgnoreChangesRemovesEntries(t *testing.T) {
	olds := resource.PropertyMap{"k": resource.NewObjectProperty(resource.PropertyMap{})}
	news := resource.PropertyMap{"k": resource.NewObjectProperty(resource.PropertyMap{
		"a": resource.NewStringProperty("A"),
	})}
	news2, err := applyIgnoreChanges(olds, news, []string{"k.a"})
	assert.NoError(t, err)
	assert.Equal(t, olds, news2)
}

```

pf/tfbridge/ignore_changes_test.go
iwahbe: It has better error messages, but it's a deep equals if that's what you're asking. It is at least as semantically rigorous as `prev.DeepEquals`.

pf/tfbridge/ignore_changes.go
iwahbe: Probably. 😄

iwahbe:
> Thinking out loud how we could write this. Perhaps something like this could be easier to reason about, except order can matter semantically perhaps which is annoying, and p.Get/p.Set/p.Delete has to be rock-solid in introducing additional ObjectProperty as needed etc.
> 
> ```go
> type PropertyPathPattern // the one with wildcards
> type PropertyPath // concrete one without wildcards
> 
> func globPropertyPath(PropertyPathPattern, PropertyValue) Set[PropertyPath]
> 
> func ignoreChanges(olds PropertyValue, news PropertyValue, ignored PropertyPathPattern) PropertyValue {
>    out := news.Copy()
>    oldPaths := globPropertyPath(ignored, olds)
>    newPaths := globPropertyPath(ignored, news)
>    for _, p := range oldPaths.Union(newPaths).ToSlice() {
>       old, gotOld := p.Get(olds)
>       new, gotNew := p.Get(news)
>       switch {
>       case gotNew && gotOld && !old.DeepEqual(new), !gotNew && gotOld:
>           p.Set(old, out)
>       case !gotNew && gotOld:
>           p.Delete(out)
>       }
>    }
>    return out
> }
> ```

I started with something very similar to this. `p.Set` doesn't introduce additional object properties at all. `p.Set` is implemented in terms of `[0:len(p-1)].Get`. The major problem for an approach based on `p.Set`, `p.Get` and `p.Delete` is that none of them handle `"*"` at all. To handle `"*"` correctly, you need to walk the whole path manually.

iwahbe:
I have added the tests you included, as well as a test for nested glob usage.

t0yv0:
Good. Thank you for fixing the application of ignoreChanges to updates!

A bit of code golf.

> The major problem for an approach based on p.Set, p.Get and p.Delete is that none of them handle "*" at all. 

I was thinking glob handling can be factored out into a glob expansion pass.

> Fully expanding the the globs requires the full walk anyway, so I'm not sure what the benefit would be

Well there is a benefit in breaking things down into algebraically simple parts: easier to reason about (faster to review for a slow-thinking clown like me), easier to compose, no space for surprising corner cases; and when you have a library of these you find yourself reaching out for existing well-tested bits to reuse.

Fusing the walks is about performance which is not the dominating factor here. 

That being said I can no longer find empirically significant problems with this algo as written so let's check it in! 

I found this one fails:

```go
func TestIgnoreChangesInsertsNested(t *testing.T) {
	olds := resource.PropertyMap{
		"k": resource.NewObjectProperty(resource.PropertyMap{
			"n1": resource.NewObjectProperty(resource.PropertyMap{
				"n2": resource.NewStringProperty("!"),
			}),
		}),
	}
	news := resource.PropertyMap{}
	news2, err := applyIgnoreChanges(olds, news, []string{"k.*"})
	assert.NoError(t, err)
	assert.Equal(t, olds, news2)
}
```

But if you switch the path to "k" it works out, so that's great. That should be fine and there's some logic to it.

t0yv0 (approved):

iwahbe:
> Good. Thank you for fixing the application of ignoreChanges to updates!
> 
> A bit of code golf.
> 
> > The major problem for an approach based on p.Set, p.Get and p.Delete is that none of them handle "*" at all.
> 
> I was thinking glob handling can be factored out into a glob expansion pass.
> 
> > Fully expanding the the globs requires the full walk anyway, so I'm not sure what the benefit would be
> 
> Well there is a benefit in breaking things down into algebraically simple parts: easier to reason about (faster to review for a slow-thinking clown like me), easier to compose, no space for surprising corner cases; and when you have a library of these you find yourself reaching out for existing well-tested bits to reuse.
> 
> Fusing the walks is about performance which is not the dominating factor here.
> 
> That being said I can no longer find empirically significant problems with this algo as written so let's check it in!
> 
> I found this one fails:
> 
> ```go
> func TestIgnoreChangesInsertsNested(t *testing.T) {
> 	olds := resource.PropertyMap{
> 		"k": resource.NewObjectProperty(resource.PropertyMap{
> 			"n1": resource.NewObjectProperty(resource.PropertyMap{
> 				"n2": resource.NewStringProperty("!"),
> 			}),
> 		}),
> 	}
> 	news := resource.PropertyMap{}
> 	news2, err := applyIgnoreChanges(olds, news, []string{"k.*"})
> 	assert.NoError(t, err)
> 	assert.Equal(t, olds, news2)
> }
> ```
> 
> But if you switch the path to "k" it works out, so that's great. That should be fine and there's some logic to it.

Thats expected behavior. We treat the ignore path as pinning the named element and everything below it, but not elements that come before the named element. We could make this test pass for objects, but I have chosen not too for consistency's sake. My concern is for paths like this:

```
old: { arr: [ 1, 2 ] }
new: { }
path: arr[1]
```

There are several different ways we could handle this, but all of them require an explanation more complicated then "we don't ignore the removal of parents".

t0yv0:
Yeah! Checked this scenario (update that removes tags):

```

// Create an AWS resource (S3 Bucket)
const bucket = new aws.s3.Bucket("my-bucket", {
    accelerationStatus: "Suspended",
    // tags: {
    //     "a": "b",
    // },
}, {
    ignoreChanges: ["accelerationStatus", "tags.*"],
});
```

So tags.* did not apply, but tags applied! I think the new code is on par with legacy then so that's great!

PR #929, opened 2023-03-30
Link ProposedNew from pulumi/terraform fork

t0yv0:
Remove home-grown code in favor of linking the canonical objchange.ProposedNew algorithm from Terraform CLI via the pulumi/terraform fork under adapters.


t0yv0:
Progress!

This passes almost all the tests except one of the new Set tests.

The other problem that needs solving is protobuf detects a conflict.

```
WARNING: proto: message tfplugin6.ReadResource.Request is already registered
        previously from: "github.com/t0yv0/terraform-proposed-new/tfplugin6"
        currently from:  "github.com/hashicorp/terraform-plugin-go/tfprotov6/internal/tfplugin6"
See https://protobuf.dev/reference/go/faq#namespace-conflict
```

But if we can work through this this would be a much better foundation to build on - rely on the same algorithm as Terraform has by linking it in is far more robust than trying to rebuilt it.

pf/go.mod
```diff
@@ -8,17 +8,31 @@ require (
...
 	github.com/stretchr/testify v1.8.2
-	google.golang.org/grpc v1.53.0
+	google.golang.org/grpc v1.54.0
 	pgregory.net/rapid v0.5.5
 )
 
+require (
+	github.com/Masterminds/semver/v3 v3.1.1 // indirect
+	github.com/Masterminds/sprig/v3 v3.2.2 // indirect
+	github.com/apparentlymart/go-versions v1.0.1 // indirect
```
t0yv0, line 26: Inheriting some new deps via pulumi/terraform

pf/internal/pfutils/attr.go
```diff
@@ -111,62 +107,3 @@ const (
 	NestingModeSet     NestingMode = 3
 	NestingModeMap     NestingMode = 4
 )
-
-// Classifies the results of LookupTerraformPath. The only valid cases are:
```
t0yv0, line 115: Gone for good.

pf/tfbridge/provider.go
```diff
@@ -76,11 +77,21 @@ func NewProvider(ctx context.Context, info ProviderInfo, meta ProviderMetadata)
...
+	// Somehow this GetProviderSchema call needs to happen at least once to avoid Resource Type Not Found in the
+	// tfServer, to init it properly to remember provider name and compute correct resource names like
+	// random_integer instead of _integer (unknown provider name).
+	getProviderSchemaResp, err := server6.GetProviderSchema(ctx, &tfprotov6.GetProviderSchemaRequest{})
 	if err != nil {
-		return nil, fmt.Errorf("Fatal failure starting a provider server: %w", err)
+		return nil, err
 	}
+
+	// TODO handle diagnostics from getProviderSchemaResp, they are actually quite important.
```
t0yv0, line 93: Reminder to self: must fix this before shipping.

pf/tests/testdata/updateprogram/Pulumi.yaml
```diff
@@ -27,4 +27,4 @@ outputs:
   testOptionalString__expect: |-
     input2
   testSetOptionals__actual: ${set.setOptionals}
-  testSetOptionals__expect: [ "${r.requiredInputStringCopy}", a, b ]
```
t0yv0, line 30: This worries me quite bit! After using the new algorithm the order changed in the test. Perhaps the values are after all reordered for sets in TF somehow?

t0yv0: @iwahbe we actually assume order preservation correct?

t0yv0:
The good news is that test pass, the bad news is that set elements might reorder.

t0yv0:
Closing in favor of https://github.com/pulumi/pulumi-terraform-bridge/pull/1276

PR #933, opened 2023-03-30
Deprecate IDFields

t0yv0:
A search across Pulumi org repos finds no usage of this field, introduced in 2017.


guineveresaenger (approved):
I love the deprecation warning 🙌

PR #943, opened 2023-03-31
Fix #941: SetAutonaming panics when used with Plugin Framework

### LINKED ISSUE #941: Nil pointer panic when using SetAutonaming with Plugin Framework
### What happened?

@tmeckel reports that this code needs to be commented out when using Plugin Framework. Autonaming should gen a "not supported yet" warning instead.

```
(prov.SetAutonaming(255, "-")), b
```

### Expected Behavior

No panic

### Steps to reproduce

Use prov.SetAutonaming(255, "-") with plugin framework 

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

t0yv0:
Fixes #941

### DIFF
--- a/pf/tfbridge/provider_info_test.go
+++ b/pf/tfbridge/provider_info_test.go
@@ -1,0 +1,31 @@
+// Copyright 2016-2023, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


iwahbe (approved):
It's not clear to me if we want this to warn or error (and exit). `SetAutonaming`, `ComputeDefaults` and `AutoAlias` all assume that `ProviderInfo.P` is non-nil. Calling them is wrong, since they do nothing. Either way, I agree that this is better then a nil panic.

We might want to abstract the above mentioned resources to work on a interface that both versions of `ProviderInfo` conform to.

PR #944, opened 2023-04-02
Add aws mapping for converter tests

Frassle:
Also fix `MarshallableElem.Unmarshal` which was exposed as incorrect when adding this mapping by the mapping to schema test.


pkg/tfbridge/info.go
```diff
@@ -529,10 +529,10 @@ func (m *MarshallableElem) Unmarshal() interface{} {
 		return nil
 	case m.Schema != nil:
 		return m.Schema.Unmarshal()
-	case m.Resource != nil:
-		return m.Resource.Unmarshal()
 	default:
-		return nil
+		// m.Resource might be nil in which case it was empty when marshalled. But Unmarshal can be called on
```
t0yv0, line 533: Ugh what value were you getting here and in what context I'd love to scroll back a bit? This value must have:

```
m != nil
m.Schema == nil
m.Resource == nil
```

That seems really weird.

I have some case analysis here:

https://github.com/pulumi/pulumi-terraform-bridge/blob/228b97f2c51a58bf3077019d9c1c5e7690f66749/pkg/tfshim/shim.go#L79-#L79

You seem to be hitting something that's neither of the cases (in the MarshallableElem projection)

Frassle: Resource is just a map, and it's tagged 'omitempty' so a resource with no fields gets written out as `{}`, which then gets read back in as both fields being nil.

Frassle: There were some list elements in the Aws schema like that. On my phone so hard to go look them up, but I can check and reply tomorrow.

Frassle:
```
                            "connector_profile_properties": {
                                "type": 5,
                                "required": true,
                                "element": {
                                    "resource": {
                                        "amplitude": {
                                            "type": 5,
                                            "optional": true,
                                            "element": {},
                                            "maxItems": 1
                                        },
```
That maps to an empty object in the codegen (see `ConnectorProfileConnectorProfileConfigConnectorProfilePropertiesAmplitude` type in the aws typescript sdk for example)

t0yv0 (changes requested):
So it turns out the suggested change stars confusing the `nil` case with the `empty resource` case, which I think is semantically dangerous - we're using that dispatch in a few places.. 

Suggest this approach instead:

https://github.com/pulumi/pulumi-terraform-bridge/pull/947

I've added a few test cases to elucidate.

Frassle:
> So it turns out the suggested change stars confusing the nil case

I don't think it does? We still have an explicit check for if the Marshalable thing is nil overall.

t0yv0:
Please check out the tests in #947. These tests do not pass on this PR.

t0yv0 (approved):
OK yes, the issue as with how the test was setup.

Let's get this in.

I've trimmed down 948 to only add the tests which confirm this fix stay fixed now:

https://github.com/pulumi/pulumi-terraform-bridge/pull/948

t0yv0:
I'll merge this up when tests pass, thanks @Frassle! Very good to fix the foundations 😓

PR #947, opened 2023-04-04
Fix Elem marshalling to support empty resource turnaround

t0yv0:
@Frassle discovered a bug where MarshalElem does not turnaround through JSON on empty resources. A resource came up in:

```
                            "connector_profile_properties": {
                                "type": 5,
                                "required": true,
                                "element": {
                                    "resource": {
                                        "amplitude": {
                                            "type": 5,
                                            "optional": true,
                                            "element": {},
                                            "maxItems": 1
                                        },
``` 

> That maps to an empty object in the codegen (see ConnectorProfileConnectorProfileConfigConnectorProfilePropertiesAmplitude type in the aws typescript sdk for example)

The problem is with this type:

```
type MarshallableElem struct {
	Schema   *MarshallableSchema  `json:"schema,omitempty"`
	Resource MarshallableResource `json:"resource,omitempty"`
}
```

Logically it has to represent three distinct cases. 1. `nil`; 2. `Schema`; 3. `Resource` - see the semantic distinction documented on `Elem()`.

However due to `omitempty` on Resource empty resources get the nil and Resource cases confused.

We expect the marshal facility to turn around all the semantic cases. 

The PR exploits the distinction between case 1 being serialized as `null` in JSON and case 3 on empty resource being serialized as `{}`. To make the distinction matter a custom UnmarshalJSON serializer is added. Test cases are added to check marshal turnaround for Resource, Schema and nil.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -15,6 +15,7 @@
 package tfbridge
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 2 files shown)


t0yv0:
Ah ok, never mind, it turns out this was a test artifact. The test needs to work on marshalling the higher level bits, then the original PR passes the tests. I'll factor out just the tests to add to the original PR.

PR #948, opened 2023-04-04
Tests Elem marshalling to support empty resources and the nil case

t0yv0:
Adding tests that confirm the fix.


pkg/tfbridge/info.go
```diff
@@ -487,6 +487,9 @@ type MarshallableResource map[string]*MarshallableSchema
 // MarshalResource converts a Terraform resource schema into a MarshallableResource.
 func MarshalResource(r shim.Resource) MarshallableResource {
 	m := make(MarshallableResource)
+	if r.Schema() == nil {
+		return m
+	}
```
iwahbe, line 492: Not a test, so maybe rename the PR.

iwahbe (approved):

pkg/tfbridge/info.go
t0yv0: True.

PR #950, opened 2023-04-04
Do not panic on happy path

t0yv0:
With this change the AWS provider example no longer panics on startup.


iwahbe (approved):
I thought this was a lint :(

t0yv0:
Yeah let me check why linter doesn't catch this.

PR #952, opened 2023-04-05
Convert dynamic blocks to ForExpressions

Zaid-Ajaj:
This PR handles `dynamic` blocks and converts them to `ForExpression`s

It relies on the `entries` function to create a list of key-value pair objects because the `for_each` expression of the `dynamic` block creates a local which a `key` and a `value` property. 

The `entries` function has the wrong type in PCL but this is addressed by https://github.com/pulumi/pulumi/pull/12607 which blocks this PR until merged and used here


pkg/tf2pulumi/convert/tf.go
```diff
@@ -1281,8 +1281,50 @@ func convertBody(sources map[string][]byte, scopes *scopes, fullyQualifiedPath s
...
-		// TODO: We need to correctly handle dynamic blocks somehow.
 		if block.Type == "dynamic" {
+			eachVar := scopes.generateUniqueName("entry", "", "")
+			dynamicTokens := hclwrite.Tokens{makeToken(hclsyntax.TokenOBrack, "[")}
+			dynamicTokens = append(dynamicTokens, makeToken(hclsyntax.TokenIdent, "for"))
+			dynamicTokens = append(dynamicTokens, makeToken(hclsyntax.TokenIdent, eachVar))
+			dynamicTokens = append(dynamicTokens, makeToken(hclsyntax.TokenIdent, "in"))
+			dynamicBody, ok := block.Body.(*hclsyntax.Body)
+			if !ok {
+				continue
```
Frassle, line 1292: This should probably panic

pkg/tf2pulumi/convert/tf.go
```diff
@@ -1281,8 +1281,50 @@ func convertBody(sources map[string][]byte, scopes *scopes, fullyQualifiedPath s
...
+			}
+
+			forEachAttr, hasForEachAttr := dynamicBody.Attributes["for_each"]
+			if !hasForEachAttr {
+				continue
```
Frassle, line 1297: Is it valid to have a dynamic block without for_each? Should it just set the attribute to `[]`?

pkg/tf2pulumi/convert/tf.go
Zaid-Ajaj: Done

PR #963, opened 2023-04-06
WIP: on top of muxer, add renames to ProviderMeta

t0yv0:
@iwahbe early preview. I got this to pass tests but go embed limitations require some file motion/movement.

The change makes Plugin Framework providers (even unmuxed) to write Renames into MetadataInfo, which is now required for PF providers.

This imposes a change on the user code, unfortunately, so that will need to be called out in release notes. 

Also looks like I need to update pf/README.md examples for this.


pkg/tfgen/generate.go
```diff
@@ -1800,3 +1803,20 @@ func ignoreMappingError(s []string, str string) bool {
 	}
 	return false
 }
+
+func addRenamesToMetadataInfo(g *Generator) error {
```
t0yv0, line 1807: This has the side-effect of adding renames to providers that don't strictly need them, but opt into MetadataInfo.

iwahbe (approved):
I think this is better. Its breaking, but future adoptions only need to embed one file, not multiple files.

t0yv0:
Yeah I think it's the right direction but I itch to rename ProviderMetadata; later.. OK I will rebase on master.

t0yv0:
Closing in fav of https://github.com/pulumi/pulumi-terraform-bridge/pull/964 cleaned up and rebased on master.

PR #964, opened 2023-04-06
Breaking: Require MetadataInfo for Plugin Framework

t0yv0:
ProviderInfo.MetadataInfo is now required to be set for Plugin Framework based providers (when bridinging with
pf/tfbridge). ProviderMetadata.BridgeMetadata is now deprecated.

To update your provider follow these steps:

- remove BridgeMetadata so that the main.go of the provider reads:

      meta := tfbridge.ProviderMetadata{
           PackageSchema: schema,
           // BridgeMetadata: bridgeMetadata, // this is now redundant
      }
      ...

- move the declaration `//go:embed bridge-metadata.json` from the main.go of the provider to the resources.go file that
configures ProviderInfo. The updated stanza might now need to read like this:

      //go:embed cmd/pulumi-resource-myproviderxyz/bridge-metadata.json
      var bridgeMetadata

- add a new field to ProviderInfo:

      import "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
      ...
      MetadataInfo: tfbridge.NewProviderMetadata(bridgeMetadata),

The change is motivated by the need to maintain providers that combine SDK-v2 and Plugin Framework based resources in
a single Pulumi bridged provider. The MetadataInfo framework helps disambiguate metadata required by pf/tfbridge from
other forms of metadata required to support such mixed providers.


pf/tests/internal/testprovider/cmd/pulumi-resource-random/#main.go#
iwahbe, line 1: I don't think we want to check this in.

iwahbe (approved):
One quick fix, otherwise LGTM!

PR #967, opened 2023-04-06
Muxer plus renames-via-metadata

t0yv0:


t0yv0:
Closing. A more up-to-date attempt is in https://github.com/pulumi/pulumi-terraform-bridge/pull/988

PR #972, opened 2023-04-10
Fix muxer so that AWS provider starts again

t0yv0:


pkg/tfgen/generate.go
```diff
@@ -1814,7 +1809,20 @@ func addRenamesToMetadataInfo(g *Generator) error {
...
-	renames, err := g.Renames()
+	var renames Renames
+
+	_, renamesAlreadySet, err := metadata.Get[Renames](g.info.MetadataInfo.Data, "renames")
+	if err != nil {
+		return fmt.Errorf("[pkg/tfgen] failed to retrieve renames from MetadataInfo: %w", err)
+	}
+
+	// Skip setting renames, for scenarios such as muxed providers that have them precomputed.
+	if renamesAlreadySet {
```
t0yv0, line 1820: Adding this early termination - this is important not to clobber in a multi-provider case. Sounds like if renames are already set adding them by tfgen should be either an error or it should resolve the conflict somehow (merge?). Here it resolves the conflict to the prior renames bit set.

pkg/tfgen/generate.go
```diff
@@ -662,20 +662,15 @@ func (of *overlayFile) Name() string { return of.name }
 func (of *overlayFile) Doc() string  { return "" }
 func (of *overlayFile) Copy() bool   { return of.src != "" }
 
-func GenerateSchemaAndRenames(info tfbridge.ProviderInfo, sink diag.Sink) (pschema.PackageSpec, Renames, error) {
```
t0yv0, line 665: I did not like this expansion of public API since we already have GenerateSchemaWithOptions that gives the caller access to renames if needed; this new API over-emphasizes Renames internals. So I inlined it.

pf/tfgen/main.go
```diff
@@ -294,25 +285,3 @@ func mergeRenames(renames []tfgen.Renames) tfgen.Renames {
 	}
 	return main
 }
-
-func writeRenames(renames tfgen.Renames, opts tfgen.GeneratorOptions) error {
```
t0yv0, line 298: So since renames are part of metadata now metadata writing to bridge-metadata.json already writes renames there. I think this method shouldn't be around as it can clobber bridge-metadata.json.

pf/tfgen/main.go
```diff
@@ -235,22 +224,24 @@ func MainWithMuxer(provider string, infos ...tfbridge.Muxed) {
 			"Must provide a metadata store when muxing providers. See ProviderInfo.MetadataInfo")
 		err = metadata.Set(muxedInfo.GetMetadata(), "mux", mapping)
 		if err != nil {
-			return fmt.Errorf("Failed to save metadata: %w", err)
+			return fmt.Errorf("[pf/tfgen] Failed to set mux data in MetadataInfo: %w", err)
+		}
+
+		// In the muxing case precompute renames by merging them and set them to MetadataInfo, this will avoid
+		// recomputing the renames in GenerateFromSchema, it will write out the mergedRenames as-is.
+		mergedRenames := mergeRenames(renames)
```
t0yv0, line 232: This bit is a bit odd. We'll be calling g.GenerateFromSchema below but we do not want that to recompute renames. We instead want it to accept the merged renames from all the tfgen.GenerateSchemaWithOptions calls done before this line is hit. So we set it upfront.

iwahbe: I agree that we need to pull out renames here. I think we need to set rename somewhere that `GenerateFromSchema` is aware of, maybe with a `opts.PrecomputedRenames = mergeRenames(renames)`. 

See my other comment for why existence checking isn't sufficient.

pf/tfgen/main.go
iwahbe: I agree. This method was only for the pre-bridge based metadata world.

t0yv0:
Hmm interesting, there's a bit of a factoring issue with renames and side-effects. I'm finding that:

- Renames get deserialized from previous version of provider in MetadataInfo which we really don't want that value of old renames to be used anywhere

- GeneratePackageFromSchema is not well founded because of imperative nature of Generator.. Like accessing Generator renamesBuilder without generating schema first is not OK; I tried to tinker with the signature to make it more obvious it needs to receive precomputed renames from a different phase

- Still, nameCheck in GeneratePackageFromSchema is not well founded (more imperative assumptions). I tried to move it up.

With these changes pulumi-aws starts. And looks like we're passing tests. Still something feels not ideal.

iwahbe:
I'm going to merge into `iwahbe/muxer` as is. I'll fix the test case and do some renaming in the main muxer branch.

PR #975, opened 2023-04-11
Assert Generated renames match existing renames

iwahbe:
It's surprisingly difficult to create/use an existing rename. This strategy allows us to recompute a rename set safely. Instead of saving it, we just check if the computed renames match existing renames, erroring out if that isn't the case.


iwahbe:
Closing in favor of `t0yv0/muxer-3`.

PR #982, opened 2023-04-12
camelCase the attributes of the default value of object-typed config variables

Zaid-Ajaj:
Also adding examples of resources of which the inputs are using members of object-typed config variables that are rewritten  as camelCase as well.

### DIFF
--- a/pkg/tf2pulumi/convert/tf.go
+++ b/pkg/tf2pulumi/convert/tf.go
@@ -1425,32 +1425,52 @@ func convertBody(sources map[string][]byte, scopes *scopes, fullyQualifiedPath s
 
 // camelCaseObjectAttributes rewrites the attributes of objects to camelCase and returns the modified value.
 // when the input is a list of objects or map of objects, those are modified recursively.
-func camelCaseObjectAttributes(value cty.Value, valueType cty.Type) cty.Value {
+func camelCaseObjectAttributes(value cty.Value) cty.Value {
 	// handle type object({...})
-	if valueType.IsObjectType() {
+	if value.Type().IsObjectType() {
 		modifiedAttributes := map[string]cty.Value{}
 		for propertyKey, propertyValue := range value.AsValueMap() {
-			modifiedValue := camelCaseObjectAttributes(propertyValue, propertyValue.Type())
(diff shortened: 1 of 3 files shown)


pkg/tf2pulumi/convert/tf.go
```diff
@@ -1423,6 +1423,43 @@ func convertBody(sources map[string][]byte, scopes *scopes, fullyQualifiedPath s
...
+
+	// handle type map(object({...}))
+	if valueType.IsMapType() && valueType.ElementType().IsObjectType() {
+		modifiedAttributes := map[string]cty.Value{}
+		for propertyKey, propertyValue := range value.AsValueMap() {
+			modifiedValue := camelCaseObjectAttributes(propertyValue, propertyValue.Type())
+			modifiedAttributes[propertyKey] = modifiedValue
+		}
+		return cty.MapVal(modifiedAttributes)
+	}
```
Frassle, line 1457: Don't we need to handle set and tuple types here?

pkg/tf2pulumi/convert/tf.go
```diff
@@ -1423,6 +1423,43 @@ func convertBody(sources map[string][]byte, scopes *scopes, fullyQualifiedPath s
 	return newAttributes
 }
 
+// camelCaseObjectAttributes rewrites the attributes of objects to camelCase and returns the modified value.
+// when the input is a list of objects or map of objects, those are modified recursively.
+func camelCaseObjectAttributes(value cty.Value, valueType cty.Type) cty.Value {
```
Frassle, line 1428: Do we really need the type here? The cty.Value knows it's type, no need to pass in another variable (that might be wrong if we mess up the call to this)

Zaid-Ajaj: I used this because `config.Type` might not be the same as `config.Default.Type()` but I guess we should go off of the `config.Default : cty.Value` in this case. I've simplified the implementation now.

pkg/tf2pulumi/convert/tf.go
Zaid-Ajaj: Done! sets are handled like lists now ✅

Frassle (approved):

PR #984, opened 2023-04-12
Fix recognizing MaxItems=1 list blocks in Plugin Framework

t0yv0:
Based on a failing example from AWS, ensure that listvalidator.SizeAtMost(1) directive is recognized at schema generation time so that Pulumi can decide correctly whether a property has plural name and a List[T] type or a singular name and a T type.

This is especially important for backwards-compatibility for providers migrating to the Plugin Framework.

### DIFF
--- a/pf/internal/pfutils/block.go
+++ b/pf/internal/pfutils/block.go
@@ -76,7 +76,10 @@ type hasListValidators interface {
 	ListValidators() []validator.List
 }
 
-var listSizeRegExp = regexp.MustCompile(`^list must contain at least (\d+) elements and at most (\d+) elements$`)
+var listSizeRegExpAtLeastAtMost = regexp.MustCompile(
+	`^list must contain at least (\d+) elements and at most (\d+) elements$`)
+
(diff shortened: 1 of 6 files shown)


pf/internal/pfutils/block.go
```diff
@@ -86,13 +87,18 @@ func detectSizeConstraints(x BlockLike) (int64, int64, bool) {
 	if listBlock, isList := x.(hasListValidators); isList {
 		for _, v := range listBlock.ListValidators() {
 			desc := v.Description(ctx)
-			if m := listSizeRegExp.FindStringSubmatch(desc); m != nil {
+			if m := listSizeRegExpAtLeastAtMost.FindStringSubmatch(desc); m != nil {
```
iwahbe, line 90: It's unfortunate that this can't be a type case.
https://github.com/hashicorp/terraform-plugin-framework-validators/blob/48349f4e12f4283dc442114e01029baf933dfee5/listvalidator/size_at_least.go#L14-L17

iwahbe (approved):
LGTM!

pf/internal/pfutils/block.go
t0yv0: reflection could work.

PR #986, opened 2023-04-12
Implement GetMapping for Muxed Providers

### LINKED ISSUE #946: [Muxer] Implement Support for GetMappings
## Hello!

## Issue details

We need to support merging the GetMappings call. This is necessary to, for example, allow `tf2pulumi` to work with AWS.

iwahbe:
Fixes #946

There are three different cases for behavior to consider:
1. No underlying provider implements GetMapping, in which case the muxer provider also doesn't implement GetMapping.
2. Only 1 underlying provider implements GetMapping, in which case the muxer provides unfiltered pass through from the provider.
3. Multiple providers respond to the key, in which case we check for a key merging function. We delegate merging the response to this function. This allows us to avoid a cyclic dependency on the rest of the bridge.

### DIFF
--- a/x/muxer/main.go
+++ b/x/muxer/main.go
@@ -79,10 +79,9 @@ import (
 //     `Mux` is replaced. If subsidiary servers where constructed with the same `host` as
 //     passed to `Mux`, then they will observe the new `host` spurred by `Attach`.
 //
-//   - GetMapping: TODO (https://github.com/pulumi/pulumi-terraform-bridge/issues/946)
-//     GetMapping returns an opaque blob. It's not clear how to merge blobs. We might
-//     require that only one provider responds with a mapping, and just forward the first
-//     provider that responds.
+//   - GetMapping: `GetMapping` dispatches on all underlerver Servers. If zero or 1 server
+//     responds with a non-empty data section, we call GetMappingHandler[Key] to merge the
+//     data sections, where Key is the key given in the GetMappingRequest.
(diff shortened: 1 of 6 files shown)


x/muxer/muxer.go
```diff
@@ -423,8 +430,77 @@ func (m *muxer) Attach(ctx context.Context, req *rpc.PluginAttach) (*emptypb.Emp
...
+			errs.Errors = append(errs.Errors, err)
+			continue
+		}
+		response := v.A
+		if len(response.Data) == 0 {
+			continue
+		}
+		if response.GetProvider() == "" {
+			errs = multierror.Append(errs,
+				m.Warnf(ctx, "GetMapping", "Missing provider name for subprovider %d", i))
```
t0yv0, line 455: AFAIK It always has to be set so can start being strict.

x/muxer/main.go
```diff
@@ -94,6 +93,8 @@ type Main struct {
 	//
 	// If set, ComputedMapping must also be set.
 	Schema string
+
+	GetMappingHandler map[string]MultiMappingHandler
```
t0yv0, line 97: The naming is terribly confusing with ComputedMapping

t0yv0 (approved):

x/muxer/muxer.go
iwahbe: Great, I'll make it strict.

PR #989, opened 2023-04-13
Fixes an failure to derive Config encoder encountered in pulumi-aws

t0yv0:
Currently pulumi-aws has a dangling reference in the schema, confusing the code in encoder derivation. The change works around this until the underlying issue is fixed.

### DIFF
--- a/pf/internal/convert/dangling_ref_test.go
+++ b/pf/internal/convert/dangling_ref_test.go
@@ -1,0 +1,91 @@
+// Copyright 2016-2023, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package convert
+
+import (
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+// AWS has a dangling enum ref, tracked in pulumi/pulumi-aws#2463 - this code needs to be able to handle this.
+func TestRegressEncodingConfigWithDanglingEnumRefs(t *testing.T) {
+	spec := PrecomputedPackageSpec(&schema.PackageSpec{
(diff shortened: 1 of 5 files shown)


pf/tfbridge/provider.go
```diff
@@ -243,37 +243,3 @@ func newProviderServer6(ctx context.Context, p pfprovider.Provider) (tfprotov6.P
 
 	return server6, nil
 }
-
-type packageSpec struct {
```
t0yv0, line 247: This just moves around to facilitate testing. Can also remove. A bit of premature abstraction perhaps.

pf/internal/convert/dangling_ref_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+// AWS has a dangling enum ref, tracked in pulumi/pulumi-aws#2463 - this code needs to be able to handle this.
+func TestRegressEncodingConfigWithDanglingEnumRefs(t *testing.T) {
```
t0yv0, line 27: Before the fix this blew up with "cannot derive Encoder for provider Config".

pf/internal/convert/encoding.go
```diff
@@ -253,7 +253,17 @@ func (e *encoding) deriveEncoder(typeSpec *pschema.TypeSpec, t tftypes.Type) (En
...
+			dangling := false
+			if strings.HasPrefix(typeSpec.Ref, "#/types/") {
+				token := strings.TrimPrefix(typeSpec.Ref, "#/types/")
+				if e.spec.Type(tokens.Type(token)) == nil {
+					dangling = true
+				}
+			}
+			if !dangling {
+				return nil, fmt.Errorf("expected Object or Tuple type but got %s", t.String())
+			}
```
iwahbe, line 266: Schema types with `Ref` fields that refer to non-existent types are called ["token types"](https://github.com/pulumi/pulumi/blob/6dd2b026916c2e999502100ae16f47faf3737690/pkg/codegen/schema/schema.go#L282-L289). For the purposes of deriving an encoding, they are equivalent to their underlying type. 

Instead of handling this in the `default` case (and causing `tftypes.Object` and `tftypes.Tuple` to fail with an unhelpful error in this case), I think we should check if the `Ref` is dangling before we enter the `switch`:

```go
if typeSpec.Ref != "" && e.validRef(typeSpec.Ref) {
    switch t := t.(type) {
        ...
    }
}
```

t0yv0: Hm that's fair. We're doing co-recursion over tftypes.Type and pschema.TypeSpec which is always a little odd to write; we could perhaps benefit from pattern-matching on tftypes.Type first. I took a pass simplifying this so we start the object/tuple cases by looking at tftypes.Type. IFF that needs an object or tuple, we require a local resolvable "ref". This seems reasonable for now. Leaving non-local refs for later.

iwahbe (approved):

PR #990, opened 2023-04-13
Fix muxer config mapping

iwahbe:
This change simplifies and corrects mapping code, notably fixing config mapping.

### DIFF
--- a/pf/tfgen/main.go
+++ b/pf/tfgen/main.go
@@ -20,7 +20,6 @@ import (
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
-	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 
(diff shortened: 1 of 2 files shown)


aq17 (approved):
LGTM but @t0yv0 has more context.

t0yv0:
So a refactor without any change in behavior? That's fine! Your comments seem to indicate as much - just mark the description REFACTOR in these cases.

iwahbe:
There were correctness changes here. I think my comment said as much:
```
This change simplifies and corrects mapping code, notably fixing config mapping.
                           ^^^^^^^^                       ^^^^^^
```

t0yv0:
AH sorry, can you highlight which ones? It's usually helpful to zoom in on things like this with another pair of eyes or a small test (in prod mode). I may still be bit confused if we're in prod mode here or continue the spike mode.

pf/tfgen/main.go
```diff
@@ -174,73 +166,42 @@ func UnstableMuxProviderInfo(
...
+				func(k string) {
+					delete(info.ExtraFunctions, k)
+					info.IgnoreMappings = append(info.IgnoreMappings, k)
+				})
+
+			// Config
+			layerUnder(&muxedInfo.Config, info.Config, muxedInfo.ExtraConfig,
+				func(k string) { delete(info.Config, k) })
+			layerUnder(&muxedInfo.ExtraConfig, info.ExtraConfig, muxedInfo.Config,
+				func(k string) { delete(info.Config, k) })
```
iwahbe, line 204: In terms of changes, we add layering for the config.

x/muxer/mapping.go
```diff
@@ -108,65 +108,54 @@ func (mapping mapping) layerSchema(dstSchema, srcSchema *schema.PackageSpec, src
...
-		}
-		// TODO: Validity check that this value matches any other config value
-		// with the same key
-		m.dstSchema.Config.Variables[k] = v
+	for k := range m.srcSchema.Config.Variables {
 		m.mapping.Config[k] = append(m.mapping.Config[k], m.srcIndex)
-		m.addType(v.TypeSpec)
 	}
+	layerMap(&m.dstSchema.Config.Variables, m.srcSchema.Config.Variables,
+		func(_ string, t schema.PropertySpec) { m.addType(t.TypeSpec) })
```
iwahbe, line 115: We layer the config instead of setting it. In terms of our tests, this means we get SDKv2 config values instead of PF config values. SDKv2 values have defaults and descriptions.

PR #992, opened 2023-04-14
Remove unused elements, not used elements when muxing ProviderInfo

iwahbe:
Building pulumi-cloudflare with this change gets us https://github.com/pulumi/pulumi-cloudflare/pull/378/files, which is the schema diff we expect.

### DIFF
--- a/x/muxer/mapping.go
+++ b/x/muxer/mapping.go
@@ -130,7 +130,9 @@ func layerMap[K comparable, V any](dst *map[K]V, src map[K]V, finalize func(k K,
 		_, skip := (*dst)[k]
 		if !skip {
 			(*dst)[k] = v
-			finalize(k, v)
+			if finalize != nil {
+				finalize(k, v)
+			}
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


t0yv0 (changes requested):
I'd like to reach alignment on whether we continue with this approach or take the alternative approach where all the layer* code (both for mixing ProviderInfo and for mixing the Schema) is made redundant (deleted) and we run tfgen in a single-pass. Perhaps we can do so promptly (today even) doing a quick design discussion. It's wonderful that you have another test case here for Cloudflare, so as part of evaluation we can make sure thing work on Cloudflare as well as AWS and GCP.

I'm starting to slightly regret we got the code on main branch before sorting out end-to-end problems and getting more confidence. It can be too much hassle with feature branches so perhaps we can merge and iterate on main branch but it gets hard demarcating prod vs experimental code. I worry we will get confused which one is which as the codebase gets complicated. 
 
Also as a gentle reminder that the PR has no tests to illustrate the Cloudflare problem and as such is suitable for experimentation but does not meet our bar for prod code - it has a deferred testing obligation to reach a level of suitable for prod use.

t0yv0 (approved):
Sounds good, let's merge then. Tracking test obligations somehow would be nice even if it's just a TODO.

PR #993, opened 2023-04-14
Handle max items one override from fields info

Zaid-Ajaj:


pkg/tf2pulumi/convert/testdata/mappings/maxItemsOne.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "maxItemsOne",
+  "provider": {
+    "resources": {
```
Frassle, line 4: We should probably check this is correct for data sources as well.
Plus it would be good to cleanup whatever is currently testing the "maxItems=1" logic and cover that here so this mapping gives us coverage of "maxItems=1 && maxItemsOne=false" + just "maxItems=1".

pkg/tf2pulumi/convert/tf.go
```diff
@@ -1994,6 +1989,90 @@ func (s *scopes) pulumiName(fullyQualifiedPath string) string {
...
+		contract.Assertf(parts[0] != "", "empty path part passed into pulumiName")
+
+		// Lookup the info for this part
+		var curSch shim.Schema
+		if sch != nil {
+			curSch = sch.Get(parts[0])
+		}
+		curInfo := info[parts[0]]
+
+		// We want the name of this part
```
Frassle, line 1942: I feel like this core recurse loop could be shared, the only part that differs is this central return part.

pkg/tf2pulumi/convert/testdata/mappings/maxItemsOne.json
Zaid-Ajaj: > Plus it would be good to cleanup whatever is currently testing the "maxItems=1" logic and cover that here so this mapping gives us coverage of "maxItems=1 && maxItemsOne=false" + just "maxItems=1".

Already did that in `convertBody` 😄 `isList := !scopes.maxItemsOne(blockPath)`

> We should probably check this is correct for data sources as well.

Will add a test case for it ✅

Frassle (approved):

PR #996, opened 2023-04-14
Implemented TfBridge PF Muxing via shim.Provider merging

iwahbe:


iwahbe:
As of https://github.com/pulumi/pulumi-terraform-bridge/pull/996/commits/cfc707909083cc386ce069527ccfffbdbb269c4f, the PR:
- Changes the implementation of `pf/tfgen.MainWithMuxer` to mux `shim.Provider` instead of the generated schema.
- Does not change the API for `pf/tfgen.MainWithMuxer`.
- Generates a full schema correctly when tested against `pulumi-cloudflare`.

pf/tfgen/extend.go
```diff
@@ -0,0 +1,182 @@
...
+	shimSchema "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
+	"github.com/pulumi/pulumi-terraform-bridge/x/muxer"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+)
+
+func SchemaOnlyPluginFrameworkProvider(ctx context.Context, provider provider.Provider) shim.Provider {
+	return schemashim.ShimSchemaOnlyProvider(ctx, provider)
+}
+
+// AugmentShimWithPF augments an existing shim with a PF provider.Provider.
```
t0yv0, line 37: Comment here perhaps since it's to obvious from the name that the resulting shim.Provider is only suitable for schema generation, not runtime use with tfbridge.

pf/tfgen/main.go
```diff
@@ -79,50 +79,62 @@ func Main(provider string, info tfbridge.ProviderInfo) {
 func MainWithMuxer(provider string, infos ...tfbridge.Muxed) {
 	ctx := context.Background()
 
-	// It is safe to pass infos[0].GetInfo() to MainWithCusomGenerate because it is
-	// only used when passed to GenerateOptions.
+	contract.Assertf(len(infos) == 2,
```
t0yv0, line 82: Oof a bit brutal, so the idea is to change the signature later? note the code expect the right order also not just len=2.

t0yv0:
Looking good so far! Please signal when things are in "complete enough" state you'd like me to try testing them on pulumi-aws.

x/muxer/muxer.go
```diff
@@ -95,40 +95,12 @@ func (m *muxer) GetSchema(ctx context.Context, req *rpc.GetSchemaRequest) (*rpc.
 	return &rpc.GetSchemaResponse{Schema: m.schema}, nil
 }
 
-func filterConfig[T any](m *muxer, i int, vars map[string]T) {
```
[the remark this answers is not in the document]
iwahbe: All providers (including `pkg/tfbridge` and `pf/tfbridge` based providers) are required to ignore unknown keys. This means that it is safe to send each provider all keys and trust it to filter to only keys it cares about.

This gives us a simpler MVP without loss of functionality.

pf/tfgen/main.go
iwahbe: Fixed by merging in #997.

pf/tfgen/extend.go
iwahbe: Fixed by merging in #997.

pf/tfgen/extend.go
```diff
@@ -0,0 +1,182 @@
...
+type mergedProviderShim struct {
+	simpleSchemaProvider
+
+	providers []shim.Provider
+}
+
+func (m *mergedProviderShim) extend(provider shim.Provider) error {
+	res, err := disjointUnion(m.resources, provider.ResourcesMap())
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
iwahbe: Good catch!

iwahbe:
Since merging in https://github.com/pulumi/pulumi-terraform-bridge/pull/997, we have made a breaking change to the existing API, altering `MainWithMuxer` to accept only a `pkg/tfbridge.ProviderInfo` for `tfgen` and that `ProviderInfo` with schema and context for `tfbridge`.

t0yv0:
MainWithMuxer may be unreleased yet.

x/muxer/muxer.go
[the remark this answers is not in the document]
t0yv0: ack 👍

t0yv0 (approved):
🚢 

Looked a this code before doing a little skim-over still looks good! It's great that it passes muster on pulumi-aws and is able to get us to the good place.

Please be mindful of project history and tidy up the commit message so it lands with a good description in the history, something like this (I think we're not changing observable behavior here at least not for the tests we've run):

```
Refactors the internals of supporting providers that combine SDKv2 and Plugin Framework based resources and functions in one provider. The updated code merges the simpler shim.Provider structures instead of merging the outputs of schema generation. This reduce the surface area of potential issues as schema generation now runs in a single-pass and can continue to make use of the global program analysis assumption.
```

pf/tfbridge/extend.go
```diff
@@ -27,7 +27,9 @@ func SchemaOnlyPluginFrameworkProvider(ctx context.Context, p pfprovider.Provide
 	return schemashim.ShimSchemaOnlyProvider(ctx, p)
 }
 
-// AugmentShimWithPF augments an existing shim with a PF provider.Provider.
-func AugmentShimWithPF(ctx context.Context, shim shim.Provider, p pfprovider.Provider) shim.Provider {
+// MuxShimWithPF initializes a shim.Provider that will server resources from both shim and p.
+//
+// To create a muxed provider, ProviderInfo.P must be the result of this function.
+func MuxShimWithPF(ctx context.Context, shim shim.Provider, p pfprovider.Provider) shim.Provider {
```
t0yv0, line 33: This one is a mouthful but I struggle to suggest a name here. I lean to something abbreviation free like `ExtendProvider` but YMMV with Go, this is definitely not .NET (in terms of longer names)...

PR #997, opened 2023-04-15
Change Muxer API to take a single tfbridge.ProviderInfo

iwahbe:
This builds on https://github.com/pulumi/pulumi-terraform-bridge/pull/996, simplify the user facing API.

Example usage looks like https://github.com/pulumi/pulumi-cloudflare/pull/379.

### DIFF
--- a/pf/tests/provider_get_mapping_test.go
+++ b/pf/tests/provider_get_mapping_test.go
@@ -25,9 +25,7 @@ import (
 
 	"github.com/pulumi/pulumi-terraform-bridge/pf/tests/internal/testprovider"
 	"github.com/pulumi/pulumi-terraform-bridge/pf/tfbridge"
-	"github.com/pulumi/pulumi-terraform-bridge/pf/tfgen"
 	tfbridge0 "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
-	tfgen0 "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen"
 )
 
 func TestGetMapping(t *testing.T) {
@@ -64,17 +62,9 @@ func TestGetMapping(t *testing.T) {
 func TestMuxedGetMapping(t *testing.T) {
 	ctx := context.Background()
 
-	servers := testprovider.MuxedRandomProvider()
+	info := testprovider.MuxedRandomProvider()
 
-	_, _, schema, _, err := tfgen.UnstableMuxProviderInfo(ctx,
-		tfgen0.GeneratorOptions{}, "muxedrandom", servers...)
-	require.NoError(t, err)
-
-	schemaBytes, err := json.Marshal(schema)
-	require.NoError(t, err)
-
-	server, err := tfbridge.MakeMuxedServer(ctx, "random",
-		tfbridge.ProviderMetadata{PackageSchema: schemaBytes}, servers...)(nil)
+	server, err := tfbridge.MakeMuxedServer(ctx, genSDKSchema(t, info), info)(nil)
 	require.NoError(t, err)
 
 	req := func(key string) (context.Context, *pulumirpc.GetMappingRequest) {
--- a/pf/tfbridge/main.go
+++ b/pf/tfbridge/main.go
@@ -27,8 +27,9 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 
+	pfmuxer "github.com/pulumi/pulumi-terraform-bridge/pf/internal/muxer"
+	"github.com/pulumi/pulumi-terraform-bridge/pf/internal/schemashim"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
-	sdkBridge "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata"
 	"github.com/pulumi/pulumi-terraform-bridge/x/muxer"
 )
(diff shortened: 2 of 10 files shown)


pf/tfbridge/main.go
```diff
@@ -105,11 +106,13 @@ func MainWithMuxer(ctx context.Context, pkg string, meta ProviderMetadata, infos
...
 func MakeMuxedServer(
-	ctx context.Context, pkg string, meta ProviderMetadata, infos ...Muxed,
+	ctx context.Context, schema []byte, info tfbridge.ProviderInfo,
 ) func(host *rprovider.HostClient) (pulumirpc.ResourceProviderServer, error) {
-	version := infos[0].GetInfo().Version
-	schema := string(meta.PackageSchema)
-	dispatchTable, found, err := metadata.Get[muxer.DispatchTable](infos[0].GetInfo().GetMetadata(), "mux")
+
+	shim, ok := info.P.(*pfmuxer.ProviderShim)
+	contract.Assertf(ok, "MainWithMuxer must have a ProviderInfo.P created with AugmentShimWithPF")
```
t0yv0, line 113: Ah ok here we go, here is the downcast trick! Thanks.

pf/tfgen/main.go
```diff
@@ -121,181 +106,3 @@ func MainWithMuxer(provider string, infos ...tfbridge.Muxed) {
...
-		contract.Assertf(!ok, "Map was not disjoint on key %s", k)
-		(*dst)[k] = v
-	}
-}
-
-// Mux multiple infos into a unified representation.
-//
-// This function exposes implementation details for testing. It should not be used outside
-// of pulumi-terraform-bridge.  This is an experimental API.
-func UnstableMuxProviderInfo(
```
t0yv0, line 144: Thinking out loud: do you want to keep this around somewhere for the generalized provider muxing case - we don't have a use case lined up now but maybe in the future? If we're mixing two providers P1 and P2 without knowing much about them (not assuming they're bridged, say). This would require mixing their schemas at least. So perhaps the code to merge ProviderInfo is not needed there, and all this code is about that, not merging the schemas.

pf/tests/provider_get_mapping_test.go
```diff
@@ -64,17 +62,9 @@ func TestGetMapping(t *testing.T) {
 func TestMuxedGetMapping(t *testing.T) {
 	ctx := context.Background()
 
-	servers := testprovider.MuxedRandomProvider()
+	info := testprovider.MuxedRandomProvider()
```
t0yv0, line 65: I forgot we have this test already, this is great!

pf/tfgen/main.go
```diff
@@ -76,29 +73,17 @@ func Main(provider string, info tfbridge.ProviderInfo) {
...
-func MainWithMuxer(provider string, infos ...tfbridge.Muxed) {
-	ctx := context.Background()
-
-	contract.Assertf(len(infos) == 2,
-		"We expect to see a SDKv2 and then a provider info for a muxed provider")
-	sdkV2 := infos[0].SDK
-	pf := *infos[1].PF
-
-	shim := AugmentShimWithPF(ctx, sdkV2.P, pf.NewProvider()).(*mergedProviderShim)
+func MainWithMuxer(info sdkBridge.ProviderInfo) {
```
t0yv0, line 76: I'm guessing this can't merge into Main yet? 

You could rename Main -> classicMain, MainWithMuxer -> mainWithMuxer and write Main that detects `info.P.(*pfmuxer.ProviderShim)` and dispatches based on that.

pf/tfbridge/main.go
```diff
@@ -91,10 +92,10 @@ func handleFlags(ctx context.Context, prov ProviderInfo, meta ProviderMetadata,
 // Implements main() or a bridged Pulumi plugin, complete with argument parsing.
 //
 // This is an experimental API.
-func MainWithMuxer(ctx context.Context, pkg string, meta ProviderMetadata, infos ...Muxed) {
-	f := MakeMuxedServer(ctx, pkg, meta, infos...)
+func MainWithMuxer(ctx context.Context, schema []byte, info tfbridge.ProviderInfo) {
```
t0yv0, line 95: Similarly to tfgen, consider if this can be merged into Main as a special case. If we don't like the signature of Main we can tweak that later on (just need to annotate "breaking" for the users and explain how to migrate). Proliferation of Main's that are linked to from outside packages is a bit unfortunate. Or do we need that for some reason?

pf/tfgen/main.go
iwahbe: This is only about merging `ProviderInfo`, which is bridge specific. Unless we envision a scenario where we are merging multiple bridged providers but can't do this by merging `shim.Provider` on a single `ProviderInfo` I don't think this will be used.

Schema mixing code in in `x/muxer`.

pf/tfgen/main.go
iwahbe: It could, and this would remove our dependence on `pf/tfbridge.ProviderInfo`. My preference would be to add the new functionality under `MainWithMuxer`, making sure that it is well tested and existing tests pass as is.

I'll merge `MainWithMuxer` into `Main` in a follow up PR that necessarily touches all our existing integration tests.

pf/tfbridge/main.go
iwahbe: This is just about not making a breaking change. As I said replying to the next comment, my plain is to finish this PR and merge. I'll follow up with a separate PR to fix the function API.

pf/tfgen/main.go
t0yv0: Thank you makes sense! We can defer that until "API cleanup" phase later on too - it does not matter for functionality per se.

pf/tfbridge/main.go
t0yv0: Perfect SGTM.

PR #1001, opened 2023-04-17
[Muxer simplify get mapping

iwahbe:


pf/tfbridge/main.go
```diff
@@ -36,7 +36,14 @@ import (
...
 // Implements main() or a bridged Pulumi plugin, complete with argument parsing.
 func Main(ctx context.Context, pkg string, prov ProviderInfo, meta ProviderMetadata) {
-	handleFlags(ctx, prov, meta, prov.Version)
+	handleFlags(ctx, prov.Version,
+		func() (*tfbridge.MarshallableProviderInfo, error) {
+			pp, err := newProviderWithContext(ctx, prov, meta)
+			if err != nil {
+				return nil, err
+			}
+			return pp.(*provider).marshalProviderInfo(ctx), nil
```
t0yv0, line 45: I was wondering why this trick is needed and then I recall is because in PF case we have prov.P == nil; so the helper function substitutes SchemaOnlyProvider() for the nil.

pf/tfbridge/main.go
```diff
@@ -120,12 +134,22 @@ func MakeMuxedServer(
 		fmt.Printf("Missing precomputed mapping. Did you run `make tfgen`?")
 		os.Exit(1)
 	}
+
+	getTFMapping := func(muxer.GetMappingArgs) (muxer.GetMappingResponse, error) {
```
t0yv0, line 138: 👍 this works.

t0yv0 (approved):
This looks good. 🚢

PR #1002, opened 2023-04-18
Fix provider configuration secrets handling for bridged providers

t0yv0:
The results of CheckConfig are written to the state by Pulumi CLI, therfore handling sensitive fields in CheckConfig is important to avoid storing sensitive data in plain-text from in the state file.

Changes:

- If a program uses a secret value when doing explicit provider configuration of a bridged provider, this value is now stored encrypted in the state file.

  See also: https://www.pulumi.com/docs/intro/concepts/resources/providers/#explicit-provider-configuration

- If a value is marked as sensitive in the upstream provider schema, it is now stored encrypted in the Pulumi statefile regardless of whether the provider is explicit or default.

- Providers with nested configuration options now support sensitive sub-properties. Due to limitatios of current design the sub-properties are supported using a conservative approximation. If any sub-properties are passed secret values by the program or marked sensitive in the upstream provider schema, the entire top-level property is stored encrypted in the state file.

The implementation relies on Pulumi engine to respect the results of CheckConfig call when storing provider state.

### DIFF
--- a/pkg/tfbridge/config_encoding.go
+++ b/pkg/tfbridge/config_encoding.go
@@ -28,7 +28,8 @@ import (
 )
 
 type configEncoding struct {
-	fieldTypes map[resource.PropertyKey]shim.ValueType
+	fieldTypes      map[resource.PropertyKey]shim.ValueType
+	sensitiveFields map[resource.PropertyKey]struct{}
 }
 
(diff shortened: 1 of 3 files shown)


t0yv0:
In addition to the included tests I've played with a modified build of pulumi-gcp that injects secret-ness into a sub-property of "batching" config value to make sure the results are encrypted in the state file.

Issue #904, opened 2023-03-22
Automatic MaxItem preservation in minor versions


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

### Context
The addition or remove of `MaxItemsOne` on an upstream property is a breaking change in the downstream Pulumi provider. 

The bridge projects `{ Type: List, Element: T, MaxItemOne: true }` as `T`, so removing `MaxItemOne` changes the projected type to `List[T]`. Adding `MaxItemOne: true` goes the other way, changing `List[T]` into `T`. 

Inspired by https://github.com/pulumi/pulumi-openstack/issues/248

### The problem
1. Removing `MaxItem: 1` is a non-breaking change in terraform, so we should make it non-breaking in pulumi.
2. Adding `MaxItem: 1` is a breaking change in terraform, but it is more disruptive in Pulumi than terraform.

### The solution

We should add `AutoAlias` style functionality to the bridge, adding appropriate alterations on minor updates, and remove then on major updates.

With regards to (2), the ecosystem team needs to decide on the correct step to remediate the break (design docs welcome), then implement it.

The playbook for doing this manually is [maxitemsone.md](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/maxitemsone.md) (Pulumi org only).

### Affected area/feature

`tokens.go` in the bridge.

PR #1007, opened 2023-04-18
Add MaxItemOne aliasing

iwahbe:
This PR augments `AutoAlias` to include support for "sticky" MaxItemOne properties.

When `AutoAlias` runs, it remembers the state of all values and sets `ProviderInfo.Resources` accordingly. This prevents upstream breaking changes from effecting Pulumi users.

When a new major version is set, `AutoAlias` discards its memory of aliases, allowing the provider to revert to its correct values.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/904

### DIFF
--- a/pkg/tfbridge/x/token_test.go
+++ b/pkg/tfbridge/x/token_test.go
@@ -23,24 +23,24 @@ import (
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
-	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/util"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata"
 	md "github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata"
 )
 
 func TestTokensSingleModule(t *testing.T) {
 	info := tfbridge.ProviderInfo{
-		P: Provider{
-			resources: map[string]shim.SchemaMap{
+		P: (&schema.Provider{
+			ResourcesMap: schema.ResourceMap{
 				"foo_fizz_buzz":       nil,
 				"foo_bar_hello_world": nil,
 				"foo_bar":             nil,
 			},
-			datasources: map[string]shim.SchemaMap{
+			DataSourcesMap: schema.ResourceMap{
 				"foo_source1":             nil,
 				"foo_very_special_source": nil,
 			},
-		},
+		}).Shim(),
 	}
 
 	makeToken := func(module, name string) (string, error) {
@@ -81,8 +81,8 @@ func TestTokensSingleModule(t *testing.T) {
 
 func TestTokensKnownModules(t *testing.T) {
 	info := tfbridge.ProviderInfo{
-		P: Provider{
-			resources: map[string]shim.SchemaMap{
+		P: (&schema.Provider{
+			ResourcesMap: schema.ResourceMap{
 				"cs101_fizz_buzz_one_five": nil,
 				"cs101_fizz_three":         nil,
 				"cs101_fizz_three_six":     nil,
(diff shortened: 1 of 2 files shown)


aq17 (approved):
LGTM 👍🏼

pkg/tfbridge/x/token_test.go
```diff
@@ -472,46 +473,333 @@ func TestAliasing(t *testing.T) {
...
+			return r.Schema()
+		},
+	}
+}
+
+func (r Resource) Schema() shim.SchemaMap {
+	return r.schema
+}
+
+func newSchemaMap(m map[string]shim.Schema) shim.SchemaMap {
```
t0yv0, line 767: FWIW there's some existing structs implementing shim.SchemaMap in the shim module, not a blocker..

pkg/tfbridge/x/tokens.go
```diff
@@ -472,20 +472,33 @@ type alias[T ~string] struct {
 }
 
 type aliasHistory struct {
-	Resources   map[string]*tokenHistory[tokens.Type]         `json:"resources"`
+	Resources   map[string]*resourceHistory                   `json:"resources"`
```
t0yv0, line 475: So this is a change to a serializable struct that does not look bw-compat.. will it break parsing existing JSON data that was serialized using the previous form?

t0yv0: Ah never mind, it might be bw-compat because tokenHistory[tokens.Type] is embedded.

pkg/tfbridge/x/token_test.go
```diff
@@ -472,46 +473,333 @@ func TestAliasing(t *testing.T) {
...
+
+	info = provider(false, true)
+
+	// Apply metadata back into the provider
+	info.Version = "1.0.0" // New major version
+	err = AutoAliasing(info, metadata)
+	require.NoError(t, err)
+
+	assert.Nil(t, info.Resources["pkg_r1"].Fields["f1"].MaxItemsOne)
+	assert.Nil(t, info.Resources["pkg_r1"].Fields["f2"].MaxItemsOne)
```
t0yv0, line 623: Could you talk me through this one? 

```info = provider(false, true)```

This makes a provider with `info.Resources["pkg_r1"].Fields["f2"].MaxItemsOne == true` before the auto-alias pass. Why does it become nil after the AutoAlias pass? Weren't we supposed to discard history and leave as-is on a new major version?

File: pkg/tfbridge/x/token_test.go [hunk 4]
```diff
@@ -472,46 +473,333 @@ func TestAliasing(t *testing.T) {
...
+	assert.Equal(t, `{
+    "auto-aliasing": {
+        "resources": {
+            "pkg_r1": {
+                "current": "pkg:index/r1:R1",
+                "majorVersion": 1,
+                "fields": {
+                    "f1": {
+                        "maxItemOne": false
```
t0yv0, line 632: Perhaps `omitempty` to skip false ones in the serial form.

t0yv0:
This is looking good, walk is the recursion over three structures that does most of the job it looks like. I had a few questions.

pkg/tfbridge/x/tokens.go
iwahbe: It is backwards compatible via embeddings, but thats good to call out.

pkg/tfbridge/x/token_test.go
iwahbe: The underlying type is a `*bool`. We actually have `omitempty` applied, but we need to keep track of previous values, both true and false.

pkg/tfbridge/x/token_test.go
iwahbe: In this test, we 
1. create a `provider(true, false)`
2. Apply blank metadata on that provider, baking the provider state into `metadata`.
3. Create a new provider with a different `.P` value for `MaxItemsOne` and a new version.
4. We apply the previous `metadata` to the new provider, but `AutoAliasing` discards the old information. Since `AutoAliasing` is applying without previous data, it remembers but doesn't set `info`. (the nil check)

I think your confusion is because `provider(false, true)` does not make a provider with `info.Resources["pkg_r1"].Fields["f2"].MaxItemsOne == true`. It makes a provider with `info.P.resources["pkg_r1"].Fields["f2"].MaxItemsOne == true`, which is different.

t0yv0: Got it. Correct but confusing. The new version wipes the "user overrides" to nil, which is the info.Resources while info.P.Resources is what the provider wants.

t0yv0 (approved):

PR #1008, opened 2023-04-18
Adds a new token mapper: TokensMappedModules

iwahbe:
This allows a user with a known list of desired modules to specify the map and let the mapper drive the process. This is the common case when applying the mapper to existing provider with an existing set of mappings between TF and Pulumi modules.

The major use case is moving larger existing providers onto auto-mapping.

### DIFF
--- a/pkg/tfbridge/x/tokens.go
+++ b/pkg/tfbridge/x/tokens.go
@@ -58,7 +58,9 @@ func TokensSingleModule(
 }
 
 func tokensKnownModules[T b.ResourceInfo | b.DataSourceInfo](
-	prefix, defaultModule string, modules []string, new func(string, string) (*T, error),
+	prefix, defaultModule string, modules []string,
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/x/tokens.go
```diff
@@ -98,15 +100,61 @@ func TokensKnownModules(
...
+	sort.Sort(sort.Reverse(sort.StringSlice(mods)))
+
+	transform := func(tf string) string {
+		s, ok := modules[tf]
+		if !ok && tf == defaultModule {
+			// We pass through the default module as is, so it might not be in
+			// `modules`. We need to catch that and return as is.
+			return tf
+		}
+		assert := "Because any mod selected must be from mods, it is garenteed to be in modules, got %#v"
```
t0yv0, line 136: `guaranteed`

t0yv0 (approved):
Looks simple 🚀

Issue #998, opened 2023-04-17
Cannot reference an attribute marked as secret in schema


frezbo:
### What happened?

Referencing an attribute marked as secret in schema.json produces an error:

```bash
    error: an unhandled error occurred: program failed:
    waiting for RPCs: unexpected secret result returned to invoke call
```

Using the sdk generated from TF PF bridge available here: https://github.com/frezbo/pulumi-provider-talos/tree/chore/update-sdk

https://github.com/frezbo/pulumi-provider-talos/blob/chore/update-sdk/provider/cmd/pulumi-resource-talos/schema.json#L1239

### Expected Behavior

Pulumi should be able to reference types marked as secret

### Steps to reproduce

Example program used:

```go
package main

import (
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
	"github.com/siderolabs/pulumi-provider-talos/sdk/v2/go/talos"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		talosMachineSecrets, err := talos.NewTalosMachineSecrets(ctx, "secret", &talos.TalosMachineSecretsArgs{})
		if err != nil {
			return err
		}

		cpConfig := talos.TalosMachineConfigurationOutput(ctx, talos.TalosMachineConfigurationOutputArgs{
			ClusterEndpoint: pulumi.Sprintf("https://%s:6443", "10.5.0.2"),
			ClusterName:     pulumi.String("talos"),
			MachineType:     pulumi.String("controlplane"),
			MachineSecrets: talos.TalosMachineConfigurationMachineSecretsArgs{
				Certs: talos.TalosMachineConfigurationMachineSecretsCertsArgs{
					Etcd: talos.TalosMachineConfigurationMachineSecretsCertsEtcdArgs{
						Cert: talosMachineSecrets.MachineSecrets.Certs().Etcd().Cert().Elem(),
						Key:  talosMachineSecrets.MachineSecrets.Certs().Etcd().Key().Elem(),
					},
					K8s: talos.TalosMachineConfigurationMachineSecretsCertsK8sArgs{
						Cert: talosMachineSecrets.MachineSecrets.Certs().K8s().Cert().Elem(),
						Key:  talosMachineSecrets.MachineSecrets.Certs().K8s().Key().Elem(),
					},
					K8sAggregator: talos.TalosMachineConfigurationMachineSecretsCertsK8sAggregatorArgs{
[...]

PR #1011, opened 2023-04-19
Fix returning secret values out of Plugin Framework based functions

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/998

The fix works around what appears to be the underlying problem in Pulumi SDKs at least for Go and YAML: https://github.com/pulumi/pulumi/issues/12710

I would love some feedback from platform team on whether we can fix it in a different way before proceeding with the workaround, as the workaround can lose secret-bits that are intentionally marked as such in the schema.

### DIFF
--- a/pf/tests/integration/program_test.go
+++ b/pf/tests/integration/program_test.go
@@ -121,9 +121,7 @@ func ensureTestBridgeProviderCompiled(wd string) error {
 func validateExpectedVsActual(t *testing.T, stack integration.RuntimeValidationStackInfo) {
 	expects := map[string]interface{}{}
 	actuals := map[string]interface{}{}
-	all := []string{}
 	for n, output := range stack.Outputs {
-		all = append(all, n)
 		if strings.HasSuffix(n, "__actual") {
 			actuals[strings.TrimSuffix(n, "__actual")] = output
 		}
--- a/pf/tests/integration/program_test.go
+++ b/pf/tests/integration/program_test.go
@@ -121,7 +121,9 @@ func ensureTestBridgeProviderCompiled(wd string) error {
 func validateExpectedVsActual(t *testing.T, stack integration.RuntimeValidationStackInfo) {
 	expects := map[string]interface{}{}
 	actuals := map[string]interface{}{}
+	all := []string{}
 	for n, output := range stack.Outputs {
+		all = append(all, n)
 		if strings.HasSuffix(n, "__actual") {
 			actuals[strings.TrimSuffix(n, "__actual")] = output
 		}
@@ -129,7 +131,11 @@ func validateExpectedVsActual(t *testing.T, stack integration.RuntimeValidationS
 			expects[strings.TrimSuffix(n, "__expect")] = output
 		}
 	}
+	keys := []string{}
 	for k := range expects {
+		keys = append(keys, k)
+	}
+	for _, k := range keys {
 		k := k
 		t.Run(k, func(t *testing.T) {
 			assert.Equal(t, expects[k], actuals[k])
--- a/pf/tests/integration/program_test.go
+++ b/pf/tests/integration/program_test.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/pulumi/pulumi/pkg/v3/engine"
 	"github.com/pulumi/pulumi/pkg/v3/testing/integration"
+	"sort"
 )
 
 func TestBasicProgram(t *testing.T) {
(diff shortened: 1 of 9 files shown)


pf/tests/testdata/basicprogram/Pulumi.yaml
```diff
@@ -171,11 +171,28 @@ outputs:
...
             port: 22
         protocol: ssh
-  tupleElementAttrRes__actual: ${tupleElementAttrRes.tuplesOptionals}
-  tupleElementAttrRes__expected:
-    - { t0: false, t1: "no" }
-    - { t0: true, t1: "yes"}
+
+  # TODO this is currently failing.
+  # tupleElementAttrRes__actual: ${tupleElementAttrRes.tuplesOptionals}
+  # tupleElementAttrRes__expect:
```
t0yv0, line 177: __expected vs __expect was not being picked up, and this test fails, unfortunately @iwahbe

pf/tests/integration/program_test.go
```diff
@@ -121,15 +121,21 @@ func ensureTestBridgeProviderCompiled(wd string) error {
...
 			actuals[strings.TrimSuffix(n, "__actual")] = output
 		}
 		if strings.HasSuffix(n, "__expect") {
 			expects[strings.TrimSuffix(n, "__expect")] = output
 		}
 	}
+	keys := []string{}
 	for k := range expects {
+		keys = append(keys, k)
+	}
```
iwahbe, line 137: I don't understand the point of `keys`. Should `keys` be sorted?

pf/tests/integration/program_test.go
```diff
@@ -121,15 +121,21 @@ func ensureTestBridgeProviderCompiled(wd string) error {
 func validateExpectedVsActual(t *testing.T, stack integration.RuntimeValidationStackInfo) {
 	expects := map[string]interface{}{}
 	actuals := map[string]interface{}{}
+	all := []string{}
 	for n, output := range stack.Outputs {
+		all = append(all, n)
```
iwahbe, line 126: `all` is created but I don't see where it is used.

pf/tfbridge/provider_invoke.go
```diff
@@ -96,6 +97,12 @@ func (p *provider) readDataSource(ctx context.Context, handle datasourceHandle,
 			handle.terraformDataSourceName, err)
 	}
 
+	// TODO[pulumi/pulumi#12710] consuming programs (at lest in Go and YAML) are unable to accept secrets from an
+	// Invoke response at the moment. Replace secrets with underlying plain values.
+	for k, v := range propertyMap {
+		propertyMap[k] = propertyvalue.RemoveSecrets(v)
```
iwahbe, line 103: This might be too heavy handed, but it feels like we should warn here. Users will expect that secrets can be returned through invokes.
```suggestion
		if propertyValue.ContainsSecrets() {
			warn(invokeName, "Secrets in return value '%s' were dropped")
		}
		propertyMap[k] = propertyvalue.RemoveSecrets(v)
```

CC @AaronFriel

iwahbe (approved):
I think we should consult with the providers team to see if they can perform the fix in the SDK. That said, this fix makes sense.

Does this problem not apply `pkg` based bridged providers?

Also: Pulumi YAML is the Pulumi Go SDK under the hood.

pf/tests/testdata/basicprogram/Pulumi.yaml
iwahbe: Tracked at https://github.com/pulumi/pulumi-terraform-bridge/issues/1012

pf/tfbridge/provider_invoke.go
t0yv0: Yes, we probably should warn.

t0yv0:
I am really curious to trace this through classic bridged providers tomorrow to find out what they do.

t0yv0:
https://github.com/pulumi/pulumi-terraform-bridge/pull/1031 looks like the SDKv2 based providers behave similarly, that is they recognize but ignore secrets passed in. So this is reasonable on bw-compat grounds.

t0yv0:
> I think we should consult with the providers team to see if they can perform the fix in the SDK. That said, this fix makes sense.

Per Justin we need this to work now as a bandaid as the comprehensive fix is not yet scheduled.

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/998

The fix works around what appears to be the underlying problem in Pulumi SDKs at least for Go and YAML: https://github.com/pulumi/pulumi/issues/12710 - before the fix, Invokes that returned secret values caused programs to fail. After the fix, the behavior is similar to SDKv2 bridged providers, secret values in Invoke outputs are replaced with their plain underlying values, which introduces the risk of accidentally storing them un-encrypted in the state but ensures that the programs at least can call the invokes without issue.

Issue #1013, opened 2023-04-19
Moving `cloud flare:index:RuleSet` from SDK to PF changes type names


iwahbe:
### What happened?

Attempting a [partial migration of cloud flare to pf via the muxer](https://github.com/pulumi/pulumi-cloudflare/pull/382), there were unexpected schema diffs. Investigating showed that the diff was caused by changing `Parameters` to `Parameter`. 

### Expected Behavior

No schema diffs without upstream changes.

### Steps to reproduce

See https://github.com/pulumi/pulumi-cloudflare/pull/382

### Output of `pulumi about`

NA

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

iwahbe:
Blocking https://github.com/pulumi/pulumi-cloudflare/pull/382.

PR #1015, opened 2023-04-20
PF nested field MaxItemsOne

iwahbe:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1013

### DIFF
--- a/pf/internal/schemashim/block_schema.go
+++ b/pf/internal/schemashim/block_schema.go
@@ -17,7 +17,6 @@ package schemashim
 import (
 	"fmt"
 
-	pfattr "github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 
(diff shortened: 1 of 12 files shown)


pf/internal/schemashim/block_schema.go
```diff
@@ -70,10 +77,19 @@ func (s *blockSchema) Elem() interface{} {
...
 	switch tt := t.(type) {
 	case types.MapType:
+		if r, ok := retObj(tt.ElemType); ok {
+			return r
+		}
 		schema = newTypeSchema(tt.ElemType, s.block.NestedAttrs())
 	case types.ListType:
+		if r, ok := retObj(tt.ElemType); ok {
+			return r
+		}
```
iwahbe, line 87: This is the change in behavior that fixed the problem.

pkg/tfgen/generate_schema_test.go
```diff
@@ -51,6 +51,15 @@ func TestRegressMiniRandom(t *testing.T) {
...
 }
 
+func TestNestedMaxItemsOne(t *testing.T) {
+	provider := testprovider.ProviderMiniCloudflare()
+	schema, err := GenerateSchema(provider, diag.DefaultSink(io.Discard, io.Discard, diag.FormatOptions{
+		Color: colors.Never,
+	}))
+	assert.NoError(t, err)
+	bridgetesting.AssertEqualsJSONFile(t, "test_data/regress-minicloudflare-schema.json", schema)
+}
```
iwahbe, line 61: This behavior didn't change. I'm just codifying my understanding of the old behavior with a test.

pf/internal/schemashim/block_schema.go
t0yv0: This is great. This is the right fix but I've simplified the function and added some targeted unit tests. There's an additional problem that set-nested blocks were still not working properly.  Consider: https://github.com/pulumi/pulumi-terraform-bridge/pull/1018

t0yv0 (approved):

pkg/tfshim/schema/schema.go
```diff
@@ -168,3 +168,85 @@ func (m SchemaMap) Set(key string, value shim.Schema) {
...
+func FromProvider(p shim.Provider) *Provider {
+	if p == nil {
+		return nil
+	}
+	return &Provider{
+		Schema:         FromSchemaMap(p.Schema()),
+		ResourcesMap:   FromResourceMap(p.ResourcesMap()),
+		DataSourcesMap: FromResourceMap(p.DataSourcesMap()),
+	}
+}
```
iwahbe, line 252: Before I merge, I don't understand where these functions are used. @t0yv0 Can you clarify?

pkg/tfshim/schema/schema.go
```diff
@@ -168,3 +168,85 @@ func (m SchemaMap) Set(key string, value shim.Schema) {
 func (m SchemaMap) Delete(key string) {
 	delete(m, key)
 }
+
+func FromSchema(s shim.Schema) *Schema {
```
t0yv0, line 172: This is not needed. Apologies. Should delete.

PR #1017, opened 2023-04-21
Correctly mux aliases

### LINKED ISSUE #1016: Muxer does not map aliased resources
### What happened?

When computing the mapping for a muxed (hybrid) pf & sdk provider, we don't attribute resources that are aliases back to their originating provider.

### Expected Behavior

Aliasing as called via bridge methods should _just work_™️. If something won't work, it should error during `make tfgen` so it doesn't become end user facing.

### Steps to reproduce

Attempt to serve a hybrid provider with an alias. For example, in pulumi-aws there is code that renames the alb resource with alias:

```

	prov.RenameResourceWithAlias("aws_alb", awsResource(legacyAlbMod, "LoadBalancer"),
		awsResource(albMod, "LoadBalancer"), legacyAlbMod, albMod, &tfbridge.ResourceInfo{
				"load_balancer_type": {Type: awsResource(albMod, "LoadBalancerType")},
				"ip_address_type":    {Type: awsResource(albMod, "IpAddressType")},
			},
			Docs: &tfbridge.DocInfo{Source: "lb.html.markdown"},
		})

```

When using a muxed build of pulumi-aws, schema generation proceeds fine but tests fail with:

```
Resource type 'aws:elasticloadbalancingv2/loadBalancer:LoadBalancer' not found.
```

The error comes from a missing entry in the DispatchTable for this resource.

### Output of `pulumi about`

NA

### Additional context

On bridge version v3.44.3

### Contributing

iwahbe:
Fixes #1016

### DIFF
--- a/pf/internal/muxer/muxer.go
+++ b/pf/internal/muxer/muxer.go
@@ -21,12 +21,14 @@ import (
 
 	"context"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/terraform-plugin-framework/provider"
 	"github.com/pulumi/pulumi-terraform-bridge/pf/internal/schemashim"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 	shimSchema "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
 	"github.com/pulumi/pulumi-terraform-bridge/x/muxer"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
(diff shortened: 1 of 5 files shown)


pf/internal/muxer/muxer.go
```diff
@@ -119,31 +121,88 @@ func newProviderShim(provider shim.Provider) ProviderShim {
...
+}
+
+// Resolve either resources or datasoruces into their originating providers.
+//
+// A resource/datasource is considered "from" a provider if the provider serves a
+// resource/datasource for the token or if the provider serves a resource that matches the
+// resource inserted into the global map of m. The latter logic is used to handle aliases
+// which are inserted via:
+//
+//	p.P.ResourcesMap().Set(legacyResourceName, p.P.ResourcesMap().Get(resourceName))
```
iwahbe, line 160: This is the explanation for how we are recovering information.

pf/internal/muxer/muxer.go
```diff
@@ -119,31 +121,88 @@ func newProviderShim(provider shim.Provider) ProviderShim {
...
+func resolveDispatchMap[T interface{ GetTok() tokens.Token }](
```
iwahbe, line 161: Cool trick: `type Foo interface { Bar() }` is really just saying `Foo` is a type that resolves to `interface { Bar() }`. Whenever you need to supply a type in Go, you can provider either a type name (`Foo`) or a type value `interface { Bar() }`. Even in generics.

t0yv0 (approved):

t0yv0:
Thank you or this 🚢

PR #1018, opened 2023-04-21
Fix block schema recognition for Plugin Framework

t0yv0:
A few more tests and fixes for recognizing schema of Plugin Framework blocks.

### DIFF
--- a/pf/internal/schemashim/convert_type.go
+++ b/pf/internal/schemashim/convert_type.go
@@ -46,8 +46,10 @@ func convertType(typ pfattr.Type) (shim.ValueType, error) {
 		return shim.TypeInvalid, nil
 	default:
 		switch tftype.(type) {
-		case tftypes.List, tftypes.Set:
+		case tftypes.List:
 			return shim.TypeList, nil
+		case tftypes.Set:
+			return shim.TypeSet, nil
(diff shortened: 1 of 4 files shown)


pf/internal/schemashim/convert_type.go
```diff
@@ -46,8 +46,10 @@ func convertType(typ pfattr.Type) (shim.ValueType, error) {
 		return shim.TypeInvalid, nil
 	default:
 		switch tftype.(type) {
-		case tftypes.List, tftypes.Set:
```
t0yv0, line 49: This was a bug, set-nested blocks were incorrectly recognized as lists.

iwahbe (approved):
Thanks for all the additional tests.

pkg/tfshim/schema/schema.go
```diff
@@ -168,3 +168,85 @@ func (m SchemaMap) Set(key string, value shim.Schema) {
 func (m SchemaMap) Delete(key string) {
 	delete(m, key)
 }
+
+func FromSchema(s shim.Schema) *Schema {
```
iwahbe, line 172: Where are the additions to this file used?

t0yv0: Oops I meant to remove it. I didn't end up using this. Was thinking it can help writing test.

PR #1022, opened 2023-04-21
Actually set runtime aliases in the underlying provider

iwahbe:
This PR ensures that the underlying providers are informed of aliases at runtime. https://github.com/pulumi/pulumi-terraform-bridge/pull/1017 ensured that the bridge's mappings dispatched correctly, but did not actually get them working at runtime.

This PR also adds a runtime test to ensure that aliased resources work correctly.

### DIFF
--- a/pf/internal/muxer/muxer.go
+++ b/pf/internal/muxer/muxer.go
@@ -121,6 +121,9 @@ func newProviderShim(provider shim.Provider) ProviderShim {
 
 // Assign each Resource and DataSource mapped in `info` whichever runtime provider defines
 // it.
+//
+// For alias based mappings to work correctly at runtime, it is necessary to call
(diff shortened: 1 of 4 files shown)


pf/internal/muxer/muxer.go
```diff
@@ -194,6 +197,7 @@ func resolveDispatchMap[T interface{ GetTok() tokens.Token }](
 			i, ok := reverseLookupMap[r]
 			if ok {
 				dispatch[string(res.GetTok())] = i
+				mapKind(m.MuxedProviders[i]).Set(tfToken, r)
```
t0yv0, line 200: A little spooky but works!

t0yv0 (approved):
Thanks for incredibly fast turnaround here! Let's try it.

PR #1023, opened 2023-04-21
Permit skipping flaky example generation

t0yv0:
Resolves an AWS build P1: https://github.com/pulumi/pulumi-aws/issues/2477

Add a hook to permit bypassing examples when flaky example generation causes entire provider build to fail, case in point being pulumi-aws builds becoming flaky because of one resource.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -102,6 +102,19 @@ type ProviderInfo struct {
 	MetadataInfo *MetadataInfo
 
 	UpstreamRepoPath string // An optional path that overrides upstream location during docs lookup
+
+	// If set, allows selecting individual examples to skip generating into the PackageSchema (and eventually API
+	// docs). The primary use case for this hook is to ignore problematic or flaky examples temporarily until the
+	// underlying issues are resolved and the examples can be rendered correctly.
+	//
+	// token will be a resource, function, or type token from Pulumi Package Schema. For instance,
+	// "aws:acm/certificate:Certificate" would indicate the example pertains to the Certificate resource in the AWS
+	// provider.
+	//
+	// examplePath will provide even more information on where the example is found. For instance,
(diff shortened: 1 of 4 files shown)


iwahbe (approved):
I'm hesitant to add this, since it allows us to move on without fixing the underlying programgen non-determinism.

I think we can merge if we mark the `SkipExamples` API as experimental. I don't think this is the kind of option we want to support long term.

t0yv0:
The alternative per @thomas11 is to patch upstream AWS, I'm trying that now. 

> I don't think this is the kind of option we want to support long term.

I think it's absolutely necessary long-term in the same way that Skip API is necessary to skip entire resources. A reason to mark it experimental is if we're not sure we get the signature right, which is reasonable, happy to do that.

iwahbe:
> The alternative per @thomas11 is to patch upstream AWS, I'm trying that now.
> 
> > I don't think this is the kind of option we want to support long term.
> 
> I think it's absolutely necessary long-term in the same way that Skip API is necessary to skip entire resources. The underlying machinery is a data pipeline converting examples content, and data pipelines need filters configurable as a way to maintain data quality. A reason to mark it experimental is if we're not sure we get the signature right, which is reasonable, happy to do that.

I understand your point. I want to hold off as long as possible because I want to make sure we don't use a data quality filter to make up for a lack of pipeline quality (which is what we are doing now AFAIK).

I'd argue that we should make sure this API is extensible, so we should pass in an `ExampleFilterArgs` type instead of individual fields. I expect we will want to thread a `context.Context` in at some point, but we don't need that yet.

t0yv0:
>  I want to make sure we don't use a data quality filter to make up for a lack of pipeline quality

Yeah that's very good, just perhaps not something to fix at API level but metrics and process. Let me see if this can interact with examples metrics.

PR #1026, opened 2023-04-22
[Muxer] Respect deeply nested MaxItemsOne properties

iwahbe:
This is a follow up on https://github.com/pulumi/pulumi-terraform-bridge/pull/1015.

### DIFF
--- a/pf/tests/internal/testprovider/cmd/pulumi-resource-testbridge/schema.json
+++ b/pf/tests/internal/testprovider/cmd/pulumi-resource-testbridge/schema.json
@@ -45,9 +45,32 @@
                 },
                 "bic": {
                     "type": "boolean"
+                },
+                "phases": {
+                    "$ref": "#/types/testbridge:index/TestnestRuleActionParametersPhases:TestnestRuleActionParametersPhases"
                 }
             },
-            "type": "object"
+            "type": "object",
+            "required": [
+                "automaticHttpsRewrites",
+                "bic",
+                "phases"
+            ]
+        },
+        "testbridge:index/TestnestRuleActionParametersPhases:TestnestRuleActionParametersPhases": {
+            "properties": {
+                "p1": {
+                    "type": "boolean"
+                },
+                "p2": {
+                    "type": "boolean"
+                }
+            },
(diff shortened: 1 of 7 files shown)


pf/internal/schemashim/type_schema.go
```diff
@@ -56,19 +57,36 @@ func (*typeSchema) Required() bool  { return false }
 func (*typeSchema) Sensitive() bool { return false }
 
 func (s *typeSchema) Elem() interface{} {
+	asObjectType := func(typ any) (shim.Resource, bool) {
```
t0yv0, line 60: This doesn't feel right. typeSchema is in the represents a Schema that's a simple type, not block or attribute, and it's testing it for types.ObjectType on line 69, and recurring into itself. Let me check this PR out and play with this a bit.

t0yv0:
There's an additional bug here that bic and phases marked as required when they're required.

```
            "required": [
                "automaticHttpsRewrites",
                "bic",
                "phases"
            ]
```

pf/tests/internal/testprovider/cmd/pulumi-resource-testbridge/schema.json
```diff
@@ -48,12 +48,31 @@
...
+                    "type": "boolean"
+                },
+                "p2": {
+                    "type": "boolean"
+                }
+            },
+            "type": "object",
+            "required": [
+                "p1",
+                "p2"
```
iwahbe, line 75: These should be optional. I'm not 100% sure if this is an existing bug or if this change introduced a new bug. My guess is existing.

t0yv0: I've got a fix, one sec.

t0yv0:
@iwahbe this passes tests on top of https://github.com/pulumi/pulumi-terraform-bridge/pull/1030 

1030 also fixes the issue with incorrectly "required" properties.

Please take a look!

I suggest we check in your PR for the awesome new test cases, but not the prod code change.

t0yv0 (approved):
Brilliant, let's take this.

PR #1027, opened 2023-04-22
Fix MaxItemsOne Recognition for Deeply Nested Blocks

iwahbe:
Building on the knowledge gained from https://github.com/pulumi/pulumi-terraform-bridge/pull/1026, this change is also necessary. We correctly apply MaxItemsOne aliasing for nested types.

While this change is breaking on master, it only breaks backward-compatibility against https://github.com/pulumi/pulumi-terraform-bridge/pull/1007, which has not been released.

### DIFF
--- a/pkg/tfbridge/x/token_test.go
+++ b/pkg/tfbridge/x/token_test.go
@@ -717,12 +717,14 @@ func TestMaxItemsOneAliasingNested(t *testing.T) {
                     },
                     "f2": {
                         "maxItemsOne": false,
-                        "fields": {
-                            "n1": {
-                                "maxItemsOne": true
-                            },
-                            "n2": {
-                                "maxItemsOne": false
+                        "elem": {
+                            "fields": {
+                                "n1": {
+                                    "maxItemsOne": true
+                                },
+                                "n2": {
+                                    "maxItemsOne": false
+                                }
                             }
                         }
                     }
@@ -740,8 +742,8 @@ func TestMaxItemsOneAliasingNested(t *testing.T) {
 	require.NoError(t, err)
 
 	assert.Equal(t, expected, string(metadata.Marshal()))
-	assert.True(t, *info.Resources["pkg_r1"].Fields["f2"].Fields["n1"].MaxItemsOne)
-	assert.False(t, *info.Resources["pkg_r1"].Fields["f2"].Fields["n2"].MaxItemsOne)
+	assert.True(t, *info.Resources["pkg_r1"].Fields["f2"].Elem.Fields["n1"].MaxItemsOne)
+	assert.False(t, *info.Resources["pkg_r1"].Fields["f2"].Elem.Fields["n2"].MaxItemsOne)
 }
 
 type Schema struct {
--- a/pkg/tfbridge/x/token_test.go
+++ b/pkg/tfbridge/x/token_test.go
@@ -538,10 +538,10 @@ func TestMaxItemsOneAliasing(t *testing.T) {
                 "current": "pkg:index/r1:R1",
                 "fields": {
                     "f1": {
-                        "maxItemOne": true
+                        "maxItemsOne": true
                     },
                     "f2": {
-                        "maxItemOne": false
+                        "maxItemsOne": false
                     }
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/x/token_test.go
```diff
@@ -587,10 +587,10 @@ func TestMaxItemsOneAliasing(t *testing.T) {
                 "current": "pkg:index/r1:R1",
                 "fields": {
                     "f1": {
-                        "maxItemOne": false
+                        "maxItemsOne": false
```
t0yv0, line 590: So this got renamed?

pkg/tfbridge/x/token_test.go
```diff
@@ -713,16 +713,18 @@ func TestMaxItemsOneAliasingNested(t *testing.T) {
...
                     "f2": {
-                        "maxItemOne": false,
-                        "fields": {
-                            "n1": {
-                                "maxItemOne": true
-                            },
-                            "n2": {
-                                "maxItemOne": false
+                        "maxItemsOne": false,
+                        "elem": {
```
t0yv0, line 720: Oof. This is intentional?

pkg/tfbridge/x/tokens.go
```diff
@@ -666,35 +666,44 @@ func aliasResource(
...
 		} else {
 			// There is no history for this value, so we bake it into the
 			// alias history.
-			h.MaxItemOne = b.BoolRef(b.IsMaxItemsOne(schema, info))
+			h.MaxItemsOne = b.BoolRef(b.IsMaxItemsOne(schema, info))
 		}
 		e := schema.Elem()
 		switch e := e.(type) {
-		case shim.Schema:
+		case shim.Resource:
```
t0yv0, line 684: Reordering shim.Resource and shim.Schema actually helps?

pkg/tfbridge/x/tokens.go
```diff
@@ -666,35 +666,44 @@ func aliasResource(
...
 				info.Elem = &b.SchemaInfo{}
 			}
-			walk(h.Elem, info.Elem, e)
-		case shim.Resource:
+
 			safeRange(e, func(k string, v shim.Schema) {
 				walk(
-					getNonNil(&h.Fields, k),
-					getNonNil(&info.Fields, k),
+					getNonNil(&h.Elem.Fields, k),
```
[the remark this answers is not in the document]
iwahbe: Yeah. I had a test case but my misunderstanding of the input was mirrored in my tests. That's why I added an integration test here.

t0yv0 (approved):
Had questions but LGTM - let's make sure this goes on master as a separate PR (I'll merge the base PR first) for the history, it's a separate change. Ty!

pkg/tfbridge/x/token_test.go
iwahbe: Yeah, for 2 reasons:
1. The property is actually called `MaxItemsOne`, so we may as well get it to match.
2. If anyone is using this in the wild (as in test PRs, this is still unreleased), this will prevent old (bad) data from infecting the next run.

pkg/tfbridge/x/token_test.go
iwahbe: Yes. See https://github.com/pulumi/pulumi-terraform-bridge/pull/1027#discussion_r1175653263.

pkg/tfbridge/x/tokens.go
```diff
@@ -666,35 +666,44 @@ func aliasResource(
 
 	var walk func(*fieldHistory, *b.SchemaInfo, shim.Schema)
 	walk = func(h *fieldHistory, info *b.SchemaInfo, schema shim.Schema) {
```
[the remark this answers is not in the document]
iwahbe: This is the comment I've been working from. AFAIK `walk` works correctly with this comment. Tests and practical application also indicate that this works as expected.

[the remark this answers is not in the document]
t0yv0: Yeah it seems to me too. Let's try this! Thank you.

PR #1031, opened 2023-04-24
Add a test for invoke secret handling

t0yv0:


iwahbe (approved):
Thanks for the detailed test/documentation.

PR #1050, opened 2023-04-27
Testing: add secret tests for PF resources

t0yv0:
Test that schema-level resource secrets are preserved in Plugin Framework resources.

### DIFF
--- a/pf/tests/provider_create_test.go
+++ b/pf/tests/provider_create_test.go
@@ -189,11 +189,7 @@ func TestCreateWithSchemaBasedSecrets(t *testing.T) {
 	      "acceptSecrets": true,
 	      "acceptResources": true
 	    },
-	    "response": {
-	      "acceptSecrets": true,
-	      "supportsPreview": true,
-	      "acceptResources": true
-	    }
+	    "response": "*"
 	  },
 	  {
 	    "method": "/pulumirpc.ResourceProvider/Create",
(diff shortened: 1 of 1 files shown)


pf/tests/provider_create_test.go
```diff
@@ -174,3 +174,61 @@ func TestCreateWithFirstClassSecrets(t *testing.T) {
...
+	      "properties": {
+		"length": 10
+	      }
+	    },
+	    "response": {
+	      "id": "none",
+	      "properties": {
+		"__meta": "{\"schema_version\":\"3\"}",
+		"bcryptHash": {
+		  "4dabf18193072939515e22adb298388d": "1b47061264138c4ac30d75fd1eb44270",
```
t0yv0, line 211: This works apparently because convert package is looking at the Schema and doing secret-aware conversion to PropertyValue.

pf/tests/provider_create_test.go
```diff
@@ -174,3 +174,61 @@ func TestCreateWithFirstClassSecrets(t *testing.T) {
...
+	    "request": {
+	      "args": {},
+	      "acceptSecrets": true,
+	      "acceptResources": true
+	    },
+	    "response": {
+	      "acceptSecrets": true,
+	      "supportsPreview": true,
+	      "acceptResources": true
+	    }
```
iwahbe, line 196: The response here actually tests our `newProviderServer`. It doesn't get to code in this repo. The rest is good.

iwahbe (approved):

pf/tests/provider_create_test.go
t0yv0: Yeah, I can mark it "*". Looks like /pulumirpc.ResourceProvider/Configure seeds the plumbing to get ready to send secrets back. We don't care about the response.

Issue #938, opened 2023-03-30
[Plugin Framework] Support populating defaults from DefaultInfo


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

In current interactions of Plugin Framework we do not do any default application as it's hidden beneath Plan Modifiers before version 1.2 of the framework, but this means we also do not do default application specified by bridged provider author specifically customizing DefaultInfo. This seems very useful to support though. We can apply these Pulumi-level defaults, possibly in the Check level method even without the full ability to apply TF-level defaults.

```
	// an optional default directive to be applied if a value is missing.
	Default *DefaultInfo
```

This is also a prerequisite for [AutoName](https://github.com/pulumi/pulumi-terraform-bridge/issues/917)

PR #1052, opened 2023-04-28
feat: Set defined defaults for provider using Terraform Plugin Framework

tmeckel:
This PR contains changes to set defined default values from the provider schema while creating a provider instance.

Closes: #938

### DIFF
--- a/pf/internal/convert/defaults.go
+++ b/pf/internal/convert/defaults.go
@@ -1,0 +1,23 @@
+package convert
+
+import (
+	"os"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+)
+
+func GetDefaultValue(d *tfbridge.DefaultInfo, resource *tfbridge.PulumiResource) (interface{}, error) {
+	if d == nil {
+		return nil, nil
+	}
+	if d.From != nil {
+		return d.From(resource)
+	} else if d.EnvVars != nil {
+		for _, n := range d.EnvVars {
+			if v := os.Getenv(n); v != "" {
+				return v, nil
+			}
+		}
+	}
+	return d.Value, nil
+}
--- a/pf/tfbridge/provider_create.go
+++ b/pf/tfbridge/provider_create.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 
 	"github.com/pulumi/pulumi-terraform-bridge/pf/internal/convert"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 )
 
 // Create allocates a new instance of the provided resource and returns its unique resource ID.
@@ -43,6 +44,23 @@ func (p *provider) CreateWithContext(
 	tfType := rh.schema.Type().TerraformType(ctx).(tftypes.Object)
 
 	priorState := newResourceState(ctx, &rh)
+
+	for key, fld := range rh.pulumiResourceInfo.Fields {
+		if _, ok := checkedInputs[resource.PropertyKey(key)]; !ok && fld.Default != nil {
+			// using default value for empty property
+			v, err := convert.GetDefaultValue(fld.Default, &tfbridge.PulumiResource{
+				URN:        urn,
+				Properties: checkedInputs,
+				Seed:       make([]byte, 0),
+			})
+			if err != nil {
+				return "", nil, 0, err
+			}
+			checkedInputs[resource.PropertyKey(key)] = resource.PropertyValue{
+				V: v,
+			}
+		}
+	}
 
 	checkedInputsValue, err := convert.EncodePropertyMap(rh.encoder, checkedInputs)
 	if err != nil {
--- a/pf/tfgen/not_supported.go
+++ b/pf/tfgen/not_supported.go
@@ -150,7 +150,6 @@ func (u *notSupportedUtil) schema(path string, schema *tfbridge.SchemaInfo) {
 	u.assertIsZero(path+".Elem", schema.Elem)
 	u.fields(path, schema.Fields)
 	u.assertIsZero(path+".Asset", schema.Asset)
-	u.assertIsZero(path+".Default", schema.Default)
 	u.assertIsZero(path+".Stable", schema.Stable)
 	u.assertIsZero(path+".SuppressEmptyMapElements", schema.SuppressEmptyMapElements)
 	u.assertIsZero(path+".ForceNew", schema.ForceNew)


pf/internal/convert/defaults.go
```diff
@@ -0,0 +1,23 @@
...
+
+func GetDefaultValue(d *tfbridge.DefaultInfo, resource *tfbridge.PulumiResource) (interface{}, error) {
+	if d == nil {
+		return nil, nil
+	}
+	if d.From != nil {
+		return d.From(resource)
+	} else if d.EnvVars != nil {
+		for _, n := range d.EnvVars {
+			if v := os.Getenv(n); v != "" {
```
t0yv0, line 17: Should this be os.GetEnv or os.LookupEnv with handling the difference between empty value and no value in the env?

pf/internal/convert/defaults.go
```diff
@@ -0,0 +1,23 @@
+package convert
+
+import (
+	"os"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+)
+
+func GetDefaultValue(d *tfbridge.DefaultInfo, resource *tfbridge.PulumiResource) (interface{}, error) {
```
t0yv0, line 9: I'd not place it in convert package as that package is dedicated to supporting conversions between PropertyValue and tftypes.Value, and this functionality does not touch tftypes.Value.

tmeckel: Is there a place you would like it to be?

pf/internal/convert/defaults.go
tmeckel: Yes, you're right. But I was lazy and copied the code around the EnvVars from the TF Bridge implementation about handling default values. I'll change this 👍🏼

pf/tfbridge/provider_create.go
```diff
@@ -44,6 +45,23 @@ func (p *provider) CreateWithContext(
 
 	priorState := newResourceState(ctx, &rh)
 
+	for key, fld := range rh.pulumiResourceInfo.Fields {
+		if _, ok := checkedInputs[resource.PropertyKey(key)]; !ok && fld.Default != nil {
+			// using default value for empty property
+			v, err := convert.GetDefaultValue(fld.Default, &tfbridge.PulumiResource{
+				URN:        urn,
+				Properties: checkedInputs,
+				Seed:       make([]byte, 0),
```
t0yv0, line 54: This `Seed` bit is probably not right. 

I've looked closely at pkg/tfbridge/provider.go to see how we currently do this in prod.

It looks like this needs to move into `CheckWithContext`. Pulumi calls Check on a resource to find out "final inputs" and it'a good place to apply Pulumi-level defaults; also what pre-PF implementation does. Then Pulumi will pass the result of Check to the Create implementation. So Create does not need to do it again.

You will notice that `randomSeed []byte` parameter is expoed to CheckWithContext, this is what's needed to be passed here.

t0yv0:
Thank you for your contribution @tmeckel ! I'm very excited to have community contributions. Please be patient as we work through the quirks here.

pf/internal/convert/defaults.go
t0yv0: I'd just throw it in a new file in top-level (under `pf/*`), like `pf/defaults.go` we can start with that and reorg later.

tmeckel:
First time I'm contributing to TF Bridge so I'm aware that this PR will not merged instantly and that it will take time to get it right. Espcially because TF Bridge in one of the core components of Pulumi, where you need to take extra precautions.

pf/tfgen/not_supported.go
```diff
@@ -150,7 +150,6 @@ func (u *notSupportedUtil) schema(path string, schema *tfbridge.SchemaInfo) {
 	u.assertIsZero(path+".Elem", schema.Elem)
 	u.fields(path, schema.Fields)
 	u.assertIsZero(path+".Asset", schema.Asset)
-	u.assertIsZero(path+".Default", schema.Default)
```
t0yv0, line 153: This is slightly complicated because there are three places where defaults apply:

Check - this is for resources
Invoke - this is for calling functions
CheckConfig - this is for the provider-level "global" configuration itself

Shall we tackle all of these in the same PR? If not we can move the warnings around so that it wants only for Invoke and CheckConfig.

t0yv0: Actually CheckConfig has some pain I'm working through now so perhaps best left out of this PR.

pf/tfbridge/provider_create.go
```diff
@@ -44,6 +45,23 @@ func (p *provider) CreateWithContext(
...
+			})
+			if err != nil {
+				return "", nil, 0, err
+			}
+			checkedInputs[resource.PropertyKey(key)] = resource.PropertyValue{
```
t0yv0, line 59: There's an implicit conversion here from `any` aka `interface{}` to `resource.PropertyValue`. I think that's not quite right, I'd suggest this.

First cast and check if this is a resource.PropertyValue already, if so use that. This will let the users define precisely what they want if needed.

Otherwise fallback to `resource.NewPropertyValue` conversion. This will take care of converting slices, arrays and so on deeply into the `PropertyValue` representation.

pf/tfbridge/provider_create.go
```diff
@@ -44,6 +45,23 @@ func (p *provider) CreateWithContext(
 
 	priorState := newResourceState(ctx, &rh)
 
+	for key, fld := range rh.pulumiResourceInfo.Fields {
```
t0yv0, line 48: We might want to abstract this into a func as we'll need to use it again.

pf/internal/convert/defaults.go
t0yv0: Eh, the incumbent code uses so perhaps that's what we should keep. "" is missing.

```
func multiEnvDefault(names []string, dv interface{}) interface{} {
	for _, n := range names {
		if v := os.Getenv(n); v != "" {
			return v
		}
	}
	return dv
}

```

t0yv0: Incumbent code does schema-aware parsing for TypeBool, TypeInt, TypeFloat:

https://github.com/pulumi/pulumi-terraform-bridge/blob/9baf57a1b25518606d53a509b10dc492d99ce229/pkg/tfbridge/schema.go#L620-#L620

t0yv0:
The trickiest missing bit will be recursion and path renames. I think we might need to push some helper combinators in first... You see, DefaultInfo can be nested inside SchemaInfo but paths in there refer to Terraform names while this change needs to mostly work in terms of Pulumi names, and those do not match; worse, during maxItems=1 flattening not only path fragments don't match but pulumi paths and tf paths have different length (Pulumi paths shorter).

I was thinking how to write it and it sounds quite messy. I tried adding Renames functionality to compute name maps but it's still clunky in terms of existing API.

Perhaps if we could have API that hides the detail of recursion as well as the detail of name mapping like this:

```
type SchemaVisitor = func(resource.PropertyPath, *SchemaInfo, shim.Schema)
func WalkSchemaMap(map[string]*SchemaInfo{}, shim.SchemaMap, visitor SchemaVisitor) 
func WalkSchema(*SchemaInfo, shim.Schema, visitor SchemaVisitor)
```

This is already in stdlib:

```
func (p PropertyPath) Set(dest, v PropertyValue) bool {
```

Then the task here could be accomplished by doing something like this:

```
        x = resource.NewObjectProperty(inputs)
	WalkSchemaMap(rh.pulumiResourceInfo.Fields, .., func(p, info, _) {
	    p.Set(x, computeDefault(info))
	})
```

CC @iwahbe recently writing walk recursion over these for any additional ideas..

iwahbe:
My recursive walk didn't have to accommodate TF~Pulumi path names, so I'm not sure how much help that will be. I agree that it's worth having an API for iteration separate from the logic for defaults themselves.

pf/internal/convert/defaults.go
tmeckel: Is this really necessary? Because in the newly introduced `SetDefaultValues` I'll use the value from the environment variable (if present) and call `resource.NewPropertyValue` to convert the input value to a proper property value, as you proposed.

pf/tfbridge/provider_create.go
tmeckel: Implemented your proposal in the new `SetDefaultValues`

pf/internal/convert/defaults.go
tmeckel: Moved the code

pf/tfbridge/provider_create.go
tmeckel: Moved the code to `CheckWithContext`

pf/tfbridge/provider_create.go
tmeckel: Introduced `SetDefaultValues`. Not sure about the method signature. Perhaps it's too narrow.

pf/tfgen/not_supported.go
tmeckel: Understand your concerns, but how to only remove the warning for `CheckConfig` what I'm currently try to implement?

pf/internal/convert/defaults.go
tmeckel: @t0yv0 Yep, you're right. I thought `resource.NewPropertyValue` will handle this and convert a `string` to a the target type if they differ. But it does not. That's why the code at https://github.com/pulumi/pulumi-terraform-bridge/blob/9baf57a1b25518606d53a509b10dc492d99ce229/pkg/tfbridge/schema.go#L620-#L620 is doing this.

Issue is that I need the target type to do the conversation and this seems to be very different between PF and the V2 or V1 implementations.

If I understand the code in PF correctly I do the following:

(1) get a `resourceHandle` to the current resource
(2) use `rh.Attr[propertyName].GetType()` to get the type
(3) use cascaded `if` statements to check for the target type

`if tftypes.Is(tftypes.String) {
  ...
} else if tftypes.Is(tftypes.Number) {
} else if tftypes.Is(tftypes.Bool) {
}

image

Is this correct?

tmeckel: Ahhh ... just spotted `AttributeAtPath(context.Context, path.Path) (Attr, diag.Diagnostics)` 😁 
Issue still: what is the path of `sql_auth -> username`? And how can it be "computed" when we walk over the fields recursively?

tmeckel: Came up with something like this

```go
func SetDefaultValues(res *tfbridge.PulumiResource, schema pfutils.Schema, fields map[string]*tfbridge.SchemaInfo) error {
  for key, fld := range fields {
    if _, ok := res.Properties[resource.PropertyKey(key)]; !ok && fld.Default != nil {
      // using default value for empty property
      v, err := getDefaultValue(fld.Default, res)
      if err != nil {
        return err
      }
      if pval, ok := v.(resource.PropertyValue); ok {
        res.Properties[resource.PropertyKey(key)] = pval
      } else {
        attr, ok := schema.Attrs()[key]
        if !ok {
          return fmt.Errorf("failed to get attribute from schema for key %s", key)
        }
        switch attr.GetType().(type) {
        case basetypes.Int64Type:

        case basetypes.Float64Type:

        case basetypes.BoolType:

        case basetypes.StringType:

        default:
          return fmt.Errorf("unable to convert value of type %T to attribute of type %T", v, attr)
        }
        res.Properties[resource.PropertyKey(key)] = resource.NewPropertyValue(v)
      }
    }
  }

  return nil
}
```

t0yv0: I think it would be very annoying to work with AttributeAtPath in this code because Pulumi paths do not map to TF paths 1-1, there's cases of nested renaming and maxitems=1 flattening to be handled. 

I'm working on a WIP PR to find some easy to use API to hide this complexity so we don't need to deal with it here, but it's not quite done yet. Hoping to finish tomorrow.

https://github.com/pulumi/pulumi-terraform-bridge/pull/1065/commits/3a89a67168bc37816f76ccc2a3c75af8eae50dc4#diff-0d24b3420725c5b1914ff5deafbaeee99714952daa359bdc2f8846aec47463f6R34

t0yv0: Regarding env-variable parsing, rh.Attr[propertyName].GetType() and inferring Pulumi type from the PF type works for simple type like int,bool,float, so that should work! However it would be also easy for me to add some plumbing so you get the shim.Type model for these types to work with also.

tmeckel: I realize more and more that I always approach the subject too simply in my mind. If you look at environment variables, then you can assume that only simple values are defined. Or do we also want to parse JSON from a string or something similar? I think that can't be the objective.  But do we need to convert a default value that was simply defined in code?

t0yv0: I think for this PR to go in we only need to match the sophistication of the incumbent v3 bridge which doesn't do objects (JSON parsing) as yet. We can backlog the rest as a feature enhancement.

pf/tfgen/not_supported.go
t0yv0: Sorry for the late reply here, I didn't have time to write sample code; basically you'd need to keep track which structure you're traversing; if resources then remove the warning; if functions or provider-level config then keep it.

t0yv0:
Some suggested changes on https://github.com/pulumi/pulumi-terraform-bridge/pull/1066 but still a little bit of work to do to get this merge-able. Thank you for your patience.

t0yv0:
This is merged now via another PR.

PR #1062, opened 2023-05-01
[Muxer] Update and standardize public API

iwahbe:
This change is necessary because some providers have a different `provider` argument then their `ProviderInfo.Name` value. The previous API assumed they were the same.

### DIFF
--- a/pf/tests/internal/testprovider/cmd/pulumi-resource-muxedrandom/main.go
+++ b/pf/tests/internal/testprovider/cmd/pulumi-resource-muxedrandom/main.go
@@ -27,6 +27,6 @@ import (
 var schema []byte
 
 func main() {
-	tfbridge.MainWithMuxer(context.Background(), schema,
-		testprovider.MuxedRandomProvider())
+	tfbridge.MainWithMuxer(context.Background(), "muxedrandom",
+		testprovider.MuxedRandomProvider(), schema)
 }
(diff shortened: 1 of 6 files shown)


t0yv0:
This can be a fine change but PF also assumes these names are the same.  

So to dig a littler further. DO they really have to be different?

Which providers are affected? 

I checked pulumi-gcp. We have "gcp"  as name but ProviderInfo.Name has "google-beta" - is that the one known use case or do we have more? Should this just be fixed? "google-beta" sounds off, should it be fixed to "gcp" I that provider? I'm wondering what's the "by-design" distinction between the two names. It doesn't seem like we have it documented so it's possible to trip over it.

t0yv0:
Umm, yes, ProviderInfo.Name is documented as this, and is probably legitimately different from the bridged provider name. IN this case LGTM but we'll need the same fix to non-muxed variant of PF providers.

```
	Name          string        // the TF provider name (e.g. terraform-provider-XXXX).
```

t0yv0 (approved):

iwahbe:
@t0yv0 Are you sure that PF also makes this assumption? PF based `Main` functions take a name string:

https://github.com/pulumi/pulumi-terraform-bridge/blob/5513a5246d101e79640e79a942c6dd95deafac31/pf/tfbridge/main.go#L38

https://github.com/pulumi/pulumi-terraform-bridge/blob/5513a5246d101e79640e79a942c6dd95deafac31/pf/tfgen/main.go#L38

t0yv0:
You're right, should have checked. Seems OK! Thank you.

PR #1063, opened 2023-05-01
[Muxer] Allow non disjoint augmented provider

iwahbe:
This is necessary to build `pulumi-gcp` due to backported resources.

This PR is based on https://github.com/pulumi/pulumi-terraform-bridge/pull/1062, but will  merge separately.

### DIFF
--- a/pf/internal/muxer/muxer.go
+++ b/pf/internal/muxer/muxer.go
@@ -45,7 +45,9 @@ func AugmentShimWithPF(ctx context.Context, shim shim.Provider, pf provider.Prov
 	return p
 }
 
-func augmentShimWithPF(ctx context.Context, shim shim.Provider, pf provider.Provider) (*ProviderShim, []string, []string) {
+func augmentShimWithPF(
+	ctx context.Context, shim shim.Provider, pf provider.Provider,
+) (*ProviderShim, []string, []string) {
 	var p ProviderShim
 	if alreadyMerged, ok := shim.(*ProviderShim); ok {
 		p = *alreadyMerged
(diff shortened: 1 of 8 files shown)


pf/internal/muxer/muxer.go
```diff
@@ -37,18 +38,43 @@ func SchemaOnlyPluginFrameworkProvider(ctx context.Context, provider provider.Pr
...
-	contract.AssertNoErrorf(err, "shim and pf provider are not disjoint")
-	return &p
+	r, d := p.extend(schemashim.ShimSchemaOnlyProvider(ctx, pf))
+	return &p, r, d
+}
+
+// AugmentShimWithDisjointPF augments an existing shim with a PF provider.Provider.
+//
+// This function asserts that there is no overlap between providers.
+func AugmentShimWithDisjointPF(ctx context.Context, shim shim.Provider, pf provider.Provider) *ProviderShim {
```
t0yv0, line 65: This new function is in an internal package. Should it be public?


The public one is in a different package.

```
// MuxShimWithPF initializes a shim.Provider that will server resources from both shim and p.
//
// To create a muxed provider, ProviderInfo.P must be the result of this function.
func MuxShimWithPF(ctx context.Context, shim shim.Provider, p pfprovider.Provider) shim.Provider {
	return muxer.AugmentShimWithPF(ctx, shim, p)
}
```

This is the one that needs a new version.

Can have two functions or can have one function with [functional options](https://golang.cafe/blog/golang-functional-options-pattern.html).

The name is still a mouthful as discussed.

t0yv0 (approved):
Approving modulo exposing the old functionality in public so I can use it in pulumi-aws builds.

PR #1064, opened 2023-05-01
Cleanup Empty AutoAliasing Objects

iwahbe:
This PR modifies token mapping to not add extra fields for otherwise empty objects. This is necessary because `AutoName` sets a full `tfbridge.SchemaInfo`, it does not modify one.

### DIFF
--- a/pkg/tfbridge/x/token_test.go
+++ b/pkg/tfbridge/x/token_test.go
@@ -514,6 +514,7 @@ func TestMaxItemsOneAliasing(t *testing.T) {
 					"pkg_r1": (&schema.Resource{Schema: schema.SchemaMap{
 						"f1": Schema{MaxItemsOne: f1},
 						"f2": Schema{MaxItemsOne: f2},
+						"f3": Schema{typ: shim.TypeString},
 					}}).Shim(),
 				},
 			}).Shim(),
@@ -652,8 +653,8 @@ func TestMaxItemsOneAliasingExpiring(t *testing.T) {
 	err = AutoAliasing(info, metadata)
 	require.NoError(t, err)
 
-	assert.Nil(t, info.Resources["pkg_r1"].Fields["f1"].MaxItemsOne)
-	assert.Nil(t, info.Resources["pkg_r1"].Fields["f2"].MaxItemsOne)
+	assert.Nil(t, info.Resources["pkg_r1"].Fields["f1"])
+	assert.Nil(t, info.Resources["pkg_r1"].Fields["f2"])
 	assert.Equal(t, `{
     "auto-aliasing": {
         "resources": {
@@ -746,9 +747,98 @@ func TestMaxItemsOneAliasingNested(t *testing.T) {
 	assert.False(t, *info.Resources["pkg_r1"].Fields["f2"].Elem.Fields["n2"].MaxItemsOne)
 }
 
+// (*ProviderInfo).SetAutonaming skips fields that have a SchemaInfo already defined in
+// their resource's ResourceInfo.Fields. We need to make sure that unless we mark a field
+// as `MaxItemsOne: nonNil` for some non-nil value, we don't leave that field entry behind
+// since that will disable SetAutonaming.
+func TestMaxItemsOneAliasingWithAutoNaming(t *testing.T) {
+	provider := func() *tfbridge.ProviderInfo {
+		prov := &tfbridge.ProviderInfo{
+			P: (&schema.Provider{
+				ResourcesMap: schema.ResourceMap{
+					"pkg_r1": (&schema.Resource{Schema: schema.SchemaMap{
+						"name":      Schema{typ: shim.TypeString},
+						"nest_list": Schema{elem: Schema{typ: shim.TypeBool}},
+						"nest_flat": Schema{
+							elem:        Schema{typ: shim.TypeBool},
+							MaxItemsOne: true,
+						},
+						"override_list": Schema{elem: Schema{typ: shim.TypeBool}},
+						"override_flat": Schema{
+							elem:        Schema{typ: shim.TypeInt},
+							MaxItemsOne: true,
+						},
+					}}).Shim(),
+				},
+			}).Shim(),
+		}
+		err := ComputeDefaults(prov, TokensSingleModule("pkg_", "index", MakeStandardToken("pkg")))
+		require.NoError(t, err)
+		return prov
+	}
+
+	assertExpected := func(t *testing.T, p *tfbridge.ProviderInfo, hist *metadata.Data) {
+		r := p.Resources["pkg_r1"]
+		assert.True(t, r.Fields["name"].Default.AutoNamed)
+
+		assert.Nil(t, r.Fields["nest_list"])
+		assert.Nil(t, r.Fields["override_list"])
+
+		t.Log(string(hist.Marshal()))
+		assert.JSONEq(t, `{
+                "auto-aliasing": {
+                    "resources": {
+                        "pkg_r1": {
+                            "current": "pkg:index/r1:R1",
+                            "fields": {
+                                "nest_flat": {
+                                    "maxItemsOne": true
+                                },
+                                "nest_list": {
+                                    "maxItemsOne": false
+                                },
+                                "override_flat": {
+                                    "maxItemsOne": true
+                                },
+                                "override_list": {
+                                    "maxItemsOne": false
+                                }
+                            }
+                        }
+                    },
+                    "datasources": {}
+                }
+            }`, string(hist.Marshal()))
+	}
+
+	t.Run("auto-named-then-aliased", func(t *testing.T) {
+		p := provider()
+
+		info, err := metadata.New(nil)
+		require.NoError(t, err)
+		p.SetAutonaming(24, "-")
+		err = AutoAliasing(p, info)
+		require.NoError(t, err)
+
+		assertExpected(t, p, info)
+	})
+
+	t.Run("auto-aliased-then-named", func(t *testing.T) {
+		p := provider()
+		info, err := metadata.New(nil)
+		require.NoError(t, err)
+		err = AutoAliasing(p, info)
+		require.NoError(t, err)
+		p.SetAutonaming(24, "-")
+
+		assertExpected(t, p, info)
+	})
+}
+
 type Schema struct {
 	shim.Schema
 	MaxItemsOne bool
+	typ         shim.ValueType
 	elem        any
 }
 
@@ -760,7 +850,13 @@ func (s Schema) MaxItems() int {
 }
 
 func (s Schema) Type() shim.ValueType {
-	return shim.TypeList
+	if s.typ == shim.TypeInvalid {
+		return shim.TypeList
+	}
+	return s.typ
 }
+
+func (s Schema) Optional() bool { return true }
+func (s Schema) Required() bool { return false }
 
 func (s Schema) Elem() any { return s.elem }
--- a/pkg/tfbridge/x/tokens.go
+++ b/pkg/tfbridge/x/tokens.go
@@ -649,8 +649,8 @@ func aliasResource(
 	if res == nil {
 		return
 	}
-
-	safeRange := func(r shim.Resource, f func(string, shim.Schema)) {
+	var walk func(*fieldHistory, *b.SchemaInfo, shim.Schema)
+	walkResource := func(r shim.Resource, hist *map[string]*fieldHistory, info *map[string]*b.SchemaInfo) {
 		if r == nil {
 			return
 		}
@@ -659,12 +659,15 @@ func aliasResource(
 			return
 		}
 		m.Range(func(k string, v shim.Schema) bool {
-			f(k, v)
+			walk(
+				getNonNil(hist, k),
+				getNonNil(info, k),
+				v,
+			)
 			return true
 		})
 	}
 
-	var walk func(*fieldHistory, *b.SchemaInfo, shim.Schema)
 	walk = func(h *fieldHistory, info *b.SchemaInfo, schema shim.Schema) {
 		if schema == nil || (schema.Type() != shim.TypeList && schema.Type() != shim.TypeSet) {
 			// MaxItemsOne does not apply, so do nothing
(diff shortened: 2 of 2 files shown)


t0yv0:
How does this connect to the user problem prompting this PR, would it be possible to elaborate? The tests are at a level of indirection, no elucidating new tests are added,  and it's not obvious how the problem is being fixed.

```
I believe you will recover old behavior by putting p.SetAutonaming before the call to x.AutoAliasing.
```

After this PR do we believe that the order of p.SetAutonaming and x.AutoAliasing does not matter anymore? 

Does editing the existing test not weaken the MaxItems one test strength?

I'm going to be able to review tomorrow morning but it's not an obvious fix to review in a tired state.

pkg/tfbridge/x/tokens.go
```diff
@@ -649,62 +649,96 @@ func aliasResource(
...
+
 			return true
 		})
+
+		return rHasH, rHasI
 	}
 
-	var walk func(*fieldHistory, *b.SchemaInfo, shim.Schema)
-	walk = func(h *fieldHistory, info *b.SchemaInfo, schema shim.Schema) {
+	walk = func(h *fieldHistory, info *b.SchemaInfo, schema shim.Schema) (hasH bool, hasI bool) {
```
guineveresaenger, line 688: I think there is a lot of state context here that gets confusing because the variable names are so similar.
For example, we return `hasH` and `hasI` here but on line 667 we assign them to `keepH` and `keepI` and it would be great to know what is prompting us to "keep" info or history.

pkg/tfbridge/x/tokens.go
```diff
@@ -649,62 +649,96 @@ func aliasResource(
...
+			return hist != nil, info != nil
 		}
+
+		var rHasH, rHasI bool
+
 		m.Range(func(k string, v shim.Schema) bool {
-			f(k, v)
+			h, hasH := getNonNil(hist, k)
+			i, hasI := getNonNil(info, k)
+			keepH, keepI := walk(h, i, v)
```
guineveresaenger, line 667: I'm not sure what is being kept here.

_Should_ we keep history/info?
Are we checking _whether_ history or info are being kept?

pkg/tfbridge/x/tokens.go
```diff
@@ -649,62 +649,96 @@ func aliasResource(
...
+
+			hasH = hasH || keepH
+			hasI = hasI || keepI
+
+			if !hasH {
+				delete(*hist, k)
```
guineveresaenger, line 673: This deletes both the type/field key and its content, correct?

pkg/tfbridge/x/tokens.go
```diff
@@ -713,27 +747,24 @@ func aliasResource(
...
-			getNonNil(&hist[tfToken].Fields, k),
-			getNonNil(&computed.Fields, k),
-			v,
-		)
-	})
+	walkResource(res, &hist[tfToken].Fields, &computed.Fields)
 }
 
-func getNonNil[K comparable, V any](m *map[K]*V, key K) *V {
+func getNonNil[K comparable, V any](m *map[K]*V, key K) (_ *V, alreadyThere bool) {
```
guineveresaenger, line 753: I'm not familiar with this Go pattern, and I could not find any quick reference to it - why are we returning a blank identifier here?

pkg/tfbridge/x/tokens.go
```diff
@@ -713,27 +747,24 @@ func aliasResource(
...
 		*m = map[K]*V{}
 	}
 	v := (*m)[key]
+
 	if v == nil {
 		var new V
 		v = &new
 		(*m)[key] = v
+	} else {
+		alreadyThere = true
```
guineveresaenger, line 765: Where is `alreadyThere` initialized?

guineveresaenger (commented):
I would like some comments that describe the recursion pattern, and the purpose of each inline function.
Can we also link the Docker PR that failed, with an explanation of why that happened?

Thank you for the quick turnaround.

pkg/tfbridge/x/tokens.go
iwahbe: This removes the map entry indexed by the key `k` (`string) from the map `*hist` (`map[string]*fieldHistory`).

Did that answer your question?

pkg/tfbridge/x/tokens.go
iwahbe: The feature I'm using is called [Named return values
](https://go.dev/tour/basics/7). I wanted to name the second returned value, since I often get these types of bool indicators confused. I didn't feel the need to name the first returned value (type `*V`). `_` is a special identifier in Go that can never be referenced.[^1]

[^1]: If you are familiar with lisp, you can conceptually replace each use of `_` with `(gensym)`.

pkg/tfbridge/x/tokens.go
iwahbe: As part of the [named return value](https://go.dev/tour/basics/7).

iwahbe:
> How does this connect to the user problem prompting this PR, would it be possible to elaborate? The tests are at a level of indirection, no elucidating new tests are added, and it's not obvious how the problem is being fixed.
> 
> ```
> I believe you will recover old behavior by putting p.SetAutonaming before the call to x.AutoAliasing.
> ```
> 
> After this PR do we believe that the order of p.SetAutonaming and x.AutoAliasing does not matter anymore?
> 
> Does editing the existing test not weaken the MaxItems one test strength?
> 
> I'm going to be able to review tomorrow morning but it's not an obvious fix to review in a tired state.

After this PR we do believe that `p.SetAutonaming` and `x.AutoAliasing` should be callable in any order.[^1] We want this property to avoid confusion.

The test was more specific than necessary. Specifically, it was testing for a property that was not necessary for the desired behavior that we want to change:

```patch
- assert.Nil(t, info.Resources["pkg_r1"].Fields["f2"].MaxItemsOne)
+ assert.Nil(t, info.Resources["pkg_r1"].Fields["f2"])
```

Previously, `info.Resources["pkg_r1"].Fields["f2"]` was equal to `&tfbridge.SchemaInfo{}`, and we only cared that `info.Resources["pkg_r1"].Fields["f2"].MaxItemsOne` was `nil`. We are tightening the requirements to ensure that the full thing is nil. 

What we actually cared about is this property (expressed in TS style pseudo-code):
```
info?.Resources?["pkg_r1"]?.Fields?["f2"]?.MaxItemsOne == nil
```

Because of how painful it is to express this in go, I abbreviated to asserting `MaxItemsOne` was `nil`.

[^1]: I believe that `SetAutonaming` will break at runtime when it is applied to a non-string field, so I'm discounting the possibility of a list type with auto-naming called `name`.

pkg/tfbridge/x/token_test.go
```diff
@@ -746,9 +747,98 @@ func TestMaxItemsOneAliasingNested(t *testing.T) {
 	assert.False(t, *info.Resources["pkg_r1"].Fields["f2"].Elem.Fields["n2"].MaxItemsOne)
 }
 
+// (*ProviderInfo).SetAutonaming only applies to fields that are not present in their
+// resource's property map. We need to make sure that unless we mark a field as
+// `MaxItemsOne: nonNil` for some non-nil value, we don't leave that field entry behind
+// since that will disable SetAutonaming.
+func TestMaxItemsOneAliasingWithAutoNaming(t *testing.T) {
```
iwahbe, line 754: @guineveresaenger @t0yv0 Sorry for the confusing PR. I have added our motivating example as a test case with an explanation of the underlying logic for the change.

pkg/tfbridge/x/token_test.go
```diff
@@ -514,6 +514,7 @@ func TestMaxItemsOneAliasing(t *testing.T) {
 					"pkg_r1": (&schema.Resource{Schema: schema.SchemaMap{
 						"f1": Schema{MaxItemsOne: f1},
 						"f2": Schema{MaxItemsOne: f2},
+						"f3": Schema{typ: shim.TypeString},
```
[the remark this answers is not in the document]
iwahbe: It was intentional. I wanted to test that we don't record the `MaxItemsOne` state for a field where the concept doesn't apply (type string).

pkg/tfbridge/x/tokens.go
iwahbe: We should keep it when (1) it already existed or (2) we stored useful information in it. I added some comments explaining what we are trying to do that will hopefully clarify.

pkg/tfbridge/x/tokens.go
iwahbe: I added that comment.

t0yv0:
> Because of how painful it is to express this in go, I abbreviated to asserting MaxItemsOne was nil.

Yes that's true. Yes. I think we try not to care anywhere about the different flavors of "missing", not to attach semantics to nil vs missing key here, so what could work is getter functions that isolate the caller from nils, like this:

```
assert.Nil(t, info.Resource("pkg_r1").Field("f2").MaxItemsOne)
```

But perhaps for another day.

pkg/tfbridge/x/tokens.go
t0yv0: I use named return values a bit. They do have a place it seems.

pkg/tfbridge/x/token_test.go
t0yv0: Brilliant. Oof, so  SetAutonaming does look at whether the SchemaInfo is present or no.

Saw this code:

```go
				if _, hasfield := res.Fields[nameProperty]; !hasfield {
					if res.Fields == nil {
						res.Fields = make(map[string]*SchemaInfo)
					}
					res.Fields[nameProperty] = AutoName(nameProperty, maxLength, separator)
				}
```

I would rephrase

> (*ProviderInfo).SetAutonaming only applies to fields that are not present in their
 // resource's property map.

With 

> (*ProviderInfo).SetAutonaming skips fields that have a SchemaInfo already defined in ResourceInfo.Fields


I always confuse the Info things with he provider maps from shim.SchemaMap.

pkg/tfbridge/x/tokens.go
```diff
@@ -650,61 +650,147 @@ func aliasResource(
...
+		populateElem := func() {
 			if h.Elem == nil {
 				h.Elem = &fieldHistory{}
+			} else {
+				hasElemH = true
 			}
 			if info.Elem == nil {
 				info.Elem = &b.SchemaInfo{}
+			} else {
+				hasElemH = true
```
t0yv0, line 763: Is this hasElemH or should it be hasElemI?

t0yv0:
Well this is interesting, it works around the immediate problem but looks a tad error prone. I'm a bit worried about this quirk we found in SetAutonaming behavior, and perhaps there's still conditions where the order of application matters.

Consider this:

- autoaliasing runs and injects a MaxItems=1 specification, creating a SchemaInfo in ResourceInfo.Fields
- then autonaming runs and finds that whoops, the user set ResourceInfo.Fields["spock"] so I'm going to skip it

To the end user this looks like we accidentally disabled auto naming.
  
Could we do better? Could we enforce the order so SetAutonaming is always before aliasing?

jazzyfresh (approved):
LGTM, especially with the new test

iwahbe:
> Well this is interesting, it works around the immediate problem but looks a tad error prone. I'm a bit worried about this quirk we found in SetAutonaming behavior, and perhaps there's still conditions where the order of application matters.


Unfortunately, I think the quirk is here to stay. I've seen plenty of `tk: { Fields: map[string]*tfbridge.SchemaInfo{ "name": {Name: "name" } } }`. AFAIK, this exists only to disable auto-naming. 

AutoNaming should have been more selective by checking if `*SchemaInfo.Defaults` was `nil` instead of `*SchemaInfo`, but I think that change needs to wait for bridge v4.0.0.

> 
> Consider this:
> 
> * autoaliasing runs and injects a MaxItems=1 specification, creating a SchemaInfo in ResourceInfo.Fields
> * then autonaming runs and finds that whoops, the user set ResourceInfo.Fields["spock"] so I'm going to skip it
> 
> To the end user this looks like we accidentally disabled auto naming.

If `AutoAliasing` saw a resource and injected `ResourceInfo["spock"].MaxItemsOne = True()`, then `SetAutonames` would not be affected, since it only effects fields called `"name"`.

If `AutoAliasing` saw a resource and injected `ResourceInfo["name"].MaxItemsOne = True()`, it would imply that `ResourceMap.Get(resourceName).Schema().Get("name").Type` is [either a `List`, `Map` or `Set` ](https://github.com/pulumi/pulumi-terraform-bridge/blob/7747fa11af1899be0b17ec3889662972b2442567/pkg/tfbridge/x/tokens.go#L732)(something that `MaxItemsOne` makes sense on). If that is the case, then `SetAutonaming` would be invalid, since it assumes that the name field is of type `String`.

The end user will never observe `AutoAliasing` disabling a valid auto name.

> Could we do better? Could we enforce the order so SetAutonaming is always before aliasing?

We could enforce the ordering, but I don't think we should.
1. We impose accidental complexity on our users. Since we can't order enforce in the type system, this is a runtime panic when developing a provider. We should avoid this where possible.
2. This cuts down the size of `bridge-metadata.json` to only keep the data it needs. This makes diffs much more meaningful.

t0yv0:
Brilliant, so the cases don't overlap based on type. Autonaming is interested in name type=string and maxitems=1 aliasing is looking at type=list[x] or type=set[x]. That's great. 

> AutoNaming should have been more selective by checking if *SchemaInfo.Defaults was nil instead of *SchemaInfo, but I think that change needs to wait for bridge v4.0.0.

I like that. Care to backlog and label say impact/breaking? We can then find all those things when contemplating 4.0.

t0yv0 (approved):

guineveresaenger (approved):
thank you for the extra comments and improved varnames! 💟

iwahbe:
> Brilliant, so the cases don't overlap based on type. Autonaming is interested in name type=string and maxitems=1 aliasing is looking at type=list[x] or type=set[x]. That's great.
> 
> > AutoNaming should have been more selective by checking if *SchemaInfo.Defaults was nil instead of *SchemaInfo, but I think that change needs to wait for bridge v4.0.0.
> 
> I like that. Care to backlog and label say impact/breaking? We can then find all those things when contemplating 4.0.

Follow up issue is https://github.com/pulumi/pulumi-terraform-bridge/issues/1069.

Issue #1032, opened 2023-04-25
Plugin-framework based providers fail to support non-string configuration properties


t0yv0:
### What happened?

It appears that for historical reasons Pulumi CLI will pass JSON-in-proto encoded provider-level configuration properties into CheckConfig and Configure properties of the provider.  This currently trips up Plugin Framework based providers.

### Expected Behavior

Abile to seamlessly use provider configuration values regardless of their type.

### Steps to reproduce

Running tests in pulumi-aws in Node yields (TestAccAlbNew):

```
 error: cannot encode provider configuration to call ConfigureProvider: objectEncoder failed on property "skip_credentials_validation": Expected a Boolean

 What else is blocking?
   One of the resources is skipped.


error: Error: invocation of aws:index/getAvailabilityZones:getAvailabilityZones returned an error:
cannot encode provider configuration to call ConfigureProvider:
objectEncoder failed on property "skip_metadata_api_check": Expected a Boolean
```

Debugging in the PULUMI_DEBUG_GRPC log we observe:

```
{
  "method": "/pulumirpc.ResourceProvider/Configure",
  "request": {
    "variables": {
      "aws:config:region": "us-west-2",
      "aws:config:skipCredentialsValidation": "false",
      "aws:config:skipMetadataApiCheck": "true",
      "aws:config:skipRegionValidation": "true"
    },
    "args": {
      "region": "us-west-2",
      "skipCredentialsValidation": "false",
      "skipMetadataApiCheck": "true",
      "skipRegionValidation": "true",
      "version": "5.36.0-alpha.1682372726+8747b20d.dirty"
    },
    "acceptSecrets": true,
    "acceptResources": true
  },
  "response": {},
  "metadata": {
    "kind": "resource",
    "mode": "client",
    "name": "aws"
  }
}
```

So Configure call actually fails because it's tripped up by "true" instead of `true`.

The program simply did this:

```
const providerOpts = { provider: new aws.Provider("prov", { region: <aws.Region>config.require("envRegion") }) };
```

But node SDK injected default true values.

### Output of `pulumi about`

N/A

bridge version: v0.8.1-0.20230425132252-0293ca674084

### Additional context

Recent PRs making fixes to pre-Plugin Framework providers have good tests and relevant code for handling JSON-in-proto encoding.

https://github.com/pulumi/pulumi-terraform-bridge/pull/991
https://github.com/pulumi/pulumi-terraform-bridge/pull/1002

### Contributing

[...]

PR #1065, opened 2023-05-01
Enable non-string config values for Plugin Framework

t0yv0:
On top of https://github.com/pulumi/pulumi-terraform-bridge/pull/1086

Fixes https://github.com/pulumi/pulumi-terraform-bridge/pull/1032

The work on getting ConfigEncoding right in the v3 version of the bridge is here reused for the pf version of the bridge.

### DIFF
--- a/pf/internal/plugin/provider_server.go
+++ b/pf/internal/plugin/provider_server.go
@@ -28,6 +28,8 @@ import (
 	pl "github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 )
 
 type providerServer struct {
@@ -36,10 +38,18 @@ type providerServer struct {
 	provider      ProviderWithContext
 	keepSecrets   bool
 	keepResources bool
+
+	configEncoding tfbridge.ConfigEncoding
 }
 
-func NewProviderServerWithContext(provider ProviderWithContext) pulumirpc.ResourceProviderServer {
-	return &providerServer{provider: provider}
+func NewProviderServerWithContext(
+	provider ProviderWithContext,
+	configEncoding *tfbridge.ConfigEncoding,
+) pulumirpc.ResourceProviderServer {
+	return &providerServer{
+		provider:       provider,
+		configEncoding: *configEncoding,
+	}
 }
 
 func (p *providerServer) unmarshalOptions(label string) pl.MarshalOptions {
@@ -175,14 +185,14 @@ func (p *providerServer) CheckConfig(ctx context.Context,
 ) (*pulumirpc.CheckResponse, error) {
 	urn := resource.URN(req.GetUrn())
 
-	state, err := pl.UnmarshalProperties(req.GetOlds(), p.unmarshalOptions("olds"))
+	state, err := p.configEncoding.UnmarshalProperties(req.GetOlds())
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("CheckConfig failed to unmarshal olds: %w", err)
 	}
 
-	inputs, err := pl.UnmarshalProperties(req.GetNews(), p.unmarshalOptions("news"))
+	inputs, err := p.configEncoding.UnmarshalProperties(req.GetNews())
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("CheckConfig failed to unmarshal news: %w", err)
 	}
 
 	newInputs, failures, err := p.provider.CheckConfigWithContext(ctx, urn, state, inputs, true)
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -55,6 +55,7 @@ require (
 	golang.org/x/text v0.8.0
 	google.golang.org/grpc v1.54.0
 	google.golang.org/protobuf v1.29.1
+	pgregory.net/rapid v0.5.5
```
t0yv0, line 58: This is because of property value helper lib tests moving under unstable/ for reuse.

pf/internal/plugin/provider_server.go
```diff
@@ -36,10 +38,18 @@ type providerServer struct {
 	provider      ProviderWithContext
 	keepSecrets   bool
 	keepResources bool
+
+	configEncoding tfbridge.ConfigEncoding
 }
 
-func NewProviderServerWithContext(provider ProviderWithContext) pulumirpc.ResourceProviderServer {
-	return &providerServer{provider: provider}
+func NewProviderServerWithContext(
```
t0yv0, line 45: Once we started modifying this plumbing it's tempting to modify further; installed config encoding here.

pkg/tfbridge/config_encoding.go
```diff
@@ -22,41 +22,35 @@ import (
...
 
-type configEncoding struct {
-	fieldTypes      map[resource.PropertyKey]shim.ValueType
-	sensitiveFields map[resource.PropertyKey]struct{}
+type ConfigEncoding struct {
+	fieldTypes map[resource.PropertyKey]shim.ValueType
 }
 
-func newConfigEncoding(config shim.SchemaMap, configInfos map[string]*SchemaInfo) *configEncoding {
+func NewConfigEncoding(config shim.SchemaMap, configInfos map[string]*SchemaInfo) *ConfigEncoding {
```
t0yv0, line 34: Public to reuse from Plugin Framework.

pkg/tfbridge/config_encoding.go
```diff
@@ -65,16 +59,7 @@ func (*configEncoding) convertStringToPropertyValue(s string, typ shim.ValueType
 	// Otherwise, we will attempt to deserialize the input string as JSON and convert the result into a Pulumi
 	// property. If the input string is empty, we will return an appropriate zero value.
 	if s == "" {
-		switch typ {
```
t0yv0, line 68: Lifting as is to zeroValue()

pkg/tfbridge/config_encoding.go
```diff
@@ -84,50 +69,35 @@ func (*configEncoding) convertStringToPropertyValue(s string, typ shim.ValueType
...
-	copy := make(resource.PropertyMap)
-	for k, v := range pm {
-		_, secret := secrets[k]
-		if secret {
-			copy[k] = resource.MakeSecret(v)
-		} else {
-			copy[k] = v
-		}
+func (enc *ConfigEncoding) unmarshalOpts() plugin.MarshalOptions {
+	return plugin.MarshalOptions{
```
t0yv0, line 86: Inline MarshalOptions into ConfigEncoding as it's not really configurable but does a specific thing.

pkg/tfbridge/config_encoding.go
```diff
@@ -164,12 +134,22 @@ func (enc *configEncoding) UnmarshalPropertyValue(key resource.PropertyKey, v *s
 		return &v, nil
 	}
 
+	// Computed sentinels are coming in as always having an empty string, but the encoding coerses them to a zero
```
t0yv0, line 137: This might someday help avoid typing mistakes in code that parses PropertyValue but for now it's not terribly important; if something's a Computed it's not looked at what's inside.

pkg/tfbridge/config_encoding_test.go
```diff
@@ -0,0 +1,385 @@
...
+	structpb "google.golang.org/protobuf/types/known/structpb"
+
+	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	sch "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConfigEncoding(t *testing.T) {
```
t0yv0, line 30: This code caused trouble so important to test it out..

pkg/tfbridge/provider_test.go
```diff
@@ -960,6 +960,8 @@ func TestCheckConfig(t *testing.T) {
 	// Secrets: it is important to handle secrets correctly since the results of CheckConfig are stored in the
 	// state. If secrets are not marked as such, they are stored as plaintext in the state.
 	t.Run("preserve_program_secrets", func(t *testing.T) {
+		t.Skip()
```
t0yv0, line 963: This is important - I don't think the bridge needs to do this, since it responds to the engine with "acceptSecrets: false" and can rely on the engine's heuristic to take care of the secrets.

pkg/tfbridge/secrets.go
```diff
@@ -12,34 +12,32 @@
...
-	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/walk"
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 )
 
 // Ensures resource.MakeSecret is used to wrap any nested values that correspond to secret properties in the schema. A
 // property is considered secret if it is declared Sensitive in the SchemaMap in the upstream provider. Users may also
 // override a matching SchemaInfo.Secret setting to force a property to be considered secret or non-secret.
-func MarkSchemaSecrets(ctx context.Context, schemaMap shim.SchemaMap, configInfos map[string]*tfbridge.SchemaInfo,
+func MarkSchemaSecrets(ctx context.Context, schemaMap shim.SchemaMap, configInfos map[string]*SchemaInfo,
```
t0yv0, line 27: This had to move back under tfbridge package due to limitations of Go complaining about recursive package dependencies.

iwahbe (approved):

PR #1066, opened 2023-05-01
Support DefaultInfo for PluginFramework providers

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/938

Implements support for DefaultInfo default value overrides for Plugin Framework based providers. These default value overrides are applied at Pulumi level before passing the data to Terraform. It is a popular feature and having it enabled for Plugin Framework simplifies migrating already providers to new Plugin Framework versions.

Unlike the implementation in v3 tfbridge, the functions only apply the default values from DefaultInfo overrides, and do not attempt to mimic or interact with Terraform level default application process, which is handled internally by TF plugin server code outside of this repository.

Co-authored by @tmeckel (#1052).


pf/internal/defaults.go
```diff
@@ -1,42 +1,139 @@
...
+		if fld == nil || fld.Default == nil {
+			continue
+		}
+		fieldSchema, knownField := resShim.Schema().GetOk(key)
+		if !knownField {
+			continue
+		}
 
-	return nil
+		pulumiName := tfbridge.TerraformToPulumiNameV2(key, resShim.Schema(), fields)
```
t0yv0, line 122: @tmeckel this name translation bit is unfortunately very important, otherwise foo_bar and fooBar names are confused and there's map mismatches.

pf/internal/defaults.go
```diff
@@ -1,42 +1,139 @@
...
+}
+
+func recoverDefaultValue(defaultValue any) resource.PropertyValue {
+	if pv, alreadyPV := defaultValue.(resource.PropertyValue); alreadyPV {
+		return pv
 	}
+	return resource.NewPropertyValue(defaultValue)
+}
+
+// TODO this needs to handle the nested case, setting defaults for properties of object-typed properties.
```
t0yv0, line 107: I'm working on some combinators here to make it easy to handle the recursive case, in a separate PR.

pf/internal/defaults.go
tmeckel: @t0yv0 shall we get in sync with the naming of tfbridge code and use `applyDefaults` as the function name?

t0yv0: applyDefaults does a lot more , this one applies only bridge-specified defaults from DefaultInfo; the other one tries to implement TF behavior of applying TF-specified defaults. So a new name is warranted but naming is hard.

pf/internal/defaults.go
t0yv0: You wouldn't be able to do it correctly so quickly because the codebase doesn't make it obvious how to deal with maxItems=1 flattening and Pulumi-terraform name mangling. I've made progress on https://github.com/pulumi/pulumi-terraform-bridge/pull/1065 and I'll try to get it in piecemeal shortly then we can use some of the same combinators to solve this as well. For example we can walk PropertyValue keeping track of PropertyPath and then use combinators from that PR to translate that to SchemaPath and lookup corresponding schema information.

t0yv0:
> @t0yv0 Kinda confused now 😁 Is my original PR https://github.com/pulumi/pulumi-terraform-bridge/pull/1052 now obsolete?

You can merge these changes into your PR if you like.. I wasn't able to base it on this PR since there's something odd with in-repo PRs vs forks, I've not used forks recently as much, need to brush up on the workflow.

pf/internal/defaults/defaults.go
```diff
@@ -0,0 +1,293 @@
...
+	TopSchemaMap shim.SchemaMap
+
+	// Toplevel SchemaInfo configuration matching TopSchemaMap.
+	TopFieldInfos map[string]*tfbridge.SchemaInfo
+
+	// Optional. Note that ResourceInstance need not be specified for Invoke or Configure processing.
+	ResourceInstance *tfbridge.PulumiResource
+
+	// Optional. If known, these are the provider-level configuration values, to support DefaultInfo.Config.
+	ProviderConfig resource.PropertyMap
```
iwahbe, line 47: `PropertyMap` is required. `ResourceInstance` and `ProviderConfig` are optional. What about `TopSchemaMap` and `TopFieldInfos`?

Why are they `Top*`?

pf/internal/defaults/defaults.go
```diff
@@ -0,0 +1,293 @@
...
+	t := &defaultsTransform{
+		resourceInstance: args.ResourceInstance,
+		topSchemaMap:     args.TopSchemaMap,
+		topFieldInfos:    args.TopFieldInfos,
+		providerConfig:   args.ProviderConfig,
+	}
+	result := t.withDefaults(ctx, make(resource.PropertyPath, 0), resource.NewObjectProperty(args.PropertyMap))
+	if !result.IsObject() {
+		contract.Failf("defaultsTransform.withDefaults returned a non-object value")
+	}
```
iwahbe, line 71:
```suggestion
	contract.Assertf(result.IsObject(), "defaultsTransform.withDefaults returned a non-object value")
```

pf/internal/defaults/defaults.go
```diff
@@ -0,0 +1,293 @@
...
+		raw, err := defaultInfo.From(res)
+		if err != nil {
+			return na, false, err
+		}
+		if raw == nil {
+			return na, false, nil
+		}
+		return recoverDefaultValue(raw), true, nil
+
+	} else if defaultInfo.EnvVars != nil {
```
iwahbe, line 97: Both for clarity and to match https://github.com/pulumi/pulumi-terraform-bridge/blob/16f3cd4597abb67da5b9a213e5e216e603fb3b26/pkg/tfbridge/schema.go#L617.
```suggestion
	} else if len(defaultInfo.EnvVars) != 0 {
```

pf/internal/defaults/defaults.go
```diff
@@ -0,0 +1,293 @@
...
+	if defaultInfo.From != nil && res != nil {
+		raw, err := defaultInfo.From(res)
+		if err != nil {
+			return na, false, err
+		}
+		if raw == nil {
+			return na, false, nil
+		}
+		return recoverDefaultValue(raw), true, nil
+
```
iwahbe, line 96: It seems like this goes at the end in priority:

https://github.com/pulumi/pulumi-terraform-bridge/blob/16f3cd4597abb67da5b9a213e5e216e603fb3b26/pkg/tfbridge/schema.go#L617

pf/internal/defaults/defaults.go
```diff
@@ -0,0 +1,293 @@
...
+	case shim.TypeBool:
+		v := false
+		if str != "" {
+			if v, err = strconv.ParseBool(str); err != nil {
+				return resource.NewNullProperty(), err
+			}
+		}
+		return resource.NewBoolProperty(v), nil
+	case shim.TypeInt:
+		v := int(0)
```
iwahbe, line 140: This suggests that the `0` is a default value instead of possessing special meaning.
```suggestion
		var v int
```

pf/internal/defaults/defaults.go
```diff
@@ -0,0 +1,293 @@
...
+		}
+	}
+
+	if defaultInfo.Value != nil {
+		return recoverDefaultValue(defaultInfo.Value), true, nil
+	}
+	return na, false, nil
+}
+
+func parseValueFromEnv(sch shim.Schema, str string) (resource.PropertyValue, error) {
```
iwahbe, line 128: This interprets `parseValueFromEnv(shim.TypeNumber, "")` as `0.0`. It feels like this should be an error or unset, but definitely not a zero value.

pf/internal/defaults/defaults.go
```diff
@@ -0,0 +1,293 @@
...
+func (du *defaultsTransform) extendPropertyMapWithDefaults(
+	ctx context.Context,
+	path resource.PropertyPath,
+	props resource.PropertyMap,
+) (resource.PropertyMap, error) {
+	schemaMap, infos, ok := du.lookupSchemaByContext(path)
+	if !ok {
+		return props, nil
+	}
+	res := props.Copy()
```
iwahbe, line 232: This is not a deep copy. It looks like this is fine, but I wanted to make sure it wasn't relied upon.

https://github.com/pulumi/pulumi/blob/ae247a81951b1eeb8b43500fdc13cc95d16be433/sdk/go/common/resource/properties.go#L195-L203

pf/internal/defaults/defaults.go
```diff
@@ -0,0 +1,293 @@
...
+			tflog.Trace(ctx, "Ignoring an error in extendPropertyMapWithDefaults",
+				map[string]any{"err": err.Error()})
+			value = resource.NewObjectProperty(pm)
+		} else {
+			value = resource.NewObjectProperty(extended)
+		}
+		return value, nil
+	}
+	transformed, err := propertyvalue.TransformPropertyValue(path, tr, value)
+	contract.AssertNoErrorf(err, "TransformPropertyValue should not return errors here")
```
iwahbe, line 291: Why not?

pf/internal/defaults/defaults.go
```diff
@@ -0,0 +1,293 @@
...
+) resource.PropertyValue {
+	tr := func(path resource.PropertyPath, value resource.PropertyValue) (resource.PropertyValue, error) {
+		if !value.IsObject() {
+			return value, nil
+		}
+		pm := value.ObjectValue()
+		// After recurring on the elements, try to apply defaults here.
+		if extended, err := du.extendPropertyMapWithDefaults(ctx, path, pm); err != nil {
+			tflog.Trace(ctx, "Ignoring an error in extendPropertyMapWithDefaults",
+				map[string]any{"err": err.Error()})
```
iwahbe, line 283: Does pkg/tfbrige log errors and continue on when they fail to parse values?

t0yv0: It fails e.g. `strconv.ParseBool: parsing "FOOBAR": invalid syntax` we can match that here.

pf/internal/defaults/defaults.go
t0yv0: Inside the module I was trying not to get confused with the recursive shim.SchemaMap and map[string]*tfbridge.SchemaInfo being looked up for the current object context (at path /x/y/z), vs the top-level (at path /) schema from which the lookups start. It's not really relevant for the consumers of the module though so I can edit that out.

pf/internal/defaults/defaults.go
t0yv0: Surprisingly in https://github.com/pulumi/pulumi-terraform-bridge/blob/a6742438179a39da5598e5ec0860d42dce51906e/pkg/tfbridge/schema.go#L617-#L617 we do EnvVars, then Config, then Value, then From. This kind of doesn't make sense as maybe a combination of EnvVars and Value could be useful. But I can copy the order from there for now.

pf/internal/defaults/defaults.go
t0yv0: Yeah, it's fine, we want exactly a shallow copy here, and we're filling out missing fields in an object. The deep recursion is taken in propertyvalue.TransformPropertyValue so this can stay at one level.

pf/internal/defaults/defaults.go
t0yv0: Because `tr` always returns `err=nil`, so this can't error.

pf/internal/defaults/defaults.go
t0yv0: I've tinkered a bit and decided to flush these down as `tflog.Error` - this should surface nicely in the engine, without stopping the program as in a catastrophic error; nice advantage is that if you have many of these, the user can be exposed to several error messages at once, instead of debugging one by one.

pf/internal/defaults/defaults.go
t0yv0: You're right; I've looked at prod code and it dismisses env vars that are set by empty; copying this behavior for now. Can add a quick test.

pf/internal/defaults/defaults.go
t0yv0: That code is strange. It does in fact fallback to support EnvVars and Value, but not Config and Value. I landed ons something slightly more permissive.

t0yv0:
Ready for another round @iwahbe

pf/internal/defaults/defaults.go
iwahbe: That is super helpful. Could we put that logic in the `Assertf`?

t0yv0:
That is https://github.com/pulumi/pulumi-terraform-bridge/issues/1095

pf/internal/defaults/defaults.go
```diff
@@ -0,0 +1,307 @@
...
+						"key":      defaultInfo.Config,
+						"property": string(property),
+					})
+				return pv, true
+			}
+		}
+	}
+
+	// Unlike v3/tfbridge the code here still falls back to Value or From specs if Config did not match, this seems
+	// resaonable to do.
```
iwahbe, line 124: I agree that this is reasonable for new providers, but I'm concerned this will bite us when switching over a resource from `/pkg` to `/pf`.

If the resource was depending on the old behavior, switching over will cause a diff.

pf/internal/defaults/defaults.go
iwahbe: Since we are no longer tolerating empty `str` values, let's (1) remove the `if str != ""` checks and (2) add a `contract.Assertf(str != "", "empty env vars should have been skipped")` at the top of the function. This makes the intended behavior of the function much easer to read at a glance.

iwahbe (approved):
One last comments. Otherwise 🚢🚢🚢!

pf/internal/defaults/defaults.go
t0yv0: That'd require someone specifying both .Config and .Value. I think it should be super rare. But I can match the old behavior.

PR #1068, opened 2023-05-02
Ensure replay tests use exact same server as prod code

t0yv0:
We started making changes to provider_server.go to customize the translation of messages from protobuf layer talking the ResourceProvider Pulumi gRPC protocol to the resource.PropertyValue layer. In particular we updated Configure to signal to the engine that the bridged providers do not accept secrets. The tests did not reflect this due to a peculiarity - the tests were not using the same gRPC server. This is now fixed in the PR so the tests are representative of real use.

### DIFF
--- a/pf/tfbridge/provider.go
+++ b/pf/tfbridge/provider.go
@@ -148,6 +148,7 @@ func newProviderWithContext(ctx context.Context, info ProviderInfo,
 	}, nil
 }
 
+// Internal. The signature of this function can change between major releases. Exposed to facilitate testing.
 func NewProviderServer(
 	ctx context.Context,
 	logSink logutils.LogSink,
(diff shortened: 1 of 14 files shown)


pf/tfbridge/provider.go
```diff
@@ -148,7 +148,7 @@ func newProviderWithContext(ctx context.Context, info ProviderInfo,
 	}, nil
 }
 
-func newProviderServer(
+func NewProviderServer(
```
t0yv0, line 151: This is exposed to facilitate testing as tests live in another go module to not pollute the main module with dependencies of test resource providers such as the random provider. I think this may be OK to expose but we owe a doc string later on.

iwahbe, line 151: Should we mark this as an unstable API. This isn't really intended for end users to consume AFAIK.

iwahbe (approved):
Thanks!

PR #1075, opened 2023-05-03
Implement tfshim addressing and tree-walking facilities

t0yv0:
Add walk.SchemaPath to represent locations within a Schema tree.

Add LookupSchemaPath to find a nested Schema.

Add VisitSchema to traverse the tree with a visitor.

These functions take care of object type encodigs, so schema paths naturally correspond to value paths for TF-based values. They are intended to use later on to simplify some recursive traversal code and reporting error locations.

### DIFF
--- a/pkg/tfshim/walk/walk.go
+++ b/pkg/tfshim/walk/walk.go
@@ -22,7 +22,7 @@ import (
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
 )
 
-// Represents locations in a Schema value as a sequence of steps to locate it.
+// Represents locations in a tfshim.Schema value as a sequence of steps to locate it.
 //
 // An empty SchemaPath represents the current location.
 //
--- a/pkg/tfshim/walk/walk.go
+++ b/pkg/tfshim/walk/walk.go
@@ -170,15 +170,15 @@ type SchemaVisitor = func(SchemaPath, shim.Schema)
 
 // Visit all nested schemas, including the current one.
 func VisitSchema(schema shim.Schema, visitor SchemaVisitor) {
-	walkSchemaInner(NewSchemaPath(), schema, visitor)
+	visitSchemaInner(NewSchemaPath(), schema, visitor)
 }
 
 // Visit all nested schemas in a SchemaMap, keeping track of SchemaPath location.
 func VisitSchemaMap(schemaMap shim.SchemaMap, visitor SchemaVisitor) {
-	walkSchemaMapInner(NewSchemaPath(), schemaMap, visitor)
+	visitSchemaMapInner(NewSchemaPath(), schemaMap, visitor)
 }
 
-func walkSchemaInner(path SchemaPath, schema shim.Schema, visitor SchemaVisitor) {
+func visitSchemaInner(path SchemaPath, schema shim.Schema, visitor SchemaVisitor) {
 	visitor(path, schema)
 	switch elem := schema.Elem().(type) {
 	case shim.Resource:
@@ -190,15 +190,15 @@ func walkSchemaInner(path SchemaPath, schema shim.Schema, visitor SchemaVisitor)
 		} else {
 			nestedPath = path.Element()
 		}
-		walkSchemaMapInner(nestedPath, elem.Schema(), visitor)
+		visitSchemaMapInner(nestedPath, elem.Schema(), visitor)
 	case shim.Schema:
-		walkSchemaInner(path.Element(), elem, visitor)
+		visitSchemaInner(path.Element(), elem, visitor)
 	}
 }
 
-func walkSchemaMapInner(path SchemaPath, schemaMap shim.SchemaMap, visitor SchemaVisitor) {
+func visitSchemaMapInner(path SchemaPath, schemaMap shim.SchemaMap, visitor SchemaVisitor) {
 	schemaMap.Range(func(key string, schema shim.Schema) bool {
-		walkSchemaInner(path.GetAttr(key), schema, visitor)
+		visitSchemaInner(path.GetAttr(key), schema, visitor)
 		return true
 	})
 }
--- a/pkg/tfshim/walk/walk.go
+++ b/pkg/tfshim/walk/walk.go
@@ -29,7 +29,7 @@ import (
 // Values of this type are immutable by convention, use Copy as necessary for local mutations.
 type SchemaPath []SchemaPathStep
 
-func (p SchemaPath) String() string {
+func (p SchemaPath) GoString() string {
 	parts := []string{"walk", "NewSchemaPath()"}
 	for _, e := range p {
 		switch ee := e.(type) {
@@ -79,7 +79,7 @@ func LookupSchemaPath(path SchemaPath, schema shim.Schema) (shim.Schema, error) 
 		}
 		nextResult, err := p[0].Lookup(result)
 		if err != nil {
-			return nil, fmt.Errorf("LookupSchemaPath failed at %s: %w", current, err)
+			return nil, fmt.Errorf("LookupSchemaPath failed at %s: %w", current.GoString(), err)
 		}
 		result, p, current = nextResult, p[1:], current.WithStep(p[0])
 	}
@@ -94,7 +94,7 @@ func LookupSchemaMapPath(path SchemaPath, schemaMap shim.SchemaMap) (shim.Schema
 //
 // This interface is closed, the only the implementations given in the current package are allowed.
 type SchemaPathStep interface {
-	String() string
+	GoString() string
 	Lookup(shim.Schema) (shim.Schema, error)
 }
 
@@ -103,7 +103,7 @@ type GetAttrStep struct {
 	Name string
 }
 
-func (step GetAttrStep) String() string {
+func (step GetAttrStep) GoString() string {
 	return fmt.Sprintf("walk.GetAttrStep{%q}", step.Name)
 }
 
@@ -111,17 +111,17 @@ func (step GetAttrStep) Lookup(s shim.Schema) (shim.Schema, error) {
 	if sm, ok := UnwrapSchemaMap(s); ok {
 		s, found := sm.GetOk(step.Name)
 		if !found {
-			return nil, fmt.Errorf("%s not found", step.String())
+			return nil, fmt.Errorf("%s not found", step.GoString())
 		}
 		return s, nil
 	}
-	return nil, fmt.Errorf("%s is not applicable", step.String())
+	return nil, fmt.Errorf("%s is not applicable", step.GoString())
 }
 
 // Drill down into a Map, Set or List element schema.
 type ElementStep struct{}
 
-func (step ElementStep) String() string {
+func (step ElementStep) GoString() string {
 	return "walk.ElementStep{}"
 }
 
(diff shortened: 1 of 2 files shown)


pkg/tfshim/walk/walk.go
```diff
@@ -0,0 +1,198 @@
...
+	String() string
+	Lookup(shim.Schema) (shim.Schema, error)
+}
+
+// Drill down into an attribute by the given attribute name.
+type GetAttrStep struct {
+	Name string
+}
+
+func (step GetAttrStep) String() string {
```
iwahbe, line 106: Again, this should be `GoString()`.

pkg/tfshim/walk/walk.go
```diff
@@ -0,0 +1,198 @@
...
+
+func wrapSchemaMap(sm shim.SchemaMap) shim.Schema {
+	return (&schema.Schema{
+		Type: shim.TypeMap,
+		Elem: (&schema.Resource{Schema: sm}).Shim(),
+	}).Shim()
+}
+
+// Utility function to recognize nested object field type schemas encoded in shim.Schema.
+func UnwrapSchemaMap(s shim.Schema) (shim.SchemaMap, bool) {
```
iwahbe, line 154: Does this need to be public?

pkg/tfshim/walk/walk.go
```diff
@@ -0,0 +1,198 @@
...
+func VisitSchema(schema shim.Schema, visitor SchemaVisitor) {
+	walkSchemaInner(NewSchemaPath(), schema, visitor)
+}
+
+// Visit all nested schemas in a SchemaMap, keeping track of SchemaPath location.
+func VisitSchemaMap(schemaMap shim.SchemaMap, visitor SchemaVisitor) {
+	walkSchemaMapInner(NewSchemaPath(), schemaMap, visitor)
+}
+
+func walkSchemaInner(path SchemaPath, schema shim.Schema, visitor SchemaVisitor) {
```
iwahbe, line 175: What is this `Inner` of? I don't see a `walkSchema`.
```suggestion
func walkSchema(path SchemaPath, schema shim.Schema, visitor SchemaVisitor) {
```

pkg/tfshim/walk/walk_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"fmt"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
+)
+
+var strSchema = (&schema.Schema{
```
iwahbe, line 27: I generally make these functions to ensure immutability. I don't think it we mutate this by mistake here so I leave this up to you.

pkg/tfshim/walk/walk_test.go
```diff
@@ -0,0 +1,136 @@
...
+func TestLookupSchemaPath(t *testing.T) {
+	s := testSchemaMap
+
+	type testCase struct {
+		name   string
+		path   SchemaPath
+		expect any
+	}
+
+	/* this is still confusing */
```
iwahbe, line 67: What is?

pkg/tfshim/walk/walk.go
```diff
@@ -0,0 +1,198 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
+)
+
+// Represents locations in a Schema value as a sequence of steps to locate it.
```
iwahbe, line 25:
```suggestion
// Represents locations in a tfshim.Schema value as a sequence of steps to locate it.
```

iwahbe (commented):
I'm having some trouble understanding the point here. In your comment, you say that these paths correspond to TF path values. If this is the desired property, can we add a test for this?

Is this equivalent to TF attribute walkers. Mentioning the sister type of `SchemaPath` in the `SchemaPath` doc comment would be great.

pkg/tfshim/walk/walk_test.go
```diff
@@ -0,0 +1,136 @@
...
+// limitations under the License.
+
+package walk
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"fmt"
```
iwahbe, line 22: This should be above `"testing"`.

t0yv0:
> I'm having some trouble understanding the point here. In your comment, you say that these paths correspond to TF path values. If this is the desired property, can we add a test for this?
> Is this equivalent to TF attribute walkers. Mentioning the sister type of SchemaPath in the SchemaPath doc comment would be great.

I've been looking at zclconf/go-cty package for inspiration; this design is either used in earlier TF versions directly or else inspires newer TF designs. There is [Path](https://pkg.go.dev/github.com/zclconf/go-cty/cty?utm_source=godoc#Path) type in go-cty that logically is kind of like resource.PropertyPath. It identifies locations of sub-values within a parent value. 

So it should be possible to convert go-cty Path to a SchemaPath "naturally". GetAttr steps are identical. As to index steps, in go-cty the  IndexStep is tagged with the value of an index to drill down to. In the schema world we just always drill down to the schema describing the element, so we do not need to know the value. So we could write this without loss of generality:

```
func FromValuePath(cty.Path) SchemaPath
```

But we could not write the reverse:

```
func ToValuePath(SchemaPath) cty.Path // cannot write this as ElementStep{} is ambiguous.
```

I think in TF world there is no dedicated type for schema paths but schema bits are sometimes indexed by the paths anyway, in that case what they do is something like `IndexStep{zeroValue}` to represent drilling down *any* element.

pkg/tfshim/walk/walk.go
t0yv0: Eh, probably not from this package and the naming is off. I will hide.

pkg/tfshim/walk/walk.go
```diff
@@ -0,0 +1,198 @@
...
+		if !found {
+			return nil, fmt.Errorf("%s not found", step.String())
+		}
+		return s, nil
+	}
+	return nil, fmt.Errorf("%s is not applicable", step.String())
+}
+
+// Drill down into a Map, Set or List element schema.
+type ElementStep struct{}
```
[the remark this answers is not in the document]
t0yv0: So this is a great question. By resource I understand you mean shim.Resource which actually is our current mis-named tfshim encoding of an Object type schema. This is the bit of the design I went back and forth on a bit as it wasn't super clear what to do.

Currently what the code does is this, if you want to find an object type, you search for it's location and you get back a shim.Schema with s.Type()=TypeMap and s.Elem()=shim.Resource with s.Elem().Schema() giving you a SchemaMap describing the resource's properties.  Although this encoding is using Elem(), the path will not us ElementPath{} because actual values of this type will typically not be drilling down through `x[y]` accessors.

This is elaborated in the tests a bit, one sec.

pkg/tfshim/walk/walk_test.go
```diff
@@ -0,0 +1,134 @@
...
+	type testCase struct {
+		name   string
+		path   SchemaPath
+		expect any
+	}
+
+	testCases := []testCase{
+		{
+			"single-nested block object",
+			NewSchemaPath().GetAttr("x"),
```
t0yv0, line 70: Here you're getting an "object type" as a shim.Schema with a shim.Resource inside.

pkg/tfshim/walk/walk_test.go
```diff
@@ -0,0 +1,134 @@
...
+			s.Get("x"),
+		},
+		{
+			"cannot do Element on an object",
+			NewSchemaPath().GetAttr("x").Element(),
```
t0yv0, line 75: This fails; although NewSchemaPath().GetAttr("x") looks up a Schema with an Elem(), we can't drill down an Element(). The motivation for this is that values of this type would look like this:

```
v = {"x": {"y": "foo"}}
```

So drilling down to "foo" is done via `v["x"]["y"]` not `v["x"][index]["y"]`.

pkg/tfshim/walk/walk_test.go
```diff
@@ -0,0 +1,134 @@
...
+				`walk.ElementStep{} is not applicable to object types`),
+		},
+		{
+			"nested x.y prop",
+			NewSchemaPath().GetAttr("x").GetAttr("y"),
+			strSchema,
+		},
+		{
+			"list elem",
+			NewSchemaPath().GetAttr("list").Element(),
```
t0yv0, line 86: Here we have a list of strings, values look like `{"list": ["foo", "bar"]}`. So drilling down uses the Element step.

pkg/tfshim/walk/walk_test.go
```diff
@@ -0,0 +1,134 @@
...
+			strSchema,
+		},
+		{
+			"regress batching.send_after",
+			NewSchemaPath().GetAttr("batching").Element().GetAttr("send_after"),
```
t0yv0, line 91: Here we have a list of objects. 

```
v = {"batching": [{"send_after": "3s"}]}
```

Drilling down looks like this: `v["batching"][0]["send_after"]` so the schema path is like written.

pkg/tfshim/walk/walk.go
```diff
@@ -0,0 +1,235 @@
...
+//
+// An empty SchemaPath represents the current location.
+//
+// Values of this type are immutable by convention, use Copy as necessary for local mutations.
+type SchemaPath []SchemaPathStep
+
+func (p SchemaPath) GoString() string {
+	parts := []string{"walk", "NewSchemaPath()"}
+	for _, e := range p {
+		switch ee := e.(type) {
```
iwahbe, line 38: Each `case` body is a new scope, so you can just shadow `e`.
```suggestion
		switch e := e.(type) {
```

pkg/tfshim/walk/walk.go
```diff
@@ -0,0 +1,235 @@
...
+			p = p.Element()
+		case cty.GetAttrStep:
+			p = p.GetAttr(s.Name)
+		}
+	}
+	return p
+}
+
+// Converts a value path to a Schema Path (zclconf package representation).
+func FromCtyPath(path hcty.Path) SchemaPath {
```
iwahbe, line 224: I think the names on these functions is backwards. 
```suggestion
func FromHCtyPath(path hcty.Path) SchemaPath {
```

pkg/tfshim/walk/walk.go
```diff
@@ -0,0 +1,235 @@
...
+
+func visitSchemaMapInner(path SchemaPath, schemaMap shim.SchemaMap, visitor SchemaVisitor) {
+	schemaMap.Range(func(key string, schema shim.Schema) bool {
+		visitSchemaInner(path.GetAttr(key), schema, visitor)
+		return true
+	})
+}
+
+// Converts a value path to a Schema Path (hashicorp/go-cty representation).
+func FromHCtyPath(path cty.Path) SchemaPath {
```
iwahbe, line 210:
```suggestion
func FromCtyPath(path cty.Path) SchemaPath {
```

iwahbe (approved):
LGTM

PR #1076, opened 2023-05-03
Implement path translation and SchemaInfo walking

t0yv0:
New function PropertyPathToSchemaPath allows converting paths describing nested locations in Pulumi values (resource.PropertyValue) to matching schema paths.

New function LookupSchemaInfoMapPath allows finding a nested SchemaInfo object.

These functions in combination are useful for schema-aware transformations on Pulumi values.

On top of https://github.com/pulumi/pulumi-terraform-bridge/pull/1075

### DIFF
--- a/pkg/tfbridge/walk/walk.go
+++ b/pkg/tfbridge/walk/walk.go
@@ -40,10 +40,11 @@ func PropertyPathToSchemaPath(
 }
 
 func propertyPathToSchemaPath(
-	basePath walk.SchemaPath,
+	basePath SchemaPath,
 	propertyPath resource.PropertyPath,
 	schemaMap shim.SchemaMap,
-	schemaInfos map[string]*tfbridge.SchemaInfo) walk.SchemaPath {
+	schemaInfos map[string]*tfbridge.SchemaInfo,
+) SchemaPath {
 
 	if len(propertyPath) == 0 {
 		return basePath
@@ -69,10 +70,11 @@ func propertyPathToSchemaPath(
 }
 
 func propertyPathToSchemaPathInner(
-	basePath walk.SchemaPath,
+	basePath SchemaPath,
 	propertyPath resource.PropertyPath,
 	schema shim.Schema,
-	schemaInfo *tfbridge.SchemaInfo) walk.SchemaPath {
+	schemaInfo *tfbridge.SchemaInfo,
+) SchemaPath {
 
 	if len(propertyPath) == 0 {
 		return basePath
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/walk/walk.go
```diff
@@ -0,0 +1,154 @@
...
+// overrides in SchemaInfos, and accounting for MaxItems=1 situations where Pulumi flattens collections to plain values.
+// and therefore SchemaPath values are longer than the PropertyPath values.
+//
+// PropertyPathToSchemaPath may return nil if there is no matching schema found. This may happen when drilling down to
+// values of unknown type, attributes not tracked in schema, or when there is a type mismatch between the path and the
+// schema.
+func PropertyPathToSchemaPath(
+	propertyPath resource.PropertyPath,
+	schemaMap shim.SchemaMap,
+	schemaInfos map[string]*tfbridge.SchemaInfo) SchemaPath {
```
iwahbe, line 38: Completely optional, but I think this is much easier to read.
```suggestion
	schemaInfos map[string]*tfbridge.SchemaInfo,
) SchemaPath {
```

pkg/tfbridge/walk/walk.go
```diff
@@ -0,0 +1,154 @@
...
+package walk
+
+import (
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/walk"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+type SchemaPath = walk.SchemaPath
```
iwahbe, line 24: I don't understand when we use the distinction between a `walk.SchemaPath` and a `SchemaPath`. This library uses both.

pkg/tfbridge/walk/walk_test.go
```diff
@@ -0,0 +1,201 @@
...
+		"single_obj": (&schema.Schema{
+			Type: shim.TypeMap,
+			Elem: xySchema,
+		}).Shim(),
+	}
+
+	type testCase struct {
+		name        string
+		pp          resource.PropertyPath
+		schemaInfos map[string]*tfbridge.SchemaInfo
```
iwahbe, line 67: It looks like this is never set. Can we delete it?

pkg/tfbridge/walk/walk_test.go
```diff
@@ -0,0 +1,201 @@
...
+			pp := PropertyPathToSchemaPath(tc.pp, schemaMap, tc.schemaInfos)
+			assert.Equal(t, tc.expected, pp)
+		})
+	}
+}
+
+func TestLookupSchemaInfoMapPath(t *testing.T) {
+	yes := true
+
+	schemaInfos := map[string]*tfbridge.SchemaInfo{
```
iwahbe, line 151: Can we have a test case for when `SchemaInfos` declares something `MaxItemsOne: &true`.

pkg/tfbridge/walk/walk.go
t0yv0: It's an alias exact same thing.

t0yv0: Well the key idea is that code outside this package does to need to import the other walk package just to reference SchemaPath. I can tidy up.

iwahbe (approved):

PR #1077, opened 2023-05-03
Implement MarkSchemaSecrets

t0yv0:
It is the responsibility of the bridged providers to ensure that any PropertyValue data returned to the engine is marked with resource.MakeSecret when it corresponds to a secret property in the schema. Previously this responsibility was carried by the serializers, but it can be more convenient to reuse if expressed as a pure PropertyValue based transformation pass. This is added now as a new function MarkSchemaSecrets.

On top of #1076

### DIFF
--- a/pkg/tfbridge/x/secrets.go
+++ b/pkg/tfbridge/x/secrets.go
@@ -1,0 +1,101 @@
+// Copyright 2016-2023, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package x
+
+import (
+	"context"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/walk"
+	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/x/secrets.go
```diff
@@ -0,0 +1,100 @@
...
+	return secret
+}
+
+func (ss *schemaSecrets) markSchemaSecretsTransform(
+	path resource.PropertyPath,
+	value resource.PropertyValue,
+) resource.PropertyValue {
+	switch {
+	case value.IsArray():
+		tvs := []resource.PropertyValue{}
```
iwahbe, line 61:
```suggestion
		tvs := make([]resource.PropertyValue, 0, len(value.ArrayValue()))
```

pkg/tfbridge/x/secrets.go
```diff
@@ -0,0 +1,100 @@
...
+			Dependencies: o.Dependencies,
+		})
+	case value.IsSecret():
+		s := value.SecretValue()
+		tv := ss.markSchemaSecretsTransform(path, s.Element)
+		if tv.IsSecret() {
+			value = tv
+		} else {
+			value = resource.MakeSecret(tv)
+		}
```
iwahbe, line 92: We don't need to nest secrets. Once a value is secret, every nested value is considered secret. We can just return here.

pkg/tfbridge/x/secrets.go
```diff
@@ -0,0 +1,100 @@
...
+		value = resource.NewArrayProperty(tvs)
+	case value.IsObject():
+		pm := make(resource.PropertyMap)
+		for k, v := range value.ObjectValue() {
+			subPath := append(path, string(k))
+			tv := ss.markSchemaSecretsTransform(subPath, v)
+			pm[k] = tv
+		}
+		value = resource.NewObjectProperty(pm)
+	case value.IsOutput():
```
iwahbe, line 76: If `value.OutputValue().Secret`, we can just return `value`. The output it contains is already secret.

iwahbe (approved):
Where do we consume the library? Is that a separate PR?

t0yv0:
>Where do we consume the library? Is that a separate PR?

Yes.

I've used it in https://github.com/pulumi/pulumi-terraform-bridge/pull/1065 to mark secrets, now breaking it down into smaller changes.

I also want to use it in https://github.com/pulumi/pulumi-terraform-bridge/pull/1066 - very similarly to MarkSecrets, a path aware transformation on PropertyValue can be written that populates nested Pulumi-level defaults with relative ease.

Issue #1074, opened 2023-05-03
AWS skip_metadata_api_check not working with Plugin Framework


t0yv0:
### What happened?

Bridged pulumi-aws provider has an interesting property defined upstream:

```go
			"skip_metadata_api_check": {
				Type:         nullable.TypeNullableBool,
				Optional:     true,
				ValidateFunc: nullable.ValidateTypeStringNullableBool,
				Description: "Skip the AWS Metadata API check. " +
					"Used for AWS API implementations that do not have a metadata api endpoint.",
			},
```

with

```go
const (
	TypeNullableBool = schema.TypeString
)
```

and these SchemaInfo overrides:

```
			"skip_metadata_api_check": {
				Type: "boolean",
				Default: &tfbridge.DefaultInfo{
					Value: true,
				},
			},
```

Note that this config property surfaces as a boolean in Pulumi Package Schema but is a String in the underlying provider. This  currently does not work with PF bridging as expected:

- converters should not look at the schema but at the TF information to make sure data is sent as a string not a boolean to TF
- looks like converters are expected to tolerate parsing resource.PropertyValue booleans into a string-typed property to seamlessly adapt this case

### Expected Behavior

skip_metadata_api_check works as with pre-PF bridge

Instead we get

```
   aws:iam:Role mylambda-role  error: cannot encode provider configuration to call ConfigureProvider: objectEncoder failed on property "skip_metadata_api_check": Expected a Boolean
```

### Steps to reproduce

https://github.com/pulumi/pulumi-aws/pulls

### Output of `pulumi about`

NA

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #1083, opened 2023-05-04
Refactor: plugin framework converters to not depend on Pulumi Package Schema 

t0yv0:
Converting between resource.PropertyValue and tftypes.Value needs to be aware of data type schemas, Pulumi-TF name mangling and MaxItemsOne=1 flattening. Prior to this refactor the code was relying on Renames infrastructure for naming and Pulumi Package Schema for detecting MaxItemsOne=1 and other schema information. 

Some reasons to change this:

- since this was written we are much more confident that Renames tables and PulumiToTerraformName/TerraformToPulumiNameV2 functions are equivalent, and prefer the later because we fixed bugs in these functions and installed runtime checks into provider builds that confirm the equivalence

- we found that Pulumi schema may diverge from the TF schema such as in the https://github.com/pulumi/pulumi-terraform-bridge/issues/1074 case; we should use the more native TF schema in the low-level converter module

- generally it's safer to drive converters by tftypes.Type as the conformance of tftypes.Value we sent do the tftypes.Type is exactly what's enforced by the receiving code

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1074


t0yv0:
This seems to empirically resolve the failing pulumi-aws test for me finally, but I'm waiting on the PR checks to officially confirm.

pf/internal/convert/encoding.go
```diff
@@ -331,140 +239,95 @@ func (e *encoding) deriveDecoder(typeSpec *pschema.TypeSpec, t tftypes.Type) (De
...
 // A generic base function for deriving tuple encoders and decoders.
 //
 // It handles reference validation and property discovery.
-func deriveTupleBase[T any](
-	f func(*pschema.TypeSpec, tftypes.Type) (T, error),
-	tok tokens.Type,
-	typ *pschema.ComplexTypeSpec,
-	t tftypes.Tuple,
-) ([]T, error) {
+func deriveTupleBase[T any](pctx *schemaPropContext, f func(*schemaPropContext, tftypes.Type) (T, error),
```
t0yv0, line 302: Do we need additional tests for tuples? I was slightly concerned here but it looks like existing tests pass.

pf/internal/convert/schema_context.go
```diff
@@ -0,0 +1,168 @@
...
+import (
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/walk"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+type schemaMapContext struct {
```
t0yv0, line 26: I can talk through this; looking over this again it looks like awkward glue code again; I wanted to take this glue code out of the main converter module code.

Looks like these types step down Element or GetAttr (which is the same thing as SchemaPath elements) to co-locate the schema information.  Instead of writing it like this, we could instead hold on to top-level shim.SchemaMap and simply walk down the SchemaPath  and always lookup from the top using walk.LookupSchemaMapPath.

It seems like in many places we want a record that holds shim.SchemaMap and a matching *tfbridge.SchemaInfo together; or else a Schema and matching tfbridge.SchemaInfo together. This is not the only place.. these pieces of information naturally join. I wonder if we could name it better and reuse.

t0yv0:
I think I got one passing CI build on pulumi-aws, with an unrelated issue (fixing now). I think this change is very beneficial but I'd like some scrutiny and suggestions on what tests to add; this is essentially a refactor and while it's passing existing tests is it enough? 

Sorry for poor commit structure the development here went a bit organically. Let me know if it would be helpful to reorganize commits or pair/walk through this.

pf/internal/convert/schema_context.go
iwahbe: I agree that this should be a separate package. 

Maybe something like `bridged.SchemaMap`?

iwahbe (commented):
Since this change gets `pulumi-aws` from red to green (mostly), it's extremely welcome. That said, it's not really a refactor since it changes behavior.

I'd really appreciate a test that indicates what changed for `pulumi-aws` such that it used to not work and now works.

iwahbe:
A rebase onto `master` will remove the failing `Test azure Downstream` and get the PR to go green.

pf/internal/convert/schema_context.go
t0yv0: Hm yes or SchemaWithInfo and SchemaMapWithInfo ? This PR is too big to take this on but perhaps another time we're touching something like this.

t0yv0:
> That said, it's not really a refactor since it changes behavior.

In my book if you remove an assumption while keeping all the functionality you need that's a refactor. If you have a bug stemming from that assumption violated but now that assumption is gone, adding tests is testing something that's gone, it's only guarding against that same assumption again :) I think the assumption was that PackageSchema has type fidelity with the TF schema. 

That being said the AWS failure case is somewhat interesting it's a TF string rebranded as boolean in PackageInfo, so I'll endeavor to add a test case after all.

pf/tests/testdata/smac-program/Pulumi.yaml
```diff
@@ -0,0 +1,42 @@
+name: smac-program
```
t0yv0, line 1: Here's the test @iwahbe 

It's still weird how this works out. The property is passed as bool. But Configure returns it as string because it's a string internally. The functions verify they receive the right string values. 

Something in the Yaml SDK must be auto-converting  for us.

iwahbe (approved):
LGTM

PR #1084, opened 2023-05-04
Internal fix: cd pf && make build.testproviders

t0yv0:
Looks like CI does not enforce yet that these schemas are in sync.

### DIFF
--- a/pf/tests/internal/testprovider/cmd/pulumi-resource-testbridge/schema.json
+++ b/pf/tests/internal/testprovider/cmd/pulumi-resource-testbridge/schema.json
@@ -50,12 +50,7 @@
                     "$ref": "#/types/testbridge:index/TestnestRuleActionParametersPhases:TestnestRuleActionParametersPhases"
                 }
             },
-            "type": "object",
-            "required": [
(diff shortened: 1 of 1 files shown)


pf/tests/internal/testprovider/cmd/pulumi-resource-testbridge/schema.json
```diff
@@ -50,12 +50,7 @@
                     "$ref": "#/types/testbridge:index/TestnestRuleActionParametersPhases:TestnestRuleActionParametersPhases"
                 }
             },
-            "type": "object",
-            "required": [
-                "automaticHttpsRewrites",
```
t0yv0, line 55:
```
									"automatic_https_rewrites": schema.BoolAttribute{
										Optional:            true,
										MarkdownDescription: "Turn on or off Cloudflare Automatic HTTPS rewrites.",
									},

```

This looks right, should be optional.

iwahbe (approved):
Can we also enforce that this is actually in sync?

t0yv0:
Yes one sec.

PR #1085, opened 2023-05-04
Testing: ensure schemas on disk stay in sync

t0yv0:
This will ask to `make build.testproviders` when some changes in the logic cause changes in the schema of the test providers.

### DIFF
--- a/pf/tests/schema_test.go
+++ b/pf/tests/schema_test.go
@@ -16,13 +16,18 @@ package tfbridgetests
 
 import (
 	"encoding/json"
+	"os"
 	"testing"
 
-	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+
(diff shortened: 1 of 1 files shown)


pf/tests/schema_test.go
```diff
@@ -59,3 +64,80 @@ func TestSchemaGen(t *testing.T) {
...
+			if tc.pf.NewProvider != nil {
+				data := genMetadata(t, tc.pf)
+				require.NoError(t, json.Unmarshal(data.PackageSchema, &actualSpec))
+			} else {
+				var err error
+				actualSpec, err = tfgen.GenerateSchema(tc.provider, testSink(t))
+				require.NoError(t, err)
+			}
+
+			// Ignoring version differences, for some obscure reason they diverge right now.
```
iwahbe, line 130: This should probably be an issue.

iwahbe (approved):
Thanks for adding this!

pf/tests/schema_test.go
t0yv0: Well I don't think it affects users in any way.

t0yv0:
And of course this fails on windows. Blimey. Will pick up tomorrow.

PR #1086, opened 2023-05-04
Refactor handling provider-level Configuration values

t0yv0:
The only intentional user-visible change is this: before the refactor, if any nested schema paths for a provider-level configuration property were sensitive, bridged providers would mark the entire property as secret, and it would be stored as a secret value in the state, even if the nested properties in question were not set by the user. After the refactor, the behavior is more precise. Bridged providers will only mark the providers properties as secret if configued with actual secrets.

The main aim of the refactor is to enable the reuse of the code for parsing provider level configuration (ConfigEncoding) and marking schema secrets (MarkSchemaSecrets) for Plugin Framework based providers. The necessary functions and types are now exported from the package.

ConfigEncoding no longer carries the secret marking responsibility as it is taken over by MarkSchemaSecrets.


pkg/tfbridge/config_encoding.go
```diff
@@ -164,12 +134,22 @@ func (enc *configEncoding) UnmarshalPropertyValue(key resource.PropertyKey, v *s
 		return &v, nil
 	}
 
+	// Computed sentinels are coming in as always having an empty string, but the encoding coerses them to a zero
+	// value of the appropriate type.
+	if pv.IsComputed() && gotShimType {
```
t0yv0, line 139: Probably being overcautious here; current prod code does not peek inside a Computed and if it's an empty string inside it's ok; still having a type-appropriate zero value seems more future-proof.

pkg/tfbridge/config_encoding.go
```diff
@@ -204,9 +181,15 @@ func (enc *configEncoding) UnmarshalProperties(props *structpb.Struct,
 }
 
 // Inverse of UnmarshalProperties, with additional support for top-level (but not nested) secrets.
-func (enc *configEncoding) MarshalProperties(props resource.PropertyMap,
-	opts plugin.MarshalOptions) (*structpb.Struct, error) {
-	opts.KeepSecrets = true
+func (enc *ConfigEncoding) MarshalProperties(props resource.PropertyMap) (*structpb.Struct, error) {
+	opts := plugin.MarshalOptions{
```
t0yv0, line 185: It wasn't really configurable; this encoding implies a certain parameter set, inline it.

pkg/tfbridge/config_encoding.go
```diff
@@ -218,17 +201,17 @@ func (enc *configEncoding) MarshalProperties(props resource.PropertyMap,
...
 
-func (enc *configEncoding) jsonEncodePropertyValue(k resource.PropertyKey,
+func (enc *ConfigEncoding) jsonEncodePropertyValue(k resource.PropertyKey,
 	v resource.PropertyValue) (resource.PropertyValue, error) {
 	if v.ContainsUnknowns() {
 		return resource.NewStringProperty(plugin.UnknownStringValue), nil
 	}
-	if v.IsSecret() {
-		encoded, err := enc.jsonEncodePropertyValue(k, v.SecretValue().Element)
+	if v.ContainsSecrets() {
```
t0yv0, line 209: This will encode any nested secrets as top-level secret. Because it can't represent nested secrets accurately in the JSON-in-proto encoding.

pkg/tfbridge/config_encoding_test.go
```diff
@@ -0,0 +1,381 @@
...
+	structpb "google.golang.org/protobuf/types/known/structpb"
+
+	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	sch "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConfigEncoding(t *testing.T) {
```
t0yv0, line 30: Trying to tread really carefully here in light of recent P1s and postmortem. Added lots of tests.

pkg/tfbridge/provider.go
```diff
@@ -348,8 +339,11 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
 		return nil, validationErrors
 	}
 
+	// Ensure propreties marked secret in the schema have secret values.
+	secretNews := MarkSchemaSecrets(ctx, p.config, p.info.Config, resource.NewObjectProperty(news)).ObjectValue()
```
t0yv0, line 343: MarkSchemaSecrets takes over from ConfigEncoding to ensure the provider marks nested secrets.

pkg/tfbridge/secrets.go
```diff
@@ -12,34 +12,32 @@
...
 type schemaSecrets struct {
 	schemaMap   shim.SchemaMap
-	schemaInfos map[string]*tfbridge.SchemaInfo
+	schemaInfos map[string]*SchemaInfo
 }
 
 func (ss *schemaSecrets) shouldBeSecret(path resource.PropertyPath) bool {
-	schemaPath := walk.PropertyPathToSchemaPath(path, ss.schemaMap, ss.schemaInfos)
-	s, info, err := walk.LookupSchemas(schemaPath, ss.schemaMap, ss.schemaInfos)
+	schemaPath := PropertyPathToSchemaPath(path, ss.schemaMap, ss.schemaInfos)
```
t0yv0, line 39: Unfortunately due to Go limitations this can't live in a separate package; getting an error about recursive package dependencies.

pkg/tfbridge/config_encoding.go
```diff
@@ -204,9 +181,15 @@ func (enc *configEncoding) UnmarshalProperties(props *structpb.Struct,
 }
 
 // Inverse of UnmarshalProperties, with additional support for top-level (but not nested) secrets.
```
iwahbe, line 183: I thought we now support nested secrets.

pkg/tfbridge/config_encoding_test.go
```diff
@@ -0,0 +1,381 @@
...
+				// When there any unknowns at all in the config, unk sentinel is sent by the engine.
+				checkMarshal(t, testCase{
+					tc.ty,
+					unk,
+					resource.MakeComputed(tc.pv),
+				})
+			})
+		}
+
+		// Checking a few extra cases for nested secrets.
```
iwahbe, line 245:
```suggestion
		// Checking a few extra cases for nested computedness.
```

pkg/tfbridge/provider_test.go
```diff
@@ -960,6 +960,8 @@ func TestCheckConfig(t *testing.T) {
 	// Secrets: it is important to handle secrets correctly since the results of CheckConfig are stored in the
 	// state. If secrets are not marked as such, they are stored as plaintext in the state.
 	t.Run("preserve_program_secrets", func(t *testing.T) {
+		t.Skip()
```
iwahbe, line 963: Why are these skipped?

iwahbe (commented):
The production code looks fine to me. The only major problem I see is that we have skipped some tests that I would like to see go green.

pkg/tfbridge/provider_test.go
t0yv0: So these are skipped because they fail. We are basically counting on these test situations never happening because the engine takes care of the "program-originating" secrets for us. So it's not the bridged provider's responsibility anymore to track these but the engine's. Test can be deleted perhaps.

pkg/tfbridge/config_encoding.go
t0yv0: Updated comment, obsolete comment.

pkg/tfbridge/provider_test.go
t0yv0: Removed these.

t0yv0:
On top of https://github.com/pulumi/pulumi-terraform-bridge/pull/1090 that should provide good integration level coverage checking what the removed tests checked.

iwahbe (approved):

PR #1088, opened 2023-05-05
Fix #1087: do not call PreConfigureCallback with unknowns during preview

### LINKED ISSUE #1087: PreConfigureCallback called with Computed values, breaking user code
### What happened?

A recent change refactored PreConfigureCallback to be invoked during CheckConfig phase instead of Configure phase in the Pulumi Provider protocol, and unfortunately this change breaks user code in corner cases. One example was reported in pulumi-aws:

https://github.com/pulumi/pulumi-aws/issues/2513 

When previewing a program like this:

```typescript
import * as rand from "@pulumi/random";

const ri = new rand.RandomInteger("ri", {min: 1, max: 10});

function show(x: number): string {
    return String(x);
}

const x: pulumi.Output<string> = ri.result.apply(show);

const devAccountProvider = new aws.Provider("devAccountProvider", {
    assumeRole: {
        roleArn: pulumi.interpolate`arn:aws:iam::devaccountid:role/${x}`,
    },
});

// Create an AWS resource (S3 Bucket)
const bucket = new aws.s3.Bucket("my-bucket", {}, {provider: devAccountProvider});

export const bid = bucket.id;
```

On pulumi-aws@5.33.0 PreConfigureCallback is not called for devAccountProvider at all, because the engine does not call Configure in case there are unknowns. However on the latest 5.39.0 the callback is called with a Computed value, since it's called from CheckConfig, which panics in the user code not expecting to encounter such a value:

```
    panic: interface conversion: interface {} is resource.Computed, not resource.PropertyMap
    goroutine 82 [running]:
    github.com/pulumi/pulumi/sdk/v3/go/common/resource.PropertyValue.ObjectValue(...)
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.64.0/go/common/resource/properties.go:435
    github.com/pulumi/pulumi-aws/provider/v5.preConfigureCallback(0xc002dd4f00?, {0x65?, 0x1000000000101?})
        /home/runner/work/pulumi-aws/pulumi-aws/provider/resources.go:361 +0xbbb
    github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge.(*Provider).CheckConfig(0xc002dd4f00, {0xfd66548, 0xc0003dda10}, 0xc002fb4240)
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.45.3/pkg/tfbridge/provider.go:327 +0x346
    github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_CheckConfig_Handler.func1({0xfd66548, 0xc0003dda10}, {0xe005560?, 0xc002fb4240})
```

A quick fix to restore previous behavior would be skip calling PreConfigureCallback if there are any unknowns. 

### Expected Behavior

As a user I expect not having to deal with Computed or unknown values in PreConfigureCallback.

### Steps to reproduce

See above

### Output of `pulumi about`

N/A

### Additional context

_No response_

### Contributing

t0yv0:
Fixes #1087

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -322,18 +322,20 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
 		return nil, errors.Wrap(validationErrors, "could not marshal config state")
 	}
 
-	if p.info.PreConfigureCallback != nil {
-		// NOTE: the user code may modify news in-place.
-		if validationErrors = p.info.PreConfigureCallback(news, config); validationErrors != nil {
-			return nil, validationErrors
+	if !news.ContainsUnknowns() {
+		if p.info.PreConfigureCallback != nil {
+			// NOTE: the user code may modify news in-place.
+			validationErrors := p.info.PreConfigureCallback(news, config)
+			if validationErrors != nil {
+				return nil, validationErrors
+			}
 		}
-	}
-
-	if p.info.PreConfigureCallbackWithLogger != nil {
-		// NOTE: the user code may modify news in-place.
-		validationErrors := p.info.PreConfigureCallbackWithLogger(ctx, p.host, news, config)
-		if validationErrors != nil {
-			return nil, validationErrors
+		if p.info.PreConfigureCallbackWithLogger != nil {
+			// NOTE: the user code may modify news in-place.
+			validationErrors := p.info.PreConfigureCallbackWithLogger(ctx, p.host, news, config)
+			if validationErrors != nil {
+				return nil, validationErrors
+			}
 		}
 	}
 
--- a/pkg/tfbridge/provider_test.go
+++ b/pkg/tfbridge/provider_test.go
@@ -1355,6 +1355,46 @@ func TestPreConfigureCallback(t *testing.T) {
 		  }
 		}`)
 	})
+	t.Run("PreConfigureCallback not called at preview with unknown values", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/tfbridge/provider.go
```diff
@@ -322,18 +322,20 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
 		return nil, errors.Wrap(validationErrors, "could not marshal config state")
 	}
 
-	if p.info.PreConfigureCallback != nil {
-		// NOTE: the user code may modify news in-place.
-		if validationErrors = p.info.PreConfigureCallback(news, config); validationErrors != nil {
-			return nil, validationErrors
+	if !news.ContainsUnknowns() {
```
thomas11, line 325: I think this would warrant a comment on why this check is there.

iwahbe (approved):
In an ideal world, we should call `PreConfigureCallback` with unknowns.

LGTM for backwards compatibility!

t0yv0:
> In an ideal world, we should call PreConfigureCallback with unknowns.

Well, I think once we have an actual user scenario where it's needed we could log this, I'm not terribly sure; it's a bit of a quirk but we've lived with this for a while it seems. It's quite inconsistent that we call it at preview if known and not call it if unknown, which makes the aws program ask for creds. Kind of surprising you make a little change to your program or state and it asks for creds. But not too terrible.

PR #1089, opened 2023-05-05
Allow Docs Rule based Overrides

iwahbe:
This PR allows providers to specify arbitrary rule based docs replace directives. The obvious use case is to avoid docs elisions by replacing phrases such as `"The resource ID in terraform of "` with `"The resource ID in Pulumi of "`.

This PR also includes 3 rules that are applied to all providers:
1. Replacing `terraform plan` with `pulumi preview`.
2. Replacing `terraform apply` with `pulumi up`.
3. Stripping markdown links to `terraform.*`.

For an example PR that consumes this API, see https://github.com/pulumi/pulumi-alicloud/pull/398.

---

Applies a stop-gap fix for https://github.com/pulumi/pulumi-terraform-bridge/issues/658.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -115,6 +115,7 @@ type ProviderInfo struct {
 }
 
 type DocRuleInfo struct {
+	// A list of replace rules to apply.
 	ReplaceRules []ReplaceRule
 
 	// A function to suggest alternative file names for a TF element.
@@ -127,7 +128,12 @@ type DocsPathInfo struct {
 
 type ReplaceRule struct {
 	// The path at which this rule applies. Paths are matched via filepath.Match.
-	Path    string
+	//
+	// To match all files, supply "*".
+	Path string
+	// The function that performs the replace on the byte string.
+	//
+	// Must not be nil.
 	Replace func(path string, content []byte) ([]byte, error)
 }
 
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -159,13 +159,6 @@ type DocRuleInfo struct {
 	AlternativeNames func(DocsPathInfo) []string
 }
 
-// PreAppendEdits is a MakeEditRules that applies custom edits before default edits.
-func PreAppendEdits(edits ...DocsEdit) MakeEditRules {
-	return func(defaults []DocsEdit) []DocsEdit {
-		return append(edits, defaults...)
-	}
-}
-
 // Information for file lookup.
 type DocsPathInfo struct {
 	TfToken string
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -101,7 +101,7 @@ type ProviderInfo struct {
 	// See NewProviderMetadata for in-place construction of a *MetadataInfo.
 	MetadataInfo *MetadataInfo
 
-	// Rules that control discovery and replacement for docs.
+	// Rules that control file discovery and edits for any subset of docs in a provider.
 	DocRules *DocRuleInfo
 
 	UpstreamRepoPath string // An optional path that overrides upstream location during docs lookup
@@ -114,27 +114,29 @@ type ProviderInfo struct {
 	SkipExamples func(SkipExamplesArgs) bool
 }
 
+// DocRuleInfo controls file discovery and edits for any subset of docs in a provider.
 type DocRuleInfo struct {
-	// A list of replace rules to apply.
-	ReplaceRules []ReplaceRule
+	// A list of edit rules to apply.
+	EditRules []DocsEdit
 
 	// A function to suggest alternative file names for a TF element.
 	AlternativeNames func(DocsPathInfo) []string
 }
 
+// Information for file lookup.
 type DocsPathInfo struct {
 	Resource string
 }
 
-type ReplaceRule struct {
+type DocsEdit struct {
 	// The path at which this rule applies. Paths are matched via filepath.Match.
 	//
 	// To match all files, supply "*".
 	Path string
-	// The function that performs the replace on the byte string.
+	// The function that performs the edit on the file bytes.
 	//
 	// Must not be nil.
-	Replace func(path string, content []byte) ([]byte, error)
+	Edit func(path string, content []byte) ([]byte, error)
 }
 
 // TFProviderLicense is a way to be able to pass a license type for the upstream Terraform provider.
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -120,18 +120,47 @@ type DocRuleInfo struct {
 	EditRules []DocsEdit
 
 	// A function to suggest alternative file names for a TF element.
+	//
+	// When the bridge loads the documentation for a resource or a datasource, it
+	// infers the name of the file that contains the documentation. AlternativeNames
+	// allows you to provide a provider specific extension to the override list.
+	//
+	// For example, when attempting to find the documentation for the resource token
+	// aws_waf_instances, the bridge will check the following files (in order):
+	//
+	//	"waf_instance.html.markdown"
+	//	"waf_instance.markdown"
+	//	"waf_instance.html.md"
+	//	"waf_instance.md"
+	//	"aws_waf_instance.html.markdown"
+	//	"aws_waf_instance.markdown"
+	//	"aws_waf_instance.html.md"
+	//	"aws_waf_instance.md"
+	//
+	// The bridge will check any file names returned by AlternativeNames before
+	// checking it's standard list.
 	AlternativeNames func(DocsPathInfo) []string
 }
 
 // Information for file lookup.
 type DocsPathInfo struct {
-	Resource string
+	TfToken string
 }
 
 type DocsEdit struct {
-	// The path at which this rule applies. Paths are matched via filepath.Match.
+	// The file name at which this rule applies. File names are matched via filepath.Match.
 	//
 	// To match all files, supply "*".
+	//
+	// All 4 of these names will match "waf_instances.html.markdown":
+	//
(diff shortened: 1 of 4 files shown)


aq17 (commented):
This is nice - will let Daniel review first/ approve

iwahbe:
The only remaining question I have is if we should aggressively expand the list of default replaces. `pulumi-alicloud` benefit greatly from:
1. "The resource ID in terraform of " -> "The resource ID in Pulumi of "
2. "Use Terraform to" -> "Use Pulumi to"
3. "but Terraform cannot destroy it" -> "but Pulumi cannot destroy it"
4. "**NOTE:** Terraform will" -> "**NOTE:** Pulumi will"

All of these items are theoretically generic.

The other option is to maintain a `var AllReplaceRules []ReplaceRules` that ecosystem providers will use, but other providers will opt into.

CC @danielrbradley

iwahbe:
> Can we consider some tests please? You will notice that the area of generating the docs is currently very hard to test for no good reason, perhaps this is our opportunity to clean up a little and make it easy to put tests in for docs-related bits of the code?

It's on my todo list for this PR. I just didn't get to it yet.

t0yv0:
Awesome! This is going to help a lot of things.

danielrbradley (approved):
This looks like a good first step to getting docs replacements into the bridge 👍 

I'm not 100% convinced about hard coding any rules by default, but maybe they're okay if there's no way they can ever replace anything incorrectly. Also having these rules stated explicitly is much simpler to comprehend why docs are changing without having to search in the bridge code. Could we just put the default rules into the config file if one doesn't exist yet so it basically auto-initializes itself?

Guessing the next step after this is being able to read the rules from a JSON file and then also to propose required changes?

pkg/tfbridge/info.go
```diff
@@ -111,6 +114,31 @@ type ProviderInfo struct {
...
+type DocRuleInfo struct {
+	// A list of edit rules to apply.
+	EditRules []DocsEdit
+
+	// A function to suggest alternative file names for a TF element.
+	AlternativeNames func(DocsPathInfo) []string
+}
+
+// Information for file lookup.
+type DocsPathInfo struct {
```
t0yv0, line 127: Curious, how will this struct grow? Adding DataSource id so AlternativeNames can find datasource?

pkg/tfbridge/info.go
```diff
@@ -111,6 +114,31 @@ type ProviderInfo struct {
 	SkipExamples func(SkipExamplesArgs) bool
 }
 
+// DocRuleInfo controls file discovery and edits for any subset of docs in a provider.
+type DocRuleInfo struct {
+	// A list of edit rules to apply.
+	EditRules []DocsEdit
+
+	// A function to suggest alternative file names for a TF element.
+	AlternativeNames func(DocsPathInfo) []string
```
t0yv0, line 123: It think I know what this does, but just because I've dug deep in the implementation before; for the casual users looking just at the docstrings this is all a bit laconic. Perhaps add a sentence or two about "file discovery". 

What may be happening is that we're trying to find markdown files for every resource and datasource inside a checkout of UpstreamRepoPath (or Repository). 

AlternativeNames here informs this process by making suggestions of possible file names to look for when finding docs for a given resource.

It's not obvious if those should have paths or not. Perhaps a quick example in the doc string?

pkg/tfbridge/info.go
```diff
@@ -111,6 +114,31 @@ type ProviderInfo struct {
 	SkipExamples func(SkipExamplesArgs) bool
 }
 
+// DocRuleInfo controls file discovery and edits for any subset of docs in a provider.
+type DocRuleInfo struct {
```
t0yv0, line 118: There's some overlap with DocInfo at least conceptually but I think I see why you'd want this at top level.

pkg/tfgen/docs.go
```diff
@@ -201,9 +201,35 @@ func findRepoPath(repoPathsEnvVar string, moduleCoordinates string) string {
...
+		rawName + ".html.markdown",
+		rawName + ".markdown",
+		rawName + ".html.md",
+		rawName + ".md",
+	}
+
+	if globalInfo != nil && globalInfo.AlternativeNames != nil {
+		// We look at user generated names before we look at default names
+		possibleMarkdownNames = append(globalInfo.AlternativeNames(tfbridge.DocsPathInfo{
+			Resource: rawName,
```
t0yv0, line 221: Doesn't this get called for data sources too?

pkg/tfgen/docs.go
```diff
@@ -218,23 +244,12 @@ func getMarkdownDetails(sink diag.Sink, repoPath, org, provider string,
...
-		withoutPackageName(resourcePrefix, rawname) + ".markdown",
-		withoutPackageName(resourcePrefix, rawname) + ".html.md",
-		withoutPackageName(resourcePrefix, rawname) + ".md",
-		// But for some providers, the prefix is included in the name of the doc file
-		rawname + ".html.markdown",
-		rawname + ".markdown",
-		rawname + ".html.md",
-		rawname + ".md",
-	}
+	possibleMarkdownNames := getMarkdownNames(resourcePrefix, rawName, globalInfo)
```
t0yv0, line 252: 👍 I personally love smaller funcs.

pkg/tfgen/docs.go
```diff
@@ -248,6 +263,66 @@ func getMarkdownDetails(sink diag.Sink, repoPath, org, provider string,
 	return markdownBytes, markdownFileName, true
 }
 
+// Create a regexp based replace rule that is bounded by non-ascii letter text.
+//
+// This function is not appropriate to be called in hot loops.
```
t0yv0, line 268: You can lift `r` to `var boundedReplaceRegExp = ..` at the module level, removing the hot loops limitation.

t0yv0: Ah never mind, it's staging the computation already.

pkg/tfgen/docs.go
```diff
@@ -456,29 +531,41 @@ func splitGroupLines(s, sep string) [][]string {
...
+
+type diagsSink interface {
+	debug(string, ...interface{})
+	warn(string, ...interface{})
+	error(string, ...interface{})
 }
 
 type tfMarkdownParser struct {
-	g                *Generator
+	sink             diagsSink
```
t0yv0, line 560: This is good.

pkg/tfgen/docs.go
```diff
@@ -456,29 +531,41 @@ func splitGroupLines(s, sep string) [][]string {
...
+
+		infoCtx: infoContext{
+			language: g.language,
+			pkg:      g.pkg,
+			info:     g.info,
+		},
+		docRules: g.info.DocRules,
 	}
-	return p.parse()
+	return p.parse(markdown)
```
t0yv0, line 550: Nit: does the functionality of replacing the markdown belong in a structure called 'markdownParser' or can it be easily expressed as a transformation on the input that happens before its passed to the parser? I guess which one's easier to test?

pkg/tfgen/docs_test.go
```diff
@@ -898,6 +897,110 @@ throw new Exception("!");
...
+	// Assert that file contents match the expected description.
+	desc := func(fileContents, expected string) testCase {
+		return tc(func(c *testCase) {
+			c.fileContents = []byte(fileContents)
+			c.expected.Description = expected
+		})
+	}
+
+	tests := []testCase{
+		desc(`
```
t0yv0, line 942: This is good consider adding some whitespace and indent for readability.. since funcs an not records are used and without indent it's hard to quickly see which arg is which :)

pkg/tfgen/docs.go
iwahbe: Yes it does.

pkg/tfbridge/info.go
iwahbe: I'm not sure. We might add a `TokenKind: Enum(Resource|DataSource)` indicator. I can also see the benefit of adding a `DocsPath: string` field.

Given how expensive these kinds of changes are, I though making it a struct preemptively would save us pain later.

pkg/tfgen/docs.go
t0yv0: That makes sense then! Ty.

pkg/tfbridge/info.go
t0yv0: Just call it token and mention its resource or datasource token in the comment? I think that' what it is, but internal variables say "resource" de facto it's R or D.

pkg/tfbridge/info.go
```diff
@@ -111,6 +114,84 @@ type ProviderInfo struct {
...
+	// A function to suggest alternative file names for a TF element.
+	//
+	// When the bridge loads the documentation for a resource or a datasource, it
+	// infers the name of the file that contains the documentation. AlternativeNames
+	// allows you to provide a provider specific extension to the override list.
+	//
+	// For example, when attempting to find the documentation for the resource token
+	// aws_waf_instances, the bridge will check the following files (in order):
+	//
+	//	"waf_instance.html.markdown"
```
t0yv0, line 148: 👍

pkg/tfbridge/info.go
```diff
@@ -111,6 +114,84 @@ type ProviderInfo struct {
...
+	//	"aws_waf_instance.html.md"
+	//	"aws_waf_instance.md"
+	//
+	// The bridge will check any file names returned by AlternativeNames before
+	// checking it's standard list.
+	AlternativeNames func(DocsPathInfo) []string
+}
+
+// PreAppendEdits is a MakeEditRules that applies custom edits before default edits.
+func PreAppendEdits(edits ...DocsEdit) MakeEditRules {
```
t0yv0, line 163: This is a funny kind of function to export perhaps users can just inline this.

PR #1096, opened 2023-05-10
Handle empty SchemaInfo during naming

iwahbe:
There was an error in our logic when accounting for pluralizing a field that has a conflict with a set name. We check confirm that the set `.Name != candidate`, but we forgot to check that `.Name != ""`.

Adding the `.Name != ""` check causes the newly added test cases to pass, and unblocks enabling `x.AutoAliasing` for `pulumi-azure`.

### DIFF
--- a/pkg/tfbridge/names.go
+++ b/pkg/tfbridge/names.go
@@ -64,11 +64,12 @@ func isTfPlural(tfs shim.Schema) bool {
 		return false
 	}
 
-	if tfs.Type() != shim.TypeList && tfs.Type() != shim.TypeSet {
+	switch tfs.Type() {
+	case shim.TypeSet, shim.TypeList:
+		return tfs.MaxItems() != 1
+	default:
 		return false
 	}
-
-	return tfs.MaxItems() != 1
 }
 
 func isPulumiMaxItemsOne(ps *SchemaInfo) bool {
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/names.go
```diff
@@ -142,7 +143,16 @@ func terraformToPulumiName(name string, sch shim.SchemaMap, ps map[string]*Schem
...
 		//
 		// The non-bijectivity will be caught at tfgen time and a warning will be emitted.
-		if _, conflict := sch.GetOk(candidate); !conflict || (ps[candidate] != nil && ps[candidate].Name != candidate) {
+
+		_, conflict := sch.GetOk(candidate)
+
+		// A conflict at the `sch` level doesn't necessarily mean that it is
+		// unsafe to pluralize. It is possible that the potentially conflicting
+		// field had its name manually set to another value.
+		conflictSafe := (ps[candidate] != nil &&
```
t0yv0, line 152: Appreciate the clarity of the new var.

pkg/tfbridge/names_test.go
```diff
@@ -248,6 +248,34 @@ func TestBijectiveNameConversion(t *testing.T) {
...
+				"galleryApplications": "gallery_applications",
+			},
+		},
+		{ // Check when SchemaInfo.MaxItemsOne: &false is set explicitly
+			schema: map[string]*schemav2.Schema{
+				"gallery_application":  {Type: schemav2.TypeList},
+				"gallery_applications": {Type: schemav2.TypeList},
+			},
+			info: map[string]*SchemaInfo{
+				"gallery_application":  {MaxItemsOne: ref(false)},
```
t0yv0, line 271: Does this do anything without schemav2.Schema also specifying MaxItems=1? or merely adding MIO=false was tripping something up?

t0yv0 (approved):

pkg/tfbridge/names_test.go
t0yv0: Yeah got it

PR #1097, opened 2023-05-10
Remove broken downstream testing for `pulumi-azure`

iwahbe:
Our downstream testing action doesn't account for patch based providers.


t0yv0 (approved):
Unfortunate but necessary. LGTM.

PR #1102, opened 2023-05-11
Docs include post argument strings

iwahbe:
I'm trying have our resource docs generation include trailing `**Note**` in our generated docs.

This PR changes how our argument parsing works so that all text after argument `n` but before argument `n+1` is part of the description for argument `n`.

This allows us to recover the argument notes such as 
Screenshot 2023-05-11 at 1 43 21 PM
(from https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/sql_firewall_rule)

With this change, the new schema description for `"endIpAddress"` looks like this:
```
Specifies the End IP Address associated with this Firewall Rule. 

> **NOTE:** The Azure feature `Allow access to Azure services` can be enabled by setting `start_ip_address` and `end_ip_address` to `0.0.0.0` which ([is documented in the Azure API Docs](https://docs.microsoft.com/rest/api/sql/firewallrules/createorupdate)).
```

Fixes https://github.com/pulumi/pulumi-azure/issues/1256

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -29,6 +29,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"unicode"
 
 	"github.com/hashicorp/go-multierror"
 	"github.com/spf13/afero"
@@ -819,9 +820,9 @@ func (p *tfMarkdownParser) parseSection(h2Section []string) error {
 		// Now process the content based on the H2 topic. These are mostly standard across TF's docs.
 		switch sectionKind {
 		case sectionArgsReference:
-			p.parseArgReferenceSection(reformattedH3Section)
+			parseArgReferenceSection(reformattedH3Section, &p.ret)
 		case sectionAttributesReference:
-			p.parseAttributesReferenceSection(reformattedH3Section)
+			parseAttributesReferenceSection(reformattedH3Section, &p.ret)
 		case sectionFrontMatter:
 			p.parseFrontMatter(reformattedH3Section)
 		case sectionImports:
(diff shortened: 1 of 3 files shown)


susanev:
@ian i totally misunderstood you before, i didnt realize you specifically meant using .. `>` i think this is fine, but will pick up our quote styles instead of our note styles. so we will eventually need to adjust things in hugo to possibly use that for notes and something else for quotes 🤷🏻 unless anyone else has a better idea

iwahbe:
> @iwahbe i totally misunderstood you before, i didnt realize you specifically meant using .. `>` i think this is fine, but will pick up our quote styles instead of our note styles. so we will eventually need to adjust things in hugo to possibly use that for notes and something else for quotes 🤷🏻 unless anyone else has a better idea

Thanks! My prior is to merge this functionality as is and do any display logic as necessary. We already convert `~>` to `>`, so this is at least consistent.

t0yv0:
This is very practical looking code, it's hard to know if something changes that's untested but I think 🚢 is a good strategy and we can pick up regressions as needed and add more tests incrementally.

Onde question I have - the codebase already references AST-based markdown processing methods like blackfriday and goldmark. Would these help here or make it more convoluted? It looks like at least some of the processing is trying to recover basic markdown structure, and on top of that we're searching the markdown.  An alternative would be searching for patterns in a goldmark tree. This may be overkill though.

t0yv0 (approved):

iwahbe:
> Onde question I have - the codebase already references AST-based markdown processing methods like blackfriday and goldmark. Would these help here or make it more convoluted? It looks like at least some of the processing is trying to recover basic markdown structure, and on top of that we're searching the markdown. An alternative would be searching for patterns in a goldmark tree. This may be overkill though.

Moving to a markdown parser would probably improve code clarity and maybe even speed, but it would require a much larger re-write of docs parsing. Markdown is one of the few languages that is parsable with regex, since it was originally written as a perl script with a bunch of regexes.

t0yv0:
Yeah. We can consider if there's more fixes in here. Speed is not an object but having this easy to edit and get "right" may be.

PR #1116, opened 2023-05-11
Toward SetAutonaming support for PF-based providers

t0yv0:


pf/tfbridge/main.go
```diff
@@ -44,6 +44,16 @@ func Main(ctx context.Context, pkg string, prov ProviderInfo, meta ProviderMetad
 			}
 			return pp.(*provider).marshalProviderInfo(ctx), nil
 		})
+
+	// Apply deferred autonaming, if any.
```
t0yv0, line 48: Deferred AutoNaming is a bit ugly, but it's possible. It could also help us outside of PF as a general strategy... make sure AutoNaming is applied *at the end*. 

If we do not want that then another good alternative is to first take a change that eliminates PF-custom ProviderInfo and unifies everything so that even in PF case, `prov.P` is populated with a schema-only provider, then SetAutonaming does not need custom conditionals for it.

pkg/tfbridge/provider.go
```diff
@@ -1340,12 +1340,28 @@ func generateResourceName(packageName tokens.Package, moduleName string, moduleM
 	return fmt.Sprintf("%s.%s.%s", packageName, moduleName, moduleMemberName)
 }
 
-// SetAutonaming will loop all resources with a name property, and will add an auto-name property.  It will skip
-// those that already have a name mapping entry, since those may have custom overrides set in the resource
-// declaration (e.g., for length).
+// SetAutonaming applies [AutoName] to every required or optional resource property that is literally called "name".
```
t0yv0, line 1343: Do we have good docs elsewhere? I found the existing docs pretty laconic, itching to expand a bit.

t0yv0:
Now this doesn't quite work as written, because of __defaults, I think. Autonaming ends up using DefaultInfo.From "computed" default function. In current naive implementation of PF, DefaultInfo.From is called every time that we're planning a change where the user program does not have any values for the "name" property. This creates a problem for autonaming since it's not deterministic and second pulumi update returns a different result.

Now the legacy bridge was different in how it treated From. I think it would consult a magic __defaults pseudo-property to realize that From need not be called again and instead copy the previous default value from state. 

```
    ~ testbridge:index/testres:AutoNameRes: (update)
        [id=some-id]
        [urn=urn:pulumi:p-it-antons-mac-autoname-p-4d326094::autoname-program::testbridge:index/testres:AutoNameRes::ar]
      ~ name: "ar-5723d91" => "ar-f857cd7"
    --outputs:--
  ~ test_autoname__actual: "ar-5723d91" => "ar-f857cd7"
```

We'd need to figure out what to do here, I'm interested in figuring out a design that does not have a need for __defaults as that was in some ways not well-founded.

Perhaps we can observe that From takes this struct:

```
// PulumiResource is just a little bundle that carries URN, seed and properties around.
type PulumiResource struct {
	URN        resource.URN
	Properties resource.PropertyMap
	Seed       []byte
}
```

We can extend this struct to pass prior state resource.PropertyMap alongside Properties (from the program). 

In the case of autonaming it can implement From callback to copy "name" from state instead of generating a new one.

t0yv0:
Case to think through: if the user has saved a custom value into the state, how to make sure, without __defaults? that this value doesn't stay in the state forever, how can the user replace that value with one actually generated by AutoName machinery.

t0yv0:
This would be cleaner on top of https://github.com/pulumi/pulumi-terraform-bridge/issues/1210

t0yv0:
Closing in favor of https://github.com/pulumi/pulumi-terraform-bridge/pull/1321

PR #1117, opened 2023-05-12
Allow users to override .MaxItems() == 1 for pluralization

iwahbe:
Pulumi automatically pluralizes list and set fields when they will be projected into the Pulumi type system as a list. For backwards compatibility reasons, Pulumi needs the `MaxItemsOne` value to be sticky between updates, this extends to type decisions `List[T]->T` as well as naming decisions `singular->plural`.

That means we cannot take into account the value of `.MaxItems()` when deciding if we should pluralize a name. This PR changes our pluralization logic to consult `.MaxItems()==1` only if `SchemaInfo.MaxItemsOne==nil`.

This brings our pluralization logic inline with our type projection logic.

### DIFF
--- a/pkg/tfbridge/names.go
+++ b/pkg/tfbridge/names.go
@@ -59,23 +59,6 @@ func PulumiToTerraformName(name string, tfs shim.SchemaMap, ps map[string]*Schem
 	return result
 }
 
-func isTfPlural(tfs shim.Schema) bool {
-	if tfs == nil {
-		return false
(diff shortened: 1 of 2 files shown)


iwahbe:
I have confirmed that this doesn't cause any downstream effects for `pulumi-azure`. This PR is necessary for a clean update of `pulumi-cloudflare`.

t0yv0 (approved):

Issue #1129, opened 2023-05-16
Convert `terraform-aws-acm`


justinvp:
Error converting https://github.com/terraform-aws-modules/terraform-aws-acm:

```
pulumi convert --from terraform --language typescript --out pulumi
warning: Plugin converters are currently experimental
        error: main.pp:13,35-56: cannot traverse value of type none;
        error: main.pp:11,23-14,5: cannot assign expression of type list({ domainName: dynamic, validationDomain: none }) to location of type   list(  { domainName: output(string) | string, validationDomain: output(string) | string }
          | output({ domainName: string, validationDomain: string }))
        | output(list({ domainName: string, validationDomain: string }))?: ;
        error: could not generate output program: failed to bind program
```

justinvp:
We think https://github.com/pulumi/pulumi-terraform-bridge/pull/1131 fixes this issue. Re-run converter.

PR #1131, opened 2023-05-16
[converter] Fix inferring default value type

Zaid-Ajaj:
The PR addresses #1129 where a dynamic variable of type `any` is converted to a type of empty object `object({})` because we try to infer the type from the default value. However, this does not work in practice because the type of default value doesn't account for how the value is used in the main program. 

With this PR we no longer try to infer the type of the default value for dynamic variables, _except_ when it is a primitive value!

After the fix, the program compiles to typescript but there is an issue in TS program-gen still. Will be addressed separately


justinvp (approved):
LGTM

Issue #821, opened 2023-02-16
[Plugin Framework] Implement CheckConfig properly


t0yv0:
Recent work on improving CheckConfig for pkg/v3 bridged providers uncovered the importance of getting this method right and its various responsibilities and quirks. The task is to port all the features to Plugin Framework.

Here is what CheckConfig needs to do:

- validate that all required properties and set and inform gracefully of missing properties

- validate that properties have reasonable values (pass validations such as regex-based validations) by invoking ValidateProviderConfig in the underlying provider and interpreting its results

- for historical reasons it needs to use a non-standard json-in-proto encoding when translating between Pulumi PropertyMap structure and protobuf structures (see https://github.com/pulumi/pulumi-terraform-bridge/issues/1032)

- ensure that every property marked as sensitive in the schema (including nested properties) is returned as a secret to the engine; due to the limitations of json-in-proto encoding this is only done as an approximation, so that if a sub-property is batching.sendAfter is secret, the entire batching blob is marked secret to protect the data

## Out of scope

https://github.com/pulumi/pulumi-terraform-bridge/issues/938 - either Configure or CheckConfig (likely the later) would be a good place to honor default values overrides specified by the user in ProviderInfo.

https://github.com/pulumi/pulumi-terraform-bridge/issues/935 tracks supporting PreConfigureCallback. It should probably be called in CheckConfig following the pkg/v3 implementation so the user code hooks can modify provider configuration.

## See also

- Fix https://github.com/pulumi/pulumi-terraform-bridge/issues/1044: explicit provider configuration may panic when outputs are passed by @t0yv0 in https://github.com/pulumi/pulumi-terraform-bridge/pull/1049
- Fix https://github.com/pulumi/pulumi-terraform-bridge/issues/1040: correctly handle secret nested configure properties by @t0yv0 in https://github.com/pulumi/pulumi-terraform-bridge/pull/1038
- Fix #974 provider configuration secrets handling for bridged providers by @t0yv0 in https://github.com/pulumi/pulumi-terraform-bridge/pull/1002
- Re-enable CheckConfig for bridged providers, and enable PreConfigureCallback to make modifications to the provider's configuration by @t0yv0 in https://github.com/pulumi/pulumi-terraform-bridge/pull/991

PR #1142, opened 2023-05-23
[x/muxer]: Log instead of failing on divergent CheckConfig results

t0yv0:
This change makes muxer less strict as it is now able to tolerate providers that do not agree on CheckConfig responses. What used to be a hard error is downgraded to logging a message at debug level.

The practical use case driving this is that before pulumi/pulumi-terraform-bridge#821 is implemented, pf vs v3 based providers may not agree on the secret value markers in the CheckConfig response.

### DIFF
--- a/x/muxer/muxer.go
+++ b/x/muxer/muxer.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/golang/glog"
+	"github.com/golang/protobuf/jsonpb"
 	"github.com/hashicorp/go-multierror"
 	"github.com/pulumi/pulumi/pkg/v3/resource/provider"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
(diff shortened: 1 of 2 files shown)


t0yv0:
This fixes https://github.com/pulumi/pulumi-gitlab/issues/273 

Arguably in the long term this still should not be an error, but perhaps a more visible warning. The blast radius of failing in CheckConfig is very large, it is difficult to be quite so strict here.

x/muxer/muxer.go
```diff
@@ -127,16 +129,16 @@ func (m *muxer) CheckConfig(ctx context.Context, req *rpc.CheckRequest) (*rpc.Ch
...
-			if has {
-				// If different servers return different check values,
-				// signal an error. We don't know how to deal with that.
-				if proto.Equal(existingValue, v) {
-					continue
-				}
-				return nil, fmt.Errorf("config value '%s' mismatch between servers %d and a previous server",
-					k, i)
+			if has && !proto.Equal(existingValue, v) {
+				// If different servers return different values, pick arbitrarily.
```
guineveresaenger, line 133: is there a way for the user to pick explicitly so they don't hit an error later on? disregard if that is an unlikely scenario.

guineveresaenger (approved):
Couple questions and a typo; LGTM otherwise.

x/muxer/muxer.go
```diff
@@ -127,16 +129,16 @@ func (m *muxer) CheckConfig(ctx context.Context, req *rpc.CheckRequest) (*rpc.Ch
...
+				glog.V(9).Infof("[muxer] CheckConfig results do not agree on the '%s' property:"+
+					"\n    server %d: %s"+
+					"\n    server %d: %s"+
+					"\nPicking the server %d response",
```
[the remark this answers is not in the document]
t0yv0: Likewise, having them numbered instead of named makes it difficult. These are the servers subject to muxing. I don't have better vocabulary at the moment but `glog.V(9)` is aimed at maintainers not users so it's not to the same level of scrutiny.

x/muxer/muxer.go
t0yv0: We have only one known use case for this code in production, muxed V3/PF providers. They really should agree.. I'm not sure the users would know what to do with this... Current code picks up the conservative choice (V3) that has been around in production more, that's the right call I think.

Issue #906, opened 2023-03-22
error.go `Error()` ignores `AttributeError` Leading to ambiguous error messages


phillipedwards:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
Currently, the `Error()` method in `pkg/tfshim/diagnositics/error.go` ignores the `AttributePath` property which leads to ambiguous error messages such as 

```
Error: could not validate provider configuration: 1 error occurred:
   * Invalid or unknown key
```

This is helpful in telling the user that there was an error, but also lacks any sort of context as to _what_ property or key cause the error in the first place.

### Affected area/feature

I'd like to see the `AttributePath` utilized when possible to give a less ambiguous error; mainly include the actual key or property that is causing the error in the first place.

t0yv0:
Thanks @phillipedwards this fits perfectly with #886 work we're lining up for upcoming quarter and should be easy to fix, I've added it to the epic.

PR #1153, opened 2023-05-24
Improve error messages for invalid keys

t0yv0:
Fixes pulumi/pulumi-terraform-bridge#906

Before the change setting a config value that was not recognized by a bridged provider resulted in a confusing error message displayed to the user.

After the change the code tries to guess whether an explicit or default provider is being configured and respond better, using this form for default providers:

    Invalid or unknown key. Check `pulumi config get someprovider:someprop`.

Or else this form for explicit providers:

    Invalid or unknown key. Examine values at 'explicitprovider.SomeProp'

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -422,7 +422,7 @@ func validateProviderConfig(
 		}
 	}
 
-	failures := p.formatCheckFailures(urn, true /*isProvider*/, []string{"provider"}, p.config,
+	failures := p.formatCheckFailures(urn, true /*isProvider*/, []string{urn.Name().String()}, p.config,
 		p.info.GetConfig(), errs)
 	return nil, failures
 }
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -32,7 +32,7 @@ import (
 	"github.com/golang/glog"
 	pbempty "github.com/golang/protobuf/ptypes/empty"
 	pbstruct "github.com/golang/protobuf/ptypes/struct"
-	multierror "github.com/hashicorp/go-multierror"
+
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/provider.go
```diff
@@ -409,16 +415,16 @@ func validateProviderConfig(ctx context.Context, p *Provider, config shim.Resour
...
+			return nil, nil
 		}
 	}
 
-	if len(errs) > 0 {
-		return nil, errors.Wrap(multierror.Append(nil, errs...), "could not validate provider configuration")
-	}
-
-	return nil, nil
+	failures := p.formatCheckFailures(urn, true /*isProvider*/, []string{"provider"}, p.config,
```
t0yv0, line 425: Hmm instead of "provider" we can extract the URN name here to narrow it down better.

pkg/tfbridge/provider.go
```diff
@@ -409,16 +415,16 @@ func validateProviderConfig(ctx context.Context, p *Provider, config shim.Resour
 	// Perform validation of the config state so we can offer nice errors.
 	warns, errs := p.tf.Validate(config)
 	for _, warn := range warns {
-		if err := p.host.Log(ctx, diag.Warning, "", fmt.Sprintf("provider config warning: %v", warn)); err != nil {
-			return nil, err
+		logErr := p.host.Log(ctx, diag.Warning, "", fmt.Sprintf("provider config warning: %v", warn))
+		if logErr != nil {
+			glog.V(9).Infof("Failed to log to the engine: %v", logErr)
```
t0yv0, line 420: Downgrading severity of failure in p.host.Log.

pkg/tfbridge/provider_test.go
```diff
@@ -840,32 +839,36 @@ func TestCheckConfig(t *testing.T) {
 
 	t.Run("invalid_config_value", func(t *testing.T) {
 		p := testprovider.ProviderV2()
-		p.Schema["propwithvalidator"] = &schema.Schema{
```
t0yv0, line 843: This was copy-pasted and is actually not being used.

t0yv0:
The original ticket didn't include a way to acceptance test but what I did is rebuilding cloudflare provider on top of this patch and setting `pulumi config set cloudflare:unk unk`  values that are not part of the schema.

aq17 (approved):

PR #1165, opened 2023-05-26
Suggest corrections for misspelled config keys

t0yv0:
This change improves presentation of error messages in providers for the case of the provider encountering configuration keys it does not understand. Known keys are checked for Levenshtein-similar keys to present a "Did you mean .." suggestion to the user. Also, the change fixes presentation of compound_keys; the errors used to mention them as `provider:compound_key` but the correct form is actually `provider:compoundKey`.


aq17 (approved):
❇️

Frassle:
Oh this is nice. We should make sure we have something like this when we get round to typed provider config. Then the engine could do this distance matching covering all provider types, rather than this being a tfbridge feature.

PR #1172, opened 2023-06-01
Add test for builtin inputs

Frassle:


Zaid-Ajaj (approved):
Looks good! Just needs a lint with a few lines that are over 120 characters

PR #1175, opened 2023-06-02
Update to pulumi 3.69

Frassle:


iwahbe (approved):
Tests are 🟢, so :shipit:

Issue #1167, opened 2023-05-30
Update Go to v1.20


rquitales:
### What happened?

When trying to update a provider that uses this library, I ran into an error:
```
../../../../go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/pf@v0.9.2/internal/muxer/muxer.go:75:33: undefined: errors.Join
```

This is because https://github.com/pulumi/pulumi-terraform-bridge/blob/902e9d5f3c181c4a64079d06e2ae9adf0d474696/pf/internal/muxer/muxer.go#L75 uses the new `errors.Join` function which was introduced in Go v1.20. Any downstream projects that uses this function without compiling with Go v1.20 will error out.

We should update our go.mod files to indicate that v1.20 of Go should be used, since this breaks backwards compatibility with Go 1.19 and below.

### Expected Behavior

Build/compilation to be successful, or a warning message indicating that the Golang compiler is a different version to the one declared by this project's go.mod.

### Steps to reproduce

Use `tfgen.MainWithMuxer` on Golang v1.19.X or below. Eg. `make tfgen` in the pulumi-gcp providers repo.

### Output of `pulumi about`

-

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

guineveresaenger:
Thank you for filing! @t0yv0 could you take a look? 🙏

PR #1181, opened 2023-06-02
remove errors.Join for backwards compat with providers using Go < 1.20

guineveresaenger:
This is a quick and dirty fix for #1167.
This PR replaces any usage of errors.Join() (available in Go 1.20+) for an alternative using strings and error formatting.

Not all providers are on Go 1.20 at the moment; this unblocks them for now.

### DIFF
--- a/pf/internal/muxer/muxer.go
+++ b/pf/internal/muxer/muxer.go
@@ -15,7 +15,6 @@
 package muxer
 
 import (
-	"errors"
 	"fmt"
 	"sort"
 	"strings"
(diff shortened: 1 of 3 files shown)


t0yv0:
This breaks ignoreChange tests, at a first glance I can't rule out that it's a semantic change - unfortunately this can't go in.

Perhaps could have a look at "github.com/hashicorp/go-multierror" used elsewhere in this repository as something predating "errors" that serves a similar person.

Alternatively if we can figure out why the tests are failing.

iwahbe:
Closing in favor of https://github.com/pulumi/pulumi-terraform-bridge/pull/1233.

Issue #1145, opened 2023-05-24
Conversion error: `unknown property '0' among [foo];`


justinvp:
Error converting https://github.com/terraform-aws-modules/terraform-aws-eks:

```
$ pulumi convert --from terraform --language typescript --out pulumi
error: main.pp:54,41-44: unknown property '0' among [endpointPublicAccess publicAccessCidrs securityGroupIds subnetIds vpcId clusterSecurityGroupId endpointPrivateAccess];
error: node_groups.pp:35,75-78: unknown property '0' among [data];
error: node_groups.pp:294,219-222: unknown property '0' among [endpointPrivateAccess endpointPublicAccess publicAccessCidrs securityGroupIds subnetIds vpcId clusterSecurityGroupId];
error: node_groups.pp:399,220-223: unknown property '0' among [publicAccessCidrs securityGroupIds subnetIds vpcId clusterSecurityGroupId endpointPrivateAccess endpointPublicAccess];
```

Errors converting https://github.com/aws-samples/hub-and-spoke-with-inspection-vpc-terraform:

```
$ pulumi convert --from terraform --language typescript --out pulumi
error: main.pp:82,20-24: unknown property 'key' among [value];
error: main.pp:125,20-24: unknown property 'key' among [value];
```

justinvp:
A user was trying to convert https://github.com/aws-ia/terraform-aws-eks-blueprints/tree/main/examples/multi-tenancy-with-teams to typescript and got similar errors:

```
[...]

PR #1182, opened 2023-06-04
Remove index from collections that have max items = 1

Zaid-Ajaj:
This PR addresses #1145 where we no longer index a collection (removing the index from the traversal) when the collection is marked with max items = 1. 

Also, when the assigning a resource attribute to an expression of which the result is a collection marked with max items = 1, we no longer apply `projectListToSingleton`

### DIFF
--- a/pkg/tf2pulumi/convert/tf.go
+++ b/pkg/tf2pulumi/convert/tf.go
@@ -935,6 +935,16 @@ func camelCaseName(name string) string {
 	return name
 }
 
+// Returns whether the fully qualified path is being applied for a property
+// for which we can get the `maxItemsOne` field.
+func (s *scopes) isPropertyPath(fullyQualifiedPath string) bool {
+	if fullyQualifiedPath == "" {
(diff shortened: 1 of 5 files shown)


pkg/tf2pulumi/convert/tf.go
```diff
@@ -935,6 +935,16 @@ func camelCaseName(name string) string {
 	return name
 }
 
+// Returns whether the fully qualified path is being applied for a property
+// for which we can get the `maxItemsOne` field.
```
Frassle, line 939: I'd remove this second line of the comment. Presumably this isn't unique to "maxItemsOne" and we may look up other attributes of properties in the future.

Frassle (approved):
Looks reasonable, tests seem happy

pkg/tf2pulumi/convert/tf.go
Zaid-Ajaj: Done ✅

Issue #866, opened 2023-03-06
Rollout DiffStrategy=PlanState across providers


t0yv0:
[Internal Design Doc](https://docs.google.com/document/d/1Zupf1LsEXKIIhtMIgsL7TryQYJdVNdzes-XlLjxvnhY/edit?usp=sharing)

The improved Diff implementation is under a feature flag DiffStrategy to make it easier to roll out across Pulumi providers gradually while mitigating any unexpected consequences. The specific risk with the new Diff implementation is that it's a sensitive method. For example, it that it might introduce unexpected replace plans to existing infrastructure on provider upgrades. The plan is:

- [ ] DiffStrategy=TryPlanState -> PlanState on 5 ecosystem providers
   - [x] hcloud
       - noting PlanState fixes https://github.com/pulumi/pulumi-hcloud/issues/170
       - but introduces https://github.com/pulumi/pulumi-hcloud/issues/175
   - [ ] aiven
       - https://github.com/pulumi/pulumi-terraform-bridge/issues/911
- [ ] DiffStrategy=TryPlanState on pulumi-gcp, pulumu-azure
  - [ ] gcp  
  - [ ] azure  
- [ ] DiffStrategy=PlanState on pulumi-gcp, pulumi-azure
  - [ ] gcp  
  - [ ] azure  
- [ ] DiffStrategy=TryPlanState on pulumi-aws
- [ ] DiffStrategy=PlanState on pulumi-aws
- [ ] pulumi-aws removes the workaround apply to make https://github.com/pulumi/pulumi-terraform-bridge/issues/791 work for the end-users, namely https://github.com/hashicorp/terraform-provider-aws/pull/25708 reintroduced into forked pulumi-aws upstream and pulumi-aws rebuilt with unmodified upstream
- [ ] DiffStrategy=PlanState by default, remove the flag

ericrudder:
maybe explicity try gcp/azure as step 3/4 (try, plan) before aws?

i'm not really trying to add "more steps" however, *if* you think the risk
is large, this might be another baby step to consider, since aws really is
the biggest customer risk

t0yv0:
Great idea, I'll ask again in the providers team channel to confirm. --A

t0yv0:
Check on https://github.com/pulumi/pulumi-aws/issues/2475#issuecomment-1517892836 possibly related here.

marco-m:
[...]

PR #1189, opened 2023-06-05
Fix DiffStrategy=PlanState handling of unknowns

t0yv0:
This change is aimed at fixing pulumi/pulumi-hcloud#175

This provider is early in adopting a feature-flagged change in https://github.com/pulumi/pulumi-terraform-bridge/issues/866 and uncovered an issue.

Due to technicalities this project is linking code from TF plugins and TF CLI proper and they do not agree on the dependency "github.com/hashicorp/go-cty/cty" vs "github.com/zclconf/go-cty/cty/json", and it looks like the bug was in the "trivial" adapter between cty.Value objects from these two packages. 

This is now fixed by expanding the adapter to handle each case explicitly and adding some turnaround cases.

### DIFF
--- a/pkg/tfshim/sdk-v2/proposed_new.go
+++ b/pkg/tfshim/sdk-v2/proposed_new.go
@@ -15,8 +15,6 @@
 package sdkv2
 
 import (
-	"reflect"
-
 	hcty "github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/zclconf/go-cty/cty"
@@ -39,7 +37,7 @@ func proposedNew(res *schema.Resource, prior, config hcty.Value) (hcty.Value, er
 func htype2ctype(t hcty.Type) cty.Type {
 	// Used t.HasDynamicTypes() as an example of how to pattern-match types.
 	switch {
-	case reflect.DeepEqual(t, hcty.NilType):
+	case t == hcty.NilType:
 		return cty.NilType
 	case t == hcty.DynamicPseudoType:
 		return cty.DynamicPseudoType
@@ -176,7 +174,7 @@ func hcty2ctyWithType(ty cty.Type, val hcty.Value) cty.Value {
 func ctype2htype(t cty.Type) hcty.Type {
 	// Used t.HasDynamicTypes() as an example of how to pattern-match types.
 	switch {
-	case reflect.DeepEqual(t, cty.NilType):
+	case t == cty.NilType:
 		return hcty.NilType
 	case t == cty.DynamicPseudoType:
 		return hcty.DynamicPseudoType
--- a/pkg/tfshim/sdk-v2/proposed_new.go
+++ b/pkg/tfshim/sdk-v2/proposed_new.go
@@ -127,6 +127,8 @@ func hcty2ctyWithType(ty cty.Type, val hcty.Value) cty.Value {
 			newAVs[name] = newAV
 		}
 		return cty.ObjectVal(newAVs)
+	case ty.IsCapsuleType():
+		contract.Assertf(false, "Capsule types are not yet supported")
 	}
 	contract.Assertf(false, "match failure on hcty.Value: %v", val.GoString())
 	return cty.DynamicVal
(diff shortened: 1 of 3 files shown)


pkg/tfshim/sdk-v2/proposed_new.go
```diff
@@ -52,39 +47,189 @@ func htype2ctype(t hcty.Type) (cty.Type, error) {
...
 	}
+	contract.Assertf(false, "match failure on hcty.Value: %v", val.GoString())
+	return cty.DynamicVal
+}
 
-	typeJSON, err := ctyjson.MarshalType(v.Type())
-	if err != nil {
-		return hcty.False, err
+func ctype2htype(t cty.Type) (hcty.Type, error) {
+	if t.GoString() == "cty.NilType" {
```
iwahbe, line 138: Can you leave a comment on why this is necessary? I'm not 100% sure I understand.

pkg/tfshim/sdk-v2/proposed_new.go
```diff
@@ -52,39 +47,189 @@ func htype2ctype(t hcty.Type) (cty.Type, error) {
...
+	if val.IsKnown() && val.Equals(cty.NilVal).True() {
+		return hcty.NilVal
+	}
+	ty, err := ctype2htype(val.Type())
+	contract.AssertNoErrorf(err, "unexpected error in ctype2htype")
+	return cty2hctyWithType(ty, val)
+}
+
+func cty2hctyWithType(ty hcty.Type, val cty.Value) hcty.Value {
+	// Pattern-matching below is based on how hcty.Transform is written.
```
iwahbe, line 158: What does "is based on" mean? Are we matching structure? the set of types? copy & paste?

pkg/tfshim/sdk-v2/proposed_new_test.go
```diff
@@ -72,3 +73,43 @@ func TestProposedNew(t *testing.T) {
 		})
 	}
 }
+
+func TestCtyTurnaround(t *testing.T) {
+	type testCase struct {
+		value cty.Value
+	}
+
+	testCases := []testCase{
```
iwahbe, line 82: Can we add some tests for nested types: `cty.ListValEmpty(cty.ListValEmpty(cry.NumberIntVal(42)))`?

pkg/tfshim/sdk-v2/proposed_new.go
t0yv0: Removing this panics in the turnaround test... Be my guest.. Some kind of insanity in the underlying lib I think.

pkg/tfshim/sdk-v2/proposed_new.go
t0yv0: The most difficult thing, coming from OCaml languages, is making sure every branch is covered. I've looked at hcty.Transform as an example of covering every container branch. After that I've extended this to primitives (str/num/bool), and I think we do not need to support Capsule types in the Pulumi context. Those are all the cases I could come up with.

pkg/tfshim/sdk-v2/proposed_new_test.go
t0yv0: Sure.

pkg/tfshim/sdk-v2/proposed_new.go
iwahbe: This seems less brittle and allows `go test ./... --run=TestCtyTurnaround` to pass (if thats the turnaround test your talking about).
```suggestion
	if t == hcty.NilType {
```

t0yv0: t.Equals(hcty.NilType) panics.

t0yv0:
PTAL @iwahbe this should be good to go.

pkg/tfshim/sdk-v2/proposed_new.go
iwahbe: That makes sense, since `t` and `hcty.NilType` are from different sets of types.

pkg/tfshim/sdk-v2/proposed_new.go
```diff
@@ -16,75 +16,296 @@ package sdkv2
...
+	case t.IsPrimitiveType():
+		switch {
+		case t.Equals(hcty.Bool):
+			return cty.Bool
+		case t.Equals(hcty.String):
+			return cty.String
+		case t.Equals(hcty.Number):
+			return cty.Number
+		default:
+			contract.Assertf(false, "Match failure on hcty.Type with t.IsPrimitiveType()")
```
iwahbe, line 53: We have [contract.Failf](https://github.com/pulumi/pulumi/blob/2f3e1657eb37b6c0980765a4e0055e5655b35c0d/sdk/go/common/util/contract/fail.go#L31) for this
```suggestion
			contract.Failf("Match failure on hcty.Type with t.IsPrimitiveType()")
```

iwahbe (approved):

PR #1217, opened 2023-06-13
Remove `pf/tfbridge.ProviderInfo`

### LINKED ISSUE #1197: `ComputeDefaults` and `AutoAliasing` don't work on pure PF providers
## Hello!

## Issue details

`ComputeDefaults` and `AutoAliasing` both work by examining `tfbridge.ProviderInfo.P` (`.P`). This works for sdk-v2 providers because `.P` holds the provider (and thus all its schema information). This also works for muxed providers since `MainWithMuxer` expects a value for `.P` with a correct schema (and `pfbridge.MuxShimWithPF` providers one).

Pure PF based providers leave `.P` empty, so `ComputeDefaults` fails.

### Affected area/feature

I believe we should fold this change in with removing `pf/tfbridge.ProviderInfo` by embedding `pf/tfbridge.ProviderInfo.NewProvider` into the `.P`. Either way, we need a schema-correct value to stick in `.P`.

### LINKED ISSUE #1210: Cleanup interfacing with Plugin Framework
## Hello!

## Issue details

Consider removing intermediate pf.ProviderInfo types and interfacing with Plugin Framework more directly similar to how traditional sdk/v2 providers were shimmed. This would be a breaking change to users, and here is how it would look like:

Before:

```go
    package myprovider

    import (
        _ "embed"
        "github.com/hashicorp/terraform-plugin-framework/provider"
        pf "github.com/pulumi/pulumi-terraform-bridge/pf/tfbridge"
    )

    //go:embed cmd/pulumi-resource-myprovider/bridge-metadata.json
    var bridgeMetadata []byte

    func MyProvider() pf.ProviderInfo {
        info := tfbridge.ProviderInfo{
            Name:    "myprovider",
            Version: "1.2.3",
            Resources: map[string]*tfbridge.ResourceInfo{
                "myresource": {Tok: "myprovider::MyResource"},
            },
            MetadataInfo: tfbridge.NewProviderMetadata(bridgeMetadata),
        }
        return pf.ProviderInfo{
            ProviderInfo: info,
            NewProvider: func() provider.Provider {
                return nil // TODO fill in Terraform Provider from Step 1
            },
        }
    }

```

After:

```go
    package myprovider

    import (
        _ "embed"
        "github.com/hashicorp/terraform-plugin-framework/provider"
        pf "github.com/pulumi/pulumi-terraform-bridge/pf/tfbridge"
    )

    //go:embed cmd/pulumi-resource-myprovider/bridge-metadata.json
    var bridgeMetadata []byte

    func MyProvider() tfbridge.ProviderInfo {
        newProvider := func() provider.Provider {
                return nil // TODO fill in Terraform Provider from Step 1
        }
        return tfbridge.ProviderInfo{
            P:        pf.NewProvider(newProvider),
            Name:    "myprovider",
            Version: "1.2.3",
            Resources: map[string]*tfbridge.ResourceInfo{
                "myresource": {Tok: "myprovider::MyResource"},
            },
            MetadataInfo: tfbridge.NewProviderMetadata(bridgeMetadata),
        }
    }

```

We should consider taking this breaking change before the code is more widely used.

iwahbe:
Fixes #1210 
Fixes #1197

This PR removes the `pf/tfbridge.ProviderInfo` struct in favor of embedding `NewProvider` in `pkg/tfbridge.ProviderInfo.P` via the newly introduced `pf/tfbridge.ShimProvider(github.com/hashicorp/terraform-plugin-framework/provider.Provider) shim.Provider` function.

This mirrors how muxed providers are created.

This is a breaking change and needs to be called out in our release notes.

This changes the bridge from accepting a `func() github.com/hashicorp/terraform-plugin-framework/provider.Provider` to accepting a plain `github.com/hashicorp/terraform-plugin-framework/provider.Provider`, contrary to the "after" example in #1210. To allow `pkg/tfbridge.ProviderInfo.P` to fulfill its contract as a `shim.Provider` (#1197), we can't make any strong guarantees on when we call the proposed NewProvider function. Since we will be treating this function like its result, I decided to just replace the function with its result.


AaronFriel (approved):
This is great; I looked at doing this yesterday and didn't see the ShimProvider constructor.

PR #1219, opened 2023-06-13
chore: Update pf dependencies

AaronFriel:
Following the release of v3.50.0, update PF bridge dependencies to latest pulumi/pkg and bridge.


AaronFriel:
The linter is also clearly wrong; it says "fmt" is not used and it is - line 118. 

I believe I've seen issues like this before on pulumi-yaml, but I don't remember what the fix was.

PR #1226, opened 2023-06-20
Allow state upgraders to modify ID

### LINKED ISSUE #923: bug: Terraform Bridge does not reliably perform Terraform state upgrades
@t0yv0 Found the issue. Inside the Pulumi stack the resource id is lower case on the `webHooks` element:

`/subscriptions/e5116c25-7bc9-4e96-9c3e-88257f68db7c/resourceGroups/RG-WKB-Test/providers/Microsoft.ContainerRegistry/registries/acr4189bf944305c36f/webhooks/b1ec43f3ce50aab6`

image

The parser inside the AzureRM Provider expects `webHooks`:

image

The team introduced the change with:

https://github.com/hashicorp/terraform-provider-azurerm/commit/821ea0a145254b3c2bfed79bd4e331c499f6ae45
https://github.com/hashicorp/terraform-provider-azurerm/pull/19507

No issue when using Terraform,  because the team added a (silent) state migration.

image

Pulumi isn't aware of state migrations because it handles the state on it's own. But in such cases this is a real problem, because in the current case, I've to update the Pulumi state manually and correct the casing from `webhooks` to `webHooks`.

_Originally posted by @tmeckel in https://github.com/pulumi/pulumi-terraform-bridge/issues/720#issuecomment-1488617011_
            
Affected versions: Upgrade from Azure Classic 5.16.0 to 5.38.0

**Steps to reproduce:**

1. Deploy Azure Container Registry with a WebHook using 5.16.0
2. Upgrade to Version 5.38.0
3. Try to change the infrastructure (e.g. add a Storage Account). `pulumi pre -r` will show an error about a the missing `webHooks` element.

t0yv0:
Before this change, Pulumi would correctly run migrations on the resource state if those are specified in the upstream provider. However, in the rare situation where the migrations would intentionally modify the ID field, the bridged provider would ignore this modification. After the change, migrations of the ID field are respected.

Fixes #923.

### DIFF
--- a/pkg/tests/regress_923_test.go
+++ b/pkg/tests/regress_923_test.go
@@ -1,0 +1,166 @@
+// Copyright 2016-2023, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"context"
+	"fmt"
+	"strings"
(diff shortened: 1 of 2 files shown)


iwahbe (approved):
Looks good to me. LGTM

PR #1230, opened 2023-06-22
Extract inputs from outputs with max items one

iwahbe:
Fixes https://github.com/pulumi/pulumi-cloudflare/issues/436

### DIFF
--- a/pkg/tfshim/schema/schema.go
+++ b/pkg/tfshim/schema/schema.go
@@ -145,7 +145,11 @@ func (m SchemaMap) Len() int {
 }
 
 func (m SchemaMap) Get(key string) shim.Schema {
-	return m[key]
+	s, ok := m[key]
+	if !ok {
+		panic("Missing key: " + key)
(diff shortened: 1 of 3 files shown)


pkg/tfshim/schema/schema.go
```diff
@@ -145,7 +145,11 @@ func (m SchemaMap) Len() int {
 }
 
 func (m SchemaMap) Get(key string) shim.Schema {
-	return m[key]
+	s, ok := m[key]
+	if !ok {
+		panic("Missing key: " + key)
+	}
```
iwahbe, line 151: This mirrors the error message for other `Get` calls. See:
https://github.com/pulumi/pulumi-terraform-bridge/blob/dde6f7ec58ab7d00a3a26c3bba867e5be527ebf9/pf/internal/schemashim/object_pseudoresource.go#L109-L115

t0yv0 (approved):

PR #1232, opened 2023-06-22
Finalize expiremental tokens

### LINKED ISSUE #1180: Promote experimental automations (token mapping, aliasing) to GA

iwahbe:
Fixes #1180 

Changes here are:
- Move the package from `x` to `tfbridge` & `tfbridge/tokens`.
- Rename `x.ComputeDefaults` to `tfbridge.ProviderInfo.ComputeTokens`.
- Rename `x.AutoAliasing` to `tfbridge.ProviderInfo.ApplyAutoAliases`.
- Add `Must*` versions of the new methods.
- Remove the second argument of `tfbridge.ProviderInfo.ApplyAutoAliases` since it can be derived from the first.
- Moved token strategies to `tfbridge/tokens`.
- Add deprecation warnings to all exposed elements of the `x` package.


pkg/tfbridge/x/apply.go
```diff
@@ -25,6 +25,9 @@ import (
 )
 
 // Add mapped resources and datasources according to the given strategies.
+//
+// Deprecated: This item has been moved to
+// "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/defaults.ApplyStrategy"
 func ComputeDefaults(info *b.ProviderInfo, opts DefaultStrategy) error {
```
t0yv0, line 31: Can it call the new implementation instead of leaving the duplicated code body here?

pkg/tfbridge/defaults/package.go
```diff
@@ -0,0 +1,5 @@
+// Experimental extensions to the tfbridge module.
+//
+// API exposed by this package is unstable and subject to change. As it stabilizes it will be moved to the parent
```
t0yv0, line 3: Is this comment accurate?

pkg/tfbridge/defaults/unmappable.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// Indicate that a token cannot be mapped.
+type UnmappableError struct {
+	TfToken string
+	Reason  error
+}
+
+func (err UnmappableError) Error() string {
+	return fmt.Sprintf("'%s' unmappable: %s", err.TfToken, err.Reason)
```
t0yv0, line 29: It would be helpful to know if this stops tfgen or the software recovers from this error somehow, and what the user should do with this. Perhaps the idea is that the user can:

- Provide a token manually before applying defaults strategies
- Mark the resource or datasource to be skipped explicitly, thus avoiding the error

If that's the case it would be wonderful to indicate these courses of action in the error that surfaces.

pkg/tfbridge/defaults/tokens.go
```diff
@@ -0,0 +1,928 @@
...
+	defaultMinimumModuleSize    = 5
+	defaultMinimumSubmoduleSize = 5
+)
+
+// A function that joins a module and name into a pulumi type token.
+//
+// For example:
+//
+//	func(module, name string) (string, error) {
+//	 return fmt.Sprintf("pkgName:%s:%s", module, name), nil
```
t0yv0, line 43: Should it be pkg_name in the example (typical for TF?)

pkg/tfbridge/defaults/tokens.go
```diff
@@ -0,0 +1,928 @@
...
+//	}
+type MakeToken func(module, name string) (string, error)
+
+// A strategy that assigns all tokens to the same module.
+//
+// For example:
+//
+//	rStrat, dStrat := TokensSingleModule("pkgName_", "index", finalize)
```
t0yv0, line 51: Should this be `TokensSingleModule("pkgName_", "index", MakeStandardToken)`?

Should this use functional options pattern and default the finalize to MakeStandardToken?

pkg/tfbridge/defaults/tokens.go
```diff
@@ -0,0 +1,928 @@
...
+				return &b.DataSourceInfo{Tok: tokens.ModuleMember(tk)}, nil
+			}, camelCase),
+	}
+}
+
+// A strategy for assigning tokens to a hand generated set of modules with an arbitrary
+// mapping from TF modules to Pulumi modules.
+//
+// If defaultModule is "", then the returned strategies will error on not encountering a matching module.
+func TokensMappedModules(
```
t0yv0, line 119: Again, quick example would be great.

Just curious - is generalizing `defaultModule string, modules map[string]string` to `func(string)string` not feasible for fundamental reasons? I thought perhaps if the system has access to the entire set of TF tokens upfront it could do this:

```
m := map[string]string{}
for _, t := range allTokens {
   m[t] = f(t)
}
return m
```

To convert a functional assignment strategy to a map.

pkg/tfbridge/defaults/tokens.go
```diff
@@ -0,0 +1,928 @@
...
+				return &b.DataSourceInfo{Tok: tokens.ModuleMember(tk)}, nil
+			}, transform),
+	}
+}
+
+// Convert a Terraform token to a Pulumi token with the standard mapping.
+//
+// The mapping is
+//
+//	(pkg, module, name) => pkg:module/lowerFirst(name):name
```
t0yv0, line 165: This doctoring is great.

pkg/tfbridge/defaults/tokens.go
```diff
@@ -0,0 +1,928 @@
...
+	//
+	//	pkg:mod:Sub1A, pkg:modSub2:B, pkg:modSub2C, pkg:mod:Sub3D
+	//
+	// < 0 -> don't bin into submodules. Only the most common prefix will be used.
+	// = 0 -> apply the default value.
+	// > 0 -> set the value.
+	MimimumSubmoduleSize int
+}
+
+// A strategy to infer module placement from global analysis of all items (Resources & DataSources).
```
t0yv0, line 209: I'm ok not documenting it in detail as it's for "advanced use cases". Can state that most users could probably get away with simpler strategies - and say this in the doc string.

pkg/tfbridge/defaults/tokens.go
```diff
@@ -0,0 +1,928 @@
...
+}
+
+type fieldHistory struct {
+	MaxItemsOne *bool `json:"maxItemsOne,omitempty"`
+
+	Fields map[string]*fieldHistory `json:"fields,omitempty"`
+	Elem   *fieldHistory            `json:"elem,omitempty"`
+}
+
+// Apply resource and function aliases to maintain backwards compatibility with previous
```
t0yv0, line 537: You can just cross-reference `[ApplyAliases]` here, godoc understands links.

pkg/tfbridge/defaults/tokens.go
```diff
@@ -0,0 +1,928 @@
...
+	err := ApplyAliases(providerInfo)
+	contract.AssertNoErrorf(err, "Failed to apply aliases")
+}
+
+// Apply resource and function aliases to maintain backwards compatibility with previous
+// versions of this provider.  Apply MaxItemsOne history to maintain backwards
+// compatibility with previous versions of this provider.
+//
+// All history is dropped when there is a major version change.
+func ApplyAliases(providerInfo *b.ProviderInfo) error {
```
t0yv0, line 554: Now there's a ton of subtlety going on here and I'd take the time to write it out.

```
Automatically applies backwards compatibility best practices.

Specifically, ApplyAliases may perform the following actions:

- Call [ResourceInfo.RenameResourceWithAlias] 
- Edit [ResourceInfo.Aliases]
- Edit [SchemaInfo.MaxItemsOne]

The goal is to always maximize backwards compatibility and reduce breaking changes for the users of the Pulumi providers. 

Resource aliases help mask TF provider resource renames or changes in mapped tokens so older programs continue to work.  See [ResourceInfo.RenameResourceWithAlias] and [ResourceInfo.Aliases] for more details.

[SchemaInfo.MaxItemsOne] changes are also important because they involve flattening and pluralizing names. Collections (lists or sets) marked with MaxItems=1 are projected as scalar types in Pulumi SDKs. Therefore changes to the MaxItems property may be breaking the compilation of programs as the type changes from `T to List[T]` or vice versa. To avoid these breaking changes, this method undoes any upstream changes to MaxItems using [SchemaInfo.MaxItemsOne] overrides. This happens until a major version change is detected, and then overrides are cleared. Effectively this makes sure that upstream MaxItems changes are deferred until the next major version.

Implementation note: to operate correctly this method needs to keep a persistent track of a database of past decision history. This is currently done by doing reads and writes to `providerInfo.GetMetadata()`, which is assumed to be persisted across provider releases. The bridge framework keeps this information written out to an opaque `bridge-metadata.json` blob which is expected to be stored in source control to persist across releases.
```

t0yv0: The name ApplyAliases is also a bit surprising given that it also applies MaxItemsOne edits. Naming is hard though. ApplyAliesesAndMaxItems is a mouthful. ApplyBackwardsCompatibility is too ambitious. OK with leaving as is given bigger docstring but if you have ideas..

pkg/tfbridge/defaults/strategy.go
```diff
@@ -0,0 +1,35 @@
...
+package defaults
+
+import (
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+)
+
+// A generic remapping strategy.
+type ElementStrategy[T tfbridge.ResourceInfo | tfbridge.DataSourceInfo] func(tfToken string) (*T, error)
+
+// Describe the mapping from resource and datasource tokens to Pulumi resources and
```
t0yv0, line 24: ```from TF resource and datasource tokens...```

pkg/tfbridge/defaults/apply.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/hashicorp/go-multierror"
+
+	b "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+)
+
+// Add mapped resources and datasources according to the given strategies.
+func ApplyStrategy(info *b.ProviderInfo, opts Strategy) error {
```
t0yv0, line 29: I would splurge on the doc string here as well.


```
Assigns resource and data source tokens unless already specified by the user. 

Context: bridging a provider requires every TF resource and data source have a matching entry in [ProviderInfo.Resources] and [ProviderInfo.DataSources]. Each entry needs to minimally specifying the Pulumi token in [ResourceInfo.Tok]. The mapping entries can be done by hand for smaller provides but become a chore for providers with 1000s of entries. 

ApplyStrategy scans TF resources and datasources to populate missing entries as needed with automatically computed Pulumi tokens. 

ApplyStrategy always respects and does not modify pre-existing entires. The user can therefore manually override the token decision by providing ResourceInfo or DataSourceInfo entry prior to calling ApplyStrategy. 

<TBD discuss on how to prevent resources from being mapped>
```

guineveresaenger:
What about naming it `tokens`? The we could have `tokens.SingleModule` etc.

I am also tempted by the idea of having a separate package for aliasing and general post-conversion doctoring - I have definitely confused it before!

aq17:
> What about naming it `tokens`? The we could have `tokens.SingleModule` etc.
> 
> I am also tempted by the idea of having a separate package for aliasing and general post-conversion doctoring - I have definitely confused it before!

`tokens` and `alias` packages?

pkg/tfbridge/x/apply.go
iwahbe: No, since it would be a cyclical dependency between `tfbridge/x` and `tfbridge`.

pkg/tfbridge/defaults/apply.go
iwahbe: ProviderInfo.IgnoreMappings is respected. If your token is there, it won't be mapped. I added that to the doc comment.

iwahbe:
Responding to feedback, I:
- Moved `x.ComputeDefaults` to `tfbridge.ProviderInfo.ComputeTokens`.
- Moved `x.AutoAliasing` to `tfbridge.ProviderInfo.ApplyAutoAliasing`.
- Moved token strategies to `tfbridge/tokens`.

iwahbe:
Can I get another look @t0yv0 @guineveresaenger @aq17?

pkg/tfbridge/defaults/package.go
iwahbe: It's no longer relevant. 😄

pkg/tfbridge/defaults/tokens.go
iwahbe: No, because this is a Pulumi token, not a TF token.

pkg/tfbridge/defaults/tokens.go
iwahbe: It can, but that is just implementing a new strategy. Having a pre-computed map allows us to implement more complicated logic that otherwise the user would need to implement.

For example:

```go
TokensMappedModules("index", map[string]string{
    "mod": "mod",
    "mod_nest": "nested",
}, MakeStandardToken("example"))
```

Consider the token `"mod_nest_one"`. `TokensMappedModules` handles the logic to determine that `mod_nest_one` maps to `example:nested/one:One` instead of `example:mod/nestOne:NestOne`, even though without `mod_nest` as a prefix, it would map to `example:mod/nestOne:NestOne`.

---

We could expose a helper function that makes it easier to implement your own strategy but it would not replace `TokensMappedModules`.

t0yv0 (approved):
🚢

aq17 (approved):
LGTM. 
Do we want to set a tentative timeline for officially removing the `x` package/ updating all of our providers to use the stable API's? Fortunately `x` is not being used anywhere externally as far as I'm aware

Issue #1234, opened 2023-06-22
GCP Incompatible with downstream tests


iwahbe:
### What happened?

All downstream tests are failing since https://github.com/pulumi/pulumi-gcp/pull/1103.

### Expected Behavior

Downstream tests pass on valid PRs.

### Steps to reproduce

NA

### Output of `pulumi about`

NA

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Issue #1166, opened 2023-05-26
Consider replacing downstream checks


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Instead of downstream checks consider automating opening draft PRs in select repositories that upgrade the repository to the WIP bridge version. Flow:

- Maintainer engages chat-ops on a bridge PR with `/try-provider-upgrades'
- Automation stands up N WIP PRs across our major providers that link to the in-progress bridge version
- The PRs cross-link to the original PR

Why:

- While not being checks in the sense of being able to block CI, this flow is going to be send a much more interesting signal to the bridge maintainers since it will run all the usual complete provider quality checks in exactly the same way as the provider is doing, and it will simulate an actual bridge update

- There were concerns raised that doing this on every bridge commit is too expensive; being selective here helps maintainers be picky about which changes we regress through complete what-if checks of this sort

- Downstream checks often drift from the repository structure; this would assume much less about the repositories and hopefully be more robust

guineveresaenger:
I really like this! Is there a feature proposal doc or related discussion?

t0yv0:
@AaronFriel noted we should probably still test a few providers on every PR, if we can't afford to test all of them.

t0yv0:
Notes from @aq17 : currently update-provider doesn't have a switch to update to pre-rel bridge but it won't be hard to add.

t0yv0:
Extra options to propagate:

1. --target-bridge-version
2. --pr-reviewers to avoid tagging team by default
3. --pr-body to cross-link to the original PR in the bridge

Propagate where:

1. add to upgrade-provider CLI
2. add to pulumi-upgrade-provider-action
3. add to upgrade-bridge workflows such as https://github.com/pulumi/pulumi-aiven/actions/runs/6387497257/workflow in individual provider repositories
4. finally add an orchestrating workflow in the bridge that invokes the individual upgrade-bridge workflows with the right parameters when asked to do so

PR #1235, opened 2023-06-22
Remove GCP from downstream testing

iwahbe:
This is necessary since our downstream testing action does not accommodate patched providers. GCP introduced patching in pulumi/pulumi-gcp#1103.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1234


iwahbe:
See https://github.com/pulumi/pulumi-terraform-bridge/issues/1166

t0yv0 (approved):

PR #1240, opened 2023-06-23
Update pulumi dependency

Frassle:
Fixes up terrform-bridge to build against a breaking API change in the latest version of pulumi/sdk. See https://github.com/pulumi/pulumi/pull/13139.


t0yv0:
Nice thank you! This breaks downstream tests temporarily but is worth it. We wanted to replace those anyway.

t0yv0 (approved):

Frassle:
> This breaks downstream tests temporarily but is worth it.

Is that because the downstream test doesn't update the pulumi dependency to match the version being used in terraform-bridge? I wonder if you could just remove the direct references from the go.mod files and just let the transitive pick up from terraform-bridge pull in sdk and pkg?

AaronFriel (approved):

PR #1248, opened 2023-06-28
Add schema post-processing hook to schema generator

mikhailshilkov:
Adds a callback function that can be passed to a `Generator` from a provider to massage the schema spec before its gets to SDK generation.

Specifically, I need it in order to be able to reshape the WafV2 types from multi-layer copied types to actual recursive types.

I'm curious if folks see a better way to express such an operation.

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -57,6 +57,11 @@ const (
 )
 
 type Generator struct {
+	// Optional function to post-process the generated schema spec after
+	// the bridge completed its original version based on the TF schema.
+	// A hook to enable custom schema modifications specific to a provider.
+	SchemaPostProcessor func(spec *pschema.PackageSpec)
+
 	pkg              tokens.Package        // the Pulum package name (e.g. `gcp`)
 	version          string                // the package version.
 	language         Language              // the language runtime to generate.
@@ -856,6 +861,11 @@ func (g *Generator) Generate() error {
 	pulumiPackageSpec, err := genPulumiSchema(pack, g.pkg, g.version, g.info, g.renamesBuilder)
 	if err != nil {
 		return errors.Wrapf(err, "failed to create Pulumi schema")
+	}
+
+	// Apply schema post-processing if defined in the provider.
+	if g.SchemaPostProcessor != nil {
+		g.SchemaPostProcessor(&pulumiPackageSpec)
 	}
 
 	// As a side-effect genPulumiSchema also populated rename tables.


danielrbradley (approved):

t0yv0 (approved):
Where is the code for actual post-processing? You don't think it belongs in this codebase? Possibly good thinking as it's likely not bridge-specific.

pkg/tfgen/generate.go
```diff
@@ -858,6 +863,11 @@ func (g *Generator) Generate() error {
 		return errors.Wrapf(err, "failed to create Pulumi schema")
 	}
 
+	// Apply schema post-processing if defined in the provider.
```
t0yv0, line 866: Depending on the surgery that's done in the callback, this may fail nameCheck pass below.

t0yv0: Perhaps move the callback application after the namecheck pass?

iwahbe (approved):

pkg/tfgen/generate.go
mikhailshilkov: I'm a bit out of my depth here. It sounds fine to nameCheck after the surgery as an extra consistency check?

t0yv0: nameCheck takes a 	spec pschema.PackageSpec and it's going to check  spec.Types. If you move around spec.Types, it will start complaining possibly... You may deal with it later as you hit the problem, I'd suggest just running namecheck on pre-modified PackageSpec.

t0yv0:
> That code is very specific to WafV2, so I think it's right to put it there.

Could you add a link please?

I'm curious why a generic solution would not work. That also would take care of QuickSight and similar issues.

t0yv0:
That's cool! Let's certainly start with shipping that and then we can have another look to see if any of this generalizes for Quicksight.

mikhailshilkov:
> whether I should add this as yet another option to ProviderInfo. Initially I thought it was out-of-place there, but maybe it's more consistent with everything else?

Any thoughts on this?

t0yv0:
> another option in ProviderInfo

Don't mind either way. That'd give it more visibility. You just need some way to call it, if the current PR gives you that that's 👍  - once we know we use this functionality a lot we can promote and document better.

pkg/tfgen/generate.go
AaronFriel: By "to the end", I mean post `g.UnstableGenerateFromSchema`

mikhailshilkov:
It can’t be post UnstableGenerateFromSchema because that method generates language SDKs, right? And I do want to affect our SDKs here.

Also, I want to affect how examples are generated, since they need to reflect the type changes, e.g. names of classes.

PR #1250, opened 2023-06-28
Reproduce suppressed tags_all diff

t0yv0:
Looking at pulumi/pulumi-aws#1655

Findings so far. https://github.com/pulumi/terraform-provider-aws/blob/a82c36c5556e2b8de91ba2a82a24bcd6074345be/docs/resource-tagging.md#L246 has a special strategy for tags_all property which is computed as a union of provider tags and default tags. https://github.com/pulumi/terraform-provider-aws/blob/a82c36c5556e2b8de91ba2a82a24bcd6074345be/internal/verify/diff.go#L23 uses a diff customizer that computes these tags_all property.

The way the bridge works, we do not in interact with TF code in Check so Check results do not see tags-all; however when calling Diff we do call diff customizers and I've verified this customizer does get run and produces a TF diff that does indicate something is changing in my test:

```
    map[tags_all.tag2:{tag2valueModified tag2value false false <nil> false false 0}]
```

However makeDetailedDiff translates this diff to an empty diff.

Starting with an attempt to reproduce this makeDetailedDiff in a unit test.

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -157,18 +157,19 @@ func makePropertyDiff(name, path string, v resource.PropertyValue, tfDiff shim.I
 	diff map[string]*pulumirpc.PropertyDiff, tfs shim.Schema, ps *SchemaInfo, finalize, rawNames bool) {
 
 	visitor := func(name, path string, v resource.PropertyValue) bool {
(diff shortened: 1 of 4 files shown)


t0yv0:
This is obsolete now but informed the fix for 1655.

PR #1251, opened 2023-06-29
Add support for computed inputs in diff

AaronFriel:
This is intended to enable closing:
- pulumi/pulumi-aws#1655

We add a new flag on fields, `ComputedInput`, which indicates that the value will only appear as a result of a tf diff computation or planning, and not in `news`. This computed _input_ would ordinarily be computed during `Check()` in our lifecycle model.

For reasons unclear to me, we don't include computed properties in detailed diffs, and in stepping through `makeDetailedDiff`, we found that the visitor never saw properties such as `tags_all.foo` when that property was new and computed.

With this change, we can mark the `tags_all` field as a `ComputedInput`, and trigger updates on default tags changing:

```
pulumi up --skip-preview
Updating (test-aws-1655)

View in Browser (Ctrl+O): https://app.pulumi.com/pulumi/simple-yaml/test-aws-1655/updates/9

     Type                     Name                       Status              Info
     pulumi:pulumi:Stack      simple-yaml-test-aws-1655
 ~   ├─ pulumi:providers:aws  aws-provider               updated (0.13s)     [diff: ~defaultTags]
 ~   ├─ aws:s3:Bucket         my-bucket                  updated (0.87s)
 ~   └─ aws:s3:BucketObject   index.html                 updated (0.24s)
```

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -157,7 +157,7 @@ func makePropertyDiff(name, path string, v resource.PropertyValue, tfDiff shim.I
 	diff map[string]*pulumirpc.PropertyDiff, forceDiff *bool,
 	tfs shim.Schema, ps *SchemaInfo, finalize, rawNames bool) {
 
-	isComputedInput := ps != nil && ps.ComputedInput
+	isComputedInput := ps != nil && ps.XComputedInput
 	visitor := func(name, path string, v resource.PropertyValue) bool {
 		switch {
 		case v.IsArray():
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -154,7 +154,8 @@ func visitPropertyValue(name, path string, v resource.PropertyValue, tfs shim.Sc
 }
 
 func makePropertyDiff(name, path string, v resource.PropertyValue, tfDiff shim.InstanceDiff,
-	diff map[string]*pulumirpc.PropertyDiff, tfs shim.Schema, ps *SchemaInfo, finalize, rawNames bool) {
+	diff map[string]*pulumirpc.PropertyDiff, forceDiff *bool,
+	tfs shim.Schema, ps *SchemaInfo, finalize, rawNames bool) {
 
 	isComputedInput := ps != nil && ps.ComputedInput
 	visitor := func(name, path string, v resource.PropertyValue) bool {
@@ -169,7 +170,7 @@ func makePropertyDiff(name, path string, v resource.PropertyValue, tfDiff shim.I
 		case v.IsObject():
 			// If this value has a diff and is considered computed by Terraform, the diff will be woefully incomplete. In
 			// this case, do not recurse into the array; instead, just use the count diff for the details.
-			if d := tfDiff.Attribute(name + ".%"); d == nil || (!d.NewComputed && !isComputedInput) {
+			if d := tfDiff.Attribute(name + ".%"); d == nil || !d.NewComputed {
 				return true
 			}
 			name += ".%"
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/diff.go
```diff
@@ -157,18 +157,19 @@ func makePropertyDiff(name, path string, v resource.PropertyValue, tfDiff shim.I
 	diff map[string]*pulumirpc.PropertyDiff, tfs shim.Schema, ps *SchemaInfo, finalize, rawNames bool) {
 
 	visitor := func(name, path string, v resource.PropertyValue) bool {
+		isComputedInput := ps != nil && ps.ComputedInput
```
t0yv0, line 160: This is incorrect for nested properties. Let's lift it out of `visitor` and make a TODO with a comment. Here `ps` refers to top-level SchemaInfo but visitor descends into  nested properties, in fact @iwahbe was recently fixing the function that runs this visitor already for MaxItemsOne handling and I think it still might not do the right thing. We need to rewrite it sometimes. I'd have a preference reusing path based methods like LookupSchemaInfoPath to keep track of the current location in the recursion and hide away MaxItemsOne details to find the machine SchemaInfo.

pkg/tfbridge/diff.go
```diff
@@ -157,18 +157,19 @@ func makePropertyDiff(name, path string, v resource.PropertyValue, tfDiff shim.I
...
-			if d := tfDiff.Attribute(name + ".#"); d == nil || !d.NewComputed {
+			if d := tfDiff.Attribute(name + ".#"); d == nil || (!d.NewComputed && !isComputedInput) {
 				return true
 			}
 			name += ".#"
 		case v.IsObject():
 			// If this value has a diff and is considered computed by Terraform, the diff will be woefully incomplete. In
 			// this case, do not recurse into the array; instead, just use the count diff for the details.
-			if d := tfDiff.Attribute(name + ".%"); d == nil || !d.NewComputed {
+			if d := tfDiff.Attribute(name + ".%"); d == nil || (!d.NewComputed && !isComputedInput) {
```
t0yv0, line 172: This change is not covered by tests. If I remove it tests still pass. Can we construct a test case to cover it?

t0yv0:
Slight pref for https://github.com/pulumi/pulumi-terraform-bridge/pull/1254 on top of this.

pkg/tfbridge/diff.go
AaronFriel: Moved up - agree on the TODO, but could you write it? Sounds like you & @iwahbe have context on the change you want to make here.

pkg/tfbridge/info.go
```diff
@@ -394,6 +394,10 @@ type SchemaInfo struct {
 
 	// whether or not to treat this property as secret
 	Secret *bool
+
+	// whether to treat this as a computed input for the purpose of diff, i.e.: should trigger an
+	// update even if it appears only as an output property.
```
iwahbe, line 399: I don't know what this means outside of the context of a `Provider`. If this is provider only, let's say and enforce that. If not, can you flesh out this explanation.

pkg/tfbridge/diff_test.go
```diff
@@ -1903,5 +2008,80 @@ func TestListNestedAddMaxItemsOne(t *testing.T) {
 		// expected
 		map[string]DiffKind{
 			"prop.nest": AR,
-		})
+		},
+		pulumirpc.DiffResponse_DIFF_SOME)
+}
+
+func TestChangingTagsAll(t *testing.T) {
```
iwahbe, line 2015: Does this fail when you remove the new changes?

pkg/tfbridge/info.go
iwahbe: Can we mark it as experimental for now?

iwahbe (approved):
The most important thing for me is having enough tests on `SchemaInfo.ComputedInput` to form a strong understanding on what it should do in all situations.
I think we can take this as long as we have (and quickly follow up on) ensuring we have good test coverage (in repo) for all relevant situations.

pkg/tfbridge/info.go
AaronFriel: Yeah, it should be experimental. Should I rename it to `XComputedInput`?

pkg/tfbridge/diff_test.go
AaronFriel: It fails when I remove the changes _except_ what Anton raised above.

PR #1253, opened 2023-06-29
Upgrade `golangci-lint` to `v1.53`

iwahbe:
I am disabling the unused-parameter rule because we violate it many many times.


t0yv0:
Excellent very timely.

t0yv0 (approved):

PR #1256, opened 2023-06-29
Regress AWS-2490

t0yv0:
Regress pulumi/pulumi-aws#2490 in a unit test.


t0yv0:
On second look it looks like an upstream/by-design issue not a diff issue. Closing for now.

PR #1258, opened 2023-06-29
Move schema post-processing hook to ProviderInfo

mikhailshilkov:
Remake of #1248

I need to use the hook from a muxed provider which doesn't play nicely with my current API. So I suppose ProviderInfo is a more natural place to do so.


iwahbe (approved):
I think this is better. LGTM

PR #1263, opened 2023-06-30
Cleanup datasource auto aliasing

### LINKED ISSUE #1261: Running `make tfgen` with AutoAliased datasources results in appending a past instance of the datasource each run
### What happened?

Running `make tfgen` on this commit (https://github.com/pulumi/pulumi-akamai/pull/231/commits/9ff705ba59c73ee46c7e82273aaad21f63b27656) always results in appending a new history for aliased datasources.

### Expected Behavior

Running `make tfgen` should only add new information to `bridge-metadata.json`. Running `make tfgen` a second time should result in no changes.

### Steps to reproduce

Run `make tfgen && make tfgen` with commit https://github.com/pulumi/pulumi-akamai/pull/231/commits/9ff705ba59c73ee46c7e82273aaad21f63b27656 checked out.

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

iwahbe:
Fixes #1261

I encountered this when I was working on https://github.com/pulumi/pulumi-akamai/pull/231. 

Resources: We were playing fast and loose with what "current" meant. Sometimes it meant the current token, other times it meant the token where we diverged from no need to alias. This makes current mean only "the current token". This wasn't a correctness problem until the second major upgrade. The solution is to set "current" when we notice a change.

DataSources: We were appending to past each time, even if there were duplicates. This is easily fixed by adding a duplicate check the same way we do for resources. This was never a correctness problem, but it removes pointless clutter from `bridge-metadata.json` as well as the need to manually fix up the file after edits.

### DIFF
--- a/pkg/tfbridge/auto_aliasing.go
+++ b/pkg/tfbridge/auto_aliasing.go
@@ -209,10 +209,7 @@ func aliasResource(
 		hist[tfToken] = &tokenHistory[tokens.Type]{
 			Current: computed.Tok,
 		}
-	} else if prev.Current != computed.Tok {
-		// It's in history, but something has changed. Update the history to reflect
-		// the new reality, then add aliases.
-
+	} else {
 		// We don't do this eagerly because aliasResource is called while
 		// iterating over p.Resources which aliasOrRenameResource mutates.
 		*applyResourceAliases = append(*applyResourceAliases,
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/auto_aliasing.go
```diff
@@ -408,12 +405,15 @@ func aliasOrRenameResource(
 			break
 		}
 	}
-	if !alreadyPresent {
+	if !alreadyPresent && res.Tok != hist.Current {
+		// The resource is in history, but the name has changed. Update the new current name
+		// and add the old name to the history.
```
aq17, line 410: LGTM

aq17 (approved):

PR #1264, opened 2023-06-30
Add a TODO for XComputedInput

t0yv0:
TO follow up on a previous PR #1251 the newly added XComputedInput property modifier does not work for nested properties, but can only be applied to a top-level resource property.


iwahbe (approved):
I would love if we could reference an associated issue.

Issue #1260, opened 2023-06-29
Strengthen `defaultTags` computed input regression test


AaronFriel:
As observed in #1251, there is a change to `diff.go` that still passes the test despite empirically breaking the programtest in the AWS pull request pulumi/pulumi-aws#2585:

_Originally posted by @t0yv0 in https://github.com/pulumi/pulumi-terraform-bridge/pull/1251#discussion_r1246630635_

The way the AWS provider and TF SDK interact to produce the `tfdiff` diverges from the test added in that pull request:

https://github.com/pulumi/pulumi-terraform-bridge/blob/739f4dc380cf49368f4efbdb38e1b76077839fe6/pkg/tfbridge/diff_test.go#L2040-L2074

However, it's not obvious how to make our test harness here behave identically to TF.

PR #1266, opened 2023-06-30
Add coverage for XComputedInputs

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1260

Follows #1251 with extending coverage for every changed line.

One line is re-added as it apparently was lost in a bad merge - but is crucial for this to work correctly.

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -170,7 +170,7 @@ func makePropertyDiff(name, path string, v resource.PropertyValue, tfDiff shim.I
 		case v.IsObject():
 			// If this value has a diff and is considered computed by Terraform, the diff will be woefully incomplete. In
 			// this case, do not recurse into the array; instead, just use the count diff for the details.
-			if d := tfDiff.Attribute(name + ".%"); d == nil || !d.NewComputed {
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/diff.go
```diff
@@ -170,7 +170,7 @@ func makePropertyDiff(name, path string, v resource.PropertyValue, tfDiff shim.I
 		case v.IsObject():
 			// If this value has a diff and is considered computed by Terraform, the diff will be woefully incomplete. In
 			// this case, do not recurse into the array; instead, just use the count diff for the details.
-			if d := tfDiff.Attribute(name + ".%"); d == nil || !d.NewComputed {
```
t0yv0, line 173: I think this got lost in the merge, very unfortunate. Again emphasizes the need for accurate tests..

AaronFriel (approved):
This test looks good; makes sense that we'd need to customize the diff on the TF side to reproduce.

t0yv0:
> This test looks good; makes sense that we'd need to customize the diff on the TF side to reproduce.

The CustomizeDiff bits got lifted out of AWS, that's how AWS does tags.. This is a very simplified version, in the real provider CustomizeDiff merges in provider-level tags in the resource-level tags.

The problem with the earlier unit test was that it's testing "changing a tag" and not testing "adding a tag". Now both are covered!

PR #1267, opened 2023-07-03
Update pf/README.md

t0yv0:
Updates pf/README.md to incorporate latest breaking changes in removing pf.ProviderInfo.

### DIFF
--- a/pf/README.md
+++ b/pf/README.md
@@ -5,12 +5,8 @@ providers](https://www.pulumi.com/docs/intro/concepts/resources/providers/) from
 Providers](https://github.com/terraform-providers) built using the [Terraform Plugin
 Framework](https://developer.hashicorp.com/terraform/plugin/framework).
 
-This bridge is in active development and has an incomplete feature set. Progress is tracked in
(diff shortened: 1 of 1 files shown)


pf/README.md
```diff
@@ -200,20 +192,20 @@ to the Plugin Framework.
...
     ```go
 
     //go:embed cmd/pulumi-resource-myprovider/bridge-metadata.json
     var bridgeMetadata []byte
 
-    func Provider() pf.ProviderInfo {
+    func Provider() tfbridge.ProviderInfo {
         info := tfbridge.ProviderInfo{
-            // Comment out P, use NewProvider below instead.
+            // Comment out P
```
iwahbe, line 204: Why is this "Comment out P"? Couldn't we just say "Replace P: shimv2.NewProvider(...) with pf.ShimProvider(shim.NewProvider())"?

iwahbe (approved):

pf/README.md
t0yv0: Good catch.

PR #1276, opened 2023-07-06
Remove custom code for computing planned state

t0yv0:
Factors out ProposedNew into its own package so both Plugin Framework providers and SDKv2-based providers that opt into DiffStrategy=PlanState call the exact same logical implementation that is linked from TF CLI sources (objchange.ProposedNew). 

```mermaid
  graph LR
      A[pfutils.ProposedNew] --> B[tfplan.ProposedNew]
      C[sdkv2.ProposedNew]   --> B[tfplan.ProposedNew]
      B --> D[objchange.ProposedNew]
```

Previously, PF-based providers used a homegrown implementation that was likely diverging from TF implementation in some of the edge cases. The PR trades off maintaining that implementation (which is no longer necessary) to maintaining some adapter boilerplate code.


pf/tests/testdata/updateprogram/Pulumi.yaml
```diff
@@ -38,7 +38,7 @@ outputs:
   testOptionalString__expect: |-
     input2
   testSetOptionals__actual: ${set.setOptionals}
-  testSetOptionals__expect: [ "${r.requiredInputStringCopy}", a, b ]
+  testSetOptionals__expect: [ a, b, "${r.requiredInputStringCopy}" ]
```
t0yv0, line 41: This change caused set elements to be reordered in the list projection to Pulumi. While this should theoretically be alright, as the subsequent Diff recognizes that sets are sets and should not be sensitive to element order, perhaps there are some unintended consequences?

Should the bridge define some guarantees on the canonical set ordering when projecting elements to Pulumi array property values?

t0yv0:
OK not ready for review quite yet, there's some more edge cases caught by the integration test suite.

pf/internal/schemav6/schemav6.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func (r *schemaExtractingResource) Read(context.Context, resource.ReadRequest, *resource.ReadResponse) {
+}
+
+func (r *schemaExtractingResource) Update(context.Context, resource.UpdateRequest, *resource.UpdateResponse) {
+}
+
+func (r *schemaExtractingResource) Delete(context.Context, resource.DeleteRequest, *resource.DeleteResponse) {
+}
```
iwahbe, line 170: Can we assert that these methods will not be called (and then panic)?

pf/tests/internal/testprovider/testbridge_resource_compopt.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2016-2023, Pulumi Corporation.
```
iwahbe, line 1: This file is called `testbridge_resource_compopt.go` If `compopt` is "component options", can we just say that? Otherwise, can we say what it is?

pf/tests/testdata/updateprogram/Pulumi.yaml
iwahbe: I think the bridge needs to define a constant ordering since we expose this data as a list. We shouldn't need to guarantee anything about the ordering, except consistency.

Its not unreasonable to assume that `new Resource("one", opt=set[0]); new Resource("two", opt=set[1])` is safe between updates, where a re-order would break this.

pkg/tfshim/sdk-v2/internal/schemav6/schemav6.go
```diff
@@ -0,0 +1,93 @@
...
+const (
+	resourceName = "xres"
+)
+
+func ResourceSchema(res *schema.Resource) (*tfprotov6.Schema, error) {
+	contract.Assertf(res != nil, "res != nil")
+
+	prov := &schema.Provider{
+		ResourcesMap: map[string]*schema.Resource{
+			"xres": res,
```
iwahbe, line 39:
```suggestion
			resourceName: res,
```

pkg/tfplan/proposed_new.go
```diff
@@ -0,0 +1,232 @@
...
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfplan/internal/tf/configs/configschema"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfplan/internal/tf/plans/objchange"
+	"github.com/zclconf/go-cty/cty"
+	ctypack "github.com/zclconf/go-cty/cty/msgpack"
+)
+
+func ProposedNew(
```
iwahbe, line 28: This is the public entry point for this package. Can we add a comment explaining why it exists and what it does.

If we don't believe another repository could consume this function, we should mark it as unstable for consumption. We should be clear on when we adopt a backwards compatibility requirement.

iwahbe (commented):
In general this looks good. I'm scared of the set re-ordering though. That seams like a meaningful breaking change that could cause replaces. The obvious example is:

```typescript
const resource = new Resource("parent")
resource.setProperty.apply(set => {
  // Create resources inside an apply, 
  // each resource is name associated with its ordering
  set.map((i, v) => new Child(`child-${i}`, {
    needsReplaceOnChanges: v,
  }))
}
```
Here a change in order would cause all children to be replaced.

pf/tests/testdata/updateprogram/Pulumi.yaml
t0yv0: I'm not sure I grok the example, could you expand. 

Since bridged providers customize Diff method sending re-ordered versions of the set should result in a no-changes diff. But I have a hunch somewhere we run into trouble anyway. Would be good to file an issue figuring out where. 

Perhaps with ignoreChanges paths is one place. 

FWIW this PR seems to move us from "user-defined" ordering to "TF-defined" consistent ordering, indirectly, based on the example changes. There's some subtleties with this. It probably is affected by TF hashing functions and when they change upstream this is likely to reorder sets.

There are also some serious issues with sets on the SDKV2 based providers. I was recently looking at this one:

https://github.com/pulumi/pulumi-terraform-bridge/pull/1255

pf/internal/schemav6/schemav6.go
t0yv0: I wonder if this would reduce branch coverage :)

pf/tests/testdata/updateprogram/Pulumi.yaml
iwahbe: https://github.com/pulumi/pulumi-terraform-bridge/pull/1276#pullrequestreview-1518268522 has a better example. Basically, external users can rely on the ordering of pulumi lists (and tf sets) such that re-ordering will always cause problems, no matter what the bridge does.

iwahbe: If we believe that TF-defined sets are better (more consistent, easier to maintain) then I'm ok moving to them as long as we (1) recognize it is a breaking change and (2) put tests and documentation locking that ordering in.

t0yv0:
Obsolete now, the objective is met without this PR.

PR #1285, opened 2023-07-11
Support `SchemaFunc`

iwahbe:
Adds support for `SchemaFunc` as introduced in https://github.com/hashicorp/terraform-plugin-sdk/pull/1218.

Note: This will require all providers who upgrade to include this version of the bridge to be using `>=v2.27.0` of Pulumi's `terraform-plugin-sdk`.

Rebased on top of https://github.com/pulumi/pulumi-terraform-bridge/pull/1287.


mikhailshilkov:
> Note: This will require all providers who upgrade to include this version of the bridge to be using >=v2.27.0 of Pulumi's terraform-plugin-sdk.

How disruptive is it going to be? Will we need to do this manually in every provider?

iwahbe:
> > Note: This will require all providers who upgrade to include this version of the bridge to be using >=v2.27.0 of Pulumi's terraform-plugin-sdk.
> 
> How disruptive is it going to be? Will we need to do this manually in every provider?

It will be minimally disruptive. New versions are reliably backwards compatible and `upgrade-provider` [handles this for us](https://github.com/pulumi/upgrade-provider/blob/7ffdf45faedbcd7b60d7c9b13bb6781dc5b047ea/upgrade/steps.go#L344-L351). I don't expect manual intervention will be necessary.

thomas11 (approved):

PR #1296, opened 2023-07-20
Tolerate MaxItemsOne mismatches

### LINKED ISSUE #1295: MaxItemsOne breaks when changed between major versions
### What happened?

See https://github.com/pulumi/pulumi-aws/issues/2626 part 3.

### Expected Behavior

We handle old state without errors when terraform would, even when the value of `MaxItemsOne(...)` changes.

### Steps to reproduce

We can run the repro directly against `pulumi-aws` by adding this file to `provider/` and running `cd provider && go test ./...`.
```go
package provider

import (
	"bytes"
	"context"
	"strings"
	"testing"

	pf "github.com/pulumi/pulumi-terraform-bridge/pf/tfbridge"
	testutils "github.com/pulumi/pulumi-terraform-bridge/testing/x"
	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
)

func provider(t *testing.T) pulumirpc.ResourceProviderServer {
	info := *Provider()
	ctx := context.Background()
	meta := genSDKSchema(t, info)
	p, err := pf.MakeMuxedServer(ctx, info.Name, info, meta)(nil)
	require.NoError(t, err)
	return p
}

func genSDKSchema(t *testing.T, info tfbridge.ProviderInfo) []byte {
	// import	"encoding/json"
	// import 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen"
	// pkg, err := tfgen.GenerateSchema(info, testSink(t))
	// require.NoError(t, err)
	// bytes, err := json.MarshalIndent(pkg, "", "  ")
	// require.NoError(t, err)
	// return bytes
	return []byte("{}")
}

func testSink(t *testing.T) diag.Sink {
	var stdout, stderr bytes.Buffer

	testSink := diag.DefaultSink(&stdout, &stderr, diag.FormatOptions{
		Color: colors.Never,
	})

	t.Cleanup(func() {
		if strings.TrimSpace(stdout.String()) != "" {
			t.Logf("%s\n", stdout.String())
		}
		if strings.TrimSpace(stderr.String()) != "" {
			t.Logf("%s\n", stderr.String())
		}
	})

	return testSink
}

func TestUpgrade(t *testing.T) {
	testutils.ReplaySequence(t, provider(t), replay)
}

const replay = `[
  {
    "method": "/pulumirpc.ResourceProvider/Configure",
    "request": {
      "variables": {
        "aws:config:ignoreTags": "{\"keys\":[\"ignore\"]}",
        "aws:config:region": "us-west-2",
        "aws:config:skipCredentialsValidation": "false",
        "aws:config:skipMetadataApiCheck": "true",
        "aws:config:skipRegionValidation": "true"
      },
      "args": {
        "ignoreTags": "{\"keys\":[\"ignore\"]}",
        "region": "us-west-2",
        "skipCredentialsValidation": "false",
        "skipMetadataApiCheck": "true",
        "skipRegionValidation": "true",
        "version": "6.0.0"
      },
      "acceptSecrets": true,
      "acceptResources": true,
      "sendsOldInputs": true
    },
    "response": {
      "supportsPreview": true
    },
    "metadata": {
      "kind": "resource",
      "mode": "client",
      "name": "aws"
    }
  },
  {
    "method": "/pulumirpc.ResourceProvider/Diff",
    "request": {
      "id": "cdb25640-3498-40f5-bd01-4b67df036220",
      "urn": "urn:pulumi:dev::vpc::aws:lambda/eventSourceMapping:EventSourceMapping::mapping",
      "olds": {
        "amazonManagedKafkaEventSourceConfig": null,
        "batchSize": 100,
        "bisectBatchOnFunctionError": false,
        "destinationConfig": {
          "onFailure": null
        },
        "documentDbEventSourceConfig": null,
        "enabled": true,
        "eventSourceArn": "arn:aws:kinesis:us-west-2:616138583583:stream/s-8e8f13a",
        "filterCriteria": null,
        "functionArn": "arn:aws:lambda:us-west-2:616138583583:function:f-e2911f7",
        "functionName": "arn:aws:lambda:us-west-2:616138583583:function:f-e2911f7",
        "functionResponseTypes": [],
        "id": "cdb25640-3498-40f5-bd01-4b67df036220",
        "lastModified": "2023-07-20T13:09:00Z",
        "lastProcessingResult": "No records processed",
        "maximumBatchingWindowInSeconds": 0,
        "maximumRecordAgeInSeconds": -1,
        "maximumRetryAttempts": -1,
        "parallelizationFactor": 1,
        "queues": ["my-queue"],
        "scalingConfig": null,
        "selfManagedEventSource": null,
        "selfManagedKafkaEventSourceConfig": null,
        "sourceAccessConfigurations": [],
        "startingPosition": "TRIM_HORIZON",
        "startingPositionTimestamp": "",
        "state": "Enabled",
        "stateTransitionReason": "User action",
        "topics": [],
        "tumblingWindowInSeconds": 0,
        "uuid": "cdb25640-3498-40f5-bd01-4b67df036220"
      },
      "news": {
        "__defaults": [
          "enabled"
        ],
        "enabled": true,
        "eventSourceArn": "arn:aws:kinesis:us-west-2:616138583583:stream/s-8e8f13a",
        "functionName": "f-e2911f7",
        "queues": "my-queue",
        "startingPosition": "TRIM_HORIZON"
      },
      "oldInputs": {
        "__defaults": [
          "enabled"
        ],
        "enabled": true,
        "eventSourceArn": "arn:aws:kinesis:us-west-2:616138583583:stream/s-8e8f13a",
        "functionName": "f-e2911f7",
        "queues": [
          "my-queue"
        ],
        "startingPosition": "TRIM_HORIZON"
      }
    },
    "response": {
      "changes": "DIFF_NONE",
      "hasDetailedDiff": true,
      "stables": "*"
    },
    "metadata": {
      "kind": "resource",
      "mode": "client",
      "name": "aws"
    }
  }
]`
```

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

iwahbe:
Tolerate:
- [x] MaxItemsOne: true -> false
- [x] MaxItemsOne: false -> true

This is an information losing operation, since it doesn't handle the case of `[[T]] <-> [T]`. To simply the fix, I have chosen to do nothing in this situation. This PR will have no effect for nested lists.

Fixes #1295 

We handle the `MaxItemsOne: false -> true` case by default, but the test locks in behavior.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -309,19 +309,36 @@ func MakeTerraformInputs(instance *PulumiResource, config resource.PropertyMap, 
 func (ctx *conversionContext) MakeTerraformInput(name string, old, v resource.PropertyValue,
 	tfs shim.Schema, ps *SchemaInfo, rawNames bool) (interface{}, error) {
 
-	// For TypeList or TypeSet with MaxItems==1, we will have projected as a scalar nested value, and need to wrap it
-	// into a single-element array before passing to Terraform.
+	// For TypeList or TypeSet with MaxItems==1, we will have projected as a scalar
+	// nested value, and need to wrap it into a single-element array before passing to
+	// Terraform.
 	if IsMaxItemsOne(tfs, ps) {
-		if old.IsNull() {
-			old = resource.NewArrayProperty([]resource.PropertyValue{})
-		} else {
-			old = resource.NewArrayProperty([]resource.PropertyValue{old})
+		wrap := func(val resource.PropertyValue) resource.PropertyValue {
+			if val.IsNull() {
+				return resource.NewArrayProperty([]resource.PropertyValue{})
+			}
+
+			// If we are expecting a value of type `[T]` where `T != TypeList`
+			// and we already see `[T]`, we see that `v` is already the right
+			// shape and return as is.
+			//
+			// This is possible when the old state is from a previous version
+			// with `MaxItemsOne=false` but the new state has
+			// `MaxItemsOne=true`.
+			if elem := tfs.Elem(); elem != nil {
+				if elem, ok := elem.(shim.Schema); ok &&
+					// If the underlying type is not a list or set,
+					// but the value is a list, we just return as is.
+					!(elem.Type() == shim.TypeList || elem.Type() == shim.TypeSet) &&
+					val.IsArray() {
+					return val
+				}
+			}
+
+			return resource.NewArrayProperty([]resource.PropertyValue{val})
 		}
-		if v.IsNull() {
-			v = resource.NewArrayProperty([]resource.PropertyValue{})
-		} else {
-			v = resource.NewArrayProperty([]resource.PropertyValue{v})
-		}
+		old = wrap(old)
+		v = wrap(v)
 	}
 
 	// If there is a custom transform for this value, run it before processing the value.
--- a/pkg/tfbridge/schema_test.go
+++ b/pkg/tfbridge/schema_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/internal/testprovider"
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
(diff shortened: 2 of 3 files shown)


pkg/tfbridge/schema.go
```diff
@@ -312,16 +312,32 @@ func (ctx *conversionContext) MakeTerraformInput(name string, old, v resource.Pr
...
+			// This is possible when the old state is from a previous version
+			// with `MaxItemsOne=false` but the new state has
+			// `MaxItemsOne=true`.
+			if elem := tfs.Elem(); elem != nil {
+				if elem, ok := elem.(shim.Schema); ok &&
+					// If the underlying type is not a list or set,
+					// but the value is a list, we just return as is.
+					!(elem.Type() == shim.TypeList || elem.Type() == shim.TypeSet) &&
+					v.IsArray() {
+					return v
```
t0yv0, line 333: This looks good, surgical.

pkg/tfbridge/schema_test.go
```diff
@@ -272,6 +273,109 @@ func TestTerraformInputs(t *testing.T) {
...
+			expected: map[string]interface{}{
+				"list":   []interface{}{"sc"},
+				"scalar": "sc",
+			},
+		},
+		"list-adding-max-items-one": {
+			oldsList: false,
+			newsList: true,
+			expected: map[string]interface{}{
+				"list":   []interface{}{[]interface{}{[]interface{}{"sc"}}},
```
t0yv0, line 352: What's going on in this test case?

The test look great in the sense that it looks like they cover all the interesting combinations, however some attention perhaps could be given to legibility for an out-of-context engineer coming to read this code.

The property we are trying to check for and ensure is that the result of MakeTerraformInput always conforms to the Terraform schema, as violating this property leads to panics or errors in the TF layer.

It would be lovely to make it obvious in the test case. For example, here we are looking at this value (I think the value under "scalar" is not meaningfully testing things in this test case, right?):

```
[[["sc"]]]
``` 

So presumably the TF schema here is `List<List<List<string>>>`, but it's hard to see that since it's hidden behind generating functions.

[Boolean blindness](https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/) comes to mind.

t0yv0 (approved):

pkg/tfbridge/schema_test.go
iwahbe: I've revised the test cases to be clearer and more readable, and added some additional tests on null handling. To avoid boolean blindness, I have removed the constructer functions.

PR #1304, opened 2023-07-26
Upgrade to pu/pu v3.76

abhinav:
Upgrade to pulumi/pulumi v3.76,
accounting for the following change:
https://github.com/pulumi/pulumi/commit/fe5e65f0f4ca0c163006b9ff50c7773f491515e1#diff-d165f3d042e491ac3687daac96a66ecb5acf0b62890c99852c6b92744f5e7c39R64


Frassle (approved):

pkg/tfgen/convert_test.go
```diff
@@ -78,7 +78,7 @@ func TestConvert(t *testing.T) {
 import * as pulumi from "@pulumi/pulumi";
 
 const config = new pulumi.Config();
-const regionNumber = config.getObject("regionNumber") || {
+const regionNumber = config.getObject<{us-east-1?: number}>("regionNumber") || {
```
abhinav, line 81: @Frassle @Zaid-Ajaj is this change expected?

Frassle: Yup that looks like the new typed config support Zaid added.
PCL defaults to assuming the "default" value fully describes the type, so this is what we'd currently expect.
Seems sub-optimal but I don't think it blocks this update.

Issue #1306, opened 2023-07-28
Expose more SDK generation options in GolangInfo and similar structures


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently bridged providers generate Pulumi Package Schema and specifically populate the "language" section that informs the SDK code generator for every target language (TypeScript, Python, Node, Java, C#). However not all options supported by the SDK code generators are exposed, making them harder to configure for bridged providers. 

For example, Go generator supports RespectSchemaVersion. But it is not to be found in https://github.com/pulumi/pulumi-terraform-bridge/blob/e0938e965dc13c3fa33bc88a699ba94fb31cb255/pkg/tfbridge/info.go#L506-%23L506.

As a workaround the user of a bridged provider can still set these options manually, for example:

```go
	prov.SchemaPostProcessor = func(spec *schema.PackageSpec) {
		var info gen.GoPackageInfo
		err := json.Unmarshal(spec.Language[string(tfgen.Golang)], &info)
		contract.Assertf(err == nil, `err == nil`)
		info.RespectSchemaVersion = true
		bytes, err := json.Marshal(info)
		contract.Assertf(err == nil, `err == nil`)
		spec.Language[string(tfgen.Golang)] = bytes
	}
```

Consider exposing these options directly in the bridge API, possibly referencing typed structs such as [GoPackageInfo](https://github.com/pulumi/pulumi/blob/29c0cf0887f19268c54f4ed0889af45d17f8409a/pkg/codegen/go/importer.go#L24-#L24)

PR #1308, opened 2023-07-28
Expose backend SDK generation options for each target language

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1306

Adding missing language-specific options so that it is easier to configure the .language section of a bridged provider's schema generation. The specific customer ask is around RespectSchemaVersion for Go, but it seems reasonable to expose all of them.

### DIFF
--- a/internal/testprovider_sdkv2/cmd/pulumi-resource-tpsdkv2/schema.json
+++ b/internal/testprovider_sdkv2/cmd/pulumi-resource-tpsdkv2/schema.json
@@ -6,13 +6,14 @@
     },
     "language": {
         "nodejs": {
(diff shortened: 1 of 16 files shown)


iwahbe (approved):
Looks good.

I would love to either use the types in `pkg/codegen` or convert to an untyped map during our next breaking change. For now, I think this is the right call.

t0yv0:
Exposing an untyped map in the interface specifically goes against the grain of the customer ask here as they appreciate IDE /code completion to easily find the option. Re-exporting types from pkg/codegen is in some ways easier to maintain at the cost of introducing some coupling; I started from trying to do that but noticed that it then runs into duplicated fields that are already exposed, and it's not pretty unless we make a breaking change so I switched to duplicating the fields.

PR #1310, opened 2023-07-31
Add resource option PreCheckCallback

iwahbe:
Add the ability to override the check behavior for resources.

### DIFF
--- a/pf/tfbridge/provider_check.go
+++ b/pf/tfbridge/provider_check.go
@@ -47,7 +47,7 @@ func (p *provider) CheckWithContext(
 	}
 
 	if info := rh.pulumiResourceInfo; info != nil {
-		if check := info.XCustomCheck; check != nil {
+		if check := info.PreCheckCallback; check != nil {
 			checkedInputs, err := check(ctx, checkedInputs, p.lastKnownProviderConfig.Copy())
 			if err != nil {
 				return checkedInputs, []plugin.CheckFailure{}, err
--- a/pf/tfbridge/provider_check.go
+++ b/pf/tfbridge/provider_check.go
@@ -46,6 +46,15 @@ func (p *provider) CheckWithContext(
 		return checkedInputs, []plugin.CheckFailure{}, err
 	}
 
+	if info := rh.pulumiResourceInfo; info != nil {
+		if check := info.XCustomCheck; check != nil {
+			checkedInputs, err := check(ctx, checkedInputs, p.lastKnownProviderConfig.Copy())
+			if err != nil {
+				return checkedInputs, []plugin.CheckFailure{}, err
+			}
+		}
+	}
+
 	// Transform checkedInputs to apply Pulumi-level defaults.
 	news := defaults.ApplyDefaultInfoValues(ctx, defaults.ApplyDefaultInfoValuesArgs{
 		SchemaMap:   rh.schemaOnlyShimResource.Schema(),
--- a/pkg/tfbridge/provider_test.go
+++ b/pkg/tfbridge/provider_test.go
@@ -495,25 +495,10 @@ func testCheckFailures(t *testing.T, provider *Provider, typeName tokens.Type) [
 		News: pulumiIns,
 	})
 	assert.NoError(t, err)
-	assert.Len(t, checkResp.Failures, 3)
 	return checkResp.Failures
 }
 
-func TestProviderCheck(t *testing.T) {
-	provider := &Provider{
-		tf:     shimv1.NewProvider(testTFProvider),
-		config: shimv1.NewSchemaMap(testTFProvider.Schema),
-	}
-	provider.resources = map[tokens.Type]Resource{
-		"SecondResource": {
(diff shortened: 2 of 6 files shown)


t0yv0:
https://go.dev/play/p/-eNaTRqOVhu perhaps consider a way to override it without making changes to this codebase? It seems Go language has the facility to do this via embedded structs.

t0yv0 (changes requested):
Can it not be overridden in Go by other means outside of making edits to this codebase?

t0yv0:
IN addition, FWIW there's PreConfigureCallback machinery that is allowed to modify the data at Pulumi layer before it drops to TF; not sure if that is sufficient for the use case you have.

iwahbe:
> https://go.dev/play/p/-eNaTRqOVhu perhaps consider a way to override it without making changes to this codebase? It seems Go language has the facility to do this via embedded structs.

I don't see how this is applicable to this case. Where would this come into play? Which interface would I be hijacking here? 

`PreConfigureCallback` is for provider configuration, not resource configuration. It is not sufficient for my use case.

t0yv0:
Ok, my bad you're right about PreConfigureCallback belonging to a different part of the system.

Regarding overriding via Go embedding I guess it can be too tricky to be useful if we rely on calling into Main/ having multiple layers.

So let's discuss making a change to this codebase to support customizing Check. One sec.

pf/tfbridge/provider_check.go
```diff
@@ -46,6 +46,15 @@ func (p *provider) CheckWithContext(
 		return checkedInputs, []plugin.CheckFailure{}, err
 	}
 
+	if info := rh.pulumiResourceInfo; info != nil {
```
t0yv0, line 49: This bypasses defaults.ApplyDefaultInfoValues and p.validateResourceConfig and tfbridge.MarkSchemaSecrets potentially dropping the features implemented in this code. Would it be worthwhile to figure out how to express the transformation in a way that still calls into that code and keeps the features? Perhaps something along the lines of PreConfigureCallback like PreCheckCallback?

t0yv0: Ah never mind. It keeps those in place.

t0yv0:
Giving it a second read, I think this can be useful to have, I wanted to reach out for something like this in other work. Perhaps let's call it PreCheckCallback? Add some quick tests to make sure it's covered, similar to PreConfigureCallback.

pf/tfbridge/provider_check.go
```diff
@@ -46,6 +46,15 @@ func (p *provider) CheckWithContext(
...
+		if check := info.XCustomCheck; check != nil {
+			checkedInputs, err := check(ctx, checkedInputs, p.lastKnownProviderConfig.Copy())
```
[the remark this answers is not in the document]
iwahbe: Yep. I would like this to see values as passed in, not as configured.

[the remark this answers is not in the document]
t0yv0: Perfect, makes sense.

t0yv0:
This is great just needs a little more test TLC.

pkg/tfbridge/provider_test.go
```diff
@@ -550,6 +521,104 @@ func TestProviderCheckV2(t *testing.T) {
...
+	callback := func(
+		_ context.Context, config, meta resource.PropertyMap,
+	) (resource.PropertyMap, error) {
+		config["arrayPropertyValues"] = resource.NewArrayProperty(
+			[]resource.PropertyValue{meta["prop"]},
+		)
+		return config, nil
+	}
+
+	for _, f := range factories {
```
t0yv0, line 601: By-reference loop variable capture :)

pkg/tfbridge/provider_test.go
```diff
@@ -550,6 +521,104 @@ func TestProviderCheckV2(t *testing.T) {
...
+			    "request": {
+			      "urn": "urn:pulumi:st::pg::testprovider:index/res:Res::r",
+			      "olds": {},
+			      "news": {},
+			      "randomSeed": "wqZZaHWVfsS1ozo3bdauTfZmjslvWcZpUjn7BzpS79c="
+			    },
+			    "response": {
+			      "inputs": {
+				"__defaults": [],
+				"arrayPropertyValues": ["global"]
```
t0yv0, line 585: Thank you!

pkg/tfbridge/provider_test.go
iwahbe: The test isn't parallelized. I can change that (https://github.com/pulumi/pulumi-terraform-bridge/commit/272e881f24744a86655da45c59f5c903d5c0de46), but the by-ref isn't actually a risk here.

t0yv0: Yeah I wanted to enable a linter that flags these but we can deal with it then.

iwahbe: Have you seen https://github.com/golang/go/discussions/56010?

PR #1316, opened 2023-07-31
Autonaming: deprecate DefaultInfo.From

t0yv0:
This PR is the first in a series of changes to get Autonaming working for Plugin Framework providers; broken down into commits for easier review. The key change is to move from DefaultInfo.From to DefaultInfo.ComputeDefault. The ComputeDefaultOptions structure is there to extend with further context on the location where defaults are being computed, which is needed in subsequent PRs.

### DIFF
--- a/pf/internal/defaults/defaults.go
+++ b/pf/internal/defaults/defaults.go
@@ -40,8 +40,8 @@ type ApplyDefaultInfoValuesArgs struct {
 	// Toplevel SchemaInfo configuration matching TopSchemaMap.
 	SchemaInfos map[string]*tfbridge.SchemaInfo
 
-	// Optional. Note that ResourceInstance need not be specified for Invoke or Configure processing.
-	ResourceInstance *tfbridge.PulumiResource
+	// Note that URN need not be specified for Invoke or Configure processing.
+	ComputeDefaultOptions tfbridge.ComputeDefaultOptions
 
 	// Optional. If known, these are the provider-level configuration values, to support DefaultInfo.Config.
 	ProviderConfig resource.PropertyMap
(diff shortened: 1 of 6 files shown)


pkg/tfbridge/info.go
```diff
@@ -455,15 +455,35 @@ type DefaultInfo struct {
...
 	// Config uses a configuration variable from this package as the default value.
 	Config string
-	// From applies a transformation from other resource properties.
+
+	// Deprecated. Use ComputeDefault.
 	From func(res *PulumiResource) (interface{}, error)
+
+	// ComputeDefault specifies how to compute a default value for the given property by consulting other properties
+	// such as the resource's URN. See [ComputeDefaultOptions] for all available information.
+	ComputeDefault func(opts ComputeDefaultOptions) (interface{}, error)
```
iwahbe, line 464: I think it's generally useful to have a `context.Context` as a side-channel.
```suggestion
	ComputeDefault func(context.Context, ComputeDefaultOptions) (interface{}, error)
```

pkg/tfbridge/info.go
```diff
@@ -920,19 +940,18 @@ func (m *MarshallableDefaultInfo) Unmarshal() *DefaultInfo {
...
-	return &DefaultInfo{
+	defInfo := &DefaultInfo{
 		AutoNamed: m.AutoNamed,
-		From:      f,
 		Value:     m.Value,
 		EnvVars:   m.EnvVars,
 	}
+
+	if m.IsFunc {
+		defInfo.ComputeDefault = func(_ ComputeDefaultOptions) (interface{}, error) {
```
iwahbe, line 950:
```suggestion
		defInfo.ComputeDefault = func(ComputeDefaultOptions) (interface{}, error) {
```

iwahbe (approved):
A couple of nits, but this looks reasonable.

pf/internal/defaults/defaults.go
```diff
@@ -273,7 +264,7 @@ func (du *defaultsTransform) extendPropertyMapWithDefaults(
 		// using default value for empty property
 		pv, gotDefault := getDefaultValue(ctx,
 			pk,
-			du.resourceByPath(path),
+			du.resourceInstance,
```
t0yv0: Calling out this is a small non-refactor but actual change happening here, From/ComputedDefaults will now apply in nested property case (see commit msg).

pkg/tfbridge/info.go
t0yv0: Ok that's updated now. Better to have it locked in the signature even if it's context.TODO() for now in one place.

t0yv0:
DefaultInfo.From is now deprecated in favor of DefaultInfo.ComputeDefault. The ComputeDefaultOptions structure is similar to what DefaultInfo.From was receiving, but is extensible and able to receive some more context.

PR #1317, opened 2023-07-31
Remove panic on imports for TF Plugin Framework

t0yv0:
This PR removes the panics on `getDefaultValue` on the Type, Attr, and Block schemas in the TF Plugin Framework schemashim, as they are causing issues when importing resources using the Plugin Framework.

Instead, we emit nil and an error, which we log using our standard logger, allowing the code to run without panic. I have build and run this against the [user reported panic](https://github.com/pulumi/pulumi-cloudflare/issues/460) and verified that this resolves the issue.

We will file a followup issue to implement properly handling default values directly inside the `pf` package.

Fixes https://github.com/pulumi/pulumi-cloudflare/issues/460.


pf/internal/schemashim/block_schema.go
```diff
@@ -120,7 +120,7 @@ func (*blockSchema) DefaultFunc() shim.SchemaDefaultFunc {
 }
 
 func (*blockSchema) DefaultValue() (interface{}, error) {
-	panic("DefaultValue() should not be called during schema generation")
+	return nil, nil
```
t0yv0, line 123: Copy the comment here from another location ..

pf/tests/internal/testprovider/testbridge_resource_testnest.go
```diff
@@ -20,18 +20,30 @@ import (
...
 var _ resource.Resource = &testnest{}
+var _ resource.ResourceWithImportState = &testnest{}
 
 func newTestnest() resource.Resource {
 	return &testnest{}
 }
 
 func (*testnest) schema() rschema.Schema {
 	return rschema.Schema{
+		Attributes: map[string]rschema.Attribute{
```
t0yv0, line 39: Adding as requirement for testing import against this resource.

pf/internal/schemashim/block_schema.go
t0yv0: And there's another place I think where we do this.

pkg/tfbridge/schema.go
```diff
@@ -1407,7 +1406,12 @@ func extractInputs(oldInput, newState resource.PropertyValue, tfs shim.Schema, p
 }
 
 func getDefaultValue(tfs shim.Schema, ps *SchemaInfo) interface{} {
-	if dv, _ := tfs.DefaultValue(); dv != nil {
+	if dv, err := tfs.DefaultValue(); dv != nil {
+		if err != nil {
+			// Log error output but continue otherwise.
+			// This avoids a panic on preview such as https://github.com/pulumi/pulumi-cloudflare/issues/460.
+			glog.V(9).Infof(err.Error())
```
iwahbe, line 1413: If a user default value function returns an error, I would expect us to halt the `pulumi up` and display the error to the user. Swallowing a generic error, especially on v9, seems dangerous.

pkg/tfbridge/schema.go
```diff
@@ -1407,7 +1406,12 @@ func extractInputs(oldInput, newState resource.PropertyValue, tfs shim.Schema, p
 }
 
 func getDefaultValue(tfs shim.Schema, ps *SchemaInfo) interface{} {
-	if dv, _ := tfs.DefaultValue(); dv != nil {
+	if dv, err := tfs.DefaultValue(); dv != nil {
```
iwahbe, line 1409: We have changed `DefaultValue` to return `nil, fmt.Errorf("default values not supported")`. The error won't even be logged here, since `dv == nil`. We need to also check that `err != nil`.

iwahbe (commented):
I don't like that this will swallow a generic error. Can we give `fmt.Errorf("default values not supported")` a concrete global value and then only swallow that value (which cannot be user generated).

```go
// internal
var SchemaTimeDefaultErr = fmt.Errorf("default values not supported")

...

// pf/internal/schemashim/attr_schema.go
func (*attrSchema) DefaultValue() (interface{}, error) {
	return nil, SchemaTimeDefaultErr
}

...

// pkg/tfbridge/schema.go
func getDefaultValue(tfs shim.Schema, ps *SchemaInfo) interface{} {
	if dv, err := tfs.DefaultValue(); dv != nil &&
	err != nil &&
	!errors.Is(err, SchemaTimeDefaultErr) {
		if err != nil {
			return nil, err
		}
		return dv
	}
	return nil
}

File: pkg/tfbridge/schema.go (hunk 3 above)
Comment (guineveresaenger) [reply pkg/tfbridge/schema.go:1413]: The issue here is that there’s no fix for the user. We need to implement proper default value handling in the plug-in framework as a followup. Adding this error but emitting it as Info I felt was more informative than simply returning nil.

**Comment by guineveresaenger** [comment]
Using a custom error is the first thing I tried; the issue is that we can't declare it inside `tf/internal` and have it available from another package. Hopefully moving it to `tfbridge`, where we look for it, makes sense here.

We've been swallowing an error on tfs.DefaultValue for quite a while. I'm somewhat concerned that explicitly returning an error here is going to have other effects I'm not aware of.

File: pkg/tfbridge/schema.go (hunk 3 above)
Comment (t0yv0) [reply pkg/tfbridge/schema.go:1419]: Nope, we can't halt pulumi up in this case! This is the entire point of this PR. The code path is used to try to make import generated code better by taking default values into account, but Plugin Framework does not at this point support detecting default values, so the code must proceed with best-effort like returning nil. I think the behavior as written is what we want (modulo code style).

File: pkg/tfbridge/schema.go [hunk 5]
```diff
@@ -190,6 +189,11 @@ const defaultsKey = "__defaults"
 // marshaled back to assets by MakeTerraformOutputs.
 type AssetTable map[*SchemaInfo]resource.PropertyValue
 
+// SchemaDefaultValueErr catches the situation where pf/schemashim.DefaultValue() is called by ExtractInputsFromOutputs.
+// To avoid a panic, we return this error and log it to Info.
+// See https://github.com/pulumi/pulumi-cloudflare/issues/460
+var SchemaDefaultValueErr = fmt.Errorf("default values not supported")
```
t0yv0, line 195: You can just say `Internal` also in the comment - not something we expect users to need to link to.

t0yv0:
I think the is good to go modulo lint issues. 

```make lint``` should get you there.

```
  Running [/home/runner/golangci-lint-1.53.3-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  Warning: error-naming: error var SchemaDefaultValueErr should have name of the form ErrFoo (revive)
```

t0yv0:
@guineveresaenger can you reopen a PR in your name since you're driving this work. Then I can approve. thanks.

PR #1319, opened 2023-08-01
Support AutoName for Plugin Framework

t0yv0:
On top of https://github.com/pulumi/pulumi-terraform-bridge/pull/1318

This is the last bit in the AutoName support. AutoName machinery is modified to consult PriorState for the previously allocated name. This makes it work for Plugin Frameworks without __defaults__. 

There is a limitation in this scenario. Suppose you've created a resource with a manual name and then you want to switch to a AutoName (provider upgrade or just user program change). This won't work as the name is too sticky in PF - any updates to the resource will persist the value in the state, unless you change the program to have a non-empty name and then it will use that. You can't "drop to AutoName". In previous discussions with Friel we thought this should be an OK corner case to keep because most of the use cases of this feature require names to persist and changing them causes replace plans which is never what you want, so in cases like this the workaround of tearing down the resource and re-provisioning it should be OK.

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -100,7 +100,9 @@ func visitPropertyValue(name, path string, v resource.PropertyValue, tfs shim.Sc
 				// fill in default values for empty fields (note that this is a property of the field reader, not of
 				// the schema) as it does when computing the hash code for a set element.
 				ctx := &conversionContext{}
-				ev, err := ctx.MakeTerraformInput(ep, resource.PropertyValue{}, e, etfs, eps, rawNames)
+				// Lost context on the current path here, unfortunate for ComputedDefaults.
+				pp := resource.PropertyPath{}
+				ev, err := ctx.MakeTerraformInput(ep, resource.PropertyValue{}, e, etfs, eps, rawNames, pp)
 				if err != nil {
 					return
 				}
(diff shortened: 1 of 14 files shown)


pkg/tfbridge/provider_test.go
```diff
@@ -733,9 +733,10 @@ func TestCheck(t *testing.T) {
 			tf:     shimv2.NewProvider(testTFProviderV2),
 			config: shimv2.NewSchemaMap(testTFProviderV2.Schema),
 		}
-		computeStringDefault := func(opts ComputeDefaultOptions1) (interface{}, error) {
+		computeStringDefault := func(opts ComputeDefaultOptions) (interface{}, error) {
+			require.Equal(t, resource.PropertyPath{"stringPropertyValue"}, opts.PropertyPath)
 			if v, ok := opts.PriorState["stringPropertyValue"]; ok {
-				return v.StringValue(), nil
+				return v.StringValue() + "!", nil
```
t0yv0, line 739: This used to be a bad test, turns out this code never ran at all until `__defaults` was added in `olds`. Instead it proactively copied the past default value. Slightly odd but learning something new - it conservatively assumes a property with DefaultInfo should copy the default value if it's not found in __defaults. With the update it's obvious that the code here must run in the test.

pkg/tfbridge/diff.go
```diff
@@ -100,7 +100,9 @@ func visitPropertyValue(name, path string, v resource.PropertyValue, tfs shim.Sc
 				// fill in default values for empty fields (note that this is a property of the field reader, not of
 				// the schema) as it does when computing the hash code for a set element.
 				ctx := &conversionContext{}
-				ev, err := ctx.MakeTerraformInput(ep, resource.PropertyValue{}, e, etfs, eps, rawNames)
+				// Lost context on the current path here, unfortunate for ComputedDefaults.
```
t0yv0, line 103: Calling this out here. The surrounding code is hard to fully understand.

PR #1320, opened 2023-08-01
Deprecate pkg/tfshim/sdk-v1

t0yv0:
This is a helpful exercise as `make lint` exposes all the tests testing SDKv1 shim. We should move them over to SDKv2 shim which we care about a lot more.


iwahbe:
Looks like we will need to keep supporting (and thus testing) `sdk-v1`: https://github.com/pulumi/pulumi-github/blob/72f00503cbf04aaab3035e7d40efdb31c152e9b2/provider/resources.go#L70.

t0yv0:
That's unfortunate.

PR #1322, opened 2023-08-02
Encode default values

iwahbe:
Fixes https://github.com/pulumi/pulumi-aws/issues/2646

This PR makes two distinct changes:
- https://github.com/pulumi/pulumi-terraform-bridge/pull/1322/commits/cd9ad00a72ac00e3f62c24431cd5b9c53ee6db2b: adds an explicit check for accepting a bool that should be a string, changing the return value to `"true"` and `"false"` instead of `true` and `false`.
- https://github.com/pulumi/pulumi-terraform-bridge/pull/1322/commits/796df14c8400440358b5b2075cf16340c2f782b7: calls `MakeTerraformInput` on values drawn from `SchemaInfo.DefaultInfo.Value`. This uses the machinery introduced in the previous commit.

WIP: Still needs tests for both features.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -365,12 +365,26 @@ func (ctx *conversionContext) MakeTerraformInput(name string, old, v resource.Pr
 	case v.IsNull():
 		return nil, nil
 	case v.IsBool():
+		if tfs != nil && tfs.Type() == shim.TypeString {
+			if v.BoolValue() {
+				return "true", nil
+			}
+			return "false", nil
+		}
 		return v.BoolValue(), nil
 	case v.IsNumber():
-		if tfs != nil && tfs.Type() == shim.TypeFloat {
-			return v.NumberValue(), nil
+		var typ shim.ValueType
+		if tfs != nil {
+			typ = tfs.Type()
 		}
-		return int(v.NumberValue()), nil // convert floats to ints.
+		switch typ {
+		case shim.TypeFloat:
+			return v.NumberValue(), nil
+		case shim.TypeString:
+			return strconv.FormatFloat(v.NumberValue(), 'f', -1, 64), nil
+		default: // By default, we return ints
+			return int(v.NumberValue()), nil
+		}
 	case v.IsString():
 		return v.StringValue(), nil
 	case v.IsArray():
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -707,7 +707,12 @@ func (ctx *conversionContext) applyDefaults(result map[string]interface{}, olds,
 					defaultValue, source = tv, "config"
 				}
 			} else if info.Default.Value != nil {
-				defaultValue, source = info.Default.Value, "Pulumi schema"
+				v := resource.NewPropertyValue(info.Default.Value)
+				tv, err := ctx.MakeTerraformInput(name, resource.PropertyValue{}, v, tfi, psi, rawNames)
+				if err != nil {
+					return err
+				}
+				defaultValue, source = tv, "Pulumi schema"
 			} else if compute := info.Default.ComputeDefault; compute != nil {
 				v, err := compute(
 					// Getting the correct context needs to refactor public methods such as
--- a/pkg/tfbridge/schema_test.go
+++ b/pkg/tfbridge/schema_test.go
@@ -928,6 +928,10 @@ func TestDefaults(t *testing.T) {
 				"x1of1": {Type: shim.TypeString, ExactlyOneOf: x1ofN, DefaultFunc: fixedDefault("x1of1-value")},
 				"x1of2": {Type: shim.TypeString, ExactlyOneOf: x1ofN, DefaultFunc: fixedDefault(nil)},
 				"x1of3": {Type: shim.TypeString, ExactlyOneOf: x1ofN, DefaultFunc: fixedDefault(nil)},
+
+				// Default value application across types
+				"x2stringxbool": {Type: shim.TypeString},
+				"x2stringxint":  {Type: shim.TypeString},
 			})
 			ps := map[string]*SchemaInfo{
 				"eee": {Default: &DefaultInfo{Value: "EEE"}},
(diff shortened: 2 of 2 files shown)


pkg/tfbridge/schema.go
```diff
@@ -718,7 +724,12 @@ func (ctx *conversionContext) applyDefaults(result map[string]interface{}, olds,
 			}
 			if defaultValue != nil {
 				glog.V(9).Infof("Created Terraform input: %v = %v (from %s)", name, defaultValue, source)
-				result[name] = defaultValue
+				pValue := resource.NewPropertyValue(defaultValue)
```
t0yv0, line 727: Well this is a massive transformation introduced here! We need to be careful. We're losing track of what the actual type is of defaultValue because it all collapses to any. There's two options it's either "any" representation of pulumi.PropertyValue or it's the "any" representation of TF value (hcty.Value).

iwahbe: Your right. This transformation should have been applied earlier. Fixed in https://github.com/pulumi/pulumi-terraform-bridge/pull/1322/commits/796df14c8400440358b5b2075cf16340c2f782b7.

t0yv0: Eh! I'm not sure! That depends on what we expect users to put in defaultValue. Just need to be careful if we're changing it need to call it out... change management..

t0yv0:
Eh, I don't love this change. 

We started from a spurious warning in pulumi-aws that defines a SchemaInfo like this:

```
			"skip_metadata_api_check": {
				Type: "boolean",
				Default: &tfbridge.DefaultInfo{
					Value: true,
				},
			},
```

And actually models this filed as a String in TF land. It sounds like Pulumi->TF conversion wasn't coalescing `true` boolean to `"true"` before this change and now it does that. So far so good, and the same is extended for numbers (is there an end-user ask/ actual use case?).

The tricker bit is this:

>  calls MakeTerraformInput on values drawn from SchemaInfo.DefaultInfo.Value.

This has far more scope than simply coalescing bools to a desired string slot. The function is interesting, it can do things like rename object fields from Pulumi to TF spelling (although maybe it tolerates both spellings?), it can expand `x` to `[x]` in MaxItems=1 situation, and so on. 

While logically it would *make sense* that we ask the users to write in values *as Pulumi understands them* into DefaultInfo.Value, it's not obvious to me that the change wouldn't break some providers that did decide to write in values *as Terraform understands them* to make things work, and perhaps those providers would break now, at runtime.

At the very least this needs to be called out very obviously in release notes. I'd also love to understand why perhaps this is not a concern, are we scanning existing providers for DefaultInfo.Value, or we believe that the change can't in principle break (and can demonstrate in tests).

Note that it is possible to solve the skip_metadata_api_check weakening the change to conservatively *only* do type coalesce operation when the type mismatches Bool != String. This opens up the possibility of taking the other improvement in a separate enhancement PR.

iwahbe:
> Eh, I don't love this change.
> 
> We started from a spurious warning in pulumi-aws that defines a SchemaInfo like this:
> 
> ```
> 			"skip_metadata_api_check": {
> 				Type: "boolean",
> 				Default: &tfbridge.DefaultInfo{
> 					Value: true,
> 				},
> 			},
> ```
> 
> And actually models this filed as a String in TF land. It sounds like Pulumi->TF conversion wasn't coalescing `true` boolean to `"true"` before this change and now it does that. So far so good, and the same is extended for numbers (is there an end-user ask/ actual use case?).

So far so good: this change prevents the warning when the user explicitly sets `skipMetadataApiCheck`. The first change does not fix the error in the common case where the default value is used, because that is not processed by `MakeTerraformInputs` (hence the second change).

I'm not aware of a concrete use case for `int->string`, but it's the same problem so we should include the change for consistency. It made writing the test easier, since we can take the existing test case for `MakeTerraformOutputs` and assert that `MakeTerraformInputs` is its inverse. (modulo null uncertainty)

> The tricker bit is this:
> 
> > calls MakeTerraformInput on values drawn from SchemaInfo.DefaultInfo.Value.
> 
> This has far more scope than simply coalescing bools to a desired string slot. The function is interesting, it can do things like rename object fields from Pulumi to TF spelling (although maybe it tolerates both spellings?), it can expand `x` to `[x]` in MaxItems=1 situation, and so on.
> 
> While logically it would _make sense_ that we ask the users to write in values _as Pulumi understands them_ into DefaultInfo.Value, it's not obvious to me that the change wouldn't break some providers that did decide to write in values _as Terraform understands them_ to make things work, and perhaps those providers would break now, at runtime.

User's are required to write in Pulumi types, since schema validation enforces that `Type` matches `DefaultInfo.Value` [here](https://github.com/pulumi/pulumi/blob/244d4265d16ffedea6db1f65090d9d87652ffca9/pkg/codegen/schema/bind.go#L955-L960).[^1] This allows us to guarantee that users have not written values in TF types instead of Pulumi types; doing otherwise would error at schema validation. User's are only allowed to write simple values, not lists of objects, so `MaxItemsOne` and field renaming are not relevant. If we ever decided to accept collection types as `DefaultInfo.Value`, users would need to write pulumi values and we would then need to process them info TF values.

> At the very least this needs to be called out very obviously in release notes. I'd also love to understand why perhaps this is not a concern, are we scanning existing providers for DefaultInfo.Value, or we believe that the change can't in principle break (and can demonstrate in tests).

We will call this out in the changelog. This change, like any bug fix, can break a provider that requires the old behavior. I don't expect to find this scenario in the wild. TF's frameworks expect that a stringly typed bool is represented as {"true", "false"} instead of {"1", "0"}. If we have any providers that require {"1", "0"}, they can provide a custom transformation here, but it will be nonstandard. This part of the bridge should work in the common case.

> Note that it is possible to solve the skip_metadata_api_check weakening the change to conservatively _only_ do type coalesce operation when the type mismatches Bool != String. This opens up the possibility of taking the other improvement in a separate enhancement PR.

That is a more conservative approach, and we could do that. In the absence of evidence[^2] that this will break providers, I don't think the caution is necessary. Doing the general conversion is the more "correct" semantic, giving TF providers the types they have specified. It provides a cleaner separation between Pulumi types and TF types, which is what we want since TF providers are not aware of Pulumi types.

[^1]: In the case were `.Type` is left empty, it is filled in with the TF value during schema generation. In that case the Pulumi type is the TF type, and `.Value` must be of that type.
[^2]: I'm aware that absence of evidence isn't evidence of absence, but it is suggestive of absence.

t0yv0:
Ok, I filed https://github.com/pulumi/pulumi-terraform-bridge/issues/1325 because it's not obvious why compound values are not supported. This was not obvious to me why that wouldn't work. Note that emitting default values into the Pulumi Package Schema seems to be an entirely optional feature to uncover more information about the provider in the documentation and generated language SDKs, as the bridge code still consults default values for runtime application. In that sense it could be possible to support compound default values just fine, which I assumed was the case. 

Anyway that can wait it's time as a feature request.

For now let's merge this, the state space is a lot more constrained if compound objects are not allowed.

t0yv0 (approved):

PR #1323, opened 2023-08-02
Autoname implementation alternative: PriorValue

t0yv0:
This is an alternative where we do not expose the Path to the changed value but expose the Value from prior state. This is perhaps more intuitive to the user and works just as well for the AutoName use case.


iwahbe (approved):
This feels cleaner to me.

PR #1327, opened 2023-08-04
Remove panic on imports for TF Plugin Framework

guineveresaenger:
This PR removes the panics on getDefaultValue on the Type, Attr, and Block schemas in the TF Plugin Framework schemashim, as they are causing issues when importing resources using the Plugin Framework.

Instead, we emit nil and an error, which we log using our standard logger, allowing the code to run without panic. 
I have built and run this against the [user reported panic](pulumi/pulumi-cloudflare#460) and verified that this resolves the issue.

We will file a followup issue to implement properly handling default values directly inside the pf package.

Fixes pulumi/pulumi-cloudflare#460.

- Return nil instead of panic from block schema
- resolve merge conflict
- fix linting
- Remove panic from Type schema
- Return a custom error message and log it during getDefaultValue
- Create global err and fix logic. Return explicit error if error is not nil or ErrSchemaDefaultValue


pf/tests/internal/testprovider/testbridge_resource_testres.go
```diff
@@ -18,25 +18,25 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/mapplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/numberplanmodifier"
```
t0yv0, line 25: Nit: imports were grouped by system first, framework later.

pkg/tfbridge/schema.go
```diff
@@ -190,6 +189,10 @@ const defaultsKey = "__defaults"
 // marshaled back to assets by MakeTerraformOutputs.
 type AssetTable map[*SchemaInfo]resource.PropertyValue
 
+// ErrSchemaDefaultValue is used internally to avoid a panic in pf/schemashim.DefaultValue().
+// See https://github.com/pulumi/pulumi-cloudflare/issues/460
```
t0yv0, line 193: Nit, this is not a great issue to link into the source code for historic reference, as it does not explain the end to-end context of the problem. Let me write an issue here in terraform-bridge and we can link it instead.

t0yv0:
Let's use this one https://github.com/pulumi/pulumi-terraform-bridge/issues/1329

PR #1330, opened 2023-08-04
Avoid spurious diffs caused by schema.NormalizeObjectFromLegacySDK

t0yv0:
This change is currently under CtyInstanceState flag and so it can start rolling out targeting just one affected resource in AWS, WAFv2 WebACL.

There is a deeper problem captured here in the test, the way we use the TF primitives conflicts.

- schema.NormalizeObjectFromLegacySDK substitutes nil with empty lists.
- NewResourceConfigShimmed substitutes in the other direction

Comparing their results with Diff may cause the same data to appear to cause the diff.

In pulumi/pulumi-aws#2664 the exact issue is with `.rule[0].action.allow.customRequestHandler` field. It is truly absent from the user program, but schema.NormalizeObjectFromLegacySDK injects it back valued as an empty list. Once it's injected in the state, set element hashing gets confused about the hash of `.rule[0]`, so we get different hash for the rule in the state from the rule in config, causing a replace plan.

Avoiding the normalization pass seems like the simplest fix in terms of LOC changed, but this warrants future work to ensure inconsistent normalization does not affect us further.

### DIFF
--- a/pkg/tests/internal/webaclschema/webacl.go
+++ b/pkg/tests/internal/webaclschema/webacl.go
@@ -83,8 +83,14 @@ func ResourceWebACL() *schema.Resource {
 				name := idParts[1]
 				scope := idParts[2]
 				d.SetId(id)
-				d.Set("name", name)
-				d.Set("scope", scope)
+				err := d.Set("name", name)
+				if err != nil {
+					return nil, err
+				}
+				err = d.Set("scope", scope)
+				if err != nil {
+					return nil, err
+				}
 				return []*schema.ResourceData{d}, nil
 			},
 		},
--- a/pkg/tests/internal/webaclschema/webacl.go
+++ b/pkg/tests/internal/webaclschema/webacl.go
@@ -1,0 +1,195 @@
+// Copied from https://github.com/pulumi/terraform-provider-aws/blob/9106d49534199fa6d12d33c27c6cbd4ff777102c/internal/service/wafv2/web_acl.go
+//
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package wafv2
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"hash/crc32"
+	"regexp"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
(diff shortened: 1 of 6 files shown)


t0yv0:
So this boils down to .allow.customRequestHandling. There is a difference, as far as the default Set element computation code is concerned, between an missing field and a field set to null.

If I change the "news" allow.customRequestHandling to be "null" to match "olds", testCase2 starts to pass.


```
		    "allow": {
		      "__defaults": [],
                      "customRequestHandling": null
		    }
```

t0yv0:
```go
func allowConfigSchema() *schema.Schema {
	return &schema.Schema{
		Type:     schema.TypeList,
		Optional: true,
		MaxItems: 1,
		Elem: &schema.Resource{
				"custom_request_handling": customRequestHandlingSchema(),
			},
		},
	}
}

```

OK so the origin of null in the state is from flattening this MaxItems=1 structure that's missing an element, this creates the original `null`.

pkg/tfshim/sdk-v2/provider_diff.go
```diff
@@ -94,6 +94,7 @@ func (p v2Provider) simpleDiff(
 			}
 			state.RawState = priorStateVal
 		}
+
```
iwahbe, line 97: Can we remove the spurious diff?

iwahbe (approved):
LGTM

pkg/tfshim/sdk-v2/provider_diff.go
t0yv0: What do you mean? Care to elaborate?

t0yv0: Oh yes the extra line. I'm confused what Diff means here I thought InstanceDiff. Of course.

iwahbe: > Avoid spurious diffs caused by schema.NormalizeObjectFromLegacySDK #1330

Oh, I can see why that would be confusing given the PR context. My bad.

pkg/tests/internal/webaclschema/webacl.go
```diff
@@ -0,0 +1,195 @@
+// Copied from https://github.com/pulumi/terraform-provider-aws/blob/9106d49534199fa6d12d33c27c6cbd4ff777102c/internal/service/wafv2/web_acl.go
```
t0yv0, line 1: @lukehoban borrowing schema declarations for testing

lukehoban (approved):

PR #1333, opened 2023-08-09
Add logging context to `PreConfigureCallback` and `PreCheckCallback`

iwahbe:
Add the ability to return structured logs from `Pre*Callback` methods.

---

Side note:

We should deprecate `PreConfigureCallbackWithLogger` for `PreConfigureCallbackWithCtx` with an attached logger. This will fix the interface (not directly exposing callers to `*provider.Host`) and allow us to attach additional context data in a consistent way.

---

This PR contains **BREAKING CHANGES** to the following functions. All three functions now take a `context.Context` as their first argument. None of the functions are used outside of the bridge, according to a GH code search:
- `tfbridge.MakeTerraformInputs`: https://github.com/search?q=%2FMakeTerraformInputs%2F+-is%3Afork&type=code
- `tfbridge.MakeTerraformState`: https://github.com/search?q=%2FMakeTerraformState%2F+-is%3Afork&type=code
- `tfbridge.MakeTerraformConfig`: https://github.com/search?q=%2FMakeTerraformConfig%2F+-is%3Afork&type=code

Given the unique context where these functions make sense, I believe they are unlikely to be used outside the bridge.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -31,6 +31,7 @@ import (
 
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/logging"
 )
 
 const (
@@ -119,6 +120,38 @@ type ProviderInfo struct {
 	// the bridge completed its original version based on the TF schema.
 	// A hook to enable custom schema modifications specific to a provider.
 	SchemaPostProcessor func(spec *pschema.PackageSpec)
+}
+
+// Send logs or status logs to the user.
+//
+// Logged messages are pre-associated with the resource they are called from.
+type Logger interface {
+	Log
+
+	// Convert to sending ephemeral status logs to the user.
+	Status() Log
+}
+
+// The set of logs available to show to the user
+type Log interface {
+	Debug(msg string)
+	Info(msg string)
+	Warn(msg string)
+	Error(msg string)
+}
+
+// Get access to the [Logger] associated with this context.
+//
+// If called on a context that does not have an associated [Logger], GetLogger will panic.
+//
+// Currently, only the context from [ResourceInfo.PreCheckCallback] and
+// [ProviderInfo.PreConfigureCallback] have an associated [Logger].
+func GetLogger(ctx context.Context) Logger {
+	logger, ok := ctx.Value(logging.CtxKey).(Logger)
+	if !ok {
+		panic("Cannot call GetLogger on a context that is not equipped with a Logger")
+	}
+	return logger
 }
 
 func (info *ProviderInfo) GetConfig() map[string]*SchemaInfo {
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -141,11 +141,6 @@ type Log interface {
 }
 
 // Get access to the [Logger] associated with this context.
-//
-// If called on a context that does not have an associated [Logger], GetLogger will panic.
-//
-// Currently, only the context from [ResourceInfo.PreCheckCallback] and
-// [ProviderInfo.PreConfigureCallback] have an associated [Logger].
 func GetLogger(ctx context.Context) Logger {
 	logger, ok := ctx.Value(logging.CtxKey).(Logger)
 	if !ok {
--- a/pkg/tfbridge/log.go
+++ b/pkg/tfbridge/log.go
@@ -21,7 +21,10 @@ import (
 
 	"github.com/pulumi/pulumi/pkg/v3/resource/provider"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/logging"
 )
 
 // LogRedirector creates a new redirection writer that takes as input plugin stderr output, and routes it to the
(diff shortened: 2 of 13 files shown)


pkg/tfbridge/info.go
```diff
@@ -121,6 +122,38 @@ type ProviderInfo struct {
 	SchemaPostProcessor func(spec *pschema.PackageSpec)
 }
 
+// Send logs or status logs to the user.
+//
+// Logged messages are pre-associated with the resource they are called from.
+type Logger interface {
```
t0yv0, line 128: Could we use the existing interface from the Go SDK here?
 https://github.com/pulumi/pulumi/blob/ee25d63598eae99b4bc170d7b0595fd8609abf50/sdk/go/pulumi/log.go#L29-#L29 

What is Status()? Is it the same as LogArgs.Ephemeral = true?

pkg/tfbridge/log.go
```diff
@@ -114,3 +118,49 @@ func (lr *LogRedirector) Write(p []byte) (n int, err error) {
...
+// hostLogger is parameterized by L to allow `Status` to return the native Log type of a
+// stable Logger.
+type hostLogger struct {
+	ctx  context.Context
+	host *provider.HostClient
+	urn  resource.URN
+
+	status bool
+
+	err *multierror.Error
```
t0yv0, line 131: The err accumulator is eyebrow-raising. Why would you do that? What kind of errors are you trying to accumulate here?

pkg/tfbridge/log.go
```diff
@@ -114,3 +118,49 @@ func (lr *LogRedirector) Write(p []byte) (n int, err error) {
...
+}
+
+func (l *hostLogger) f(severity diag.Severity, msg string) {
+	f := l.host.Log
+	if l.status {
+		f = l.host.LogStatus
+	}
+	err := f(l.ctx, severity, l.urn, msg)
+	if err != nil {
+		l.err.Errors = append(l.err.Errors, err)
```
t0yv0, line 141: Looks like we're accumulating errors from being unable to send a log message to the engine, perhaps because the engine is shutting down. Why? Would we ever want to change control flow behavior based on not being able to reach the engine? An alternative here is to dump it to glog low-level stream at verbose 9 level and proceed as if nothing happened.

pkg/tfbridge/provider.go
```diff
@@ -324,10 +325,19 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
...
 			// NOTE: the user code may modify news in-place.
-			validationErrors := p.info.PreConfigureCallbackWithLogger(ctx, p.host, news, config)
-			if validationErrors != nil {
-				return nil, validationErrors
+			var logErr multierror.Error
+			ctx := ctxWithHostLogger(ctx, p.host, resource.URN(req.GetUrn()), &logErr)
+			err := p.info.PreConfigureCallbackWithLogger(ctx, p.host, news, config)
+			// If the only error is from the check itself, we let that through as-is,
+			// because it might be a grpc aware error.
+			if err != nil && len(logErr.Errors) == 0 {
```
t0yv0, line 333: I'm not sure I understand error handling here, are we trying to promote accumulated errors to an error result from Check? What's an example where we need to do this?

t0yv0:
This is great directionally, love utilizing Context more for propagating the logger interface. Had some questions above.

pkg/tfbridge/info.go
iwahbe: We can't use that interface because it expects `Resource`s, not `URN`s. It's at the wrong level.

`Status()` is the same `LogArgs.Ephemeral = true`. The normal `provider.Host.Log` has `Debug`, `DebugStatus`, etc arguments. That's where I took the name from.

pkg/tfbridge/log.go
t0yv0: Why ever collect them if we don't want to expose them? What do you want to do with them? Can they just be flushed to glog and forgotten?

pf/tfbridge/provider_checkconfig.go
```diff
@@ -72,6 +74,8 @@ func (p *provider) CheckConfigWithContext(
 			if !ok {
 				hc = &rprovider.HostClient{}
 			}
+
+			ctx := context.WithValue(ctx, unstable.LoggerCtxKey, p.logger(urn))
```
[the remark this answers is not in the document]
iwahbe: We do `PreCheckCallback` for sdk and pf. We do `PreConfigureCallback` for sdk and pf. I'm adding `ComputeDefault`. Can you think of any other places?

pkg/tfbridge/info.go
t0yv0: Ah yes. Well there's something to be said about making codebases more independent by not borrowing interfaces anyway. Thanks!

t0yv0:
The breaking changes are no-go I think. We have large customer codebases that use the bridge heavily that are not part of public GH code search. If we need to do this we can add tfbridge.MakeTerraformInputsWithContext variants and move prod to use them, deprecating previous ones.

pf/tfbridge/provider_checkconfig.go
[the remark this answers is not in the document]
iwahbe: I have injected the logger at the root context of all provider gRPC calls. This means that it should be everywhere (that is correctly threaded from the original context).

t0yv0:
```
error: resource or variable named "unfairoRandomNumber" could not be found

  on Pulumi.yaml line 61:
  61:   unfairRandomNumberFair__actual: ${unfairoRandomNumber.fair}

```

I think this depends on https://github.com/pulumi/pulumi-terraform-bridge/pull/1336 then. Let's see.

pkg/tfbridge/info.go
```diff
@@ -121,6 +122,33 @@ type ProviderInfo struct {
...
+	Log
+
+	// Convert to sending ephemeral status logs to the user.
```
t0yv0, line 131: Nit: if there's any pointers on what these are or how they're different, perhaps a link, that'd be great; but not blocking on this.

t0yv0 (approved):
LGTM once we get the tests to pass. Double-checked the dependent code corpus analysis and it seems like we're OK taking a minor break on the functions that de-facto are used internally - the MakeTerraformInput and so on. We will need to call it out in real notes and revert if there's any complaints.

pkg/tfbridge/info.go
iwahbe: Opened https://github.com/pulumi/pulumi-terraform-bridge/issues/1338 to track. It's late in the day and I want to get an alpha out today, so I'll come back to this later.

Issue #1335, opened 2023-08-10
CI Broken: `TestBasicProgram` failing in CI


iwahbe:
### What happened?

See https://github.com/pulumi/pulumi-terraform-bridge/pull/1334#issuecomment-1673586108

### Expected Behavior

CI is green unless a bug is introduced.

### Steps to reproduce

See https://github.com/pulumi/pulumi-terraform-bridge/pull/1334#issuecomment-1673586108.

### Output of `pulumi about`

NA

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

iwahbe:
Tests pass on Pulumi-yaml v1.1.1, but break on Pulumi-yaml v1.2.0. Manual bisection shows that https://github.com/pulumi/pulumi-yaml/pull/483 is the breaking change.

PR #1336, opened 2023-08-10
Restore CI

iwahbe:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1335

### DIFF
--- a/pf/tests/testdata/basicprogram/Pulumi.yaml
+++ b/pf/tests/testdata/basicprogram/Pulumi.yaml
@@ -124,8 +124,11 @@ outputs:
   testMinimal__expect: |-
     input1
 
-  testComputedNull__actual:
-    fn::toJSON: ${minimalRes.optionalInputStringCopy}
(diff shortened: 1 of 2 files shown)


t0yv0:
Do we know what the cause of  FAIL: TestBasicProgram (16.30s) is? Is there a workaround?

iwahbe:
> Do we know what the cause of FAIL: TestBasicProgram (16.30s) is? Is there a workaround?

We test against the latest install of the pulumi binary, and it hit https://github.com/pulumi/pulumi-yaml/issues/489. The workaround is to skip the test or wait for a good release (which has happened).

t0yv0:
https://github.com/pulumi/pulumi-terraform-bridge/issues/1337

t0yv0:
🚢 we'll pin later.

iwahbe:
I raised https://github.com/pulumi/pulumi-yaml/issues/492 for after the break.

PR #1340, opened 2023-08-19
Avoid NaN in JSON serialization

lukehoban:
When there are no conversions, we would compute NaN and then try to serialize that to JSON which fails.

Fixes https://github.com/pulumi/pulumi-rke/issues/54.


t0yv0 (approved):
Brilliant, thank you!

Issue #1280, opened 2023-07-07
[Epic] Switch all tfgen uses to the new converter


Frassle:
This epic tracks rollout of Terraform to Pulumi example converter to improve coverage and quality of examples in Registry.

https://docs.google.com/document/d/1FRQAJd17sTjfKYVhfuN9N9AtZNfRLlQv37-sIwEnDyA/edit?usp=sharing

### Work Items ⚙️

- [x] pulumi/pulumi-terraform-bridge#1352
- [x] Release bridge version with converter inside
- [x] Script comparing baseline/experimental https://github.com/pulumi/pulumi-terraform-bridge/pull/1349
- [x] https://github.com/pulumi/pulumi-terraform-bridge/issues/1413
- [x] Re-triage example quality issues and close the ones fixed
  - [x] Confirm if closed pulumi/pulumi-converter-terraform#56
  - [x] Confirm if closed pulumi/pulumi-converter-terraform#57
  - [x] Confirm if closed pulumi/pulumi-terraform-bridge#408
  - [x] TBD find more issues here
- [x] https://github.com/pulumi/ci-mgmt/issues/889
- [ ] Remove the PULUMI_CONVERT gate and commit to the new converter
- [ ] Follow up: check if conversion works on Windows, remove Windows-specific test skips
- [ ] Share a migration note with Pulumiverse and paying customers mentioning the need to pin Pulumi CLI
- [ ] Make PULUMI_CONVERT=true the default value
- [ ] Delete dead code

### Tier 2 Provider Upgrade Procedure

Since there are many of these we will use a simplified procedure. ci-mgmt will be updated to set PULUMI_CONVERT=true and ensure Pulumi CLI is pinned; then these providers will receive the improved converter during routine updates as they pull in new tfbrdige versions. Apparently there is a system in place that tracks example cov metrics  on S3, so we can monitor the process using this system.

### Associated Docs Bugs
Fixes pulumi/pulumi-terraform-bridge#1399

t0yv0:
Please.. let's do this.

t0yv0:
Starting to look at this. All the meaningful changes are scoped to tf2pulumi package. It turns out
tfgen package depends on tf2pulumi package through a fairly small set of APIs.
  
Option 1:

- add interfaces between tfgen and tf2pulumi
- move tf2pulumi to tfpulumilegacy implementing the interfaces
[...]

PR #1349, opened 2023-08-31
Script converter metrics comparison

t0yv0:
I'd like to add this script temporarily to keep it checked in while we're rolling out the new PULUMI_EXPERIMENTAL converter. It will help computing the metrics difference between baseline/experimental, something I can attach to PRs during rollout.


iwahbe (approved):
Looks good. Can we add a comment explaining that the script is temporary or put it in `unstable/scripts/...`?

Tests look like they are failing because of https://github.com/pulumi/ci-mgmt/pull/537 (fixed by https://github.com/pulumi/pulumi-terraform-bridge/pull/1350).

t0yv0:
https://github.com/pulumi/pulumi-terraform-bridge/issues/1280 is the context. Yes will do.

unstable/scripts/exconverter/main.go
```diff
@@ -0,0 +1,168 @@
...
+		s.exampleByHCL[entry.ExampleName] = entry
+	}
+
+	ss, err := os.ReadFile(filepath.Join(dir, "shortSummary.txt"))
+	noerr(err)
+	s.shortSummary = string(ss)
+	return s
+}
+
+func tfgen(convert int) {
```
iwahbe, line 153: I'm getting strong C vibes here. Why isn't convert a `bool`?

unstable/scripts/exconverter/main.go
```diff
@@ -0,0 +1,168 @@
...
+	s := stats{
+		exampleByHCL: map[string]flattenedExample{},
+	}
+
+	dec := json.NewDecoder(f)
+	for {
+		var entry flattenedExample
+		err := dec.Decode(&entry)
+		if err != nil || entry.ExampleName == "" {
+			break
```
iwahbe, line 141: It feels like this should be a `continue`. Why are we stopping on the first error?
Why don't we report the error?

iwahbe (approved):
It looks like this script needs to be run in the directory of the provider that it is testing. That would be an excellent thing to put in a comment.

PR #1355, opened 2023-09-06
Introduce a hook for editing state (outputs)

t0yv0:
When working on https://github.com/pulumi/pulumi-aws/issues/2778 I would like to be able to remove "tagsAll" from Pulumi state file before it gets persisted, although the underlying TF provider keeps populating this output property. Unfortunately it's really non-trivial to do this in the provider itself without changes to the bridge. The proposed change here introduces a hook for editing the state right before it gets sent to the engine which includes these gRPC methods:

- Create
- Update
- Read

In particular the hook edits the `outputs` bits of the state. The `inputs` bits can be normally modified in PreCheckCallback, although there is no mechanism yet to modify them in the Read scenario (pulumi refresh or import). It's left for future work as not really required by the scenario being addressed here.

### DIFF
--- a/pf/tfbridge/provider_read.go
+++ b/pf/tfbridge/provider_read.go
@@ -65,7 +65,7 @@ func (p *provider) ReadWithContext(
 		return result, ignoredStatus, err
 	}
 
-	if result.Outputs != nil {
+	if result.Outputs != nil && rh.pulumiResourceInfo.TransformOutputs != nil {
 		var err error
 		result.Outputs, err = rh.pulumiResourceInfo.TransformOutputs(ctx, result.Outputs)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


t0yv0:
Going to throw a few unit tests in here.

pf/tfbridge/provider_read.go
```diff
@@ -65,6 +65,14 @@ func (p *provider) ReadWithContext(
 		return result, ignoredStatus, err
 	}
 
+	if result.Outputs != nil {
```
iwahbe, line 68: Tests are great.
```suggestion
	if result.Outputs != nil && rh.pulumiResourceInfo.TransformOutputs != nil {
```

iwahbe (approved):
This looks good to me. 

Let's make sure there are unit tests here before merging.

PR #1364, opened 2023-09-08
Use coverpkg=./... to pull in coverage of more files

mjeffryes:
by default, Go coverage reports ignore files in directories with no tests. Setting coverpkg=./... tells go that you want coverage on all go files under the current directory


iwahbe (approved):

iwahbe:
I don't see how this would cause test failures. I'm retrying to see if that improves things.

t0yv0:
That's good but I bet we're missing still the entire pf/go.mod module

Aslo note https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pf/Makefile#L18

When one module covers another module do we need to do something special to capture that or it doesn't matter?

mjeffryes:
Latest attempt; it's pulling in pf files now, there's a few files supporting tests that it would be nice to exclude though; I lean toward doing that in codecov's config rather than with go's coverpkg setting.

t0yv0 (approved):

PR #1366, opened 2023-09-14
stringEncoder fromPropertyValue will now format numbers

iwahbe:
This is necessary for backwards compatibility with SDKv{1,2} based provider. It is
dangerous; there are many ways to represent bools and numbers so we are unable to round trip.

This is the bridge side fix for https://github.com/pulumi/pulumi-aws/issues/2806.

### DIFF
--- a/pf/internal/convert/convert_test.go
+++ b/pf/internal/convert/convert_test.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"math/big"
 	"reflect"
+	"strconv"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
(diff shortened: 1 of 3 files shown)


pf/internal/convert/convert_test.go
```diff
@@ -27,17 +28,28 @@ import (
...
+	prop     resource.PropertyValue
+	normVal  func(tftypes.Value) any
+	normProp func(resource.PropertyValue) any
 }
 
 func TestConvertTurnaround(t *testing.T) {
 	t.Parallel()
 
 	cases := convertTurnaroundTestCases(tftypes.String, resource.NewStringProperty, "", "test-string")
+	cases = append(cases, convertTurnaroundTestCases(tftypes.String, func(x string) resource.PropertyValue {
```
t0yv0, line 43: Nit: since these don't really turnaround, we could have just added parsing test cases rather than extending the turnaround test setup. But this works.

t0yv0 (approved):
This is as fine fix as any we can take in the bridge; we become tolerant of mis-typed values and coalesce bools and numbers to strings instead of failing with an error message. The converter applies to *everything* - config values, resource inputs, function inputs.

I think I've seen this type of coalescing before in the bridge and I think given bw compact requirements I'm leaning to accept that this is the bridge's job.

A strong alternative to consider is to push back on the original ticket and ask folks to write proper YAML so their strings don't get garbled through numeric conversion. @mikhailshilkov perhaps to make a quick call. There's always the argument that "this used to work" so I lean to include the fix.

mikhailshilkov:
> There's always the argument that "this used to work" so I lean to include the fix.

Same

pf/internal/convert/convert_test.go
iwahbe: I wanted us to confirm that we turn around the correct "value", even if the types don't line up.

Issue #1365, opened 2023-09-14
tfgen doesn't detect existing TF Docs


ringods:
### What happened?

I'm working towards getting https://github.com/pulumiverse/pulumi-acme/ fully working, docs and all. However, if you clone the repo and run make tfgen, it complains that the upstream docs can't be found:

```
/Users/ringods/Projects/pulumiverse/pulumi-acme/bin/pulumi-tfgen-acme schema --out provider/cmd/pulumi-resource-acme
warning: could not find docs for resource 'acme_certificate'. Override the Docs property in the resource mapping. See type tfbridge.DocInfo for details.
warning: could not find docs for resource 'acme_registration'. Override the Docs property in the resource mapping. See type tfbridge.DocInfo for details.
```

### Expected Behavior

The upstream TF repo has the docs here: 

https://github.com/vancluever/terraform-provider-acme/tree/main/docs/resources

I would expect these are found and processed when running my `tfgen` binary.

### Steps to reproduce

* Clone https://github.com/pulumiverse/pulumi-acme/
* Execute `make tfgen`

### Output of `pulumi about`

Not applicable

### Additional context

According the tfgen code, this path should match the checkIfNewDocsExist validation:
https://github.com/pulumi/pulumi-terraform-bridge/blob/ca7fd3a6261828e23280232800ddeec36b46df2e/pkg/tfgen/docs.go#L436-L441
The filenames in there do not have the acme prefix, so IMO these should be picked up automatically by this markdown name:
https://github.com/pulumi/pulumi-terraform-bridge/blob/ca7fd3a6261828e23280232800ddeec36b46df2e/pkg/tfgen/docs.go#L192
Even that aside, I tried to set the DocInfo and use the Source property in there, but to no avail.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

iwahbe:
@ringods You need to specify `TFProviderModuleVersion: "v2"`. (https://github.com/pulumiverse/pulumi-acme/pull/6) I'm going to leave this issue open a tracking issue for the error message.

ringods:
@iwahbe Tnx. That indeed fixes it for me.

PR #1367, opened 2023-09-14
Separate file acquisition from file parsing

iwahbe:
For a given resource, the bridge:
1. Finds the file associated with the resource.
2. Parses the file into `entityDocs`

Currently, steps (1) and (2) are tied together in one function: `getDocsForResource`. This PR encapsulates (1) behind an interface (`DocsSource`) which is then passed to `getDocsForResource`. This makes testing much easier, and also simplifies our code.

---

The only user observable change in functionality is a much more obvious and useful error message when unable to find the upstream repository. The new warning looks like this:

```
warning: Unable to find the upstream provider's documentation:

The upstream repository is expected to be at "github.com/vancluever/terraform-provider-acme".

If the expected path is not correct, you should check the values of these fields (current values shown):

tfbridge.ProviderInfo{
        GitHubHost:              "github.com",
        GitHubOrg:               "vancluever",
        Name:                    "acme",
        TFProviderModuleVersion: "",
}

The original error is: error running 'go mod download -json github.com/vancluever/terraform-provider-acme' in "/Users/ianwahbe/go/src/github.com/pulumiverse/pulumi-acme/provider" dir for module: exit status 1

Output: {
        "Path": "github.com/vancluever/terraform-provider-acme",
        "Error": "module github.com/vancluever/terraform-provider-acme: not a known dependency"
}
```

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1365

### DIFF
--- a/pf/tests/internal/testprovider/tls.go
+++ b/pf/tests/internal/testprovider/tls.go
@@ -54,13 +54,14 @@ func TLSProvider() tfbridge.ProviderInfo {
 	}
 
 	return tfbridge.ProviderInfo{
-		Name:        "tls",
-		P:           tfpf.ShimProvider(tlsshim.NewProvider()),
-		Description: "A Pulumi package to create TLS resources in Pulumi programs.",
-		Keywords:    []string{"pulumi", "tls"},
-		License:     "Apache-2.0",
-		Homepage:    "https://pulumi.io",
-		Repository:  "https://github.com/pulumi/pulumi-tls",
+		Name:             "tls",
+		P:                tfpf.ShimProvider(tlsshim.NewProvider()),
+		Description:      "A Pulumi package to create TLS resources in Pulumi programs.",
+		Keywords:         []string{"pulumi", "tls"},
+		License:          "Apache-2.0",
+		Homepage:         "https://pulumi.io",
+		Repository:       "https://github.com/pulumi/pulumi-tls",
+		UpstreamRepoPath: ".",
 		Resources: map[string]*tfbridge.ResourceInfo{
 			"tls_cert_request":        {Tok: tlsResource(tlsMod, "CertRequest")},
 			"tls_locally_signed_cert": {Tok: tlsResource(tlsMod, "LocallySignedCert")},
--- a/pkg/tfgen/generate_schema_test.go
+++ b/pkg/tfgen/generate_schema_test.go
@@ -17,7 +17,11 @@ package tfgen
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 	"text/template"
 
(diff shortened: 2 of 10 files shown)


pf/tests/internal/testprovider/tls.go
```diff
@@ -54,13 +54,14 @@ func TLSProvider() tfbridge.ProviderInfo {
...
-		Homepage:    "https://pulumi.io",
-		Repository:  "https://github.com/pulumi/pulumi-tls",
+		Name:             "tls",
+		P:                tfpf.ShimProvider(tlsshim.NewProvider()),
+		Description:      "A Pulumi package to create TLS resources in Pulumi programs.",
+		Keywords:         []string{"pulumi", "tls"},
+		License:          "Apache-2.0",
+		Homepage:         "https://pulumi.io",
+		Repository:       "https://github.com/pulumi/pulumi-tls",
+		UpstreamRepoPath: ".",
```
mjeffryes, line 64: Is this a new required field that all bridged providers will need to add?

iwahbe: No. It is an existing field, and it is not required. I wanted some of our test providers to have their docs lookup path in valid go repos, while unaltered providers have their docs lookup path in invalid go repos.

Existing providers will not need to take any action after this PR merges.

pkg/tfgen/generate_schema_test.go
```diff
@@ -41,24 +45,34 @@ import (
...
-	bridgetesting.AssertEqualsJSONFile(t, "test_data/regress-minirandom-schema.json", schema)
+	bridgetesting.AssertEqualsJSONFile(t, expectedFile, schema)
 }
 
 func TestNestedMaxItemsOne(t *testing.T) {
 	provider := testprovider.ProviderMiniCloudflare()
+
+	expectedFile := abs(t, "test_data/regress-minicloudflare-schema.json")
+	enterProviderModule(t,
+		"github.com/terraform-providers/terraform-provider-cloudflare v1.0.0")
```
mjeffryes, line 74: Why do these tests now have to change their working dir to appear as if they are in the provider directory?

pkg/tfgen/source.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (gh *gitRepoSource) getResource(rawname string, info tfbridge.ResourceOrDataSourceInfo) (*DocFile, error) {
+	return gh.getFile(rawname, info, ResourceDocs)
+}
+
+func (gh *gitRepoSource) getDatasource(rawname string, info tfbridge.ResourceOrDataSourceInfo) (*DocFile, error) {
+	return gh.getFile(rawname, info, DataSourceDocs)
+}
+
+func (gh *gitRepoSource) getFile(
```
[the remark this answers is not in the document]
mjeffryes: Since you've gone to the trouble of extracting this to it's own interface, it would be nice to add some units for it.

mjeffryes (commented):
It feels like we're missing a test that clearly shows the new behavior (ie. x used to err, but now it just prints a warning)?

I know this is mostly just moving functions around to different locations, but there's enough small changes that I don't feel qualified to give the give the approving review given that I'm not super familiar with the context yet.

t0yv0:
Can we please have refactors and fixes/behavior changes as separate PRs and try to keep them smaller? Thank you.

pkg/tfgen/source.go
```diff
@@ -0,0 +1,198 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+)
+
+type DocsSource interface {
+	getResource(rawname string, info tfbridge.ResourceOrDataSourceInfo) (*DocFile, error)
```
t0yv0, line 29: Is it not strange that a DocSource needs to know `info`? What is it reading from it? I see it's reading `info.GetDocs()` - is it a good idea for DocsSource's job to be merging overrides from info? Consider if you had Git-based and some other say AI-based DocsSource, both would have to implement overrides.

iwahbe:
Closing in favor of https://github.com/pulumi/pulumi-terraform-bridge/pull/1387 and https://github.com/pulumi/pulumi-terraform-bridge/pull/1388.

Issue #1368, opened 2023-09-15
SIGSEGV in tfbridge.parseCheckError


zhangyoufu:
This bug was triggered during `pulumi import`.

```
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x84797bf]
    goroutine 23 [running]:
    github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge.parseCheckError({0xae3bf28, 0xc0001b56b0}, 0xc001fd3a70?, {0xae07880?, 0xc003148f10})
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.59.0/pkg/tfbridge/adapt_check_failures.go:94 +0x31f
    github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge.(*Provider).adaptCheckFailures(0xc0008a8000, {0x9ec1edc?, 0xd40010?}, {0xc0010dd340, 0x3e}, 0x0?, {0xae3bf28, 0xc0001b56b0}, 0x0?, {0xc003114800, ...})
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.59.0/pkg/tfbridge/adapt_check_failures.go:46 +0x113
    github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge.(*Provider).Check(0xc0008a8000, {0xae353f0?, 0xc003108fc0?}, 0xc000ca7680)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.59.0/pkg/tfbridge/provider.go:599 +0x925
    github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Check_Handler.func1({0xae353f0, 0xc003108fc0}, {0x9417220?, 0xc000ca7680})
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.81.0/proto/go/provider_grpc.pb.go:537 +0x72
    github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc.OpenTracingServerInterceptor.func1({0xae353f0, 0xc003108ea0}, {0x9417220, 0xc000ca7680}, 0xc0031145c0, 0xc000c1ea38)
[...]

PR #1369, opened 2023-09-15
Prevent panic on parseCheckError non-validation error

mikhailshilkov:
Fix https://github.com/pulumi/pulumi-terraform-bridge/issues/1368 by not dereferencing `*diagnostics.ValidationError` when  `errors.As` failed.

I admit I haven't looked at the larger context of the problem, just hoping to avoid the panic in this specific case.

### DIFF
--- a/pkg/tfbridge/adapt_check_failures.go
+++ b/pkg/tfbridge/adapt_check_failures.go
@@ -77,8 +77,7 @@ func parseCheckError(
 		pp := NewCheckFailurePath(schemaMap, schemaInfos, name)
 		return &pp, MiscFailure, err.Error()
 	}
-	var d *diagnostics.ValidationError
-	if errors.As(err, &d) {
+	if d := (*diagnostics.ValidationError)(nil); errors.As(err, &d) {
 		failType := MiscFailure
 		if strings.Contains(d.Summary, "Invalid or unknown key") {
 			failType = InvalidKey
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/adapt_check_failures.go
```diff
@@ -91,7 +91,7 @@ func parseCheckError(
 		return pp, failType, s
 	}
 	// If there is no way to identify a propertyPath, still report a generic CheckFailure.
-	return nil, MiscFailure, d.Error()
+	return nil, MiscFailure, err.Error()
```
iwahbe, line 94: `d` is guaranteed to be `nil` at this point. Could we change the scope of `d` to avoid this kind of mistake?
```
	if d := (*diagnostics.ValidationError)(nil); errors.As(err, &d) {
```

That makes L94 a compile error.

iwahbe (approved):
LGTM!

Nit: We can prevent this class of error by better scoping `d`.

That we had the panic indicates a hole in our testing, so an added test would be great. It's not 100% necessary to merge, since this is obviously a bug.

pkg/tfbridge/adapt_check_failures.go
t0yv0: Gah, err.Error() is the right thing to do here! Good catch!

I can give more context around this if needed beyond the existing comments. Were trying to parse out diagnostics out of TF and ensure they look better in Pulumi here.

t0yv0 (approved):

t0yv0:
Needs unit test ...

mikhailshilkov:
Added a simple unit test

t0yv0:
Awesome 🚢

PR #1372, opened 2023-09-15
Fix #1370 - do not promote warnings to check failures

### LINKED ISSUE #1370: Deprecation warnings promoted to errors in Plugin Framework bridge
### What happened?

When a field is deprecated, PF bridge promotes a WARNING diagnostic from TF to a Check failure causing the Pulumi program to fail. This makes using a deprecated field a breaking change instead of a failure. 

https://github.com/pulumi/pulumi-random/issues/415 is where this got detected

### Example

https://github.com/pulumi/pulumi-random/issues/415 has an example

### Output of `pulumi about`

N/A

### Additional context

Extra printf diagnostics:

```
    DETECTED CHECK FAILURE &{WARNING Attribute Deprecated **NOTE**: This is deprecated, use `numeric` instead. AttributeName("number")}
```

Check trace:

```
{
  "method": "/pulumirpc.ResourceProvider/Check",
  "request": {
    "urn": "urn:pulumi:dev::repro4::random:index/randomString:RandomString::roleName",
    "olds": {
      "length": 10,
      "number": false,
      "special": false,
      "upper": false
    },
    "news": {
      "length": 10,
      "number": false,
      "special": false,
      "upper": false
    },
    "randomSeed": "VH4PJ8AOASPoh8mshTYZPo6RdU29VWv8V2MziXjG4f4="
  },
  "response": {
    "inputs": {
      "length": 10,
      "number": false,
      "special": false,
      "upper": false
    },
    "failures": [
      {
        "reason": "Attribute Deprecated. **NOTE**: This is deprecated, use `numeric` instead.. Examine values at 'roleName.number'."
      }
    ]
  },
  "metadata": {
    "kind": "resource",
    "mode": "client",
    "name": "random"
  }
}

```

### Contributing

t0yv0:
Fixes #1370 

TF Warnings are no longer translated to Pulumi CheckFailures in Check so that the program does not fail-fast.

Curiously these warnings are then dropped.

### DIFF
--- a/pf/tfbridge/detect_check_failures_test.go
+++ b/pf/tfbridge/detect_check_failures_test.go
@@ -1,0 +1,58 @@
+// Copyright 2016-2023, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tfbridge
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+func TestDetectCheckFailures(t *testing.T) {
+	ctx := context.Background()
+	p := &provider{}
+	urn := resource.NewURN("stack", "project", "", "typ", "name")
+
+	t.Run("ignores warnings", func(t *testing.T) {
+		cf := p.detectCheckFailure(ctx, urn, false, nil, nil, &tfprotov6.Diagnostic{
+			Severity:  tfprotov6.DiagnosticSeverityWarning,
+			Attribute: tftypes.NewAttributePath().WithAttributeName("a"),
+		})
+		assert.Nil(t, cf)
+	})
+
+	t.Run("propagates errors", func(t *testing.T) {
+		cf := p.detectCheckFailure(ctx, urn, false, nil, nil, &tfprotov6.Diagnostic{
+			Severity:  tfprotov6.DiagnosticSeverityError,
+			Attribute: tftypes.NewAttributePath().WithAttributeName("a"),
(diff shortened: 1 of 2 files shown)


pf/tfbridge/detect_check_failures_test.go
```diff
@@ -0,0 +1,58 @@
...
+		assert.NotNil(t, cf)
+		assert.Equal(t, "Bad things happening. Examine values at 'name.a'.", cf.Reason)
+	})
+
+	t.Run("ignores errors not specific to an attribute", func(t *testing.T) {
+		cf := p.detectCheckFailure(ctx, urn, false, nil, nil, &tfprotov6.Diagnostic{
+			Severity: tfprotov6.DiagnosticSeverityError,
+			Summary:  "Bad things happening",
+		})
+		assert.Nil(t, cf)
```
iwahbe, line 56: This behavior seems questionable, but it doesn't look new. Do you know why we are doing this?

iwahbe (approved):
I'd like the warnings to be propagated correctly, but I'm ok with that being out of scope. LGTM

pf/tfbridge/detect_check_failures_test.go
t0yv0: When I was writing this I couldn't quite figure out what the right reporting model is.

The CheckFailure structure is especially helpful to associate failures with particular properties (Property) or their paths, this displays nicely in Pulumi CLI with a specific property pointer:

```go
// CheckFailure indicates that a call to check failed; it contains the property and reason for the failure.
type CheckFailure struct {
	Property resource.PropertyKey // the property that failed checking.
	Reason   string               // the reason the property failed to check.
}
```

I think that whatever does not get picked up by detechCheckFailure drops to the generic processor. I guess detectCheckFailure is only interested to filter out specific special cases.

```go
	if err := p.processDiagnostics(remainingDiagnostics); err != nil {
		return nil, err
	}
```

Now that generic processor will short-circuit the entire processing at Go level if there is a DiagnosticSeverityError.

```
func (p *provider) processDiagnostics(diagnostics []*tfprotov6.Diagnostic) error {
	// Format and flush to diagSink.
	if p.diagSink != nil {
		for _, d := range diagnostics {
			p.logDiagnostic(d)
		}
	}

	// Check for errors and return non-nil if there is an error.
	for _, d := range diagnostics {
		if d.Severity == tfprotov6.DiagnosticSeverityError {
			prefix := ""
			if d.Attribute != nil {
				prefix = fmt.Sprintf("[%s] ", d.Attribute.String())
			}
			return fmt.Errorf("%s%s: %s", prefix, d.Summary, d.Detail)
		}
	}

	return nil
}
```

Which will also display in Pulumi. 

We could do a bit better here, it's a little wrong to mix error diagnostics in the underlying and `err` return channel at the bridge level. For example this code would throw only the first error instead of showing all of them to Pulumi. But we need to be careful making changes to ensure things like https://github.com/pulumi/pulumi-terraform-bridge/issues/1370  don't drop errors as well as warnings.

PR #1374, opened 2023-09-15
Update Plugin Framework refs

t0yv0:
IN preparation for an updated PF release, I need to make sure the references here are updated; I also am considering removing replace directives. This has a big upside and a few downsides.

The upside: with this change, CI tests exactly what users get that refer to the pf module. Because it's a library, and Go does not propagate replace directives, we can be in a state that CI checks against some local modifications but users will be broken if they install the new release of this library.

The downside is for us internally. In some small PRs this can get really annoying because it would require staging releases to github.com/pulumi/pulumi-terraform-bridge/v3 and muxer before they can be used here, instead of checking all the code at current commit as-is.

We could consider doing something with go workspaces in CI to additionally verify that the current code works with itself.

### DIFF
--- a/pf/go.mod
+++ b/pf/go.mod
@@ -2,10 +2,6 @@ module github.com/pulumi/pulumi-terraform-bridge/pf
 
 go 1.19
 
-replace github.com/pulumi/pulumi-terraform-bridge/v3 => ./..
-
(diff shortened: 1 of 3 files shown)


t0yv0:
Per conversation with @iwahbe  I'm going to drop replace directive removal till we consider the options carefully.

iwahbe (approved):

PR #1382, opened 2023-09-19
Fix #1381

### LINKED ISSUE #1381: PreCheckCallback cannot modify inputs for Plugin Framework
### What happened?

Looks like due to a simple omission, PreCheckCallback is not allowed to modify inputs for Plugin Framework resources, like it does for normal resources. This functionality is relied on, among other things, for default tags application in pulumi-aws and is not working as expected for PF resources.

### Example

N/A

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

t0yv0:
Fixes #1381  - with this change PreCheckCallback starts working the same way for Plugin Framework resources as normal resources, that is it can return a modify inputs map to be used in place of the original map. This arose in the context of investigating tags secrets leaks for pulumi-aws and in particular how tags/tagsAll behavior is different between normal and PF based resources in the currently released version.

### DIFF
--- a/pf/tfbridge/provider_check.go
+++ b/pf/tfbridge/provider_check.go
@@ -48,7 +48,8 @@ func (p *provider) CheckWithContext(
 
 	if info := rh.pulumiResourceInfo; info != nil {
 		if check := info.PreCheckCallback; check != nil {
-			checkedInputs, err := check(ctx, checkedInputs, p.lastKnownProviderConfig.Copy())
+			var err error
+			checkedInputs, err = check(ctx, checkedInputs, p.lastKnownProviderConfig.Copy())
 			if err != nil {
 				return checkedInputs, []plugin.CheckFailure{}, err
 			}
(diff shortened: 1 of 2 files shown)


pf/tfbridge/provider_check.go
```diff
@@ -48,7 +48,8 @@ func (p *provider) CheckWithContext(
 
 	if info := rh.pulumiResourceInfo; info != nil {
 		if check := info.PreCheckCallback; check != nil {
-			checkedInputs, err := check(ctx, checkedInputs, p.lastKnownProviderConfig.Copy())
+			var err error
+			checkedInputs, err = check(ctx, checkedInputs, p.lastKnownProviderConfig.Copy())
```
t0yv0, line 52: Shadowing considerer harmful.. it got me here again.

iwahbe: I think the pain point is that the difference between a declaration and a mutation is `:`. It somehow bothers me less then its `let ` or `auto ` or `(let ((`.

iwahbe (approved):

PR #1385, opened 2023-09-20
Faster metadata parsing

t0yv0:
Benchmarks here:

https://github.com/pulumi/pulumi-aws/pull/2819

Use json-iterator as a drop-in replacement for the stdlib. Should already be a Pulumi dependency.

It gives a quick speedup for the large AWS use case with very little effort.

Approximate saving is 300ms.

### DIFF
--- a/unstable/metadata/metadata.go
+++ b/unstable/metadata/metadata.go
@@ -16,7 +16,7 @@ package metadata
 
 import (
 	"encoding/json"
-
+	"github.com/json-iterator/go"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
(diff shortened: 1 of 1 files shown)


mjeffryes:
> Hm looks like it's not byte-for-byte compatible but may influence formatting, as evidenced by the test suite. We can still consider going forward. I can try it out on AWS and see the formatting at scale degrades or is still acceptable for text-based diffs.

Since the diffs are just missing new lines, it seems like we must just be missing something small that would get the new library to emit them. Or if we can't convince it to do that and hate not having the new lines, we could even use encoding/json for writing and jsoniter for reading since we don't care about compile time slowness nearly as much.

t0yv0:
Still a speedup with continuing to do json.MasrhalIndent.

```
(cd provider && go test -bench . -benchtime=25s)
goos: darwin
goarch: amd64
pkg: github.com/pulumi/pulumi-aws/provider/v6
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkProviderBare-12                     328         102685357 ns/op
BenchmarkProviderMetadataInfo-12             248         122396172 ns/op
BenchmarkProviderAliases-12                  100         262561475 ns/op
BenchmarkProvider-12                         105         273964600 ns/op
PASS
ok      github.com/pulumi/pulumi-aws/provider/v6        163.577s
```

PR #1387, opened 2023-09-20
Separate docs source from parsing v2

iwahbe:
This PR refactors the code used to acquire the markdown files we source our docs from. The goal is to make it easy to test file acquisition and parsing separately.

The new interface looks like this:
https://github.com/pulumi/pulumi-terraform-bridge/blob/856055a0cd3238780c9c35bbe605282f75cc2de7/pkg/tfgen/source.go#L30-L47

This PR is best reviewed as two separate commits:
- d46360529f7198cf9f3f37df3756e3308726e10e is a byte for byte copy of the functionality that we want to refactor from `docs.go` to the newly created `source.go`.
- 856055a0cd3238780c9c35bbe605282f75cc2de7 implements and consumes the interface. 

This PR does not have *any* effect on downstream users. It is purely a refactor for maintainability.

### DIFF
--- a/pkg/tfgen/source.go
+++ b/pkg/tfgen/source.go
@@ -1,0 +1,151 @@
+// Copyright 2016-2022, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tfgen
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sync"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
+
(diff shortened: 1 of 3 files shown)


pkg/tfgen/source.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+)
+
+// A source of documentation bytes.
+type DocsSource interface {
+	// Get the bytes for a resource with TF token rawname.
+	getResource(
+		rawname string, info tfbridge.ResourceOrDataSourceInfo,
```
t0yv0, line 34: Repeating the comment before, the signature is very permissive, does it really need to know tfbridge.ResourceOrDataSourceInfo and why?

t0yv0 (approved):
Love pure refactoring PRs, easy approve. Directionally this is very good we need to untangle this code here gradually.

mjeffryes (approved):
I'll also echo my comment from the previous version that it would be nice to add some units on the new interfaces, but the refactor is straightforward here and seems like an improvement overall.

pkg/tfgen/source.go
iwahbe: I reduced `info` to `*tfbridge.DocInfo`, which it really does need.

iwahbe:
> I'll also echo my comment from the previous version that it would be nice to add some units on the new interfaces, but the refactor is straightforward here and seems like an improvement overall.

I'll merge unit tests with https://github.com/pulumi/pulumi-terraform-bridge/pull/1388.

PR #1388, opened 2023-09-20
Actionable error message for a missing repo

iwahbe:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1365

The current warning for failing to find the source repo (always due to misconfigured `ProviderInfo`) looks like this (1 line per resource/datasoruce):
```
$ pulumi-acme/bin/pulumi-tfgen-acme schema --out provider/cmd/pulumi-resource-acme
warning: could not find docs for resource 'acme_certificate'. Override the Docs property in the resource mapping. See type tfbridge.DocInfo for details.
warning: could not find docs for resource 'acme_registration'. Override the Docs property in the resource mapping. See type tfbridge.DocInfo for details.
```

This PR changes the error message to look like this:

```
warning: Unable to find the upstream provider's documentation:

The upstream repository is expected to be at "github.com/vancluever/terraform-provider-acme".

If the expected path is not correct, you should check the values of these fields (current values shown):

tfbridge.ProviderInfo{
        GitHubHost:              "github.com",
        GitHubOrg:               "vancluever",
        Name:                    "acme",
        TFProviderModuleVersion: "",
}

The original error is: error running 'go mod download -json github.com/vancluever/terraform-provider-acme' in "/Users/ianwahbe/go/src/github.com/pulumiverse/pulumi-acme/provider" dir for module: exit status 1

Output: {
        "Path": "github.com/vancluever/terraform-provider-acme",
        "Error": "module github.com/vancluever/terraform-provider-acme: not a known dependency"
}
```

This error message will appear once, regardless of how many resources/datasources are in the provider.

---

For review: 
- e568cc4a1f346485ba7234be67321c75d0b629cb creates the new error.
- 1ae3dd68e14a359c272a4068873ee50b4a489d8a catches the new error and displays the warning.

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -75,6 +75,10 @@ type Generator struct {
 	editRules        editRules
 
 	convertedCode map[string][]byte
+
+	// Set if we can't find the docs repo and we have already printed a warning
+	// message.
+	noDocsRepo bool
 }
 
 type Language string
@@ -1207,10 +1211,11 @@ func (g *Generator) gatherResource(rawname string,
 	if !isProvider {
 		source := NewGitRepoDocsSource(g)
 		pd, err := getDocsForResource(g, source, ResourceDocs, rawname, info)
-		if err != nil {
+		if err == nil {
+			entityDocs = pd
+		} else if !g.checkNoDocsError(err) {
 			return nil, err
 		}
-		entityDocs = pd
 	} else {
 		entityDocs.Description = fmt.Sprintf(
 			"The provider type for the %s package. By default, resources use package-wide configuration\n"+
@@ -1400,7 +1405,7 @@ func (g *Generator) gatherDataSource(rawname string,
 	// Collect documentation information for this data source.
 	source := NewGitRepoDocsSource(g)
 	entityDocs, err := getDocsForResource(g, source, DataSourceDocs, rawname, info)
-	if err != nil {
+	if err != nil && !g.checkNoDocsError(err) {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/tfgen/source.go
```diff
@@ -98,22 +93,41 @@ func (gh *gitRepoSource) getFile(
...
 
+	defer func() {
+		if err == nil {
+			return
+		}
+		err = GetRepoPathErr{
+			Expected:   moduleCoordinates,
+			Underlying: err,
+		}
+	}()
```
mjeffryes, line 129: I find this a bit opaque (vs just having a an inner function that returns vanilla errors and then wrapping them), but I assume this is just some sort of go convention, and if that's the case I can live with it.

pkg/tfgen/source.go
```diff
@@ -178,15 +192,24 @@ func getMarkdownNames(packagePrefix, rawName string, globalInfo *tfbridge.DocRul
...
+			return &DocFile{markdownBytes, name}, nil
+		} else if !os.IsNotExist(err) && !errors.Is(err, &os.PathError{}) {
+			// Missing doc files are expected and OK.
+			//
+			// If the file we expect is actually a directory (PathError), that
+			// is also OK.
+			//
+			// Other errors (such as permission errors) indicate a problem
+			// with the host system, and should be reported.
+			return nil, fmt.Errorf("%s: %w", location, err)
```
mjeffryes, line 211: Thanks for the nice comment here :)

pkg/tfgen/generate.go
```diff
@@ -1537,6 +1542,49 @@ func (g *Generator) gatherOverlays() (moduleMap, error) {
...
+If the expected path is not correct, you should check the values of these fields (current values shown):
+tfbridge.ProviderInfo{
+	GitHubHost:              %q,
+	GitHubOrg:               %q,
+	Name:                    %q,
+	TFProviderModuleVersion: %q,
+}`, g.info.GetGitHubHost(), g.info.GetGitHubOrg(), g.info.Name, g.info.GetProviderModuleVersion())
+	}
+
+	g.sink.Warningf(&diag.Diag{Message: msg}, err.Expected, correction, err.Underlying)
```
mjeffryes, line 1585: My only concern is whether this warning will be missed in a pile of other warnings (especially since it's only emitted once). I do think this is a big improvement on the status quo however.

mjeffryes (approved):

pkg/tfgen/generate.go
iwahbe: There shouldn't be that many warnings. Most warnings are docs warnings and if we see this warning, there will not be other docs warnings (since there are no docs).

PR #1390, opened 2023-09-20
Skip running validators in the presence of unknowns

t0yv0:
This is motivated by https://github.com/pulumi/pulumi-aws/issues/2818

The panic is triggered from a simple CheckConfig:

```
{
  "method": "/pulumirpc.ResourceProvider/CheckConfig",
  "request": {
    "urn": "urn:pulumi:dev::aws-2818::pulumi:providers:aws::provider",
    "olds": {},
    "news": {
      "assumeRole": "04da6b54-80e4-46f7-96ec-b56ff0331ba9",
      "region": "us-west-2",
      "skipCredentialsValidation": "false",
      "skipMetadataApiCheck": "true",
      "skipRegionValidation": "true",
      "version": "6.0.2"
    }
  },
  "errors": [
    "rpc error: code = Unavailable desc = error reading from server: EOF"
  ],
  "metadata": {
    "kind": "resource",
    "mode": "client",
    "name": "aws"
  }
}

```

With assumeRole having an unknown, the bridge is not passing the right thing into the validator for a custom field duration. It's passing an "invalid tftypes.Value of unknown type" which subsequently causing a panic.

For starters disabling validators in presence of unknown value bypasses the issue, but this is far from ideal. I'm noticing that string validators are completely fine with unknowns. So I suspect there's some gap in CustomType support at play here.

### DIFF
--- a/pf/tfbridge/provider_checkconfig.go
+++ b/pf/tfbridge/provider_checkconfig.go
@@ -83,9 +83,13 @@ func (p *provider) CheckConfigWithContext(
 	// Store for use in subsequent ApplyDefaultInfoValues.
 	p.lastKnownProviderConfig = news
 
-	checkFailures, err := p.validateProviderConfig(ctx, urn, news)
-	if err != nil {
-		return nil, nil, err
+	var checkFailures []plugin.CheckFailure
+	if !news.ContainsUnknowns() {
+		var err error
+		checkFailures, err = p.validateProviderConfig(ctx, urn, news)
+		if err != nil {
+			return nil, nil, err
+		}
 	}
 
 	// Ensure propreties marked secret in the schema have secret values.
--- a/pf/tests/duration.go
+++ b/pf/tests/duration.go
@@ -1,0 +1,273 @@
+// Obtained from https://raw.githubusercontent.com/hashicorp/terraform-provider-aws/main/internal/framework/types/duration.go
+// Retaining original LICENCE notice below.
+//
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package tfbridgetests
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/attr/xattr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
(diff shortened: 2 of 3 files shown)


t0yv0:
This does get me past the AWS issue but I'm unable to repro full example since aws.organizations.Account does not provision in our test setup.

t0yv0:
@iwahbe for ideas. I'm going to look at this with fresh eyes tomorrow, but we can chose to deploy it this for an AWS quick fix before root-causing the CustomType issue.

pf/tfbridge/provider_checkconfig.go
```diff
@@ -83,9 +83,13 @@ func (p *provider) CheckConfigWithContext(
 	// Store for use in subsequent ApplyDefaultInfoValues.
 	p.lastKnownProviderConfig = news
 
-	checkFailures, err := p.validateProviderConfig(ctx, urn, news)
-	if err != nil {
-		return nil, nil, err
+	var checkFailures []plugin.CheckFailure
+	if !news.ContainsUnknowns() {
+		var err error
+		checkFailures, err = p.validateProviderConfig(ctx, urn, news)
```
iwahbe, line 89: I really wish we were able to take a more precise action here. This gives up validation on required as well as types.

I'll let you know if I think of something else.

iwahbe: Does TF run validators on unknown values?

t0yv0: We still need to fix the bug, this would only be a workaround if we cannot fix the bug fast enough. It seems yes TF runs validators on objects in presence of unknowns, because the validators have clauses testing if a given tftypes.Value is unknown or not; I'll give it a few hours tomorrow to try to see why we are passing an invalid-typed value there.

iwahbe:
`assumeRole` is passed as an `unknown`, but that's not accurate. It should be a `map[string]unknown` in the example. I wonder if we deserialize it into `nil` when it should be `map[string]string{}` or vice versa. I know that the bridge can't actually receive nested unknowns in CheckConfig, but I would be curious to see what happens if we send in `{"assumeRole": {"roleArn": unknown} }`

t0yv0:
That's how config encoding passes it. That's a peculiarity of the engine we need to live with. 

>  I would be curious to see what happens if we send in {"assumeRole": {"roleArn": unknown} }

I don't think that's supported currently.

t0yv0:
Possibly. I'll dive deeper and report.

t0yv0:
The has solved in a different way, closing.

PR #1392, opened 2023-09-21
Add `Generics` option to `GolangInfo`

### LINKED ISSUE #1391: Expose the new `Generics` option in the GolangInfo struct
Pulumi v3.84.0 adds a new Go [schema option for generics](https://github.com/pulumi/pulumi/blob/79271cf543dc6b02baeb560e90c97d52258c7eba/pkg/codegen/go/importer.go#L95-L100):

```go
	// Specifies how to handle generating a variant of the SDK that uses generics.
	// Allowed values are the following:
	// - "none" (default): do not generate a generics variant of the SDK
	// - "side-by-side": generate a side-by-side generics variant of the SDK under the x subdirectory
	// - "only-generics": generate a generics variant of the SDK only
	Generics string `json:"generics,omitempty"`
```

This should be exposed for bridged providers here:

https://github.com/pulumi/pulumi-terraform-bridge/blob/87b8d9e75bcb80fb414ce61fea5dede5c075dc8a/pkg/tfbridge/info.go#L613-L614

justinvp:
This is a new Go schema option in Pulumi v3.84.0 that should be exposed for bridged providers. Commits:
1. Update deps to v3.84.0
2. Expose the `Generics` option

Fixes #1391

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -680,6 +680,13 @@ type GolangInfo struct {
 	// InternalDependencies are blank imports that are emitted in the SDK so that `go mod tidy` does not remove the
 	// associated module dependencies from the SDK's go.mod.
 	InternalDependencies []string
+
+	// Specifies how to handle generating a variant of the SDK that uses generics.
+	// Allowed values are the following:
+	// - "none" (default): do not generate a generics variant of the SDK
+	// - "side-by-side": generate a side-by-side generics variant of the SDK under the x subdirectory
+	// - "only-generics": generate a generics variant of the SDK only
+	Generics string
 }
 
 // CSharpInfo contains optional overlay information for C# code-generation.
--- a/Makefile
+++ b/Makefile
@@ -48,5 +48,6 @@ tidy::
 	cd pf/tests/testdatagen/genrandom && go mod tidy
 	cd pkg/tests && go mod tidy
 	cd x/muxer && go mod tidy
+	cd x/muxer/tests && go mod tidy
 	cd testing && go mod tidy
 	go mod tidy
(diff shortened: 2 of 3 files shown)


t0yv0 (approved):
This LGTM. Is there a link to a list of more tasks around generics rollout?

pkg/tfbridge/info.go
```diff
@@ -672,6 +672,13 @@ type GolangInfo struct {
 	// InternalDependencies are blank imports that are emitted in the SDK so that `go mod tidy` does not remove the
 	// associated module dependencies from the SDK's go.mod.
 	InternalDependencies []string
+
+	// Specifies how to handle generating a variant of the SDK that uses generics.
+	// Allowed values are the following:
+	// - "none": do not generate a generics variant of the SDK
+	// - "side-by-side": generate a side-by-side generics variant of the SDK under the x subdirectory
+	// - "only-generics": generate a generics variant of the SDK only
```
iwahbe, line 680: I would love this comment to explain what `""` is interpreted as. 

Otherwise LGTM

iwahbe (approved):

pkg/tfbridge/info.go
justinvp: Updated the comment. `"none"` is currently the default when not specified. Although, at some point in the near future we intend to flip the default to `"side-by-side"` so all providers get a generic SDK by default.

justinvp:
Update: I addressed the feedback to add a comment on the default value, but also wanted to upgrade to v3.85.0 as there is a Python sdkgen regression in v3.84.0 that is fixed in v3.85.0. But we've tweaked the `Provider` interface in v3.85.0 and need at react to those changes, which I haven't done yet.

t0yv0:
Y'all don't need to take on updating p/p can delegate to providers; I think https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/6261833982/job/17002725064 the automated upgrade job is not doing it right again, we should fix it up.

PR #1394, opened 2023-09-21
Provider init benchmarks

mjeffryes:
Adding some benchmarks so we can observe changes in provider init time in the future. I broke out some metrics for loading a large metadata file as well as some around `MustApplyAutoAliases` based on the significance in our current investigations. I also find it interesting that initializing a Muxed provider is so much slower than initiating a vanilla PF provider. 

Sample before Anton's jsoniter change:
```
❯ go test -run="^$" -bench=. -benchtime=10s
goos: darwin
goarch: arm64
pkg: github.com/pulumi/pulumi-terraform-bridge/pf/tests
BenchmarkRandomProviderInit-12                              			 1000000	     11075 ns/op
BenchmarkMuxedRandomProviderInit-12                         			  209493	     57241 ns/op
BenchmarkSizedMuxedRandomProviderInit/resources_size_100-12 			  165622	     72675 ns/op
BenchmarkSizedMuxedRandomProviderInit/resources_size_10000-12       	  	   10000	   1158911 ns/op
BenchmarkMuxedRandomProviderMustApplyAutoAliases-12                	   	  286514	     42117 ns/op
BenchmarkSizedMuxedRandomProviderMustApplyAutoAliases/resources_size_100-12    	  299841	     40039 ns/op
BenchmarkSizedMuxedRandomProviderMustApplyAutoAliases/resources_size_10000-12 	  321598	     37014 ns/op
BenchmarkLoadMetadata-12                                                 	     178	  66922526 ns/op
```

After jsoniter:
```
❯ go test -run="^$" -bench=. -benchtime=10s
goos: darwin
goarch: arm64
pkg: github.com/pulumi/pulumi-terraform-bridge/pf/tests
BenchmarkRandomProviderInit-12                              			 2512153	      4779 ns/op
BenchmarkMuxedRandomProviderInit-12                         			  293250	     40511 ns/op
BenchmarkSizedMuxedRandomProviderInit/resources_size_100-12 			  218238	     56443 ns/op
BenchmarkSizedMuxedRandomProviderInit/resources_size_10000-12         		   10000	   1177436 ns/op
BenchmarkMuxedRandomProviderMustApplyAutoAliases-12                   		  316081	     38707 ns/op
BenchmarkSizedMuxedRandomProviderMustApplyAutoAliases/resources_size_100-12	  328318	     36714 ns/op
BenchmarkSizedMuxedRandomProviderMustApplyAutoAliases/resources_size_10000-12 	  350518	     34213 ns/op
BenchmarkLoadMetadata-12                                                 	     751	  16085014 ns/op
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,8 @@ jobs:
             **/go.sum
       - name: Build
         run: make build
+      - name: Build PF
+        run: cd pf && make build
       - name: Install golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
--- a/pf/tests/internal/testprovider/aws.go
+++ b/pf/tests/internal/testprovider/aws.go
@@ -1,0 +1,22 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package testprovider
+
+import (
+	_ "embed"
+)
+
+//go:embed cmd/pulumi-resource-aws/bridge-metadata.json
+var BigMetadata []byte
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -33,6 +33,8 @@ jobs:
             **/go.sum
       - name: Build
         run: make build
+      - name: Build PF
+        run: cd pf && make build
```
mjeffryes, line 37: This is not strictly necessary, but I played with pulling more of the aws provider into our tests and it revealed that our lint step could time out trying to load packages for `pf/*` because we don't do a build first.

iwahbe (approved):
The benchmark looks good. 

It doesn't seem to be hooked up to anything, so I'm not sure what the use will be. I generally don't trust us to maintain benchmarks that don't flag when exceeded.

mjeffryes:
Open to hooking this up to monitoring of some form; I mostly just wanted to have it handy for the next time we have to dig here. (It sounds like we'll be coming back to this in M95 at least)

PR #1398, opened 2023-09-25
Normalize first class support for nested arguments

iwahbe:
This PR changes how nested arguments are represented to reduce duplication and allow modeling richer nesting.

Previously, parsed TF arguments were stored in as:

```go
type entityDocs struct {
	...
	Arguments map[string]*argumentDocs
        ...
}

type argumentDocs struct {
        // The description for this argument.
	description string

	// (Optional) The names and descriptions for each argument of this argument.
	arguments map[string]string

	// Whether this argument was derived from a nested object. Used to determine
	// whether to append descriptions that have continued to the following line
        isNested bool
}
```

As a basic example, lets parse the following:

```md
* `top1` - First top level argument
* `top2` - Second top level argument
  * `nest1` - First nested argument
```

We get:

```go
Arguments: map[string]*argumentDocs{
	"top1": {
        	description: "First top level argument",
                arguments: map[string]string{},
                isNested: false,
        },
        "top2": {
        	description: "Second top level argument"
                arguments: map[string]string{
                	"nest1": "First nested argument"
		},
                isNested: false,
        },
        "nest1": {
        	description: "First nested argument",
                arguments: map[string]string{},
                isNested: true,
	},
}
```

This provided a clear parse when arguments were not nested. When arguments are nested, there is duplication tracked by the `isNested` bit.

Nesting with `depth > 2` is possible, but never tracked correctly in the code base. It looks like this:

```md
* `t1` - D1
  * `t2` - D2
    * `t3` - D3
```

Parsing the above (in theory) gives:

```go
Arguments: map[string]*argumentDocs{
	"t1": {
        	description: "D1",
                arguments: map[string]string{
	        	"t2": "D2",
		},
                isNested: false,
        },
        "t2": {
        	description: "D2",
                arguments: map[string]string{
	        	"t3": "D3",
		},
                isNested: true,
        },
        "t3": {
        	description: "D3",
                arguments: map[string]string{},
                isNested: true,
        },
}
```

Observe that this scheme immediately falls down when there are "any" fields with the same name.

When we parse this example (adapted from the AWS), we get a incorrect result:

```md
* `b1` - Some field
  * `options` - Options for `b1`
    * `type` - One of "A" or "B"
* `b2` - Another field
  * `options` - Options for `b2`
    * `type` - One of "C" or "D"
```

```go
Arguments: map[string]*argumentDocs{
	"b1": {
           	description: "Some field",
        	arguments: map[string]string{
			"options": "Options for `b1`",
                },
                isNested: false,
        },
        "options": {
           	description: "Options for `b1`",
        	arguments: map[string]string{
			"type": "One of "A" or "B"",
                },
                isNested: true,
        },
        "type": {
           	description: "One of "A" or "B"",
        	arguments: map[string]string{},
                isNested: true,
        },
        "b2": {
           	description: "Some field",
        	arguments: map[string]string{
			"options": "Options for `b2`",
                },
                isNested: false,
        },
}
```

Notice that the description for `b2.options.type` is unrepresentatable in the current schema. I have replaced this schema with a single map from path to value. This gives us a complete (sparse) representation without loosing information.

This makes it much simpler to work with arguments. It also makes it much easier to iterate on parsing, since there is an unambiguous correct parse for a given document.


iwahbe:
https://github.com/pulumi/pulumi-cloudflare/pull/493 is an example PR.

pkg/tfgen/docs.go
```diff
@@ -47,32 +47,34 @@ import (
...
+
+func (d docsPath) nested() bool                 { return strings.Contains(string(d), ".") }
+func (d docsPath) leaf() string                 { s := d.parts(); return s[len(s)-1] }
+func (d docsPath) join(segment string) docsPath { return docsPath(string(d) + "." + segment) }
+func (d docsPath) parts() []string              { return strings.Split(string(d), ".") }
+func (d docsPath) withOutRoot() docsPath        { s := d.parts(); return docsPath(strings.Join(s[1:], ".")) }
+
+type docsPathArr []docsPath
+
+var _ = (sort.Interface)((docsPathArr)(nil))
```
mjeffryes, line 71: I'm guessing this is somehow declaring that docsPathArr implements sort.Interface, but I don't quite follow how it works.

mjeffryes (commented):
This certainly feels like it's more straightforward. I don't have enough context here to give a great review, but I did review a good chunk of the schema changes in the example PR. Overall, this does look like it adds a bunch of docs that were missing before and correct docs in some places. However, there are some places where it looks like it might have regressed too.

pkg/tfgen/docs.go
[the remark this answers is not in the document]
iwahbe: This is assigning the special discard variable `_` to nil, where `nil` is created by casting a nil pointer of type `docsPathArr` to `sort.Interface`.

It is the common go idiom for asserting that `docsPathArr` conforms with the interface `sort.Interface`.

pkg/tfgen/docs.go
```diff
@@ -299,13 +296,66 @@ func getDocsForResource(g *Generator, source DocsSource, kind DocKind,
...
+		a.docs = v
+	}
+	return topLevel.children
+}
+
+func (t expandedArguments) collapse() map[docsPath]*argumentDocs {
+	m := make(map[docsPath]*argumentDocs, len(t))
+
+	var set func(docsPath, expandedArguments)
+	set = func(path docsPath, args expandedArguments) {
```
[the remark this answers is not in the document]
iwahbe: `set` is declared and then used because the implementation of `set` calls `set`. By declaring before use, the body of `set` can refer to the `set` function, allowing recursion.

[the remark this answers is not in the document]
mjeffryes: Thanks; very slowly wrapping my head around go scopes.

pkg/tfgen/docs.go
```diff
@@ -47,32 +47,34 @@ import (
...
-		Arguments:   ad.arguments,
-		IsNested:    ad.isNested,
 	})
-	if err != nil {
-		return nil, err
-	}
-	return j, nil
 }
 
+type docsPath string
```
t0yv0, line 61: Interesting choice, so presumably "." is not allowed in fragments; but this is not []string because using string repr buys automatic hashing/equality/printing in Go?

Please document what the fragments of this path are. Are they TF property names like "foo_bar"? What happens for a schema like this:

```
type X {
    foo []Y
}

type Y {
   bar string
}
```

What is the path of bar?

pkg/tfgen/docs_test.go
```diff
@@ -1010,53 +965,21 @@ This is a test for CUSTOM_REPLACES.`)
...
 			}
 
-			c.providerInfo.DocRules = &tfbridge.DocRuleInfo{
+			tc.providerInfo.DocRules = &tfbridge.DocRuleInfo{
 				EditRules: func(defaults []tfbridge.DocsEdit) []tfbridge.DocsEdit {
 					return append([]tfbridge.DocsEdit{rule}, defaults...)
 				},
 			}
 		}),
-
```
t0yv0, line 1019: This tc was redundant?

pkg/tfgen/parse_markdown_test.go
```diff
@@ -2,15 +2,21 @@ package tfgen
...
+	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 
 	bf "github.com/russross/blackfriday/v2"
 	"github.com/stretchr/testify/assert"
 )
 
+var pulumiAccept = func() bool {
```
t0yv0, line 15: Why is this not `func pulumiAccept() bool`?

t0yv0 (approved):
I think this is a great simplification! Cursorily this LGTM, can I request that we try it out on one of the larger providers perhaps GCP and observe a no-op diff over the schema or else satisfy ourselves that the schema diff is what we want. Then we can ship.

These steps are necessary since proper downstream checks we want installed here are not yet in place.

iwahbe:
> I think this is a great simplification! Cursorily this LGTM, can I request that we try it out on one of the larger providers perhaps GCP and observe a no-op diff over the schema or else satisfy ourselves that the schema diff is what we want. Then we can ship.
> 
> These steps are necessary since proper downstream checks we want installed here are not yet in place.

I expect that there will be nontrivial changes to gcp. There were to [cloudflare](https://github.com/pulumi/pulumi-cloudflare/pull/493) and to aws (tested locally). The changes were on net positive. I can do a through check on gcp before merging as a final validation before merging.

pkg/tfgen/parse_markdown_test.go
iwahbe: `pulumiAccept` is a `bool`, not a `func() bool`. Notice the invocation at the end.

pkg/tfgen/docs.go
iwahbe: I'll add some comments on what to expect.

`"."` is not allowed in fragments. `docsPath` segments are TF names, as written in TF docs. The path of `bar` (assuming that `X` is the top level resource) would be `foo.bar`. TF doesn't alter path names for collections (as opposed to objects).


Part of the reason is that representing as `string` gets us immutability, hashing and comparison for free. Another part is that sometimes TF docs will document a block as:
```
### `foo`
- `bar` (Optional String) a field of `foo`
```

Other providers will document the same field as:
```
### `x.foo`
- `bar` (Optional String) a field of `foo`
```

This representation lets our "detect enclosing objects from title" logic work for free.

pkg/tfgen/docs_test.go
iwahbe: This test case was kept on L954:
```
		test("azurerm-sql-firewall-rule"),
```

I have moved the expected block to `test_data/azurerm-sql-firewall-rule/expected.json`.

t0yv0:
> . There were to https://github.com/pulumi/pulumi-cloudflare/pull/493 and to aws (tested locally).

Ah so you went through a round of AWS testing. It's super helpful in situations like this to include a summary of what you found there. In that case due diligence has been done, I think we should ship this.

pkg/tfgen/docs.go
t0yv0: The fact that it's "foo.bar" instead of "foo.$.bar" or some such is surprising, worth calling out, but it may be the right choice. The rest makes sense!

PR #1400, opened 2023-09-26
Add ResourceInfo.TransformFromState

iwahbe:
This provides a makeshift migration mechanism for TF based providers. `PreCheckCallback` is run before an input is baked into the state, but sometimes we need to make changes after a value is read from the state: when we didn't have the `PreCheckCallback` in place before hand.

- [X] SDK
- [X] PF
- [ ] Tests

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -329,8 +329,16 @@ type ResourceInfo struct {
 	// outputs before they are stored. In particular, it can be used as a last resort hook to
 	// make corrections in the default translation of the resource state from TF to Pulumi.
 	// Should be used sparingly.
-	TransformOutputs func(ctx context.Context, outputs resource.PropertyMap) (resource.PropertyMap, error)
+	TransformOutputs PropertyTransform
+
+	// Check, Diff, Read, Update and Delete refer to old inputs sourced from the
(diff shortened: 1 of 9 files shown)


pkg/tfbridge/info.go
```diff
@@ -329,9 +329,17 @@ type ResourceInfo struct {
...
 	// make corrections in the default translation of the resource state from TF to Pulumi.
 	// Should be used sparingly.
-	TransformOutputs func(ctx context.Context, outputs resource.PropertyMap) (resource.PropertyMap, error)
+	TransformOutputs PropertyTransform
+
+	// Check, Diff, Read, Update and Delete refer to old inputs sourced from the
+	// Pulumi statefile. TransformFromState lets providers edit these outputs before they
+	// are accessed by other provider functions or by terraform. In particular, it can
+	// be used to perform upgrades on old pulumi state.  Should be used sparingly.
+	TransformFromState PropertyTransform
```
t0yv0, line 338: This pairs well with TransformOutputs as the two together work toward allowing a provider to biject between state representations that the engine sees and saves into state files and the ones that the provider sees and feeds to TF.

t0yv0 (approved):
This LGTM.

Issue #1352, opened 2023-09-01
Decouple example conversion to avoid linking converters into providers


t0yv0:
There were some concerns that due to the module structure, landing 
https://github.com/pulumi/pulumi-terraform-bridge/pull/677 will expose all bridged providers to pulumi/terraform dependency. We need to explore options that make providers and tfgen not build-depend on the converter code, this should be in principle doable since converter code is only needed to render Pulumi Package Schema with pre-rendered TS, Py, C#, etc examples.

Performance is a concern, that is building large providers like pulumi-aws should not degrade.

I will be exploring something along the following lines:

```

   in tfgen:

   func Convert(example, convertConfig) string:
        uuid = newuud()
        db.record(json.Marshal(x{uuid, example, convertConfig})
        return uuid

   this generates a schema with uuids

   then as last step of tfgen we do
        db.Write("examples.json")
        tf2pulumi --bulk examples.json --out renderedExamples.json
        r := read("renderedExamplesJson")
        schema := replaceUuds(rawSchema, db, r)
        patchExampleMetrics(r)
```

Instead of tf2pulumi binary we could maybe interface with `pulumi convert` directly and have it manage that; we just need a batched high-throughput way to call `pulumi covert`.

If this succeeds this also opens the door of moving tf2pulumi out of this repo to a dedicated one.

t0yv0:
Referencing @Frassle work: https://github.com/pulumi/pulumi-converter-terraform/pull/29

The idea is to have the bridge call into `pulumi convert` for the initial stage of the conversion, and have that use this new bulk conversion protocol.

PR #1401, opened 2023-09-26
Allow tfgen to use pulumi convert for examples conversion

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1352

Requires a modified  pulumi/pulumi-converter-terraform#29 

A new option is added for tfgen, via the environment variable PULUMI_CONVERT. IF this option is truthy, `make tfgen` will now be using `pulumi convert` when converting examples, specifically for the TF HCL to Pulumi PCL phase of the conversion.

How this was tested. Tried on these providers:

- pulumi-wavefront
- pulumi-gcp

Ensured conversion rates improve. Ensured COVERAGE_OUTPUT_DIR, if set, continues to populate sensible conversion metrics.  Ensured printed warnings and example conversion metrics out of `make tfgen` are still sensible.

When rolling this out to providers, recommend making sure that Pulumi CLI in use is pinned first, as the repositories are sensitive to small changes in the generated examples and a floating Pulumi CLI reference can introduce breaking CI checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,9 @@ This repo on its own isn't particularly interesting, until it is used to create 
 
 We use git tags and [GitHub Releases](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository).
 1. Maintainers will push a new semver [tag](https://github.com/pulumi/pulumi-terraform-bridge/tags) when appropriate
-1. Maintainers will then generate a Release with Changelog using [GitHub Releases](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository), using the tag pushed in the 
+1. Maintainers will then generate a Release with Changelog using [GitHub Releases](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository), using the tag pushed in the
 first step.
-1. Finally, maintainers will roll out bridge updates across Pulumi providers via manually running the [bridge update 
+1. Finally, maintainers will roll out bridge updates across Pulumi providers via manually running the [bridge update
 workflow in CI Management](https://github.com/pulumi/ci-mgmt/actions/workflows/update-bridge-ecosystem-providers.yml).
 
 ### Adapting a New Terraform Provider
@@ -106,3 +106,4 @@ tfgen, the command that generates Pulumi schema/code for a bridged provider supp
 * `PULUMI_SKIP_MISSING_MAPPING_ERROR`: If truthy, tfgen will not fail if a data source or resource in the TF provider is not mapped to the Pulumi provider. Instead, a warning is printed. Default is `false`.
 * `PULUMI_SKIP_EXTRA_MAPPING_ERROR`: If truthy, tfgen will not fail if a mapped data source or resource does not exist in the TF provider. Instead, warning is printed. Default is `false`.
 * `PULUMI_MISSING_DOCS_ERROR`: If truthy, tfgen will fail if docs cannot be found for a data source or resource. Default is `false`.
+* `PULUMI_CONVERT`: If truthy, tfgen will shell out to `pulumi convert` for converting example code from TF HCL to Pulumi PCL
--- a/pkg/tfgen/convert_cli.go
+++ b/pkg/tfgen/convert_cli.go
@@ -340,7 +340,9 @@ func (*cliConverter) convertViaPulumiCLI(
 
 	outputFileBytes, err := os.ReadFile(outputFile)
 	if err != nil {
-		return nil, fmt.Errorf("convertViaPulumiCLI: failed to read output file: %w", err)
+		return nil, fmt.Errorf("convertViaPulumiCLI: failed to read output file: %w, "+
+			"check if your Pulumi CLI version is recent enough to include pulumi-converter-terraform v1.0.9",
+			err)
 	}
 
 	var result map[string]translatedExample
--- a/pkg/tfgen/convert_cli.go
+++ b/pkg/tfgen/convert_cli.go
@@ -363,8 +363,7 @@ func (cc *cliConverter) convertPCL(
 		return "", diagnostics, nil
 	}
 
-	opts := cc.opts
-	if opts == nil {
+	if cc.opts == nil {
 		var opts []pcl.BindOption
 		opts = append(opts, pcl.AllowMissingProperties)
 		opts = append(opts, pcl.AllowMissingVariables)
@@ -372,7 +371,6 @@ func (cc *cliConverter) convertPCL(
 		if cc.pluginHost != nil {
 			opts = append(opts, pcl.PluginHost(cc.pluginHost))
 			loader := newLoader(cc.pluginHost)
-			//loader.BindCurrentPackage(*spec, cc.packageName)
 			opts = append(opts, pcl.Loader(loader))
 		}
 		if cc.packageCache != nil {
@@ -381,7 +379,7 @@ func (cc *cliConverter) convertPCL(
 		cc.opts = opts
 	}
 
-	program, programDiags, err := pcl.BindProgram(pulumiParser.Files, opts...)
+	program, programDiags, err := pcl.BindProgram(pulumiParser.Files, cc.opts...)
 	if err != nil {
 		return "", diagnostics, fmt.Errorf("pcl.BindProgram failed: %w", err)
 	}
--- a/pkg/tfgen/convert_cli.go
+++ b/pkg/tfgen/convert_cli.go
@@ -97,12 +97,10 @@ func (g *Generator) cliConverter() *cliConverter {
 		return g.cliConverterState
 	}
 	packageName := string(g.pkg)
-	infoCopy := g.info
-	infoCopy.Name = packageName // correct "google-beta" to "gcp"
 	g.cliConverterState = &cliConverter{
 		generator:    g,
 		hcls:         map[string]struct{}{},
-		info:         infoCopy,
+		info:         g.info,
 		packageCache: g.packageCache,
 		packageName:  packageName,
 		pluginHost:   g.pluginHost,
@@ -194,7 +192,15 @@ func (cc *cliConverter) bulkConvert() error {
 		examples[fileName] = hcl
 		n++
 	}
-	result, err := cc.convertViaPulumiCLI(examples, []tfbridge.ProviderInfo{cc.info})
+	result, err := cc.convertViaPulumiCLI(examples, []struct {
+		name string
+		info tfbridge.ProviderInfo
+	}{
+		{
+			name: cc.packageName,
+			info: cc.info,
+		},
+	})
 	if err != nil {
 		return err
 	}
@@ -220,7 +226,10 @@ func (cc *cliConverter) bulkConvert() error {
 // memory, for example run 4 instances of `pulumi convert` on 25% of examples each.
 func (*cliConverter) convertViaPulumiCLI(
 	examples map[string]string,
-	mappings []tfbridge.ProviderInfo,
+	mappings []struct {
+		name string
+		info tfbridge.ProviderInfo
+	},
 ) (
 	output map[string]translatedExample,
 	finalError error,
@@ -266,8 +275,8 @@ func (*cliConverter) convertViaPulumiCLI(
 
 	mappingsDir := filepath.Join(outDir, "mappings")
 
-	mappingsFile := func(i int, p tfbridge.ProviderInfo) string {
-		return filepath.Join(mappingsDir, fmt.Sprintf("%s.json", p.Name))
+	mappingsFile := func(name string) string {
+		return filepath.Join(mappingsDir, fmt.Sprintf("%s.json", name))
 	}
 
 	// Write out mappings files if necessary.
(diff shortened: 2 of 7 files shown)


t0yv0:
@Frassle @iwahbe @mjeffryes @mikhailshilkov 

This is rough around the edges but it's promising. I've got it to render pulumi-gcp example set through the new converter engaged out of process via `pulumi convert`. It seems viable.

`pulumi-converter-terraform` spins CPU for about 3min, but is fairly low on memory utilization (<50mb).

We can parallelize it to stand up N*CPU copies.

Before parallelization as things stand the total timings for schema generation with this on GCP are:

```
      187.20 real       254.00 user        15.11 sys
```

That's not bad we can take that.  The appeal of the new converter is high.

t0yv0:
The annoying bits: 

- I hit "limit reached on GitHub API" because something was hitting up GitHub.com/pulumi/pulumi-gcp to find latest release too often; I've modified the loader to try to recognize that it already has the partial schema for GCP in-place

- The GCP provider names itself "google-beta" but there's also "google" and "gcp" and it's really finicky which one is the provider identifier; when talking to `pulumi-convert` apparently mapping files sent via --mappings have to be called by convention "gcp.json" but this is not obvious should it be "gcp.json" or "google.json" or..

- It's extremely annoying that the code retains the package loader to perform pcl binding; this loader has to match what the underlying loader in pulumi convert is going to do; definitely for the current provider such as gcp, but really also for any dynamically loaded providers that show up in examples like "std"; there's no consistent pinning

- by virtue of above, we also still build-depend on every language's codegen here

t0yv0:
On the positive side, the code shouldn't panic so we don't need to inject recover and lose track panics. I will hook up example metrics to get them right (that's on the TODO).  Perhaps could also push through a bulk examples.json interface for all the other converters (ts,py,yaml,java,cs)? Or is that too ambitious?

pkg/tfgen/convert_cli.go
```diff
@@ -0,0 +1,427 @@
...
+	for n := range generatedFiles {
+		key = n
+	}
+	return string(generatedFiles[key]), diagnostics
+}
+
+// Encodes an example conversion request into a special form that needs to survive being encoded
+// into JSON and plain text and recovered back later on.
+func (cc *cliConverter) encodeConversionJob(hcl string, languageName string) string {
+	return fmt.Sprintf("{example:%s:%s}", languageName,
```
t0yv0, line 382: This is a creative solution that avoids refactoring too much code to achieve a batched/bulk interface with pulumi convert. The examples instead of getting rendered get spliced as is. This splice travels all the way into the generated PackageSpec JSON. A subsequent pass regexes it out and actually performs bulk conversion.

pkg/tfgen/convert_cli.go
```diff
@@ -0,0 +1,427 @@
...
+	cmd := exec.Command(pulumiPath, append([]string{"convert",
+		"--from", "terraform",
+		"--language", "pcl",
+		"--out", outDir,
+		"--generate-only",
+	}, mappingsArgs...)...)
+
+	cmd.Dir = filepath.Dir(examplesJSON.Name())
+	cmd.Env = append(os.Environ(),
+		"PULUMI_CONVERT_EXAMPLES="+filepath.Base(examplesJSON.Name()))
```
t0yv0, line 179: Had to use an ENV var since the --convert-examples arg actually fails to parse.

Frassle: You we're probably missing a double dash:
`pulumi convert --from terraform --lanugage pal -- --convert-examples examples.json`
Note the -- before --convert-examples.

t0yv0: I've tried that and it didn't work. I can try again.

t0yv0: Nope @Frassle that makes the test fail. There's no error from pulumi but it doesn't generate the output file.

t0yv0:
limit reached on GitHub API is resolved; was a bug in this code. A few more low level issues are resolved. This is working better.

pkg/tfgen/convert_cli.go
```diff
@@ -0,0 +1,453 @@
...
+
+	// Write out mappings files if necessary.
+	for i, m := range mappings {
+		if i == 0 {
+			if err := os.MkdirAll(mappingsDir, 0755); err != nil {
+				return nil, fmt.Errorf("convertViaPulumiCLI: failed to write "+
+					"mappings folder: %w", err)
+			}
+		}
+		copy := m
```
Frassle, line 281: Don't think you need `copy` here, just pass m directly in the line below.

pkg/tfgen/convert_cli.go
```diff
@@ -0,0 +1,453 @@
...
+			"to JSON: %w", err)
+	}
+	if err := os.WriteFile(examplesJSON.Name(), examplesBytes, 0600); err != nil {
+		return nil, fmt.Errorf("convertViaPulumiCLI: failed to write a temp "+
+			"bridge-examples.json file: %w", err)
+	}
+
+	mappingsDir := filepath.Join(outDir, "mappings")
+
+	mappingsFile := func(i int, p tfbridge.ProviderInfo) string {
```
Frassle, line 269: `i` isn't used?

pkg/tfgen/convert_cli.go
```diff
@@ -0,0 +1,453 @@
...
+	var mappingsArgs []string
+	for i, m := range mappings {
+		mappingsArgs = append(mappingsArgs, "--mappings", mappingsFile(i, m))
+	}
+
+	cmd := exec.Command(pulumiPath, append([]string{"convert",
+		"--from", "terraform",
+		"--language", "pcl",
+		"--out", outDir,
+		"--generate-only",
```
Frassle, line 310: `, "--", "--convert-examples", examplesJSON.Name()` should work here. What parse failure are you seeing when you try that?

pkg/tfgen/convert_cli.go
```diff
@@ -0,0 +1,453 @@
...
+}
+
+// Get or create the cliConverter associated with the Generator.
+func (g *Generator) cliConverter() *cliConverter {
+	if g.cliConverterState != nil {
+		return g.cliConverterState
+	}
+	packageName := string(g.pkg)
+	infoCopy := g.info
+	infoCopy.Name = packageName // correct "google-beta" to "gcp"
```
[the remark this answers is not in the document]
iwahbe: `Provider.Name` is the TF provider name. It should be `"google-beta"`.

pkg/tfgen/convert_cli.go
t0yv0: It doesn't work. I've tried. `pulumi convert` does not generate the output file.

pkg/tfgen/convert_cli.go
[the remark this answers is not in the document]
t0yv0: Well I seem to need that; without it the examples are all incorrect as they render as if the provider is unknown. I'll check again. I think this may have more to do with how the mappings files are named, apparently when passing `--mappings` to `pulumi convert`, the filename is significant and has to identify the provider/package. I might be conflating the two variables here by using info.Name for the filename. Checking up on this.

pkg/tfgen/convert_cli.go
Frassle: OK lemme check this out, I'm really not keen on us throwing env vars everywhere and this _should_ work.

Frassle: Yeh turns out it helps if you actually try to tell the flags library to parse the arguments 🤦 
I'll get this fixed up

pkg/tfgen/convert_cli.go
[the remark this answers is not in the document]
t0yv0: OK I factored out how to compute --mappings names and it's fixed up now.

pkg/tfgen/convert_cli.go
t0yv0: Works now. thank you.

pkg/tfgen/convert_cli.go
```diff
@@ -0,0 +1,467 @@
...
+// Calls pulumi convert to bulk-convert examples.
+//
+// To facilitate high-throughput conversion, an `examples.json` protocol is employed to convert
+// examples in batches. See pulumi/pulumi-converter-terraform#29 for where the support is
+// introduced.
+//
+// Source examples are passed in as a map from ID to raw TF code.
+//
+// This may need to be coarse-grain parallelized to speed up larger providers at the cost of more
+// memory, for example run 4 instances of `pulumi convert` on 25% of examples each.
```
iwahbe, line 226: I would prefer if this kind of speedup could happen in the engine, where we just pass the number of concurrent processes to run at once. It is safe to parallelize within process because there is no shared mutable state.[^1]

[^1]: If there is shared mutable state, then parallelizing by process splitting will be non-deterministic.

pkg/tfgen/convert_cli.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+
+	// Write out mappings files if necessary.
+	for i, m := range mappings {
+		if i == 0 {
+			if err := os.MkdirAll(mappingsDir, 0755); err != nil {
+				return nil, fmt.Errorf("convertViaPulumiCLI: failed to write "+
+					"mappings folder: %w", err)
+			}
+		}
```
iwahbe, line 289: I would raise this out of the loop:

```go
if len(mappings)>0 { ...
```

I think it makes the code read easier.

pkg/tfgen/convert_cli_test.go
```diff
@@ -0,0 +1,178 @@
...
+	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
+)
+
+func TestConvertViaPulumiCLI(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		// Currently there is a test issue in CI/test setup:
+		//
+		// convertViaPulumiCLI: failed to clean up temp bridge-examples.json file: The
```
iwahbe, line 43: Have we manually tested running the new convert workflow on windows? We do support authoring bridged providers on windows AFAIK.

iwahbe:
I would really like a plan for how this change will roll out.

README.md
```diff
@@ -106,3 +106,4 @@ tfgen, the command that generates Pulumi schema/code for a bridged provider supp
 * `PULUMI_SKIP_MISSING_MAPPING_ERROR`: If truthy, tfgen will not fail if a data source or resource in the TF provider is not mapped to the Pulumi provider. Instead, a warning is printed. Default is `false`.
 * `PULUMI_SKIP_EXTRA_MAPPING_ERROR`: If truthy, tfgen will not fail if a mapped data source or resource does not exist in the TF provider. Instead, warning is printed. Default is `false`.
 * `PULUMI_MISSING_DOCS_ERROR`: If truthy, tfgen will fail if docs cannot be found for a data source or resource. Default is `false`.
+* `PULUMI_CONVERT`: If truthy, tfgen will shell out to `pulumi convert` for converting example code from TF HCL to Pulumi PCL
```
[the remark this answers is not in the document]
t0yv0: Yes. The idea is to roll this out to Tier 1 providers as soon as we can. In terms of retiring this flag, we'd need to make sure our entire fleet has migrated (including Tier 2-3) and then stakeholders and community have migrated, then we can retire it and make it the default.

pkg/tfgen/convert_cli.go
t0yv0: >  It is safe to parallelize within process because there is no shared mutable state.

Famous last words. Prove there's no shared mutable state in large go systems (think static `var`).

t0yv0:
The plan is included in the epic https://github.com/pulumi/pulumi-terraform-bridge/issues/1280

pkg/tfgen/convert_cli_test.go
t0yv0: Nope, we haven't. It's not on the critical path. I can add it to the bottom of the epic.

pkg/tfgen/convert_cli.go
iwahbe: Hence the footnote: It will be either safe to parallelize within a single process or it isn't, in which case I'll be scared to parallelize across processes because each update shares mutable state.

pkg/tfgen/convert_cli_test.go
iwahbe: It's necessary for removing the old convert logic, but shouldn't be blocking for us to use the new logic. We might want to add a warning for users who turn on the new converter on windows computers.

pkg/tfgen/convert_cli.go
t0yv0: At a first approximation, yes, but it's not apples to apples. 

If you're parallelizing in-process, you can get issues related to say panics on concurrent map element access for shared state.

If you're parallelizing outside of process, you can get issues related to using shared OS resources such as file handles or folders.

It's just a guess at this point, I don't know if any of those are actual issues. We could try and see at some point.

PR #1402, opened 2023-09-27
Dependencies: pulumi v3.86.0

t0yv0:
Updating the Pulumi PKG dependency to v3.86.0 

Since there are breaking changes, borrowing @Frassle work from https://github.com/pulumi/pulumi-terraform-bridge/pull/1392 to accommodate.

Doing some test maintenance on top.

If we can land this it should simplify rebasing 1392 for @justinvp so we can land that as well.


iwahbe (approved):
LGTM

I'm inferring that `GetMappings` returns a list of mappings that the provider can provide, and `GetMapping` returns the requested mapping, erroring if the requested mapping is not present. For legacy support, we accept "" to key to the one mapping that we support.

t0yv0:
This seems to be the spec. I'll merge @Frassle @justinvp pipe in if any concerns, we can correct before releasing a new bridge version.

Frassle:
> For legacy support, we accept "" to key to the one mapping that we support.

One or default mapping. Like if kubernetes is extended to support helm as well it would only be able to for new engines, old engines it would just return the default mapping of "kubernetes". I don't think any bridge providers map two Terraform providers though.

Issue #1358, opened 2023-09-06
Documentation Issue in Python SDK NetworkArgs


sfenzke:
### What happened?

The docstring for the mode states " :param pulumi.Input[str] mode: One of:" but doesn't list the valid values

### Expected Behavior

docstring contains a list of valid values

### Steps to reproduce

Read the docs

### Output of `pulumi about`

not relevant

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mikhailshilkov:
Thank you @sfenzke for highlighting this. I'll add it to the list of know doc issues.

For reference:
Link to our docs: https://www.pulumi.com/registry/packages/libvirt/api-docs/network/#mode_nodejs
Link to upstream docs: https://registry.terraform.io/providers/dmacvicar/libvirt/latest/docs/resources/network

guineveresaenger:
This is a bug related to docs discovery parsing. 
Because there's a bulleted list of values, the bridge interprets these as properties, not potential values.
The fix here is, unfortunately, a patch to the upstream docs. We don't to daye have a good way of differentiating a bulleted list of subproperties from a bulleted list of possible values for a property.

guineveresaenger:
This kind of issue should, in my opinion, at this point get fixed at the provider level, for two main reasons:

- We have fixed a large subcategory of these issues in #1507 so these types of issues are hopefully few and far between
- Adding more regex to attempt to discover value lists of this format is brittle unless we know the standard ways in which they present (and at this point in time, there is not really such a standard).

[Read the PR description for pulumi-libvirt for details](https://github.com/pulumi/pulumi-libvirt/pull/334)

I'm happy to file a followup and/or move this issue back to pulumi-libvirt.

PR #1408, opened 2023-09-28
Structured argument parsing

iwahbe:
This is part of the fix for #1358.

2620ecfe2b876116cb1776ce2b436a244ad246e6 adds the new tests on the old parser. This lets 00a1cdea96ccae2bf909f3d556b17bb9987f5f2d show the change in the parse. 

#1358 needs `mode.*` fields to be parsed as associated with `mode`. Currently, each `mode` option is parsed as a sub-field. A future PR will roll unexpected arguments into their parents (actually fixing #1358), but that requires the correct that this PR gives.

---

The effects of this PR can be observed from this test run: https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/6910135157.

Note: this run happened before https://github.com/pulumi/pulumi-terraform-bridge/pull/1408/commits/3583b12595bcc3e0e871ad10f69c1d8cebb1088a

- https://github.com/pulumi/pulumi-aiven/pull/436: No change
- https://github.com/pulumi/pulumi-akamai/pull/334: No change
- https://github.com/pulumi/pulumi-alicloud/pull/503: Most changes are improvements. 
  - The major regression worth calling out is: https://github.com/pulumi/pulumi-alicloud/pull/503#discussion_r1398012963
- https://github.com/pulumi/pulumi-null/pull/52: No change
- https://github.com/pulumi/pulumi-fastly/pull/378: No change
- https://github.com/pulumi/pulumi-ec/pull/195: no change
- https://github.com/pulumi/pulumi-github/pull/502/files: minimal changes, equivalent quality
- https://github.com/pulumi/pulumi-azuredevops/pull/215/: This is blocking
  - https://github.com/pulumi/pulumi-azuredevops/pull/215#discussion_r1398021349
- https://github.com/pulumi/pulumi-oci:
  - Blocked on a panic: https://github.com/pulumi/pulumi-oci/actions/runs/6910194354

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1099,14 +1099,16 @@ func prefixMatcher(prefix string) func(string) int {
 var cleanDescriptionPrefixFns = []func(string) int{
 	// Cut Leading ':' and '-'
 	prefixMatcher("-"),
-	prefixMatcher(":"),
 	prefixMatcher("–"), // This is an en-dash (Unicode 8211)
+	prefixMatcher(":"),
 	// Cut leading "(Optional.*)"
 	//
 	// This needs to be a function and not a regex because we want the final ')' cut
 	// to balance with the first '('.
 	nestedParensMatcher("Optional"),
 	nestedParensMatcher("Required"),
+	nestedParensMatcher("Computed"),
+	nestedParensMatcher("Forces new resource"),
 }
 
 func cleanDescription(path docsPath, desc string) string {
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1064,8 +1064,9 @@ func renderMdNode(src []byte, n ast.Node) []byte {
 
 func nestedParensMatcher(trigger string) func(string) int {
 	trigger = "(" + trigger
+	lower := strings.ToLower(trigger)
 	return func(s string) int {
-		if !strings.HasPrefix(s, trigger) {
+		if !strings.HasPrefix(s, trigger) && !strings.HasPrefix(s, lower) {
 			return 0
 		}
 
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -872,10 +872,26 @@ func parseArgReferenceSection(subsection []string, ret *entityDocs) {
 	type namedNode struct {
 		node     ast.Node
 		name     string
-		writable bool
+		listItem bool
 	}
-	var parent, child, latest, latestParent *namedNode
-	setLatest := func() { latest, latestParent = child, parent }
+	var current []*namedNode
+	var latest []*namedNode
+	setLatest := func() {
+		latest = current[:0] // Zero out latest without re-allocation
+		for _, n := range current {
+			latest = append(latest, n)
+		}
+	}
+	keyOf := func(path []*namedNode) docsPath {
+		if len(path) == 0 {
+			return ""
+		}
+		key := docsPath(path[0].name)
+		for _, n := range path[1:] {
+			key = key.join(n.name)
+		}
+		return key
+	}
 
 	skipPrint := map[ast.Node]bool{}
 
@@ -896,16 +912,9 @@ func parseArgReferenceSection(subsection []string, ret *entityDocs) {
 			unbaked = map[docsPath]struct{}{}
 		case ast.KindListItem:
 			// We have entered a new item
-			if name := nodeItemName(src, node); name != "" && child == nil {
-				var key docsPath
-				if parent == nil {
-					child = nil
-					parent = &namedNode{node, name, true}
-					key = docsPath(name)
-				} else {
-					child = &namedNode{node, name, true}
-					key = docsPath(parent.name).join(name)
-				}
+			if name := nodeItemName(src, node); name != "" {
+				current = append(current, &namedNode{node, name, true})
+				key := keyOf(current)
 				setLatest()
 
 				_, isBaked := baked[key]
@@ -917,8 +926,8 @@ func parseArgReferenceSection(subsection []string, ret *entityDocs) {
 		case ast.KindHeading:
 			target := isHeadingTarget(src, node.(*ast.Heading))
 			if target != "" {
-				child = &namedNode{node, target, true}
-				parent = child
+				current = append(current[:0], &namedNode{node, target, false})
+				skipPrint[node] = true
 				setLatest()
 				return ast.WalkSkipChildren
 			}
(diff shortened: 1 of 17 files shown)


pkg/tfgen/test_data/resources/libvirt_network/docs.json
```diff
@@ -0,0 +1,75 @@
...
+    },
+    "dns.forwarders": {
+      "description": "Either `address`, `domain`, or both must be set"
+    },
+    "dns.hosts": {
+      "description": "a DNS host entry block. You can have one or more of these\nblocks in your DNS definition. You must specify both `ip` and `hostname`.\n\nAn advanced example of round-robin DNS (using DNS host templates) follows:\n\n```hcl\nresource \"libvirt_network\" \"my_network\" {\n...\ndns {\nhosts { flatten(data.libvirt_network_dns_host_template.hosts.*.rendered) }\n}\n...\n}\n\ndata \"libvirt_network_dns_host_template\" \"hosts\" {\ncount    = var.host_count\nip       = var.host_ips[count.index]\nhostname = \"my_host\"\n}\n```\n\nAn advanced example of setting up multiple SRV records using DNS SRV templates is:\n\n```hcl\ndata \"libvirt_network_dns_srv_template\" \"etcd_cluster\" {\ncount    = var.etcd_count\nservice  = \"etcd-server\"\nprotocol = \"tcp\"\ndomain   = var.discovery_domain\ntarget   = \"${var.cluster_name}-etcd-${count.index}.${discovery_domain}\"\n}\n\nresource \"libvirt_network\" \"k8snet\" {\n...\ndns {\nsrvs = [ flatten(data.libvirt_network_dns_srv_template.etcd_cluster.*.rendered) ]\n}\n...\n}\n```"
```
t0yv0: These are `hcl` blocks so this is before example rendering? Trying to wrap my head around sequencing.

pkg/tfgen/docs_test.go
```diff
@@ -34,12 +34,70 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
-	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen/internal/testprovider"
 )
 
+func TestEntityDocsParsing(t *testing.T) {
```
t0yv0, line 45: Much needed!

pkg/tfgen/test_data/resources/libvirt_network/docs.json
iwahbe: Yes. We have a parse phase and then a render phase, with intermediate data represented in `entityDocs`. This change only effects the parse: raw markdown (`[]byte`) to `entityDocs`.

pkg/tfgen/docs.go
```diff
@@ -867,69 +876,259 @@ func getNestedBlockName(line string) string {
...
+					// this until the list ends (and we reset).
+					p := node.Parent()
+					child = &namedNode{p, target, false}
+					parent = child
+					setLatest()
+					return ast.WalkSkipChildren
+				}
+			}
+		case ast.KindText:
+			if genericNestedRegexp.Match(node.Text(src)) {
```
t0yv0: What does node.Text() do?  It's been a while since I touched goldmark, I more recently wrestled with blackfriday. It can be really difficult to work with.

pkg/tfgen/docs.go
```diff
@@ -867,69 +876,259 @@ func getNestedBlockName(line string) string {
...
+	contract.AssertNoErrorf(err, "An error is never returned")
+
+	for k, v := range ret.Arguments {
+		v.description = cleanDescription(k, v.description)
 	}
+}
 
-	for _, v := range ret.Arguments {
-		v.description = strings.TrimRightFunc(v.description, unicode.IsSpace)
+func renderMdNode(src []byte, n ast.Node) []byte {
```
t0yv0, line 1057: I've run into this before; so I can guess. why are we rendering back? Trying to run regexes on a subtree? 

I think I had to do something like this before an it was very disappointing that markdown library didn't provide a full featured AST->markdown renderer.

pkg/tfgen/docs.go
iwahbe: `node`s are flyweight constructs. They don't have access to the text that they represent. `node.Text(src []byte) []byte` returns the text from `src` that `node` represents.

t0yv0:
GCP changes don't look good. Some undesired changes are being added to GCP.

```
» $schema diff resources gcp:accesscontextmanager/egressPolicy:EgressPolicy properties egressPolicyName description
+ 
+ 
+ - - -
```

t0yv0:
This kind of code is going to be very scary/ difficult to roll out based on the strength of unit tests alone. These are heuristic codes that clean up textual data. We don't have good automated data quality checks. 

We need to slow down to figure out how we're rolling this out, which probably involves eyeballing the change carefully for every Tier 1 major provider. We can talk about automated DQ checks but we don't have those yet so this is probably going to need eyeballs anyway for the rollout.

I'm tempted to do some weekend improvements to my `pus` tool that helps categorize JSON diffs, or we could try Daniel's explode tools. 

It's reasonable to budget 1d for every major provider to make sure it rolls out safe. We  should probably also flag this so rollout is per provider. 

That sounds reasonable? Or we're feeling lucky here. CC @mjeffryes

iwahbe:
> GCP changes don't look good. Some undesired changes are being added to GCP.
> 
> ```
> » $schema diff resources gcp:accesscontextmanager/egressPolicy:EgressPolicy properties egressPolicyName description
> + 
> + 
> + - - -
> ```

I don't know what those are, I *think* they are thematic breaks. `schema` has the diff backwards. This change is removing the `- - -`, not adding them.

t0yv0:
Oh we have that in the docs currently :) that's terrible. Getting back to this shortly, I'd like to eyeball the diff some more.

pkg/tfgen/docs.go
iwahbe: Basically, the MD parser doesn't provide clean spans back into the original text. Having walked the parsed nodes to find what we want to extract, we need to render back to text the selected bit.

If it was possible, I would like this function to read `return src[n.Span().Start:n.Span().End]`, but that isn't supported.

iwahbe:
This PR is out of date and has been largely superseded by @guineveresaenger's doc work.

PR #1409, opened 2023-09-29
Use a realistic example to increase convertExamples branch coverage

t0yv0:
Covers this condition:

    skippedExamples && stripSubsectionsWithErrors && !isFrontMatter

Also adds a golden test for the generated content.

Note that although convertExamples shows up as covered in our reporting right now, the coverage is very low quality, for example tweaking the generated text still passes. The new test pins this a little better.

To discover tests like these, build a provider against a local bridge version and inject panics into uncovered lines to recover function parameters. The technique is a bit tedious but seems to work here.

### DIFF
--- a/pkg/tfgen/docs_test.go
+++ b/pkg/tfgen/docs_test.go
@@ -39,6 +39,7 @@ import (
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen/internal/testprovider"
+	"os/exec"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


t0yv0:
Ah this test needs wafevfront plugin to be installed in pulumi otherwise it doesn't produce the right result; that's too bad. I'll ned to tighten this tomorrow.

iwahbe (commented):
The test looks fine. I don't see any reference to `skippedExamples` or `!isFrontMatter`. Can you call out where these are set?

t0yv0:
It's inside convertExamples function. This function is fairly complicated, and I need to perform some edits on it to enable bulk-rendering of examples via `pulumi covert` as part of updated converter rollout, but it badly needs this kind of coverage before attempting changes.

t0yv0:
@mjeffryes d'you know if something went wrong with codecov or it's not installed here yet?

mjeffryes:
I think I might still have the codecov PR outstanding

iwahbe (approved):

Issue #1410, opened 2023-09-29
Optimize the performance of `MustApplyAutoAliases` on provider init


mjeffryes:
https://github.com/pulumi/pulumi-aws/issues/2799 identified the `MustApplyAutoAliases` method in the AWS provider startup as a significant source of overhead for large providers. Due to the design of running this code both during tfgen and during provider init, we do some work that's not strictly needed on provider init:
1) Iterate the whole metadata to identify the last version
2) Iterate *all* resources on the provider (rather than just those with aliases) and modifying the metadata as we go
3) Serializing the updated metadata back to json

If we can come up with a clever way to separate the runtime concerns from the gen time concerns, we can probably get a noticeable speed up in init for lager providers

iwahbe:
(1) should only occur if `ProviderInfo.Version` was not set, which shouldn't be the case for pulumi-aws.

t0yv0:
Related: https://github.com/pulumi/pulumi-terraform-bridge/pull/1544 will take it a lot further if we can land it safely.

pgavlin:
#1544 landed; closing.

PR #1414, opened 2023-10-03
Test potential MustApplyAutoAliases speedup

mjeffryes:
toward https://github.com/pulumi/pulumi-terraform-bridge/issues/1410

Uses the build path for the binary to identify when we are in plugin runtime and  disable sections of `ApplyAutoAliases` that are only needed for generating the bridge metadata file. 

```
❯ (cd provider; go test -run="^$" -bench="BenchmarkProvider.*Gen$" -benchtime=10s)
goos: darwin
goarch: arm64
pkg: github.com/pulumi/pulumi-aws/provider/v6
BenchmarkProviderGen-12      	      58	 204080193 ns/op
BenchmarkProviderNoGen-12    	      68	 174488096 ns/op
PASS
```

I've extracted the bit of trying to guess whether we're in a tfgen binary or a plugin binary to it's own module so we could change out the implementation more easily in the future.

### DIFF
--- a/pkg/tfbridge/auto_aliasing.go
+++ b/pkg/tfbridge/auto_aliasing.go
@@ -16,8 +16,6 @@ package tfbridge
 
 import (
 	"github.com/Masterminds/semver"
-	"runtime/debug"
-	"strings"
 
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 	md "github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata"
@@ -125,9 +123,6 @@ func (info *ProviderInfo) ApplyAutoAliases() error {
 		return err
 	}
 
-	buildInfo, _ := debug.ReadBuildInfo()
-	isTfgen := buildInfo != nil && strings.Contains(buildInfo.Path, "pulumi-tfgen")
-
 	var currentVersion int
 	// If version is missing, we assume the current version is the most recent major
 	// version in mentioned in history.
@@ -167,7 +162,7 @@ func (info *ProviderInfo) ApplyAutoAliases() error {
 	for tfToken, computed := range info.Resources {
 		r, _ := rMap.GetOk(tfToken)
 		aliasResource(info, r, &applyAliases, hist.Resources,
-			computed, tfToken, currentVersion, isTfgen)
+			computed, tfToken, currentVersion)
 	}
 
 	for tfToken, computed := range info.DataSources {
@@ -180,7 +175,7 @@ func (info *ProviderInfo) ApplyAutoAliases() error {
 		f()
 	}
 
-	if isTfgen {
+	if isTfgen() {
 		if err := md.Set(artifact, aliasMetadataKey, hist); err != nil {
 			// Set fails only when `hist` is not serializable. Because `hist` is
 			// composed of marshallable, non-cyclic types, this is impossible.
@@ -212,11 +207,10 @@ func aliasResource(
 	applyResourceAliases *[]func(),
 	hist map[string]*tokenHistory[tokens.Type], computed *ResourceInfo,
 	tfToken string, version int,
-	isTfgen bool,
 ) {
 	prev, hasPrev := hist[tfToken]
 	if !hasPrev {
-		if isTfgen {
+		if isTfgen() {
 			// It's not in the history, so it must be new. Stick it in the history for
 			// next time.
 			hist[tfToken] = &tokenHistory[tokens.Type]{
@@ -271,7 +265,7 @@ func applyResourceMaxItemsOneAliasing(
 		hasH = hasH || fieldHasHist
 		hasI = hasI || fieldHasInfo
 
-		if !hasH {
+		if !hasH && isTfgen() {
 			delete(*hist, k)
 		}
 		if !hasI {
@@ -330,13 +324,15 @@ func applyMaxItemsOneAliasing(schema shim.Schema, h *fieldHistory, info *SchemaI
 		// MaxItemsOne does not apply, so do nothing
 	} else if info.MaxItemsOne != nil {
 		// The user has overwritten the value, so we will just record that.
-		h.MaxItemsOne = info.MaxItemsOne
-		hasH = true
+		if isTfgen() {
+			h.MaxItemsOne = info.MaxItemsOne
+			hasH = true
+		}
 	} else if h.MaxItemsOne != nil {
 		// If we have a previous value in the history, we keep it as is.
 		info.MaxItemsOne = h.MaxItemsOne
 		hasI = true
-	} else {
+	} else if isTfgen() {
 		// There is no history for this value, so we bake it into the
 		// alias history.
 		h.MaxItemsOne = BoolRef(IsMaxItemsOne(schema, info))
@@ -349,10 +345,12 @@ func applyMaxItemsOneAliasing(schema shim.Schema, h *fieldHistory, info *SchemaI
 	// If the .Elem existed before this function, we mark it as unsafe to cleanup.
 	var hasElemH, hasElemI bool
 	populateElem := func() {
-		if h.Elem == nil {
-			h.Elem = &fieldHistory{}
-		} else {
-			hasElemH = true
+		if isTfgen() {
+			if h.Elem == nil {
+				h.Elem = &fieldHistory{}
+			} else {
+				hasElemH = true
+			}
 		}
 		if info.Elem == nil {
 			info.Elem = &SchemaInfo{}
@@ -367,7 +365,7 @@ func applyMaxItemsOneAliasing(schema shim.Schema, h *fieldHistory, info *SchemaI
 	cleanupElem := func(elemHist, elemInfo bool) {
 		hasElemH = hasElemH || elemHist
 		hasElemI = hasElemI || elemInfo
-		if !hasElemH {
+		if !hasElemH && isTfgen() {
 			h.Elem = nil
 		}
 		if !hasElemI {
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/auto_aliasing.go
```diff
@@ -175,11 +175,13 @@ func (info *ProviderInfo) ApplyAutoAliases() error {
...
-		// composed of marshallable, non-cyclic types, this is impossible.
-		contract.AssertNoErrorf(err, "History failed to serialize")
-	}
+	/*
+		if err := md.Set(artifact, aliasMetadataKey, hist); err != nil {
+			// Set fails only when `hist` is not serializable. Because `hist` is
+			// composed of marshallable, non-cyclic types, this is impossible.
+			contract.AssertNoErrorf(err, "History failed to serialize")
+		}
+	*/
```
mjeffryes, line 184: This change by itself yields:
```
❯ go test -run="^$" -bench="BenchmarkProvider$" -benchtime=10s
goos: darwin
goarch: arm64
pkg: github.com/pulumi/pulumi-aws/provider/v6
BenchmarkProvider-12                  94         126865278 ns/op
PASS
```

pkg/tfbridge/auto_aliasing.go
```diff
@@ -226,7 +236,7 @@ func aliasResource(
 	// Note: If the user explicitly sets a MaxItemOne value, that value is respected
 	// and overwrites the current history.'
 
-	if res == nil {
+	if res == nil || hist[tfToken] == nil {
```
mjeffryes, line 239: This change means we skip traversing all the fields of the resource which seems to account for ~60% of the speedup

pkg/tfbridge/auto_aliasing.go
```diff
@@ -16,6 +16,8 @@ package tfbridge
 
 import (
 	"github.com/Masterminds/semver"
+	"runtime/debug"
+	"strings"
```
iwahbe, line 20: Nit: Our style has std lib modules in a different block then foreign modules.
```suggestion
	"runtime/debug"
	"strings"
	
	"github.com/Masterminds/semver"
```

pkg/tfbridge/auto_aliasing.go
```diff
@@ -123,6 +125,9 @@ func (info *ProviderInfo) ApplyAutoAliases() error {
 		return err
 	}
 
+	buildInfo, _ := debug.ReadBuildInfo()
+	isTfgen := buildInfo != nil && strings.Contains(buildInfo.Path, "pulumi-tfgen")
```
iwahbe, line 129: I'd really prefer that our generate entry point sets some internal global variable then us trying to parse the build path.

mjeffryes: Yes, I'm exploring that in https://github.com/pulumi/pulumi-terraform-bridge/pull/1415, but so far I can't figure out how to only set the flag for tfgen without modifying the provider itself (which we could obviously do for AWS, but that doesn't help any of our other providers).

iwahbe: I see the problem. We can leave that until the next time we change how our main function works.

mjeffryes:
Boo. Can't reproduce these test failures locally for some reason 😢

pkg/tfbridge/runtime.go
```diff
@@ -0,0 +1,48 @@
...
+		} else if strings.Contains(buildInfo.Path, "pulumi-resource") {
+			stage = ResourceStage
+		}
+	}
+	return RuntimeInfo{
+		Stage: stage,
+	}
+}
+
+func ReadRuntimeInfo() RuntimeInfo {
```
t0yv0, line 38: I'd be a bit careful exposing public API in pkg/tfbridge that's top visibility.

iwahbe: Do we want providers to use this? Until there is a need for providers to use this functionality, I would keep it private (or `internal/`).

mjeffryes: I was using it from aws to benchmark each way, but I agree we should not make it visible

mjeffryes:
Learned on a different thread that we may be initializing the plugin even more than 4 times in some programs; given that context, I agree with @t0yv0 that it's probably worth shipping this at least as a temporary measure until we have a chance to more carefully separate runtime and gen-time concerns.

pkg/tfbridge/runtime.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+	}
+	return runtimeInfo{
+		stage: stage,
+	}
+}
+
+func getRuntimeStage() runtimeStage {
+	return runtime.stage
+}
```
iwahbe, line 40: Nit: I would inline this helper function.

iwahbe (approved):
LGTM

pkg/tfbridge/auto_aliasing.go
```diff
@@ -508,3 +521,7 @@ func aliasOrRenameDataSource(
 	}
 
 }
+
+func isTfgen() bool {
+	return getRuntimeStage() != resourceStage
+}
```
iwahbe, line 527: Nit: this function should be in the `runtime.go` file.

pkg/tfbridge/auto_aliasing.go
```diff
@@ -369,11 +377,17 @@ func applyMaxItemsOneAliasing(schema shim.Schema, h *fieldHistory, info *SchemaI
 	switch e := e.(type) {
 	case shim.Resource:
 		populateElem()
-		eHasH, eHasI := applyResourceMaxItemsOneAliasing(e, &h.Elem.Fields, &info.Elem.Fields)
+		var eHasH, eHasI bool
+		if h.Elem != nil {
+			eHasH, eHasI = applyResourceMaxItemsOneAliasing(e, &h.Elem.Fields, &info.Elem.Fields)
```
t0yv0, line 382: Are we sure this here is safe? These functions edit SchemaInfo.MaxItemsOne to ensure backwards compatibility. If these edits are dropped at runtime, but only performed at compile-time, we'll get unexpected runtime serializer errors possibly.

mjeffryes: `applyResourceMaxItemsOneAliasing` and `applyMaxItemsOneAliasing` panic if  `h.Elem` is `nil`

The change in line 368 above means that sometimes `h.elem` will be `nil` here (ie. in the case that there was nothing in the metadata for this field), so we have to add the test. At runtime, it's safe to just skip because there's nothing in the metadata.

Although, it writing this I can see a cleaner way to handle this with a lot fewer conditionals, so I'll try that.

mjeffryes:
Cleaned up the logic in `applyMaxItemsOneAliasing`, and consolidated helper functions into runtime.go

pkg/tfbridge/auto_aliasing.go
t0yv0: I'm worried that this changes semantics. I don't know what conditions precisely get us to "sometimes h.elem will be nil" but it sounds like it shouldn't be? 

Are we quite certain that under both versions of the flag `isTfgen() = {true,false}` we're getting the *same* providerInfo back?

Something perhaps checkable with:

```
reflect.DeepEqual(a.Resources, b.Resouces)`
```

If this holds on a big provider I'd be less worried.

t0yv0: Hm the last version looks a lot cleaner! It seems to me only history edits are skipped, and there's short-cicuiting of edits when we know there's no history so there definitionally shouldn't be edits. This looks safer. Think about the invariant above but if you're confident we don't need further testing.

t0yv0 (approved):

pkg/tfbridge/auto_aliasing.go
mjeffryes: I'll run that test on aws before I merge, just to be sure.

mjeffryes: So the Resources are not quite identical, but I think semantically the same. When running with `isTfGen() == false` we sometimes produce empty field maps, where the version with `isTfgen() == true` produces `nil`
```
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -79007,4 +79007,3 @@
        	            	      Elem: (*tfbridge.SchemaInfo)(<nil>),
        	            	-     Fields: (map[string]*tfbridge.SchemaInfo) {
        	            	-     },
        	            	+     Fields: (map[string]*tfbridge.SchemaInfo) <nil>,
        	            	      Asset: (*tfbridge.AssetTranslation)(<nil>),
        	            	@@ -79124,4 +79123,3 @@
        	            	      Elem: (*tfbridge.SchemaInfo)(<nil>),
        	            	-     Fields: (map[string]*tfbridge.SchemaInfo) {
        	            	-     },
        	            	+     Fields: (map[string]*tfbridge.SchemaInfo) <nil>,
        	            	      Asset: (*tfbridge.AssetTranslation)(<nil>),
        	            	@@ -79168,4 +79166,3 @@
        	            	      Elem: (*tfbridge.SchemaInfo)(<nil>),
        	            	-     Fields: (map[string]*tfbridge.SchemaInfo) {
        	            	-     },
        	            	+     Fields: (map[string]*tfbridge.SchemaInfo) <nil>,
        	            	      Asset: (*tfbridge.AssetTranslation)(<nil>),
        	            	@@ -79212,4 +79209,3 @@
        	            	      Elem: (*tfbridge.SchemaInfo)(<nil>),
        	            	-     Fields: (map[string]*tfbridge.SchemaInfo) {
        	            	-     },
        	            	+     Fields: (map[string]*tfbridge.SchemaInfo) <nil>,
        	            	      Asset: (*tfbridge.AssetTranslation)(<nil>),
        	Test:       	TestProviderSame
```

t0yv0: Yes, nil vs empty should be fine here.

PR #1416, opened 2023-10-05
Add detailed documentation for auto aliasing

iwahbe:


pkg/tfbridge/auto_aliasing.go
```diff
@@ -95,20 +95,75 @@ func (info *ProviderInfo) MustApplyAutoAliases() {
 // - Edit [SchemaInfo.MaxItemsOne]
 //
 // The goal is to always maximize backwards compatibility and reduce breaking changes for
-// the users of the Pulumi providers.
+// the users of the Pulumi providers. The basic functionality behind each action is
+// identical; ApplyAutoAliases keeps a record of which TF token maps to which Pulumi
+// token, which fields have MaxItemsOne (true or false), and what version the record is
+// from.
```
guineveresaenger, line 101: I like this summary.

pkg/tfbridge/auto_aliasing.go
```diff
@@ -95,20 +95,75 @@ func (info *ProviderInfo) MustApplyAutoAliases() {
...
 //
-// Resource aliases help mask TF provider resource renames or changes in mapped tokens so
-// older programs continue to work.  See [ResourceInfo.RenameResourceWithAlias] and
-// [ResourceInfo.Aliases] for more details.
+// For example, this is the (abbreviated & modified) history for GCP's compute autoscalar:
```
guineveresaenger, line 103: There's some missing context here that I think would be beneficial to go into first.

We are specifically solving the problem of backcompat with aliasing  👍  I would like an explanation/diagram of the specific mechanics of auto aliasing.

- where does this example come from/where do we keep this record?
- how is this history maintained and modified?

pkg/tfbridge/auto_aliasing.go
```diff
@@ -95,20 +95,75 @@ func (info *ProviderInfo) MustApplyAutoAliases() {
...
+//
+// # Call [ProviderInfo.RenameResourceWithAlias] or [ProviderInfo.RenameDataSource]
+//
+// "google_compute_autoscaler.majorVersion" tells us that this record was last updated at
+// major version 6. One of the previous names is also at major version 6, so we want to
+// keep full backwards compatibility.
+//
+// ApplyAutoAliases will call [ProviderInfo.RenameResourceWithAlias] to create a SDK entry
+// for the old Pulumi token ("gcp:auto/autoscalar:Autoscalar"). This token will no longer
+// be created when `make tfgen` is run on version 7.
```
guineveresaenger, line 144: Is this because we don't have an entry for v7? because v7 is getting rid of this resource upstream?

pkg/tfbridge/auto_aliasing.go
```diff
@@ -95,20 +95,75 @@ func (info *ProviderInfo) MustApplyAutoAliases() {
...
+//	                    "cpu_utilization": {
+//	                        "maxItemsOne": false
+//	                    }
+//	                }
+//	            }
+//	        }
+//	    }
+//	}
+//
+// I will address each action as it applies to `"google_compute_autoscaler" in turn:
```
guineveresaenger, line 134: Can we mention what triggers these actions?

pkg/tfbridge/auto_aliasing.go
```diff
@@ -95,20 +95,75 @@ func (info *ProviderInfo) MustApplyAutoAliases() {
...
+//
+// # Edit [ResourceInfo.Aliases]
+//
+// In this history, we have recorded two prior names for "google_compute_autoscaler":
+// "gcp:auto/autoscalar:Autoscalar" and "gcp:auto/scaler:Scaler". Since
+// "gcp:auto/scaler:Scaler" was from a previous major version, we don't need to maintain
+// full backwards compatibility. Instead, we will apply a type alias to
+// `.Resources["gcp:compute/autoscalar:Autoscalar"].Aliases`: `AliasInfo{Type:
+// ""gcp:auto/scaler:Scaler"}`. This makes it easy for consumers to upgrade from the old
+// name to the new.
```
guineveresaenger, line 154: If I'm reading this correctly, there's a difference between a "type alias" == helpful when upgrading; needs customer to update the resource name but nothing else, and what results from "RenameResourceWithAlias" == a fully supported alternate name for a resource; needs the customer to do nothing. It would be helpful to disambiguate explicitly.

pkg/tfbridge/auto_aliasing.go
```diff
@@ -95,20 +95,75 @@ func (info *ProviderInfo) MustApplyAutoAliases() {
...
+//
+// "google_compute_autoscaler.majorVersion" tells us that this record was last updated at
+// major version 6. One of the previous names is also at major version 6, so we want to
+// keep full backwards compatibility.
+//
+// ApplyAutoAliases will call [ProviderInfo.RenameResourceWithAlias] to create a SDK entry
+// for the old Pulumi token ("gcp:auto/autoscalar:Autoscalar"). This token will no longer
+// be created when `make tfgen` is run on version 7.
+//
+// # Edit [ResourceInfo.Aliases]
```
guineveresaenger, line 146: Where are we storing the result of this edit? Does this go straight into the schema?

pkg/tfbridge/auto_aliasing.go
```diff
@@ -95,20 +95,75 @@ func (info *ProviderInfo) MustApplyAutoAliases() {
...
+//
+// # Edit [SchemaInfo.MaxItemsOne]
+//
+// The provider has been shipped with fields that could have `MaxItemsOne` applied. Any
+// change here is breaking to our users, so we prevent it. As long as the provider's major
+// version is 6, ApplyAutoAliases will override the MaxItemsOne status of
+// "autoscaling_policy" (to true) and "autoscaling_policy.elem.cpu_utilization" (to
+// false), regardless of what upstream does. This will be dropped when gcp v7 is
+// released. Effectively this makes sure that upstream MaxItems changes are deferred until
+// the next major version.
```
guineveresaenger, line 164: ah! the intent is clear and makes sense, I think pointing out the effective end goal is super helpful.

pkg/tfbridge/auto_aliasing.go
```diff
@@ -95,20 +95,75 @@ func (info *ProviderInfo) MustApplyAutoAliases() {
...
+// ""gcp:auto/scaler:Scaler"}`. This makes it easy for consumers to upgrade from the old
+// name to the new.
+//
+// # Edit [SchemaInfo.MaxItemsOne]
+//
+// The provider has been shipped with fields that could have `MaxItemsOne` applied. Any
+// change here is breaking to our users, so we prevent it. As long as the provider's major
+// version is 6, ApplyAutoAliases will override the MaxItemsOne status of
+// "autoscaling_policy" (to true) and "autoscaling_policy.elem.cpu_utilization" (to
+// false), regardless of what upstream does. This will be dropped when gcp v7 is
```
guineveresaenger, line 162: How does dropping the deferral at v7 work? Will this token get dropped from the record at tfgen time?

guineveresaenger (commented):
I think the main thing I'd like here is a bit more situational context, and an arch diagram that shows the interaction between the upstream resources, the logic in the provider's `resources.go` , where aliasing history is kept, and how /when the bridge generates it.

I also don't entirely see the "auto" part here - how is the need for an alias auto-decided (since we're also automatically generating tokens as well, I assume the answer is "when upstream decides to fix a typo or spell `Iam` as `IAM`).

Is there ever a need for manual aliasing? How does that interact with the record of past aliases?

I would prefer to see this in a separate doc for the diagram reasons but I'll take ASKII art as well.

iwahbe:
> I also don't entirely see the "auto" part here - how is the need for an alias auto-decided (since we're also automatically generating tokens as well, I assume the answer is "when upstream decides to fix a typo or spell Iam as IAM).

The "auto" part is in contrast to you manually applying aliases. 

The logic for when an automatic alias should exist is:

An alias exists for a resource if:

1. `make tfgen` was run at time $t_n$ with token mapping $tk\to p_1$:

$$ (tk \to p_1, t_n) $$

2. And there is another mapping:

$$ (tk \to p_2, t_{n+k}) | k>0, p_1\neq p_2 $$

If both mappings exist, then $p_2$ will alias $p_1$. If $majorVersion(t_n)=majorVersion(t_{n+k})$ then there will be a hard (non-breaking) alias.

I'll write a less mathy version in the doc. I'm using this as a scratch pad for explaining.

pkg/tfbridge/auto_aliasing.go
iwahbe: The token won't be dropped, but the `MaxItemsOne: ${VALUE}` will be reset. I changed the language to make it clearer.

pkg/tfbridge/auto_aliasing.go
iwahbe: I added a section at the beginning to highlight the goals here.

pkg/tfbridge/auto_aliasing.go
iwahbe: I addressed this in the next paragraph.

pkg/tfbridge/auto_aliasing.go
iwahbe: Done

pkg/tfbridge/auto_aliasing.go
iwahbe: Done

pkg/tfbridge/auto_aliasing.go
iwahbe: I would like to avoid going too deep into that. The bridge has a system for maintaining these kinds of changes `github.com/pulumi/pulumi-terraform-bridge/unstable/metadata`. I mention that we use this system, but how that works is a separate thing, and not necessary for understanding `ApplyAutoAliases`.


Brief summary:

The data winds up in `bridge-metadata.json` (`."auto-aliasing"`) and its lifecycle is managed as part of `make tfgen`.

Retrieving from history:
https://github.com/pulumi/pulumi-terraform-bridge/blob/64ac968bae388b3e941ce688d447e65c197a48e7/pkg/tfbridge/auto_aliasing.go#L277-L290

Saving to history:
https://github.com/pulumi/pulumi-terraform-bridge/blob/64ac968bae388b3e941ce688d447e65c197a48e7/pkg/tfbridge/auto_aliasing.go#L266-L270

pkg/tfbridge/auto_aliasing.go
iwahbe: Clarified in https://github.com/pulumi/pulumi-terraform-bridge/pull/1416/commits/f29d643740970b44f18a118e4e9e3d02d1aaa274.

iwahbe:
@mjeffryes Do you know why we are seeing a [drop in code coverage](https://github.com/pulumi/pulumi-terraform-bridge/pull/1416#issuecomment-1749652253) in this PR. The only thing this PR changes is a doc comment. There is no runtime effect.

mjeffryes:
> @mjeffryes Do you know why we are seeing a [drop in code coverage](https://github.com/pulumi/pulumi-terraform-bridge/pull/1416#issuecomment-1749652253) in this PR. The only thing this PR changes is a doc comment. There is no runtime effect.

I can't explain it. Clicking through the UI doesn't give any indications which files it thinks have less coverage either. The only tidbit I see is that there are  extra reports for this change (6 vs normal of 3). Codecov does not inspire a lot of confidence here.

pkg/tfbridge/README.md
```diff
@@ -0,0 +1,148 @@
+# tfbridge
+
+## Automatic Aliasing (`ApplyAutoAliases`)
+
+Automatically applies backwards compatibility best practices.
```
guineveresaenger, line 5: LOVE this summary. It clearly states the purpose and establishes context. ❤️

pkg/tfbridge/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+The goal is to prevent breaking changes from Pulumi maintainers or from the upstream
```
guineveresaenger, line 7:
```suggestion
The goal is to prevent breaking changes by Pulumi maintainers or by the upstream
```
Wanted to disambiguate the prepositions a bit here

pkg/tfbridge/README.md
```diff
@@ -0,0 +1,148 @@
...
+    D -->|"make tfgen or runtime"| E["prov.ApplyAutoAliasing*\nfrom pulumi-terraform-bridge/pkg\n\nIterates through all resources and datasources,\nto applying and recording:\n1. Hard aliases**\n2. Soft aliases***\n3. MaxItemsOne corrections"]
+    E -->|"Iterate through resources and datasources to apply"| F["New aliasing history"] 
+    E -->|"Always (ignored during tfgen time)"| G["Runtime aliasing information"]
+    F -->|"save history (make tfgen ONLY)"| A
+```
+
+> (\*)   This gets called anytime Provider() gets called.
+>
+> (\*\*) A "hard alias" is a fully schematized copy of a resource, where we
+> ensure. backwards compatibility for new token names. Mainly used during minor upgrades.
```
guineveresaenger, line 64:
```suggestion
> ensure backwards compatibility for new token names. Mainly used during minor upgrades.
```

pkg/tfbridge/README.md
```diff
@@ -0,0 +1,148 @@
...
+>
+> (\*\*\*) A "soft alias" is a plain rename of a token, which requires a user to rename
+> the resource (but not do an stack surgery). Only a resource rename is needed. Acceptable
```
guineveresaenger, line 67:
```suggestion
> the resource (but not do any stack surgery). Only a resource rename is needed. Acceptable
```

pkg/tfbridge/README.md
```diff
@@ -0,0 +1,148 @@
...
+> as a result of major upgrades
```
guineveresaenger, line 68:
```suggestion
> as a result of major upgrades.
```

pkg/tfbridge/README.md
```diff
@@ -0,0 +1,148 @@
...
+	                        "maxItemsOne": false
+	                    }
+	                }
+	            }
+	        }
+	    }
+	}
+```
+
+I will address each action as it applies to `"google_compute_autoscaler" in turn:
```
guineveresaenger, line 107:
```suggestion
Here are the details of each action as it applies to `"google_compute_autoscaler" in turn:
```

pkg/tfbridge/README.md
```diff
@@ -0,0 +1,148 @@
...
+the old name from the schema.
+
+- The upstream provider has added or removed `MaxItems: 1` from a field.
+
+`ApplyAutoAliases` applies three mitigation strategies: one for each breaking change it
+is attempting to mitigate.
+
+- Call `ProviderInfo.RenameResourceWithAlias` or `ProviderInfo.RenameDataSource`: This
+creates a "hard alias", preventing the user from experiencing a breaking change between
+major versions.
```
guineveresaenger, line 27: Should say "minor versions", or needs more words to describe that we use hard aliases for all the minor and patch versions that happen, e.g after v1.x.x and before v2.x.x.

guineveresaenger (approved):
Just a few minor nits and this looks good to go.

PR #1420, opened 2023-10-09
Don't error on already mapped tokens

iwahbe:
Token mapping should not error when it is unable to map a token that was mapped manually. Apparently, this is not a common scenario, but I encountered it as part of https://github.com/pulumi/pulumi-spotinst/issues/436.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1421

### DIFF
--- a/pkg/tfbridge/tokens/known_modules.go
+++ b/pkg/tfbridge/tokens/known_modules.go
@@ -26,13 +26,14 @@ import (
 
 func knownModules[T b.ResourceInfo | b.DataSourceInfo](
 	prefix, defaultModule string, modules []string,
-	apply func(string, string, *T) error,
+	apply func(string, string, *T, error) error,
 	moduleTransform func(string) string,
 ) b.ElementStrategy[T] {
 	return func(tfToken string, elem *T) error {
 		tk := strings.TrimPrefix(tfToken, prefix)
 		if len(tk) == len(tfToken) {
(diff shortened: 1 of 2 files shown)


t0yv0 (approved):
LGTM. I'm a bit miffed did not catch this at plan/review time. Having ComputeTokens never override the tokens manually specified is a very important property as this is our "user override" story - actually can you elaborate a bit how did this behave before the PR?

t0yv0:
Would there be a situation where this change would cause a change in token allocation? As opposed to simply turning failing programs into working programs?

rquitales (approved):

iwahbe:
> LGTM. I'm a bit miffed did not catch this at plan/review time. Having ComputeTokens never override the tokens manually specified is a very important property as this is our "user override" story - actually can you elaborate a bit how did this behave before the PR?

Before this PR, we would error on missing modules for already mapped tokens:

`"pkg_mod_res"` would error unless `"mod"` was in the list of mapped modules. This would happen regardless of if `"pkg_mod_res"` was manually mapped. In effect, we would return an error when we were unable to compute information that we would end up not needing (the module for a manually mapped resource).

> Would there be a situation where this change would cause a change in token allocation? As opposed to simply turning failing programs into working programs?

I don't expect so. We will not experience any mapping changes from this PR.

It is always possible that user's will rely on the incorrect behavior, but I can't imagine a sensible usage that will break with this change.

PR #1440, opened 2023-10-13
Speedup example converter

t0yv0:
This loads all cores. I experimented with batch sizes on pulumi-aws measuring wall clock time for `make tfgen` on my machine that reports 12 NumCPU to go runtime. Seems to like smaller batches a little better but it's not very significant.

```
|  P |   B | time  |
|----+-----+-------|
| 12 |  32 | 7:40s |
| 12 |  64 | 8:18s |
| 12 | 128 | 8:41s |
```

Let me also try to put the parmap inside the converter.

Sister PR is https://github.com/pulumi/pulumi-converter-terraform/pull/73

### DIFF
--- a/pkg/tfgen/convert_cli.go
+++ b/pkg/tfgen/convert_cli.go
@@ -85,6 +85,12 @@ type cliConverter struct {
 	opts []pcl.BindOption             // options cache; do not set
 }
 
+// Represents mappings that need to be passed to the converter.
+type mapping struct {
+	name string
+	info tfbridge.ProviderInfo
+}
+
 // Represents a partially converted example. PCL is the Pulumi dialect of HCL.
 type translatedExample struct {
 	PCL         string          `json:"pcl"`
@@ -191,10 +197,7 @@ func (cc *cliConverter) bulkConvert() error {
 		examples[fileName] = hcl
 		n++
 	}
-	result, err := cc.convertViaPulumiCLI(examples, []struct {
-		name string
-		info tfbridge.ProviderInfo
-	}{
+	result, err := cc.convertViaPulumiCliBatchedParallel(examples, []mapping{
 		{
 			name: cc.packageName,
 			info: cc.info,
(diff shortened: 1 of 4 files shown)


pkg/tfgen/par.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"fmt"
+	"runtime"
+	"sync"
+)
+
+// Transforms a large map in batches of up to batch elements, using workers number of goroutines. If
+// workers is -1, uses one worker per CPU.
+func parTransformMap[K comparable, T any, U any](
```
t0yv0, line 25: Admittedly there must be a library for this.. it's something very generic.

pkg/tfgen/convert_cli.go
```diff
@@ -213,6 +216,34 @@ func (cc *cliConverter) bulkConvert() error {
...
+func (cc *cliConverter) convertViaPulumiCliBatchedParallel(
+	examples map[string]string,
+	mappings []mapping,
+) (
+	map[string]translatedExample,
+	error,
+) {
+	batch := 36
+	if b, ok := os.LookupEnv("PULUMI_CONVERT_BATCH"); ok {
+		if n, err := strconv.Atoi(b); err == nil {
```
iwahbe, line 228: Can we fail when `PULUMI_CONVERT_BATCH` is set but not parsable as a number.

pkg/tfgen/par.go
```diff
@@ -0,0 +1,106 @@
...
+) (map[K]U, error) {
+	if batch < 1 {
+		return nil, fmt.Errorf("batch cannot be less than 1")
+	}
+	n := workers
+	if workers < 1 {
+		n = runtime.NumCPU()
+		if n < 2 {
+			n = 2
+		}
```
iwahbe, line 39: Can you explain why we are doing this? Does this not work with `n = 1`?

iwahbe (commented):
If testing indicates that going from B=128 to B=64 is 23 seconds faster, and going from B=64 to B=32 is 38 seconds faster, why not try a batch size of 16, or 8, or 1.

t0yv0:
Best result so far on my machine: 2 converter processes, batch 256, each fully parallel inside (https://github.com/pulumi/pulumi-converter-terraform/pull/73), this gives me 100% core util, somehow I can't seem to get only one of them to fully load the cores.

This gives me 6.36total.

But. The converter processes are RAM-hungry. I've seen both peak at 1.2GB RAM which is a little surprising. IN CI contexts we don't have that much RAM.

Let me try one last time. Run just one converter process, see where this gets us.

t0yv0:
```
make tfgen  3479.98s user 82.55s system 883% cpu 6:43.23 total

```

Ok this is the way to go. Least RAM. One big converter process, 1.3GB, parallelize inside it. Closing this PR.

PR #1442, opened 2023-10-16
Downstream check replacement

t0yv0:
Implements a replacement for downstream checks.

Requires https://github.com/pulumi/ci-mgmt/pull/691

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1166

Testing. Starting from this branch at commit e787c3.

I started this:
https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/6539012605

Which started this:
https://github.com/pulumi/pulumi-aiven/actions/runs/6539014556/job/17756157909

Which opened this: https://github.com/pulumi/pulumi-aiven/pull/401

### DIFF
--- a/.github/workflows/update-providers-test.yml
+++ b/.github/workflows/update-providers-test.yml
@@ -7,16 +7,16 @@ on:
   workflow_dispatch:
     inputs: {}
 
-  # Alternatively, they can be triggered by opening a feature-* branch. Not having these enabled on
-  # normal branches and PRs is intentional as the checks consume a lot of resources.
-  push:
-    branches:
-    - feature-**
-    paths-ignore:
-    - CHANGELOG.md
-    tags-ignore:
-    - "*"
-    - "**"
+  # # Alternatively, they can be triggered by opening a feature-* branch. Not having these enabled on
+  # # normal branches and PRs is intentional as the checks consume a lot of resources.
+  # push:
+  #   branches:
+  #   - feature-**
+  #   paths-ignore:
+  #   - CHANGELOG.md
+  #   tags-ignore:
+  #   - "*"
+  #   - "**"
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.github/workflows/update-providers-test.yml
```diff
@@ -0,0 +1,90 @@
...
+  # Alternatively, they can be triggered by opening a feature-* branch. Not having these enabled on
+  # normal branches and PRs is intentional as the checks consume a lot of resources.
+  push:
+    branches:
+    - feature-**
+    paths-ignore:
+    - CHANGELOG.md
+    tags-ignore:
+    - "*"
+    - "**"
```
iwahbe, line 19: It seems pretty aggressive to trigger on push, especially if we want to turn on the rest of the providers. Should we restrict it to manual triggers?

t0yv0: It's limited to branches called `feature-*`.

I think I've sent these used elsewhere in Pulumi.

So if we're building a feature that requires extra scrutiny, we start a feature-xyz branch and PRs based on that branch. When it merges it runs all the checks and reports back. That was the idea.

I'm happy to start with just manual triggers though, the policy here is up for debate.

t0yv0: One thing to note is that there's no built-in monitoring currently. When you trigger this you don't know where to look for results necessarily. The feature-* branch has this slightly curious solution where PR descriptions cross-link back to the originating PR, that's serve as a poor-man's monitoring. I'd be curious if the works for manual triggers off PR or not. Can try that out

iwahbe: We can try this for now, but I don't think an on-push trigger will scale to 50 PRs.

iwahbe (approved):

PR #1455, opened 2023-10-18
Simplify AutoAliasing datasource check

iwahbe:
This change correctly modifies `prev.Current` when we detect a mismatch between `prev.Current` and the actively set value. This improves correctness.

### DIFF
--- a/pkg/tfbridge/tokens_test.go
+++ b/pkg/tfbridge/tokens_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata"
 	md "github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata"
+	ptokens "github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 )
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/tokens_test.go
```diff
@@ -1022,6 +1022,66 @@ func TestMaxItemsOneDataSourceAliasing(t *testing.T) {
...
+                        {
+                            "name": "pkg:index:getD1",
+                            "inCodegen": false,
+                            "majorVersion": 1
+                        }
+                    ]
+                }
+            }
+        }
+    }`
```
mjeffryes, line 1074: It seems like we should have a few more test cases, that capture what happens when there's more in the "past". Eg. I'm particularly interested in what would happen if the Tok changed back to `pkg:index:getD1`.

mjeffryes:
Side note, we keep getting that same coverage delta (-4.06%); I suspect this is a product of running slightly different tests in master vs in the acceptance tests.

mjeffryes (approved):

pkg/tfbridge/tokens_test.go
iwahbe: Previously, this would result in multiple copies. I have added pruning, and a test case that addresses this.

PR #1459, opened 2023-10-21
Pin latest TF SDK

iwahbe:
We have providers using this version of the TF SDK already. We should test against it.


Makefile
```diff
@@ -41,13 +41,18 @@ generate_builtins_test::
...
-	cd pf/tests && go mod tidy
-	cd pf/tests/internal/randomshim && go mod tidy
-	cd pf/tests/internal/tlsshim && go mod tidy
-	cd pf/tests/testdatagen/genrandom && go mod tidy
-	cd pkg/tests && go mod tidy
-	cd x/muxer && go mod tidy
-	cd x/muxer/tests && go mod tidy
-	cd testing && go mod tidy
-	go mod tidy
+	find . -name go.mod -execdir go mod tidy \;
```
t0yv0, line 44: Nice, good shel-fu . I should stop writing Go programs for something of this nature.

t0yv0:
Similar to https://github.com/pulumi/pulumi-terraform-bridge/issues/918 we should double-check the upstream release notes.

t0yv0 (approved):

iwahbe:
The only thing I see is https://github.com/pulumi/pulumi-terraform-bridge/issues/1465, which is low priority.

PR #1471, opened 2023-10-25
Fix release automation

t0yv0:
Editing the GitHub actions workflow to propagate new bridge releases to all providers.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1470
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1474


t0yv0:
Example trigger receipt: https://github.com/pulumi/pulumi-aiven/actions/runs/6634481479/job/18023972713

t0yv0:
Example PR: https://github.com/pulumi/pulumi-mailgun/pull/208

iwahbe (approved):
This is fine as it stands, but I would prefer that we pull the matrix from `ci-mgmt`. Ideally, we wouldn't maintain multiple lists of all our providers.

.github/workflows/update-providers-auto.yml
```diff
@@ -0,0 +1,79 @@
...
+          # Not specifying target-bridge-version in the payload will make it upgrade to the latest.
+          client-payload: |-
+            {
+               "pr-reviewers": "t0yv0",
+               "automerge": true
+            }
+    strategy:
+      fail-fast: false
+      matrix:
+        provider:
```
iwahbe, line 32: This matrix is missing pulumi-oci. I really think we should draw this list from ci-mgmt.

t0yv0:
You are right. Scripted drawing from ci-mgmt. Test-driving.

PR #1475, opened 2023-10-25
RFC: record aliases in provider

mjeffryes:
This needs some cleaning still, but outlines a possible approach for recording the aliases in the provider shim so we can fully pack and unpack the provider info. 

I've tested this locally using the aws provider (https://github.com/pulumi/pulumi-aws/pull/2902). It appears to be working, and to offer a fairly significant speed up (>150ms in my local testing). 

Feedback that would be most helpful:
1) ideas for how to test
2) Any suggestions for how to inject the aliasing with a lighter touch

### DIFF
--- a/pf/internal/muxer/muxer.go
+++ b/pf/internal/muxer/muxer.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 	shimSchema "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/schema"
+	shimUtil "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/util"
 	"github.com/pulumi/pulumi-terraform-bridge/x/muxer"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
@@ -128,8 +129,8 @@ func (m *ProviderShim) extend(provider shim.Provider) ([]string, []string) {
 
 	data, conflictingDataSources := union(m.dataSources, provider.DataSourcesMap())
 
-	m.resources = res
-	m.dataSources = data
+	m.resources = shimUtil.NewAliasingResourceMap(res)
+	m.dataSources = shimUtil.NewAliasingResourceMap(data)
 	m.MuxedProviders = append(m.MuxedProviders, provider)
 	return conflictingResources, conflictingDataSources
 }
@@ -138,8 +139,8 @@ func newProviderShim(provider shim.Provider) ProviderShim {
 	return ProviderShim{
 		simpleSchemaProvider: simpleSchemaProvider{
 			schema:      provider.Schema(),
-			resources:   provider.ResourcesMap(),
-			dataSources: provider.DataSourcesMap(),
+			resources:   shimUtil.NewAliasingResourceMap(provider.ResourcesMap()),
+			dataSources: shimUtil.NewAliasingResourceMap(provider.DataSourcesMap()),
 		},
 		MuxedProviders: []shim.Provider{provider},
 	}
@@ -238,8 +239,8 @@ func resolveDispatchMap[T interface{ GetTok() tokens.Token }](
 type simpleSchemaProvider struct {
 	schemashim.SchemaOnlyProvider
 	schema      shim.SchemaMap
-	resources   shim.ResourceMap
-	dataSources shim.ResourceMap
+	resources   shimUtil.AliasingResourceMap
+	dataSources shimUtil.AliasingResourceMap
 }
 
 func (p *simpleSchemaProvider) Schema() shim.SchemaMap {
@@ -297,4 +298,10 @@ func copyMap[K comparable, V any](m map[K]V) map[K]V {
 		out[k] = v
 	}
 	return out
+}
+
+type AliasingProviderShim struct {
+	ProviderShim
+	resources   shimUtil.AliasingResourceMap
+	datasources shimUtil.AliasingResourceMap
 }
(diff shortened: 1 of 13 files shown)


pf/internal/muxer/muxer.go
```diff
@@ -238,8 +239,8 @@ func resolveDispatchMap[T interface{ GetTok() tokens.Token }](
 type simpleSchemaProvider struct {
 	schemashim.SchemaOnlyProvider
 	schema      shim.SchemaMap
-	resources   shim.ResourceMap
-	dataSources shim.ResourceMap
+	resources   shimUtil.AliasingResourceMap
+	dataSources shimUtil.AliasingResourceMap
```
mjeffryes, line 243: This was the most expedient way to make the aliases reachable for aws.

pf/internal/muxer/muxer.go
```diff
@@ -298,3 +299,9 @@ func copyMap[K comparable, V any](m map[K]V) map[K]V {
 	}
 	return out
 }
+
+type AliasingProviderShim struct {
+	ProviderShim
+	resources   shimUtil.AliasingResourceMap
+	datasources shimUtil.AliasingResourceMap
+}
```
mjeffryes, line 307: The idea here was to introduce a wrapper for `muxer.ProviderShim`, just as I did  for the base `shim.Provider`, but I took a short-cut and just modified ProviderShim directly for the POC

pf/internal/schemashim/resource_map.go
```diff
@@ -55,3 +55,7 @@ func (m *schemaOnlyResourceMap) Range(each func(key string, value shim.Resource)
 func (*schemaOnlyResourceMap) Set(key string, value shim.Resource) {
 	panic("Set not supported - is it possible to treat this as immutable?")
 }
```
mjeffryes, line 57: This experiment has given me pretty good confidence that we're only using `Set` to add aliases today (Although since it's a public interface, it's possible for random providers to be calling it somewhere. It would be good to remove it in a future major release of the bridge.)

pf/tfbridge/packed_provider_info.go
```diff
@@ -0,0 +1,70 @@
...
+	var datasources shimUtil.AliasingResourceMap
+	if p, ok := info.P.(*shimUtil.AliasingProvider); ok {
+		resources = p.ResourceMap
+		datasources = p.DataSourceMap
+	} else if p, ok := info.P.(*pfmuxer.ProviderShim); ok {
+		resources = p.ResourcesMap().(shimUtil.AliasingResourceMap)
+		datasources = p.DataSourcesMap().(shimUtil.AliasingResourceMap)
+	} else {
+		panic("can't find aliases from this provider")
+	}
```
mjeffryes, line 43: This is ugly, but lets us discover aliases without adding a `RangeAliases` method to the base `shim.Provider`

pf/tfbridge/packed_provider_info.go
```diff
@@ -0,0 +1,70 @@
...
+	jsoni := jsoniter.ConfigCompatibleWithStandardLibrary
+	return jsoni.Marshal(marshalled)
+}
+
+type WithProviderAliases struct {
+	Info              tfbridge.MarshallableProviderInfo `json:"info"`
+	ResourceAliases   map[string]string                 `json:"resource_aliases,omitempty"`
+	DatasourceAliases map[string]string                 `json:"datasource_aliases,omitempty"`
+}
```
mjeffryes, line 70: as discussed on https://github.com/pulumi/pulumi-aws/pull/2902, a further improvement would be to just codegen the serialized form so the runtime can just load a bunch of maps directly into memory.

pkg/tfbridge/provider.go
```diff
@@ -1237,7 +1237,7 @@ func (p *ProviderInfo) RenameResourceWithAlias(resourceName string, legacyTok to
 			currentInfo.Tok.Name().String()))
 	p.Resources[resourceName] = &currentInfo
 	p.Resources[legacyResourceName] = &legacyInfo
-	p.P.ResourcesMap().Set(legacyResourceName, p.P.ResourcesMap().Get(resourceName))
+	p.P.ResourcesMap().AddAlias(legacyResourceName, resourceName)
```
mjeffryes, line 1240: This and the change below seem to eliminate the runtime use of Set on the bridge

mjeffryes:
Ian pointed out that this is probably a fruitless direction to pursue because there are too many non-serializable hooks in the ResourceInfo that are needed at runtime. We will have to take a more piecemeal approach.

pkg/tfbridge/provider.go
t0yv0: Can we save this observation in a ticket? That'd be a good change to deprecate Set - except for this use case, which can be better named.

t0yv0:
You're so close @mjeffryes 

I think a lot of this PR would still be needed. 

It's an excellent observation that ResourceInfos have non-serializable things like Transformers, I apologize for missing that crucial bit in earlier design discussions. So we cannot turn them around through a serde boundary.

However as you're showing here it's entirely viable to capture side-effects at tfgen time, serialize them, and replay these side-effects at init-time to modify ResourceMap and ResourceInfos.

In case of ResourceMap, you've done most of the work here - it's AddAlias. Now it has to spy on these calls and do a `type AddAlias struct { alias string,  target string }` trick to save aliasing at tfgen and replay at init-time.

Now, we need to also capture side-effects performed against ResourceInfo & friends.

It seems like we could try refactoring aliasing code to trace the effects.

Note the comment:
 
// Specifically, [ApplyAutoAliases] may perform the following actions:
//
// - Call [ProviderInfo.RenameResourceWithAlias] or [ProviderInfo.RenameDataSource]
// - Edit [ResourceInfo.Aliases]
// - Edit [SchemaInfo.MaxItemsOne]

If we could turn these into a data structure.. Basically ApplyAutoAliases could be doing two things. At tfgen time it does what it currently does, plus spies on the above effects and records them; at init-time it simply replays pre-recorded effects.

mjeffryes:
> I think a lot of this PR would still be needed.
I do think there is some value in this code just by virtue of working toward shrinking the mutable surface of `shim.ResourceMap`, I would be happy to refine this for that purpose if there's interest.

However, I think that it's not likely to be useful for the init latency problem. 


>If we could turn these into a data structure.. Basically ApplyAutoAliases could be doing two things. At tfgen time it does what it currently does, plus spies on the above effects and records them; at init-time it simply replays pre-recorded effects.

Yes, Ian and I came to the same conclusion. In fact, the first two steps (the actual aliases are already captured as a list of effects and then run in a second pass in `ApplyAutoAliases`, so we just need to make these data rather than lambdas and convert the MaxItemsOne transformations to use a similar approach. (In fact Ian pointed out that just switching MaxItemsOne to a two pass approach will save a lot of GC pressure from creating empty maps and discarding them.)

PR #1493, opened 2023-10-28
Fix `muxer.Attach`

iwahbe:
The muxer's attach method is broken when setup with a nil host. It panics with a nil pointer exception. This replaces the previous implementation with a straight dispatch call to subsidiary servers.

### DIFF
--- a/x/muxer/muxer.go
+++ b/x/muxer/muxer.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"sync"
 
 	"github.com/golang/glog"
@@ -60,7 +61,7 @@ type server = rpc.ResourceProviderServer
 type muxer struct {
 	rpc.UnimplementedResourceProviderServer
 
-	host *provider.HostClient
+	host hostClient
 
 	dispatchTable dispatchTable
 
@@ -69,6 +70,12 @@ type muxer struct {
 	servers []server
 
 	getMappingByKey map[string]MultiMappingHandler
+}
+
+// An interface to make *provider.HostClient test-able.
+type hostClient interface {
+	io.Closer
+	Log(context.Context, diag.Severity, urn.URN, string) error
 }
 
 type GetMappingArgs interface {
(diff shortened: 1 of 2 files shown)


x/muxer/muxer.go
```diff
@@ -394,19 +395,16 @@ func (m *muxer) GetPluginInfo(ctx context.Context, e *emptypb.Empty) (*rpc.Plugi
...
+	for i, s := range m.servers {
+		s := s
+		attach[i] = func() error {
+			_, err := s.Attach(ctx, req)
+			return err
 		}
 	}
-	// Here we override the underlying host. This should replace the host instance of
-	// each subsidiary provider.
-	*m.host = *host
```
mjeffryes, line 408: does it matter that we're no longer setting `m.host`? Do we still need that field?

mjeffryes (commented):
So it looks like before this change we were not muxing the Attach call, but now we are? The new behavior and tests look good, but I don't think I have full context to determine if the `NewHostClient` bits are safe to remove. (I did note that the one other place the `host` field is uses is in `WarnF` around line 513.)

x/muxer/muxer.go
iwahbe: Good catch. I believe that it does matter at the margin. Fixed in the next commit.

x/muxer/muxer.go
```diff
@@ -394,19 +395,18 @@ func (m *muxer) GetPluginInfo(ctx context.Context, e *emptypb.Empty) (*rpc.Plugi
 }
 
 func (m *muxer) Attach(ctx context.Context, req *rpc.PluginAttach) (*emptypb.Empty, error) {
-	host, err := provider.NewHostClient(req.GetAddress())
-	if err != nil {
-		return nil, err
-	}
-	if m.host != nil {
-		if err := m.host.Close(); err != nil {
-			return nil, err
```
mjeffryes, line 403: Do we need to close the old host if it exists?

mjeffryes (approved):

PR #1498, opened 2023-10-31
Fix adding empty values to maps

t0yv0:
TBD

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -189,7 +189,17 @@ func makePropertyDiff(ctx context.Context, name, path string, v resource.Propert
 				name += ".%"
 			}
 		}
-		if d := tfDiff.Attribute(name); d != nil && d.Old != d.New {
+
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/diff.go
```diff
@@ -189,7 +189,17 @@ func makePropertyDiff(ctx context.Context, name, path string, v resource.Propert
 				name += ".%"
 			}
 		}
-		if d := tfDiff.Attribute(name); d != nil && d.Old != d.New {
+
+		if d := tfDiff.Attribute(name); d != nil &&
```
t0yv0, line 193: Instead of going with this, I could consider also detecting changes to map counts. The diff has an entry for `tags.%` that's changing, indicating the number of tags is changing, but is being ignored by makeDetailedDiff. Speaking of which this needs test cases also for changing a non-empty entry to empty and vise versa, and for removing an empty entry, the whole CRUD.

t0yv0:
Closing in favor of #1502

Issue #1491, opened 2023-10-27
tfbridge/diff.go does not consider terraform diff


guineveresaenger:
While investigating behavior around `defaultLabels` for GCP, we discovered that some updates did not correctly propagate to resources.

Looking into this with @iwahbe, we narrowed it down to a Diff bug where we look at all of the Pulumi-supplied `olds` and `news` resource property maps, but in our example case, the diff _only_ exists on the terraform diff. In [`makeDetailedDiff](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/diff.go#L277) we do not iterate over the Terraform diff at all, losing update information that should be passed to the bridge.

For the purposes of GCP Labels, we should be able to add logic to `makeDetailedDiff` that iterates over the TF diff Attributes, compares them to what exists in the Pulumi diff, and update the Pulumi diff accordingly.

For the future, we may want to alter our approach to detecting diffs against the terraform diff to begin with, as https://github.com/pulumi/pulumi-gcp/issues/1275 shows it is possible to miss important information.

PR #1499, opened 2023-10-31
Refer to Terraform diff to generate a more precise diff

guineveresaenger:
This pull request looks at the Terraform Diff Attributes, checks if they exist in the Pulumi diff generated by `olds` and `news`, and adds any additional diff information to Pulumi diff.

Points of concern: 
- This change is not yet under test; suggestions or help modeling a schema map and schema info are welcome.
- How detailed we can make the property diff on each schema path? There's room for improvement here.

Addresses #1491.

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -313,7 +313,7 @@ func makeDetailedDiff(
 	// We want to make sure that we do not ignore these changes, if any.
 	// In order to look up whether our attributes have been added to the Pulumi `diff`, we first need to parse the
 	// Terraform diff paths (of the format `foo.0.bars.1.baz[...]`) to Pulumi (of the format `foo[0].bar[1].baz[...]`.
-	for path, _ := range tfDiff.Attributes() {
+	for path := range tfDiff.Attributes() {
 
 		pulumiPath, specificDiff := parseTFPath(path, tfs, ps)
 		// Look up the resulting path and register any diffs if not present in Pulumi diff.
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -309,9 +309,76 @@ func makeDetailedDiff(
 		makePropertyDiff(ctx, en, string(k), v, tfDiff, diff, forceDiff, etf, eps, true, useRawNames(etf))
 	}
 
+	// In some cases, the terraform diff will have additional attributes not covered by Pulumi's olds and news.
+	// We want to make sure that we do not ignore these changes, if any.
+	// In order to look up whether our attributes have been added to the Pulumi `diff`, we first need to parse the
+	// Terraform diff paths (of the format `foo.0.bars.1.baz[...]`) to Pulumi (of the format `foo[0].bar[1].baz[...]`.
+	for path, _ := range tfDiff.Attributes() {
+
+		pulumiPath, specificDiff := parseTFPath(path, tfs, ps)
+		// Look up the resulting path and register any diffs if not present in Pulumi diff.
+		if _, ok := diff[pulumiPath]; !ok {
+			if specificDiff {
+				// TODO: Using PropertyDiff_ADD is not entirely accurate in every case.
+				// TODO: (cont) We may want to pass on a more generic UPDATE strategy here - less precise, less error prone.
+				diff[pulumiPath] = &pulumirpc.PropertyDiff{Kind: pulumirpc.PropertyDiff_ADD}
+			} else {
+				diff[pulumiPath] = &pulumirpc.PropertyDiff{Kind: pulumirpc.PropertyDiff_UPDATE}
+			}
+		}
+	}
 	changes := pulumirpc.DiffResponse_DIFF_NONE
 	if len(diff) > 0 || *forceDiff {
 		changes = pulumirpc.DiffResponse_DIFF_SOME
 	}
 	return diff, changes
+}
+
+func parseTFPath(path string, tfs shim.SchemaMap, ps map[string]*SchemaInfo) (string, bool) {
+	var pulumiPath string
+	specificDiff := false
+	pathFields := strings.Split(path, ".")
+	for _, field := range pathFields {
+		// check if the field is a schema key
+		nestedSchema, found := tfs.GetOk(field)
+		if found {
+			pulumiField := TerraformToPulumiNameV2(field, tfs, ps)
+			// fencepost the first entry in the path
+			if pulumiPath == "" {
+				pulumiPath = pulumiPath + pulumiField
+			} else {
+				pulumiPath = pulumiPath + "." + pulumiField
+			}
+			// reset the schema map to point at the nested schema
+			if nestedSchema != nil {
+				if res, isres := nestedSchema.Elem().(shim.Resource); isres {
+					tfs = res.Schema()
+				}
+			}
+		} else {
+			// check if the field is a number
+			// the error value for strconv.Atoi is 0, so we look at the error.
+			_, err := strconv.Atoi(field) // the error value here is 0, so we look at the error.
+			if err == nil {
+				// we found an index
+				pulumiPath = pulumiPath + "[" + field + "]"
+			} else {
+				// Ignore special characters from Terraform
+				if field != "%" || field != "#" {
+					// These characters denote, in terraform, a schema object (in the case of "%"),
+					// or a list (in the case of "#").
+					// We drop the special character; this case is a no-op.
+					// These chars are also always final.
+					continue
+				} else {
+					// we do not have a number but a user set obj key.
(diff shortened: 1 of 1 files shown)


pkg/tfbridge/diff.go
```diff
@@ -309,9 +309,71 @@ func makeDetailedDiff(
...
+			pulumiField := TerraformToPulumiNameV2(field, tfs, ps)
+			// fencepost the first entry in the path
+			if pulumiPath == "" {
+				pulumiPath = pulumiPath + pulumiField
+			} else {
+				pulumiPath = pulumiPath + "." + pulumiField
+			}
+			// reset the schema map to point at the nested schema
+			if nestedSchema != nil {
+				if res, isres := nestedSchema.Elem().(shim.Resource); isres {
```
iwahbe, line 354: If `!isres`, then `tfs` should be set to `nil`. A misalignment doesn't mean that no traversal happened.

Also, it's not actually a resource:
```suggestion
				if res, isObj := nestedSchema.Elem().(shim.Resource); isObj {
```

pkg/tfbridge/diff.go
```diff
@@ -309,9 +309,71 @@ func makeDetailedDiff(
...
 }
+
+func parseTFPath(path string, tfs shim.SchemaMap, ps map[string]*SchemaInfo) (string, bool) {
+	var pulumiPath string
+	specificDiff := false
+	pathFields := strings.Split(path, ".")
+	for _, field := range pathFields {
+		// check if the field is a schema key
+		nestedSchema, found := tfs.GetOk(field)
+		if found {
```
iwahbe, line 344: Tightening the scoping makes it harder to misuse these variables.
```suggestion
		if nestedSchema, found := tfs.GetOk(field); found {
```

pkg/tfbridge/diff.go
```diff
@@ -309,9 +309,71 @@ func makeDetailedDiff(
...
+			}
+		} else {
+			// check if the field is a number
+			// the error value for strconv.Atoi is 0, so we look at the error.
+			_, err := strconv.Atoi(field) // the error value here is 0, so we look at the error.
+			if err == nil {
+				// we found an index
+				pulumiPath = pulumiPath + "[" + field + "]"
+			} else {
+				// Ignore special characters from Terraform
```
iwahbe, line 366: I think this would read clearer if we reduce the nesting.
```suggestion
			continue
		}
		
		// check if the field is a number
		// the error value for strconv.Atoi is 0, so we look at the error.
		_, err := strconv.Atoi(field) // the error value here is 0, so we look at the error.
		if err == nil {
			// we found an index
			pulumiPath = pulumiPath + "[" + field + "]"
			continue
		}
		
		// Ignore special characters from Terraform
```

pkg/tfbridge/diff.go
```diff
@@ -309,9 +309,71 @@ func makeDetailedDiff(
...
+	// In order to look up whether our attributes have been added to the Pulumi `diff`, we first need to parse the
+	// Terraform diff paths (of the format `foo.0.bars.1.baz[...]`) to Pulumi (of the format `foo[0].bar[1].baz[...]`.
+	for path := range tfDiff.Attributes() {
+
+		pulumiPath, specificDiff := parseTFPath(path, tfs, ps)
+		// Look up the resulting path and register any diffs if not present in Pulumi diff.
+		if _, ok := diff[pulumiPath]; !ok {
+			if specificDiff {
+				// TODO: Using PropertyDiff_ADD is not entirely accurate in every case.
+				// TODO: (cont) We may want to pass on a more generic UPDATE strategy here - less precise, less error prone.
```
iwahbe, line 323: Can you give an example for the case there `PropertyDiff_ADD` is the wrong case? What is the effect of getting the kind wrong (is it just display)?

iwahbe (commented):
This will need to be under test to merge. To start, I would write table-driven unit tests for `parseTFPath`. You can use `shim/schema.*` to mock `shim.Schema*` interfaces:

https://github.com/pulumi/pulumi-terraform-bridge/blob/b83370c2fe7eaf28a5d7741fc52dcd31d4abde49/pkg/tfshim/schema/schema.go#L140-L142

pkg/tfbridge/diff.go
guineveresaenger: It could be a PropertyDiff_UPDATE, on an already-existing object key.

t0yv0 (changes requested):
Thank you for looking into this! 

Making changes to this code is currently an extremely high-risk area, and existing coverage is very inadequate to verify changes. If you're contributing to the expectation is to have 100% coverage on changes but also back-filling missing coverage. 

While it is possible to continue with the chosen approach and get there, it is going to be require some work.

Can we discuss alternative approaches that avoid high-risk edits? Perhaps we can start with a repro of the original problem?

pkg/tfbridge/diff.go
```diff
@@ -309,9 +309,71 @@ func makeDetailedDiff(
...
+		}
+	}
 	changes := pulumirpc.DiffResponse_DIFF_NONE
 	if len(diff) > 0 || *forceDiff {
 		changes = pulumirpc.DiffResponse_DIFF_SOME
 	}
 	return diff, changes
 }
+
+func parseTFPath(path string, tfs shim.SchemaMap, ps map[string]*SchemaInfo) (string, bool) {
```
t0yv0, line 337: Do we understand the semantics of paths here, both the TF paths and Pulumi paths?

What happens with set elements? (represented as hashes in TF paths).  Should these be mapped to Pulumi numbers as steps are represented as lists?

What happens with flattening (when Pulumi paths are shorter than TF paths with MaxItems=1).

pkg/tfbridge/diff.go
```diff
@@ -309,9 +309,71 @@ func makeDetailedDiff(
...
+			if specificDiff {
+				// TODO: Using PropertyDiff_ADD is not entirely accurate in every case.
+				// TODO: (cont) We may want to pass on a more generic UPDATE strategy here - less precise, less error prone.
+				diff[pulumiPath] = &pulumirpc.PropertyDiff{Kind: pulumirpc.PropertyDiff_ADD}
+			} else {
+				diff[pulumiPath] = &pulumirpc.PropertyDiff{Kind: pulumirpc.PropertyDiff_UPDATE}
+			}
+		}
+	}
 	changes := pulumirpc.DiffResponse_DIFF_NONE
```
t0yv0, line 330: What would happen if we considered changing the `changes` only without populating `diff[pulumiPath]` at all? Is there a way such an approach could get us to passing the bug, possibly with incorrect detailed diff displayed to the user?

t0yv0:
Linking related issue https://github.com/pulumi/pulumi-terraform-bridge/pull/1498

pkg/tfbridge/diff.go
```diff
@@ -309,9 +309,71 @@ func makeDetailedDiff(
 		makePropertyDiff(ctx, en, string(k), v, tfDiff, diff, forceDiff, etf, eps, true, useRawNames(etf))
 	}
 
+	// In some cases, the terraform diff will have additional attributes not covered by Pulumi's olds and news.
```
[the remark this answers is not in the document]
guineveresaenger: Here is a tfDiff.Attributes() when adding the new GCP labels:

```tfDiff.Attributes()=map[string]shim.ResourceAttrDiff{
           "effective_labels.%": {
               Old:         "3",
               New:         "4",
               NewComputed: false,
               NewRemoved:  false,
               NewExtra:    nil,
               RequiresNew: false,
               Sensitive:   false,
               Type:        0x0,
           },
           "effective_labels.new": {
               Old:         "",
               New:         "defaultLabel",
               NewComputed: false,
               NewRemoved:  false,
               NewExtra:    nil,
               RequiresNew: false,
               Sensitive:   false,
               Type:        0x0,
           },
           "terraform_labels.%": {
               Old:         "3",
               New:         "4",
               NewComputed: false,
               NewRemoved:  false,
               NewExtra:    nil,
               RequiresNew: false,
               Sensitive:   false,
               Type:        0x0,
           },
           "terraform_labels.new": {
               Old:         "",
               New:         "defaultLabel",
               NewComputed: false,
               NewRemoved:  false,
               NewExtra:    nil,
               RequiresNew: false,
               Sensitive:   false,
               Type:        0x0,
           },
       }

```

EvanBoyle:
Should we close this given that we merged https://github.com/pulumi/pulumi-terraform-bridge/pull/1502?

t0yv0:
I have filed https://github.com/pulumi/pulumi-terraform-bridge/issues/1504 detailing remaining work to consider here, and why we should consider finishing this up. I will close for now. I think there's a substantial amount of work required to get this right; the code in this PR will still be linked to 1504 if we need to find it again.

PR #1500, opened 2023-11-01
Add missing docs override

iwahbe:
During provider updates, we occasionally have missing docs that we accept. We also set `PULUMI_MISSING_DOCS_ERROR=true` during updates so we are made aware. Right now, the solution is to set `tfbridge.DocInfo{Markdown: []byte(" ")}`, which prevents docs from ever showing up.

This PR adds a flag for this scenario, that mitigates the error for missing docs but doesn't preclude the doc showing up in the future.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -490,6 +490,11 @@ type DocInfo struct {
 	// this document will satisfy the criteria `docs/pulumiToken.md`
 	// The examples need to wrapped in the correct shortcodes
 	ReplaceExamplesSection bool
+
+	// Don't error when this doc is missing.
+	//
+	// This applies when PULUMI_MISSING_DOCS_ERROR="true".
+	AllowMissing bool
 }
 
 // GetImportDetails returns a string of import instructions defined in the Pulumi provider. Defaults to empty.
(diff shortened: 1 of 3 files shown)


mjeffryes (approved):
I wonder if we should have some indicator when does that are marked as `AllowMissing` are actually found so we can remove the flag and lock in the improvement.

iwahbe:
> I wonder if we should have some indicator when does that are marked as `AllowMissing` are actually found so we can remove the flag and lock in the improvement.

I'm not sure we want to "lock in" the docs, since it would require more manual interaction during upgrades. Can we punt that to ideation?

mjeffryes:
> > I wonder if we should have some indicator when does that are marked as `AllowMissing` are actually found so we can remove the flag and lock in the improvement.
> 
> I'm not sure we want to "lock in" the docs, since it would require more manual interaction during upgrades. Can we punt that to ideation?

We didn't discuss this in specifically in ideation, but I think we can apply the principals from there here. (ie. Create some sort of overall monitoring for docs quality and alert on that rather than on individual docs being present or not). It does seem like in the common case resources with docs should not lose them.

iwahbe:
> > > I wonder if we should have some indicator when does that are marked as `AllowMissing` are actually found so we can remove the flag and lock in the improvement.
> > 
> > 
> > I'm not sure we want to "lock in" the docs, since it would require more manual interaction during upgrades. Can we punt that to ideation?
> 
> We didn't discuss this in specifically in ideation, but I think we can apply the principals from there here. (ie. Create some sort of overall monitoring for docs quality and alert on that rather than on individual docs being present or not). It does seem like in the common case resources with docs should not lose them.

This PR won't exempt docs from reporting, and it is more of a stopgap until we get reporting in. I want to stop doing this: `tfbridge.DocInfo{Markdown: []byte(" ")}`.

PR #1502, opened 2023-11-02
Use TF Diff Attributes to set diff response changes when SkipDetailedDiffForChanges feature flag is set

### LINKED ISSUE #1491: tfbridge/diff.go does not consider terraform diff

guineveresaenger:
This PR presents an alternative fix for #1491.

Because `makeDetailedDiff` can distort the expected changes for a diff set via custom diffs in the upstream provider logic, we should not, in all cases, rely on it to provide us with an accurate `pulumirpc.DiffResponse.Changes` result. This PR does two things:

- Rely on the terraform Diff Attributes length to tell the engine that there are changes via the diff
- Gate this behavior behind a new feature flag, `SkipDetailedDiffForChanges`.

This code will continue to rely on `makeDetailedDiff` for setting the `DetailedDiff` field in all cases.

I have verified that this succeeds against [these test cases](https://github.com/pulumi/pulumi-gcp/pull/1310/files#diff-2e4fffe32720b3d310b4336cfdde6519ab1bf6d9e0d82b418299817fdf42a29fR191) when the GCP provider sets `SkipDetailedDiffForChanges` to True, and fails when set to False.

Fixes #1491 in the context of unblocking the GCP release.

### DIFF
--- a/internal/testprovider/schema.go
+++ b/internal/testprovider/schema.go
@@ -881,7 +881,7 @@ func AssertProvider(f func(data *schemav2.ResourceData)) *schemav2.Provider {
 	}
 }
 
-func AssertCustomizedDiffProvider(f func(data *schemav2.ResourceData)) *schemav2.Provider {
+func CustomizedDiffProvider(f func(data *schemav2.ResourceData)) *schemav2.Provider {
 	return &schemav2.Provider{
 		Schema: map[string]*schemav2.Schema{},
 		ResourcesMap: map[string]*schemav2.Resource{
@@ -890,44 +890,11 @@ func AssertCustomizedDiffProvider(f func(data *schemav2.ResourceData)) *schemav2
 					"labels": {Type: schemav2.TypeString, Optional: true, Computed: true},
 				},
 				SchemaVersion: 1,
-				MigrateState: func(v int, is *terraformv2.InstanceState, p interface{}) (*terraformv2.InstanceState, error) {
-					return is, nil
-				},
-				Create: func(data *schemav2.ResourceData, p interface{}) error {
-					data.SetId("0")
-					f(data)
-					return nil
-				},
-				Read: func(data *schemav2.ResourceData, p interface{}) error {
-					f(data)
-					return nil
-				},
-				Update: func(data *schemav2.ResourceData, p interface{}) error {
-					f(data)
-					return nil
-				},
-				Delete: func(data *schemav2.ResourceData, p interface{}) error {
-					f(data)
-					return nil
-				},
-				Timeouts: &schemav2.ResourceTimeout{
-					Create: Timeout(time.Second * 120),
-				},
-				Importer: &schemav2.ResourceImporter{
-					StateContext: func(_ context.Context, state *schemav2.ResourceData,
-						_ interface{}) ([]*schemav2.ResourceData, error) {
-
-						return []*schemav2.ResourceData{state}, nil
-					},
-				},
 				CustomizeDiff: func(ctx context.Context, diff *schemav2.ResourceDiff, i interface{}) error {
 					err := diff.SetNew("labels", "1")
 					return err
 				},
 			},
-		},
-		ConfigureFunc: func(data *schemav2.ResourceData) (interface{}, error) {
-			return nil, nil
 		},
 	}
 }
--- a/internal/testprovider/schema.go
+++ b/internal/testprovider/schema.go
@@ -880,3 +880,54 @@ func AssertProvider(f func(data *schemav2.ResourceData)) *schemav2.Provider {
 		},
 	}
 }
+
+func AssertCustomizedDiffProvider(f func(data *schemav2.ResourceData)) *schemav2.Provider {
+	return &schemav2.Provider{
+		Schema: map[string]*schemav2.Schema{},
+		ResourcesMap: map[string]*schemav2.Resource{
+			"test_resource": {
+				Schema: map[string]*schemav2.Schema{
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/provider.go
```diff
@@ -661,6 +661,15 @@ func (p *Provider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*pulum
 	doIgnoreChanges(ctx, res.TF.Schema(), res.Schema.Fields, olds, news, req.GetIgnoreChanges(), diff)
 	detailedDiff, changes := makeDetailedDiff(ctx, res.TF.Schema(), res.Schema.Fields, olds, news, diff)
 
+	// There are some providers/situations which `makeDetailedDiff` distorts the expected changes, leading
+	// to changes being dropped by Pulumi.
+	// Until we address https://github.com/pulumi/pulumi-terraform-bridge/issues/1501, it is safer to refer
```
t0yv0, line 666: This change literally addresses 1501

pkg/tfbridge/info.go
```diff
@@ -128,6 +128,11 @@ type ProviderInfo struct {
 	//
 	// See also: pulumi/pulumi-terraform-bridge#1448
 	SkipValidateProviderConfigForPluginFramework bool
+
+	// Disables using detailed diff to determine diff changes and falls back on the length of TF Diff Attributes.
+	//
+	// See https://github.com/pulumi/pulumi-terraform-bridge/issues/1501
+	SkipDetailedDiffForChanges bool
```
iwahbe, line 135: I don't think we want to support this feature until v4.0.0. When we close #1501 for real, we will want to remove this flag.
```suggestion
	// Disables using detailed diff to determine diff changes and falls back on the length of TF Diff Attributes.
	//
	// See https://github.com/pulumi/pulumi-terraform-bridge/issues/1501
	//
	// This flag is experimental and may be removed in a future bridge release.
	XSkipDetailedDiffForChanges bool
```

pkg/tfbridge/provider.go
```diff
@@ -661,6 +661,15 @@ func (p *Provider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*pulum
...
+	// to the Terraform Diff attribute length for setting the DiffResponse.
+	// We will still use `detailedDiff` for diff display purposes.
```
iwahbe, line 668: I really appreciate the comment explaining what is going on!

iwahbe (commented):
I'd like at least 1 synthetic test case to make sure this is hooked up (and to prevent regressions).

pkg/tfbridge/provider.go
guineveresaenger: It does; but I'm under the impression we'll want to keep it open so we can fix the bugs in `makeDetailedDiff` "for real".

I'll amend the working to reflect more nuance.

t0yv0: I don't think so, I'm happy to close. We can open new issues for makeDetailedDiff referencing the closed issue.

pkg/tfbridge/provider.go
```diff
@@ -661,6 +661,15 @@ func (p *Provider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*pulum
...
+	if p.info.SkipDetailedDiffForChanges && len(diff.Attributes()) > 0 {
+		changes = pulumirpc.DiffResponse_DIFF_SOME
```
t0yv0, line 670: Noting that it doesn't do the inverse, that is if len(diff.Attributes()) == 0 but makeDetailedDiff returned CHANGES, it does not override it back to NO_CHANGES. Is that perhaps a degenerate/ non-existent situation? Reading the code seems to be the case, so we're probably fine.

guineveresaenger: We have not verified such a situation as of today to the best of my knowledge.

t0yv0:
Also fixes AWS P1.. LGTM once tested, consider https://github.com/pulumi/pulumi-terraform-bridge/pull/1498 test cases I believe this change makes them pass, adding empty tags to map at least certainly should pass.

internal/testprovider/schema.go
```diff
@@ -880,3 +880,54 @@ func AssertProvider(f func(data *schemav2.ResourceData)) *schemav2.Provider {
 		},
 	}
 }
+
+func AssertCustomizedDiffProvider(f func(data *schemav2.ResourceData)) *schemav2.Provider {
```
iwahbe, line 884: Since we don't care about calling `f` here, we can just remove it.
```suggestion
func CustomizedDiffProvider() *schemav2.Provider {
```

iwahbe (approved):
1 nit on the test, but LGTM!

t0yv0 (approved):

PR #1503, opened 2023-11-02
Respect schema secrets in Check

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1380 

Similarly to Plugin Framework resources, SDKv2 based resources will now proactively mark properties as secrets in the results of Check method if the upstream schema says that these properties are sensitive or else if SchemaInfo in the bridged provider specifies the Secret true option.


iwahbe (approved):
Good catch. LGTM

I was worried that setting a value as secret will cause downstream consumers to register a diff, but it doesn't happen in the simple case.

It's probably worth calling out, since any user who checks if a property is secret will see a difference here.

PR #1509, opened 2023-11-07
fix docs builder for examples overriding use case

### LINKED ISSUE #1497: Bug: tfgen does not include descriptions for resources with ReplaceExamplesSection = true
### What happened?

in Equinix's provider, we are replacing the tf examples with the option
https://github.com/pulumi/pulumi-terraform-bridge/blob/b83370c2fe7eaf28a5d7741fc52dcd31d4abde49/pkg/tfbridge/info.go#L306

I noticed that now the resources don't have their description in the registry, just the examples. I don't know how the rendering works in the registry site, but checking other providers I see that the difference in `schema.json` is that in the resource `description` field, the description goes before the tag `{{% examples %}}`.

It seems this is happening because of https://github.com/pulumi/pulumi-terraform-bridge/blob/b83370c2fe7eaf28a5d7741fc52dcd31d4abde49/pkg/tfgen/docs.go#L1272-L1278

### Example

- Resource:
https://github.com/equinix/pulumi-equinix/blob/ba8962c48cd99f3ffa0c108a785506280cf92443/provider/resources.go#L137-L139

- Examples file:
https://github.com/equinix/pulumi-equinix/blob/ba8962c48cd99f3ffa0c108a785506280cf92443/docs/resource/equinix_fabric_connection.examples.md

- Generated schema:
https://github.com/equinix/pulumi-equinix/blob/ba8962c48cd99f3ffa0c108a785506280cf92443/provider/cmd/pulumi-resource-equinix/schema.json#L9851

- Registry:
https://www.pulumi.com/registry/packages/equinix/api-docs/fabric/connection/

Here's another example without a description in AWS classic provider:

- https://www.pulumi.com/registry/packages/aws/api-docs/acm/certificatevalidation/
- https://github.com/pulumi/pulumi-aws/blob/master/docs/resource/aws_acm_certificate_validation.examples.md

### Output of `pulumi about`

CLI          
Version      3.89.0
Go Version   go1.21.1
Go Compiler  gc

Host     
OS       darwin
Version  13.6
Arch     x86_64

### Additional context

If that's the case, I would expect something like

```
if strings.Contains(docs, "```typescript") || strings.Contains(docs, "```python") ||
		strings.Contains(docs, "```go") || strings.Contains(docs, "```yaml") ||
		strings.Contains(docs, "```csharp") || strings.Contains(docs, "```java") {
		// to reconvert them. But we need to surround them in the examples shortcode for rendering on the registry

		// Find the index of "## Example Usage"
		exampleIndex := strings.Index(docs, "## Example Usage")

		// Build the new description
		newDescription := fmt.Sprintf("%s\n\n{{%% examples %%}}\n%s\n{{%% /examples %%}}",
			docs[:exampleIndex],
			docs[exampleIndex:])

		return newDescription
	}
```

### Contributing

ocobles:
Fixes #1497 

separates the resource description from the examples section

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1285,7 +1285,7 @@ func (g *Generator) convertExamples(docs string, path examplePath, stripSubsecti
 
 		// Separate resource description and surround the examples
 		return fmt.Sprintf("%s\n\n{{%% examples %%}}\n%s\n{{%% /examples %%}}",
-			docs[:exampleIndex],
+			strings.TrimRightFunc(docs[:exampleIndex], unicode.IsSpace),
 			docs[exampleIndex:])
 	}
 
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1274,7 +1274,19 @@ func (g *Generator) convertExamples(docs string, path examplePath, stripSubsecti
 		strings.Contains(docs, "```csharp") || strings.Contains(docs, "```java") {
 		// we have explicitly rewritten these examples and need to just return them directly rather than trying
 		// to reconvert them. But we need to surround them in the examples shortcode for rendering on the registry
-		return fmt.Sprintf("{{%% examples %%}}\n%s\n{{%% /examples %%}}", docs)
+
+		// Find the index of "## Example Usage"
+		exampleIndex := strings.Index(docs, "## Example Usage")
+
+		// if not found surround all content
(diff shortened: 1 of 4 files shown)


iwahbe:
Hi @ocobles. Thanks for your PR! I'm running tests now.

iwahbe (approved):
This looks good to merge.

I added a test to lock in behavior here.

PR #1512, opened 2023-11-08
WIP Create a dynamic provider that can run any tf provider

mjeffryes:


pf/internal/pfutils/provider.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+)
+
+// Creates a pf provider that's backed by a protov6.ProviderServer
+func SchemaOnlyProvider(server *tfprotov6.ProviderServer) provider.Provider {
+	return schemaOnlyProvider{server: server}
+}
```
mjeffryes, line 15: This is one possible way of adapting to the existing pf/tbridge, but we could also just write a some new constructors on tfbridge/provider.go to work directly with the `*tfprotov6.ProviderServer`

PR #1520, opened 2023-11-14
Augment `makeDetailedDiff` when replace is specified

iwahbe:
This will fix (upon adoption) https://github.com/pulumi/pulumi-aws/issues/2971.

The design discussion for this PR is https://pulumi.slack.com/archives/C02FXTZEZ6W/p1699989102568939.

### Changes

If there is a replace on the underlying TF object (as indicated by `diff.RequiresNew() || diff.Destroy()`) but no associated replace in the wire return (`pulumirpc.DiffResponse.Replaces`), then we set `pulumirpc.DiffResponse.Replaces = pulumirpc.DiffResponse_DIFF_SOME` and insert `__meta` as needing a replace.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -718,6 +718,21 @@ func (p *Provider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*pulum
 	deleteBeforeReplace := len(replaces) > 0 &&
 		(res.Schema.DeleteBeforeReplace || nameRequiresDeleteBeforeReplace(news, olds, res.TF.Schema(), res.Schema))
 
+	// If the upstream diff object indicates a replace is necessary and we have not
+	// recorded any replaces, that means that `makeDetailedDiff` failed to translate a
+	// property. This is known to happen for computed input properties:
+	//
+	// https://github.com/pulumi/pulumi-aws/issues/2971
+	if (diff.RequiresNew() || diff.Destroy()) &&
+		// In theory, we should be safe to set __meta as replaces whenever
+		// `diff.RequiresNew() || diff.Destroy()` but by checking replaces we
+		// limit the blast radius of this change to diffs that we know will panic
+		// later on.
+		len(replaces) == 0 {
+		replaces = append(replaces, "__meta")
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/provider.go
```diff
@@ -977,8 +992,11 @@ func (p *Provider) Update(ctx context.Context, req *pulumirpc.UpdateRequest) (*p
 	// ignoring changes to the keys that would result in replacement/deletion.
 	doIgnoreChanges(ctx, res.TF.Schema(), res.Schema.Fields, olds, news, req.GetIgnoreChanges(), diff)
 
-	contract.Assertf(!diff.Destroy() && !diff.RequiresNew(),
-		"Expected diff to not require deletion or replacement during Update of %s", urn)
+	if diff.Destroy() || diff.RequiresNew() {
+		return nil, fmt.Errorf("internal: expected diff to not require deletion or replacement"+
+			" during Update of %s. Found delete=%t, replace=%t. This indicates a bug in provider.",
+			urn, diff.Destroy(), diff.RequiresNew())
```
iwahbe, line 998: We know this can happen, so we shouldn't panic here. Instead we can return a helpful error here.

pkg/tfbridge/provider_test.go
```diff
@@ -1876,6 +1879,65 @@ func TestSkipDetailedDiff(t *testing.T) {
...
+		  "method": "/pulumirpc.ResourceProvider/Diff",
+		  "request": {
+		    "id": "0",
+		    "urn": "urn:pulumi:dev::teststack::Replace::exres",
+		    "olds": {},
+		    "news": {}
+		  },
+		  "response": {
+		    "changes": "DIFF_SOME",
+		    "replaces": ["__meta"],
```
t0yv0, line 1903: Yes

t0yv0 (approved):

pkg/tfbridge/provider.go
t0yv0: Will it still happen after the change here?

iwahbe: It *shouldn't*, but the assert *shouldn't* have happened either.

PR #1521, opened 2023-11-14
Fix raw plan availability

t0yv0:
Making a surgical change to avoid nil pointer panics on accessing RawPlan, coming from the AWS panic in https://github.com/pulumi/pulumi-aws/issues/2904

I have verified that this resolves 2904.

### DIFF
--- a/pkg/tfshim/sdk-v2/provider_diff.go
+++ b/pkg/tfshim/sdk-v2/provider_diff.go
@@ -74,7 +74,7 @@ func (p v2Provider) simpleDiff(
 	c *terraform.ResourceConfig,
 	rawConfigVal hcty.Value,
 	meta interface{},
-) (*terraform.InstanceDiff, error) {
+) (finalInstanceDiff *terraform.InstanceDiff, finalError error) {
 	ctx := context.TODO()
 
 	switch diffStrat {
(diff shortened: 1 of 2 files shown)


pkg/tfshim/sdk-v2/provider_diff_test.go
```diff
@@ -0,0 +1,66 @@
...
+		resource: r,
+		tf:       instanceState,
+	}
+
+	id, err := wp.Diff("myres", ss, v2ResourceConfig{
+		tf: resourceConfig,
+	})
+	require.NoError(t, err)
+
+	require.False(t, id.(v2InstanceDiff).tf.RawPlan.IsNull(), "RawPlan should not be Null")
```
iwahbe, line 65: Nit
```suggestion
	assert.False(t, id.(v2InstanceDiff).tf.RawPlan.IsNull(), "RawPlan should not be Null")
```

pkg/tfshim/sdk-v2/provider_diff.go
```diff
@@ -165,7 +165,28 @@ func simpleDiffViaPlanState(
...
+	//
+	// See for example this code in PlanResourceChange:
+	//
+	// See https://github.com/hashicorp/terraform-plugin-sdk/blob/
+	//         28e631776d97f0a5a5942b3524814addbef90875/helper/schema/grpc_provider.go#L797
+	//
+	// In TF this is communicated from PlanResourceChange to ApplyResourceChange; unlike TF, in
+	// the current codebase InstanceDiff is passed directly to Apply. If RawPlan is not set on
+	// the diff it may cause nil panics in the provider.
+	if diff != nil && len(diff.Attributes) == 0 {
```
iwahbe, line 184: I don't understand why it is impossible that `diff.RawPlan` will hold some other useful value when `len(diff.Attributes) == 0`. That said this mirrors upstream, so I'm inclined to trust it.

iwahbe (approved):

pkg/tfshim/sdk-v2/provider_diff.go
t0yv0: Yup. I think it may be assuming that "when there are no changes", the plan is to keep things as -is, so PLAN = STATE. But it might be overriding a value in RawPlan that's non-nil sometimes. Hopefully that value was already equal to STATE.

PR #1522, opened 2023-11-14
Fix missing nil check on docInfo for doc discovery

guineveresaenger:
This change adds a nil check on docInfo for handling of missing docs settings, without which an upgrade in GCP panicked on creation of a new resource whose docs could not be found.

With this change, `getDocsForResource` behaves in the following way:

1. Check if `PULUMI_MISSING_DOCS_ERROR` is set to true - if no, continue with a warning
2. if yes, verify the resource has any docs info - if no, return a docs error as per env var setting
3. if yes, there is docs info, check if `AllowMissing` is set to true - if no, return a docs error as docs are missing and we expect an error as per env var setting
4. if yes, continue with a warning, as we allow docs to be missed.


iwahbe (approved):
It would be great to throw a quick test in. That said, this is pretty straight forward.

iwahbe:
Thanks!!!

PR #1534, opened 2023-11-20
Additional perf instrumentation

t0yv0:
IN https://github.com/pulumi/pulumi-aws/issues/2987 a customer is experiencing very long Configure and CheckConfig times but our traces and logs provide limited visibility into what's going on.

This PR adds a little more information. This is complicated by the fact that there's no root span for the provider and the provider process can be killed by the engine at any moment, so finalizing the root span is not reliable. Therefore the strategy of collecting mem stats in a background thread and tagging them to the root span, as in pulumi/pulumi does not work very well. 

Here're the suggested changes:

- add spans for key sdkv2 methods GetPluginInfo Configure, CheckConfig, Diff, Create, Update, Delete, Invoke
- all the spans sample the go memory stats for the running process and report them as tags on the span
- since we are investigating Configure and CheckConfig specifically, additional tags are added to show how much time is spent tin PreConfigureCallbacks and in provider configuration validation
- the latter is also instrumented for the PF version of Configure

With these changes we have a little more visibility into PF vs SDKv2 Configure costs and have some memory stat samples to get an idea of peak memory use for the provider.  This kind of hopes that sampling memory at the beginning of each span is sufficient; perhaps can be improved upon by sampling every 1s for long-running spans.


pf/tfbridge/provider_checkconfig.go
```diff
@@ -48,35 +57,22 @@ func (p *provider) CheckConfigWithContext(
 		ProviderConfig: inputs,
 	})
 
+	if !news.DeepEquals(inputs) {
```
iwahbe, line 60: This is not safe on arbitrary values, as the tests indicate.

This is why https://github.com/pulumi/pulumi-go-provider/blob/b60247b0e97434ac7be8b9a344deb4cbad524daa/infer/internal/ende/util.go#L95-L104 exists.

iwahbe (commented):
This looks good to merge once tests pass.

pf/tfbridge/provider_checkconfig.go
t0yv0: Ouch. Very well, good catch.

VenelinMartinov (approved):

PR #1537, opened 2023-11-21
Fix log interception to respect TF_LOG 

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1536

It turns out the situation is not as drastic as it originally seemed. A lot of TF code still emits logs into `log` package, they continue to be intercepted. For example, `"S3 bucket create: my-bucket-16-00fff33."` logs on bucket creation are showing through. We were however missing out on a few log messages coming from either the SDKv2 framework itself or specific AWS provider code. For examples see:

- https://github.com/hashicorp/terraform-provider-aws/blob/b5629785d0652fb19502460d56d7b39662fe22dc/internal/conns/config.go#L170
- https://github.com/hashicorp/terraform-provider-aws/blob/b5629785d0652fb19502460d56d7b39662fe22dc/internal/service/ec2/vpc_security_group_ingress_rule.go#L208

The log interceptor code is moving from an internal package in the pf module to the root module so it can be shared across tfbridge proper and SDKv2.

Instances of context.TODO() in SDKv2 providers are eliminated in favor of propagating the right Context object which is crucial for sharing Logger objects as payload values on Context.

Subsystem adapters are added to capture errors and warnings emitted by SDKv2 framework.

Fixed a bug in filtering where TF_LOG=TRACE would actually filter out TRACE level, and propagate DEBUG level only.


unstable/logging/logging.go
```diff
@@ -106,13 +179,177 @@ func (l *Host[L]) f(severity diag.Severity, msg string) {
...
+	ctx = context.WithValue(ctx, sdkOptionsKey, sdkLoggerOptions)
+	providerLoggerOptions := makeLoggerOptions("provider", desiredLevel, output)
+	ctx = context.WithValue(ctx, providerKey, hclog.New(providerLoggerOptions))
+	ctx = context.WithValue(ctx, providerOptionsKey, providerLoggerOptions)
+	return ctx
+}
+
+// Choose the default level carefully: logs at this level or higher (more severe) will be shown to the user of Pulumi
+// CLI directly by default. Experimentally it seems that Info is too verbose, for example Cloudflare provider emits
+// routine authentication messages at INFO level.
```
iwahbe, line 211: What does `terraform` do here? It feels like we should mirror what they do.

unstable/logging/logging.go
```diff
@@ -106,13 +179,177 @@ func (l *Host[L]) f(severity diag.Severity, msg string) {
...
+	if level == hclog.NoLevel {
+		level = defaultTFLogLevel()
+	}
+	return &hclog.LoggerOptions{
+		Name:              name,
+		Output:            output,
+		Level:             level,
+		IndependentLevels: true,
+		IncludeLocation:   true,
+		TimeFormat:        " ",
```
iwahbe, line 226:
```suggestion
		TimeFormat:        " ", // Do not print time
```

unstable/logging/logging.go
```diff
@@ -106,13 +179,177 @@ func (l *Host[L]) f(severity diag.Severity, msg string) {
...
+	err = w.sink.Log(w.ctx, severity, urn, raw)
+	return
+}
+
+func parseTfLogEnvVar() hclog.Level {
+	env, present := os.LookupEnv(tfLogEnvVar)
+	if !present {
+		return hclog.NoLevel
+	}
+	switch strings.ToUpper(env) {
```
iwahbe, line 277: We don't actually care if `present` is true, so I think we can simplify this block.

```suggestion
	switch strings.ToUpper(os.Getenv(tfLogEnvVar)) {
```

pkg/tfshim/shim.go
```diff
@@ -207,3 +208,73 @@ type Provider interface {
 	NewResourceConfig(object map[string]interface{}) ResourceConfig
 	IsSet(v interface{}) ([]interface{}, bool)
 }
+
+type ProviderWithContext interface {
```
iwahbe, line 212: I'm a little fuzzy on the goal of `ProviderWithContext`. This is intended as a backwards compatibility bridge within the repo?

Is the desired end state that everybody implements `ProviderWithContext` and then `Provider` methods are never called?

Would it be more correct to change the definition of `Provider` to have methods that take `context.Context` as their first arguments?

iwahbe (commented):
This change makes sense but it's pretty hard to review. It seems like we are making 3 changes:

- Adding `ProviderWithContext`
- Moving `logging`
- Adding subsystems to logging.

This would be much easier to review if these were their own stacked PRs or commits.

> Fixed a bug in filtering where TF_LOG=TRACE would actually filter out TRACE level, and propagate DEBUG level only.

I don't know where this is. Is there an associated issue?

t0yv0:
> This change makes sense but it's pretty hard to review

Let's try this one more time. I've edited history to group related changes into commits together.

t0yv0:
> I don't know where this is. Is there an associated issue?

The last commit has the issue and the tests. I noticed that setting TF_LOG=TRACE actually excludes the TRACE-level logs, consequently setting TF_LOG=WARN excludes WARN level logs, which is likely unexpected.

unstable/logging/logging.go
t0yv0: This is an interesting question. I'd call it out of scope for this PR, note that this code is NOT net new but it's moving around the code that used to do exactly this for Plugin Framework and installing it in the SDKv2 context. I'd also say we're not bound by what TF does necessarily, though it can inform it. 

There's also https://github.com/pulumi/pulumi-terraform-bridge/issues/97 that has some different ideas of tying things to the debug flag.  

Actually coming to think of it, because of https://github.com/pulumi/pulumi-terraform-bridge/pull/1537/commits/dc75cc06ebd4f0cd8d3a1ac6e81e65bc10169a91 there will be an observable change here that WARNING level logs will now start showing up that didn't show up previously, even in PF case that's not suppose to be changing. I'm open to tightening the default level to ERROR so that doesn't happen. It's quite possible that if we do not do this, we'll receive requests to close off un-actionable TF warning logs  right after we ship this. Thoughts?

pkg/tfshim/shim.go
t0yv0: Yea this is a very mechanical refactor that avoids making the change breaking. 

> Would it be more correct to change the definition of Provider to have methods that take context.Context as their first arguments?

That'd be our desired end state. I'm curious if we can just take that change without breaking any dependent code (that is the shim.Provider is fully internal interface although it's exported). That might be the case, but I'm not 100% sure. Doing so would also fix context propagation issues for SDKv1 providers if any (which seemingly is still in production). Though I don't know if there'd be any end-user benefit.

This should probably be a follow-up change though. I've avoided going this way in this PR to avoid some additional yak shaving necessary to get SDKv2, protov5 (unused in prod but hanging around in the repo), and test providers line up with the change.

unstable/logging/logging.go
t0yv0: Slight preference for leaving it explicit. One day perhaps deploy static analysis that prohibits os.Getenv.

PR #1542, opened 2023-11-27
Make assert actionable

iwahbe:
I was triggering this assert as part of upgrading pulumi-alicloud. The message should include enough information to debug the issue, which means at least the *module name* and duplicate member name.

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -273,7 +273,8 @@ func (m *module) config() bool {
 func (m *module) addMember(member moduleMember) {
 	name := member.Name()
 	for _, existing := range m.members {
-		contract.Assertf(existing.Name() != name, "unexpected duplicate module member %s", name)
+		contract.Assertf(existing.Name() != name, "unexpected duplicate module member %q in %q",
+			name, m.name.String())
 	}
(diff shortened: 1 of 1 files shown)


pkg/tfgen/generate.go
```diff
@@ -273,7 +273,8 @@ func (m *module) config() bool {
 func (m *module) addMember(member moduleMember) {
 	name := member.Name()
 	for _, existing := range m.members {
-		contract.Assertf(existing.Name() != name, "unexpected duplicate module member %s", name)
+		contract.Assertf(existing.Name() != name, "unexpected duplicate module member %q in %q",
+			name, m.name.String())
```
t0yv0, line 277: So log as m.name.String() never panics (e.g. m.name == nil)  LGTM.

t0yv0 (approved):

pkg/tfgen/generate.go
iwahbe: It doesn't. `m` cannot be `nil` (or L275 would panic) and `m.name` is `type Module Token` (`Token` is `type Token string`). `Module.String()` is `func (tok Module) String() string { return string(tok) }`.

PR #1544, opened 2023-11-28
Aliasing phase separation

t0yv0:
Rewrite automatic aliasing to allow easy phase separation between build-time and runtime logic. As much work as possible is done at build-time. This change is instrumental in getting startup overhead of larger providers down, in particular it's the key to get pulumi-aws from 600ms to 60ms on startup.

Application to pulumi-aws: https://github.com/pulumi/pulumi-aws/pull/3072


pkg/tfbridge/tokens_test.go
```diff
@@ -506,6 +510,7 @@ func testTokenAliasing(t *testing.T) {
 	}
 
 	simple := provider()
+	simple.Version = "1.0.0"
```
t0yv0, line 513: This test is weird, do we have a bug in the prod version of aliasing or am I missing something obvious? I think that without this addition, we're going from version 0 to version 1.0.0. In this case of a major version change the policy is to stop doing RenameResourceWithAlias and simply add to r.Aliases, and the new implementation does that but the asserts are not happy. Is version 0 special?

pkg/tfbridge/tokens_test.go
```diff
@@ -699,8 +755,13 @@ func TestMaxItemsOneAliasing(t *testing.T) {
 
 	autoAliasing(info, metadata)
 	assert.False(t, *info.Resources["pkg_r1"].Fields["f1"].MaxItemsOne)
-	assert.False(t, *info.Resources["pkg_r1"].Fields["f2"].MaxItemsOne)
-	assert.Equal(t, `{
+
+	f2Schema := info.P.ResourcesMap().Get("pkg_r1").Schema().Get("f2")
+	f2Overrides := info.Resources["pkg_r1"].Fields["f2"]
+
+	assert.False(t, tfbridge.IsMaxItemsOne(f2Schema, f2Overrides))
```
t0yv0, line 762: The new implementation does not inject MaxItemsOne override here because it's not necessary, as evidenced by the new assert, without the override the system still decides on False.

t0yv0:
Got existing tests passing locally with heavy use of autogold -update .. that might need careful review so it does not throw out implicit asserts.

Reminder to self, before landing this need to cover this branch:

```
	if !isTfgen() {
		autoSettings, err := loadAutoSettings(info.GetMetadata())
		if err != nil {
			return err
		}
		autoSettings.apply(info)
		return nil
	}
```

and make sure that the runtime results match what we get at gen time.

Next steps would be playing with pulumi-aws to see if this can result in startup speedups.

pkg/tfbridge/auto_aliasing.go
```diff
@@ -125,409 +281,422 @@ func (info *ProviderInfo) ApplyAutoAliases() error {
...
+		}
 
-	for tfToken, computed := range info.Resources {
-		r, ok := rMap.GetOk(tfToken)
-		if !ok {
-			continue
+		actual := IsMaxItemsOne(s, info)
+		prev := fh.lookup(sp)
+
+		// Preserve historical MaxItemsOne if it is available and disagrees with current.
```
t0yv0, line 308: Still thinking about rollout and correctness properties of this one..  It sounds like what we'd like to know from history, is the actual value (upstream+override) of MaxItemsOne, and that's it. It's currently has the *bool model but really we could be just storing the fields for which it's true, and assume it's false otherwise. 

That might not be the same semantic of what is stored there by the previous implementation. We'll need to check.. If that's the case we need some kind of migrate-on-read associated with this bridge change to roll out invisibly.

t0yv0:
Not quite ready yet but tagging @iwahbe @mjeffryes and @blampe as you worked on this in some adjacent PRs. My hope is that we can make a distinction between aliasHistory and autoSettings, make autoSettings much smaller (minimal), and drop aliasHistory processing from runtime altogether.

t0yv0:
AWS passed cleanly without modifying schema but its use of auto settings is minimal. I'm wondering if I should try this on a more gnarly provider that has not recently underwent a major upgrade.

pkg/tfbridge/auto_aliasing.go
```diff
@@ -125,409 +282,444 @@ func (info *ProviderInfo) ApplyAutoAliases() error {
...
+		return nil
+	}
+	over := make(maxItemsOneOverrides)
+	fh := &fieldHistory{Fields: history}
+	walk.VisitSchemaMap(schemaMap, func(sp walk.SchemaPath, s shim.Schema) {
+		info := LookupSchemaInfoMapPath(sp, schemaInfos)
+
+		// If the user already set a MaxItemsOne override, respect it.
+		if info != nil && info.MaxItemsOne != nil {
+			return
```
iwahbe, line 303: As I understand it, this means that if I remove a manual `MaxItemsOne` override, the next time we run `make tfgen` we will observe a change?

I believe that this is a divergence from existing behavior.

pkg/tfbridge/auto_aliasing.go
```diff
@@ -125,409 +282,444 @@ func (info *ProviderInfo) ApplyAutoAliases() error {
...
 		}
-		prev.Past = removeIndexes(prev.Past, duplicates)
 	}
-	for _, a := range prev.Past {
-		if a.MajorVersion != currentVersion {
-			continue
+	return false
+}
+
+func encodeSchemaPath(sp walk.SchemaPath) string {
```
iwahbe, line 670: It feels like this and `decodeSchemaPath` should live in `walk`.

iwahbe (commented):
This looks really good. 

To test, how hard would it be to do a "test bridge update" job for all providers that asserts that there are no schema changes?

pkg/tfbridge/auto_aliasing.go
t0yv0: Hmm. Really appreciate this kind of analysis! Let me add tests to make explicit what happens here.

My intention here was this. If the user set in MaxItemsOne in current version, we should ever be changing that by automatic aliasing. That's what the code says. Note that maxItemsOneOverrides  are not saved to history. They're only for the current version, for moving data from build-time to runtime.

If you set MaxItemsOne=true in V=1.0.0 and you're building V=1.0.1 and you removed MaxItemsOne  in V=1.0.1, then what's happening here I believe is that effective MaxItemsOne=true will be loaded from aliasHistory, so `prev.MaxItemsOne == &true`, this code will compare it to actual, and if actual=false it will decide to put the override back to match the previous value. So presumably the schema of the provider will not change.

t0yv0:
> To test, how hard would it be to do a "test bridge update" job for all providers that asserts that there are no schema changes?

This is an interesting idea. We could figure something out indeed.

pkg/tfbridge/auto_aliasing.go
iwahbe: Great. It sounds like your describing the behavior we want:
- Overrides in `resources.go` are respected.
- They remain sticky until the next major version.

VenelinMartinov (approved):

t0yv0:
I'm digging further into this and I think there's a small bug on master after all. This code is problematic:

```
	if !alreadyPresent && res.Tok != hist.Current {
		// The resource is in history, but the name has changed. Update the new current name
		// and add the old name to the history.
		hist.Past = append(hist.Past, alias[tokens.Type]{
			Name:         hist.Current,
			InCodegen:    true,
			MajorVersion: currentVersion,
		})
```

Specifically it adds a Past entry at currentVersion, but it probably should be adding a Past entry at hist.MajorVersion.

How it plays out in the test.. This code finds that "pkg:index/mod1R1:Mod1R1" old name introduced in version 0, and creates an entry for it at version 1 (current). This subsequently confuses this code:

	// Only respect hard aliases introduced in the same major version
	if a.InCodegen && a.MajorVersion == currentVersion {

It now thinks that "pkg:index/mod1R1:Mod1R1" was introduced in v1, but it really was introduced in v0. It makes the wrong decision then and calls RenameResourceWithAlias, where the "Only respect hard aliases" seems to indicate that we want to append to res.Aliases instead because this the major version change. 

The behavior in this PR seems better, but it needs to edit the test to avoid it being a major version change (0->1).

t0yv0:
Downstream checks pass tests and there's no Tier1 schema changes except for GCP:

https://github.com/pulumi/pulumi-gcp/pull/1425

Actually GCP looks curious. It added a resource for some reason.

t0yv0:
OK this looks like a side-effect of https://github.com/pulumi/pulumi-gcp/commit/a78947ca06778eaca548d0ffa49215a170441f46 - it looks like Autoscalar is still hanging around in the metadata history although it's no longer represented in the provider at all. I think I can correct up bridge-metadata and fix this.

t0yv0:
Actually bridge-metadata.json says this:

```
            "google_compute_autoscaler": {
                "current": "gcp:compute/autoscalar:Autoscalar",
                "majorVersion": 7,
                "past": [
                    {
                        "name": "gcp:compute/autoscalar:Autoscalar",
                        "inCodegen": true,
                        "majorVersion": 6
                    }
                ],

```

The way I interpret this I think the data is erroneous, but I might be reading this with the wrong semantics for current in mind (which also this code represents). I'm reading this as, "the previously released GCP version is v7 and has the token of gcp:compute/autoscalar:Autoscalar" which is not correct because current release has already corrected to "Autoscaler".

This makes the new code think that we're at present doing the rename from Autoscalar to Autoscaler.

I'll wait till @iwahbe has a chance to look at this before merging, is it reasonable to correct up the bridge-metadata.json for GCP manually and move forward, or am I misunderstanding the intent of "current" and therefore there's a bug in the PR. Thanks.

iwahbe:
> ```json
>            "google_compute_autoscaler": {
>                "current": "gcp:compute/autoscalar:Autoscalar",
>                "majorVersion": 7,
>                "past": [
>                    {
>                        "name": "gcp:compute/autoscalar:Autoscalar",
>                        "inCodegen": true,
>                        "majorVersion": 6
>                    }
>                ],
> ```

I think the data here is erroneous. There doesn't need to be a past entry if `.current = .past[0].name`. That said, I'm concerned about needing a manual fix-up. I think this needs to work correctly for existing metadata files.

t0yv0:
This does not make any sense to me. Can you answer my question please, what does `"current": "gcp:compute/autoscalar:Autoscalar", "majorVersion": 7` mean to you? What I am looking to find, from this data is the information on the latest GCP release preceding the current release candidate, specifically in that release, what is the major version of that release and what is the token assigned to google_compute_autoscaler?  How do I find it from bridge-metadata.json or do we conclude based on the above finding that we simply cannot trust it.

A few options from here. 

- We could put a fix in the code to read the existing data better (though it looks like it may not be possible to fix the GCP issue in that way)
- We could perform manual correction/checks for all bridged providers and roll code as is, assisted by manual data edits.
- We can stop reading bridge-metadata.json entirely for the purposes of recovering aliasing history and instead code up the specification directly. Find latest release of the provider, find its major version, extract its schema, find token allocations and aliases, and use that instead of bridge-metadata.json to inform the algorithm about the history. I think the combination of Schema and MarshallableProviderInfo from the past provider should contain all the info needed for this, in theory, though it's a bit of elbow grease to extract it.

iwahbe:
> This does not make any sense to me. Can you answer my question please, what does "current": "gcp:compute/autoscalar:Autoscalar", "majorVersion": 7 mean to you? 

Considering this example:

```json
           "google_compute_autoscaler": {
               "current": "gcp:compute/autoscalar:Autoscalar",
               "majorVersion": 7,
               "past": [
                   {
                       "name": "gcp:compute/autoscalar:Autoscalar",
                       "inCodegen": true,
                       "majorVersion": 6
                   }
               ],
```

At some point in the past, `make tfgen` was run and observed the TF resource `"google_compute_autoscaler"`. When the resource was observed, the major version of the provider was `7` and the associated pulumi token was `"gcp:compute/autoscalar:Autoscalar"`.

t0yv0:
Got it, thank you! That resonates but is also clearly a bug then. I've done spelunking. Last released provider version that allocates "Autoscalar" token is v6.67.1, all the 7.x series allocate "Autoscaler" token and alias the old one, but they all have this in the metadata:

```
 "current": "gcp:compute/autoscalar:Autoscalar",
 "majorVersion": 7,
``` 

So one more option we could ship a fix before this one that writes actual current token and version into bridged-metadata. Let me first correlate how many more providers are affected by schema change in the fleet.

t0yv0:
```
|------------------------------------------------------------------+-----------------------------------------------------------|
| SUCCESS https://github.com/pulumi/pulumi-gcp/pull/1425           | Schema changed: Autoscalar                                |
| SUCCESS https://github.com/pulumi/pulumi-aws/pull/3135           | No schema changes                                         |
| SUCCESS https://github.com/pulumi/pulumi-cloudflare/pull/587     | No schema changes                                         |
| SUCCESS https://github.com/pulumi/pulumi-datadog/pull/401        | No schema changes                                         |
| SUCCESS https://github.com/pulumi/pulumi-venafi/pull/239         | No schema changes                                         |
| SUCCESS https://github.com/pulumi/pulumi-rabbitmq/pull/249       | No schema changes                                         |
| SUCCESS https://github.com/pulumi/pulumi-auth0/pull/373          | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-vault/pull/355          | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-gitlab/pull/452         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-tls/pull/367            | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-akamai/pull/372         | 13 resources added!                                       |
| SUCCESS https://github.com/pulumi/pulumi-null/pull/67            | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-mongodbatlas/pull/347   | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-aiven/pull/454          | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-ec/pull/210             | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-okta/pull/475           | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-external/pull/79        | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-artifactory/pull/541    | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-http/pull/65            | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-docker/pull/891         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-slack/pull/199          | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-random/pull/534         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-archive/pull/76         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-libvirt/pull/207        | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-onelogin/pull/187       | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-dnsimple/pull/262       | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-signalfx/pull/355       | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-confluentcloud/pull/289 | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-wavefront/pull/282      | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-hcloud/pull/357         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-splunk/pull/230         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-azuread/pull/621        | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-alicloud/pull/524       | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-civo/pull/351           | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-minio/pull/241          | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-kong/pull/270           | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-f5bigip/pull/339        | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-sumologic/pull/224      | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-snowflake/pull/421      | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-vsphere/pull/385        | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-pagerduty/pull/341      | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-consul/pull/302         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-rancher2/pull/332       | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-databricks/pull/279     | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-cloudinit/pull/182      | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-tailscale/pull/282      | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-openstack/pull/454      | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-azure/pull/1508         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-cloudamqp/pull/318      | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-spotinst/pull/509       | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-newrelic/pull/576       | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-keycloak/pull/335       | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-fastly/pull/394         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-oci/pull/292            | Weird Name DomainsUserNameInput change unrelated possibly |
| SUCCESS https://github.com/pulumi/pulumi-rke/pull/212            | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-mailgun/pull/250        | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-kafka/pull/282          | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-github/pull/517         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-postgresql/pull/331     | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-nomad/pull/182          | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-digitalocean/pull/562   | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-linode/pull/442         | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-opsgenie/pull/256       | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-ns1/pull/309            | No                                                        |
| SUCCESS https://github.com/pulumi/pulumi-mysql/pull/281          | No                                                        |
```

iwahbe:
Thanks for doing the legwork to see the effects for all of our providers! After that, I'm comfortable with either manually fixing the 2 (maybe 3) bad metadata files or shipping an automatic migration.

LGTM

t0yv0:
Checked in data corrections in Akamai and GCP.

PR #1548, opened 2023-11-30
Omit omitted types from the schema

iwahbe:
`tfbridge.SchemaInfo` has an `Omit bool` field, which will omit a property from the schema. Unfortunately, omitted fields don't omit their associated types from the schema, leading to unnecessary bloat. This leads to code such as this (from aws):

```go
// We removed the `definition` property from quicksights.Template, see
// https://github.com/pulumi/pulumi-terraform-bridge/issues/1118
// But the types are still present in the schema, which pollutes the Go SDK
// specifically. This function removes those types from the schema.
func removeUnusedQuicksightTypes(pulumiPackageSpec *schema.PackageSpec) {
        var elidedTypes []string
        for tok := range pulumiPackageSpec.Types {
                if strings.Contains(tok, ":quicksight/AnalysisDefinition") ||
                        strings.Contains(tok, ":quicksight/DashboardDefinition") ||
                        strings.Contains(tok, ":quicksight/TemplateDefinition") {
                        elidedTypes = append(elidedTypes, tok)
                }
        }
        for _, tok := range elidedTypes {
                delete(pulumiPackageSpec.Types, tok)
        }
}
```

`Omit: true` is only used in 2 providers: `pulumi-aws` and `pulumi-alicloud` ([source](https://github.com/search?q=org%3Apulumi%20Omit%3A%20true&type=code)).

Datadog has recenlty added some hefty (+900k lines) types, and I am considering using `Omit: true` to trim unwanted types. This is a necessary pre-cursor for that usage. It also has the advantage of making more intuitive sense to future users of the bridge.


iwahbe:
Applying this PR to AWS yields: https://github.com/pulumi/pulumi-aws/pull/3067, removing 88k lines of `bridge-metadata.json`.

PR #1550, opened 2023-12-01
Allow the `Type` override to effect downstream types

iwahbe:
This PR changes how `tfbridge.SchemaInfo.Type` is interpreted when applied to object properties.

Previously, type changed the `schema.TypeSpec.{Type,Ref}` fields on the generated resource. It did not effect any nested types.

With this PR, `Type` is interpreted as a command to rename the relevant object and its entire nested structure. This allows multiple fields or objects to share a type definition when specified, and will allow us to reduce the size of our SDKs.

---

This is technically a breaking change. Bridge users who have specified `Type: someNewType` on object types will see different behavior. In practice, setting `Type` on an object would generate an invalid schema without a careful fix-up. I suspect that no-one does this.

I have tested this change on `pulumi-ns1`, `pulumi-aws`, `pulumi-gcp` and `pulumi-azure`. None of these providers are effected by this change.

### Implementers Notes

`paths.TypePath` is used to determine the module of derived types. Since we want subtypes of a moved type to themselves move to the module of their parent, we need to track that in the paths used. To allow inserting a type at an arbitrary module, it was necessary to create a new `paths.TypePath`; `paths.RawTypePath` projects a `tokens.Type` into `paths.TypePath` so it can be used for module discovery.

### DIFF
--- a/pkg/tfgen/generate_schema.go
+++ b/pkg/tfgen/generate_schema.go
@@ -254,13 +254,23 @@ func (g *schemaGenerator) genPackageSpec(pack *pkg) (pschema.PackageSpec, error)
 	spec.Attribution = fmt.Sprintf(attributionFormatString, g.info.Name, g.info.GetGitHubOrg(), g.info.GetGitHubHost())
 
 	var config []*variable
+	declaredTypes := map[string]*schemaNestedType{}
 	for _, mod := range pack.modules.values() {
 		// Generate nested types.
 		for _, t := range gatherSchemaNestedTypesForModule(mod) {
 			tok := g.genObjectTypeToken(t)
 			ts := g.genObjectType(t, false)
-			spec.Types[tok] = pschema.ComplexTypeSpec{
-				ObjectTypeSpec: ts,
+
+			if existing, ok := declaredTypes[tok]; ok {
+				if !t.typ.equals(existing.typ) {
+					return pschema.PackageSpec{},
+						fmt.Errorf("%s: conflicting type definitions", tok)
+				}
+			} else {
+				declaredTypes[tok] = t
+				spec.Types[tok] = pschema.ComplexTypeSpec{
+					ObjectTypeSpec: ts,
+				}
 			}
 		}
 
--- a/pkg/tfgen/generate_schema.go
+++ b/pkg/tfgen/generate_schema.go
@@ -180,8 +180,12 @@ func (nt *schemaNestedTypes) gatherFromProperties(pathContext paths.TypePath,
 			name = inflector.Singularize(name)
 		}
 
-		nt.gatherFromPropertyType(paths.NewProperyPath(pathContext, p.propertyName),
-			declarer, namePrefix, name, p.typ, isInput)
+		var path paths.TypePath = paths.NewProperyPath(pathContext, p.propertyName)
+		if t := p.typ.typ; t != "" && p.typ.kind == kindObject {
+			namePrefix = ""
+		}
+
+		nt.gatherFromPropertyType(path, declarer, namePrefix, name, p.typ, isInput)
 	}
 }
 
@@ -195,6 +199,9 @@ func (nt *schemaNestedTypes) gatherFromPropertyType(typePath paths.TypePath, dec
 				declarer, namePrefix, name, typ.element, isInput)
 		}
 	case kindObject:
+		if typ.typ != "" {
+			namePrefix = ""
+		}
 		baseName := nt.declareType(typePath, declarer, namePrefix, name, typ, isInput)
 		nt.gatherFromProperties(typePath, declarer, baseName, typ.properties, isInput)
 	}
--- a/pkg/tfgen/generate_schema.go
+++ b/pkg/tfgen/generate_schema.go
@@ -123,6 +123,10 @@ func (nt *schemaNestedTypes) declareType(typePath paths.TypePath, declarer decla
 		typeName = typ.nestedType.Name().String()
 	}
 
+	if typ.typ != "" && typ.kind == kindObject {
+		typeName = typ.typ.Name().String()
+	}
+
 	typ.name = typeName
 
 	required := codegen.StringSet{}
@@ -781,11 +785,14 @@ func (g *schemaGenerator) genObjectTypeToken(typInfo *schemaNestedType) string {
 	}
 
 	mod := modulePlacementForTypeSet(g.pkg, typInfo.typePaths)
-	token := fmt.Sprintf("%s/%s:%s", mod.String(), name, name)
+	token := tokens.Type(fmt.Sprintf("%s/%s:%s", mod, name, name))
+	if typ.typ != "" {
+		token = typ.typ
+	}
 
-	g.renamesBuilder.registerNamedObjectType(typInfo.typePaths, tokens.Type(token))
-
-	return token
+	g.renamesBuilder.registerNamedObjectType(typInfo.typePaths, token)
+	fmt.Printf("Generated token %s (mod = %s, name = %s)\n", token, mod, name)
+	return token.String()
 }
 
 func (g *schemaGenerator) genObjectType(typInfo *schemaNestedType, isTopLevel bool) pschema.ObjectTypeSpec {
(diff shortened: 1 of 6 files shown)


t0yv0:
I like the spirit of this change very much. To paraphrase what I am seeing, it looks like we allow
the user to trigger deliberate collisions when allocating Pulumi type tokens to nested object paths.
These collisions be used to improved sharing, instead of allocating 2 pulumi tokens (and TypeScript
classes and so forth), the user can now have one.

The chief question I have is this: does this work for pseudo-recursion scenario such as wafv2 types? TLDR is that TF model does not support observable sharing recursion so they encode recursive types with "up to depth N" with say N=5, this causes type explosion and bloat at many levels. This PR relies on pre-written `.equals()` function to decide structural equality of "deliberately-colliding types". For the Wafv2 scenario will it work? Naively I'd think that genWafType(4) != genWafType(5) deliberately and this scenario will hit the case where the bridge things the user is making a mistake assigning an identical token to these supplementary types.

I also have a few nits on the implementation level that make it harder for me to follow the change, which I'd love to talk through, as perhaps they point to a gap in my understanding how this codebase works. Let me add some comments.

pkg/tfgen/generate_schema.go
```diff
@@ -764,11 +821,13 @@ func (g *schemaGenerator) genObjectTypeToken(typInfo *schemaNestedType) string {
 	}
 
 	mod := modulePlacementForTypeSet(g.pkg, typInfo.typePaths)
-	token := fmt.Sprintf("%s/%s:%s", mod.String(), name, name)
-
-	g.renamesBuilder.registerNamedObjectType(typInfo.typePaths, tokens.Type(token))
+	token := tokens.Type(fmt.Sprintf("%s/%s:%s", mod, name, name))
```
t0yv0, line 824: IMPL nit:  naively this function is the key function that I'd expected to change in this PR. It freely decides where to put a TF type (in conjunction with the helper modulePlacementFor*). The placement is based on module and name. I'd expect we are not changing the type name. In that case we are down to:

modulePlacementForType

The way this worked before the change is looking at prefixes of paths.TypePath. So when allocating for TypePath = "res_r1.prop_1.element.foo_bar" it would recur upward "res_r1.prop_1.element", "res1_r1.prop_1", "res_r1" until it realized it's in the context of allocating a type for resource "res_r1" and pick the appropriate module based on that.

Presumably this traversal could stop early at "res_r1.prop_1" if prop_1 was given an explicit token Type by the user, and decide to allocate the module according to the user's Type token? I guess what I am saying is why does this necessitate a new clause in TypePath when this information could be looked up from map[string]*SchemaInfo and existing TypePath as written?

pkg/tfgen/generate_schema.go
```diff
@@ -247,13 +284,33 @@ func (g *schemaGenerator) genPackageSpec(pack *pkg) (pschema.PackageSpec, error)
 	spec.Attribution = fmt.Sprintf(attributionFormatString, g.info.Name, g.info.GetGitHubOrg(), g.info.GetGitHubHost())
 
 	var config []*variable
+	declaredTypes := map[string]*schemaNestedType{}
```
t0yv0, line 287: I like this deconfliction check! QQ. what if there's conflicts via some "other" keys in spec.Types than the ones generated for schemaNestedTypes or this is a non-sequitur (they're one and the same)? Thanks!

pkg/tfgen/generate_schema.go
```diff
@@ -180,11 +194,23 @@ func (nt *schemaNestedTypes) gatherFromProperties(pathContext paths.TypePath,
...
+			path = newPath
+		}
+
+		nt.gatherFromPropertyType(path, declarer, namePrefix, name, p.typ, isInput)
 	}
 }
 
+// Insert a type path redirect, setting src to point to dst.
+func (nt *schemaNestedTypes) correctPath(src, dst paths.TypePath) {
+	nt.pathCorrections[src.UniqueKey()] = dst
```
t0yv0, line 211: IMPL nit: I'm losing track a little bit of the lifetime of this table and the need to track it. Is this to power through failing renames functionality? Does it not overlap with info already stored in paths.NewRawPath , e.g. could we get by with not using it, or else not using paths.NewRawPath, or indeed not using either that'd make me the most happy.

iwahbe:
> it looks like we allow the user to trigger deliberate collisions when allocating Pulumi type tokens to nested object paths. These collisions be used to improved sharing, instead of allocating 2 pulumi tokens (and TypeScript classes and so forth), the user can now have one.

Exactly!
> 
> The chief question I have is this: does this work for pseudo-recursion scenario such as wafv2 types? 

No, and it isn't designed to. This PR gets us closer to handling the recursive case, but doesn't apply yet.

> Naively I'd think that genWafType(4) != genWafType(5) deliberately and this scenario will hit the case where the bridge things the user is making a mistake assigning an identical token to these supplementary types.

That is correct.

--- 

This feature will be immediately useful for providers such as [pulumi-akamai](https://github.com/pulumi/pulumi-akamai/issues/313), where many structs share the *same* complex field (`GetPropertyRulesBuilderRulesV20230920Args`).

pkg/tfgen/generate_schema.go
iwahbe: > naively this function is the key function that I'd expected to change in this PR. It freely decides where to put a TF type (in conjunction with the helper modulePlacementFor*). The placement is based on module and name. I'd expect we are not changing the type name. 

We are changing both the module and the name. Consider the following scenario:

```
resource res_r1:
  prop:
    sub:
     str: string
```

With `SchemaInfo{Type: "res:custom:Obj"}` applied to `prop`. 

Direct name change:
The schema object generated for `prop` needs to have its name changed from `R1Prop` to `Obj` and its module changed from `index` to `custom`.

Nested name change:
The schema object generated for `sub` needs to have its name changed from `R1PropSub` to `ObjSub` and its module changed from `index` to `custom`.

> The way this worked before the change is looking at prefixes of paths.TypePath. So when allocating for TypePath = "res_r1.prop_1.element.foo_bar" it would recur upward "res_r1.prop_1.element", "res1_r1.prop_1", "res_r1" until it realized it's in the context of allocating a type for resource "res_r1" and pick the appropriate module based on that.

The way I conceptualize this, when we allocate a type path, we see a path from a root anchor to a property. Root anchors determine the stem of the type name and the module. Previously, a root object was a resource, datasource or config property. My change adds a new root object: `RawTypePath`, which also determines the name stem and module for types nested below.

> Presumably this traversal could stop early at "res_r1.prop_1" if prop_1 was given an explicit token Type by the user, and decide to allocate the module according to the user's Type token? 

In effect, this is what happens. We just encode the stop within the path itself, instead of a look aside map.

> I guess what I am saying is why does this necessitate a new clause in TypePath when this information could be looked up from map[string]*SchemaInfo and existing TypePath as written?

We need more information than is contained in `map[string]*SchemaInfo`, we need the full path of the object.

pkg/tfgen/generate_schema.go
iwahbe: The rename builder logic expects that it can derive the name of an object type from its structural path. This is true up to these corrections, so we need to store a mapping from the old (fully structural) name to the new (user supplied & arbitrary) path.

When the rename builder walks a resource: `res_r1.prop1.nested1`, it needs to build the associated property path. It builds `[resource: res_r1][property: prop1][property: nested1]`. If `prop1` was renamed, we need to recover the correction during the rename traversal so we store `{[resource: res_r1][property: prop1]: [raw: userRename]}` in `nt.pathCorrections`.

This is then used in https://github.com/pulumi/pulumi-terraform-bridge/blob/a0f16392a4c265766fc4de152155019d81a15212/pkg/tfgen/renames.go#L241.

t0yv0: See comment above. The name sems to be a function from TypePath and top-level `map[string]SchemaInfo` but this may be a good form of caching. No matter, I'm leaning toward removing renames from the codebase altogether and if we make that happen we can probably drop this bit as well.

PR #1553, opened 2023-12-04
Upgrade pulumi/pulumi to v3.95.0

### LINKED ISSUE #1552: Upgrade to p/p 3.95.0
## Hello!

## Issue details

Routine upgrade to p/p 3.95.0 needs to happen this week. Users are reporting a minor breaking change that we might need to adjust the code for here:

```
Error: ../../../../go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.66.0/pkg/tfbridge/check_failures.go:159:77: urn.Name().String undefined (type string has no field or method String)
make: *** [Makefile:126: tfgen] Error 1
```

blampe:
Upgrades pulumi/pulumi to v3.95.0.

Fixes #1552.


internal/testing/schemas.go
```diff
@@ -60,14 +60,14 @@ func AssertEqualsJSONFile[T any](
 	assertDataMatchesFile := func(actualData T, file string) {
 		expectedData, err := readTFromFile(file)
 		assert.NoError(t, err)
-		assert.Equal(t, tToString(expectedData), tToString(actualData))
+		assert.JSONEq(t, tToString(expectedData), tToString(actualData))
```
blampe, line 63: This would fail for equal JSON if fields weren't serialized in exactly the same order, which is incorrect behavior.

t0yv0 (approved):
Looking great! 


AS discussed elsewhere the steps around this are in 

https://github.com/pulumi/platform-providers-
team/blob/main/playbooks/Release%3A%20Terraform%20Bridge.md

t0yv0:
The first round of downstream testing detected regressions. We've reverted the offending PR. Doing another round of testing.

t0yv0:
The second round of downstream testing came back clean. Mering! Thank you.

PR #1558, opened 2023-12-07
Fixes unknown handling in makeResourceRawConfig

t0yv0:
AWS opts into PlanState DiffStrategy but it appears that under certain conditions involving unknowns this may cause panics in objchange.ProposedNew. This fix takes apart schema.JSONMapToStateValue that we used to employ for the purpose of massaging config values to the desired form. The original method does not handle unknowns properly translating them to strings of the unknown sentinel value as part of the any->json->cty.Value conversion. The replacement method does a more explicit any->cty.Value conversion taking the unknown sentinels into account, but retains CoerceValue method from schema.JSONMapToStateValue that can help to continue to be permissive and automatically recognize values of a slightly different type (auto converting bool to string etc). 

See also pulumi/pulumi-aws#3094

### DIFF
--- a/pkg/tfshim/sdk-v2/cty.go
+++ b/pkg/tfshim/sdk-v2/cty.go
@@ -19,9 +19,12 @@ import (
 	"math/big"
 
 	"github.com/golang/glog"
+
 	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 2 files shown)


VenelinMartinov (approved):

t0yv0:
Downstream checks came OK with the exception of

FAILED https://github.com/pulumi/pulumi-local/pull/63
FAILED https://github.com/pulumi/pulumi-libvirt/pull/200

I think this is good to go. These failures are not related.

Issue #1556, opened 2023-12-05
 unable to finish polling, context has been cancelled from a bridge update


t0yv0:
### What happened?

Pre-testing bridge candidate we found that acceptance tests started failing in https://github.com/pulumi/pulumi-gcp/pull/1390

Tracking the investigation here.

### Example

```
 TestAccLoadbalancer (1m5.84s)
      program.go:1903: sample: /home/runner/work/pulumi-gcp/pulumi-gcp/examples/loadbalancer
      program.go:1908: pulumi: /opt/hostedtoolcache/pulumi/3.95.0/x64/pulumi
      program.go:1068: acquired yarn install lock
      command.go:34: **** Invoke '/usr/local/bin/yarn install' in '/tmp/p-it-runner-loadbalanc-a0080cef-3274804897'
      command.go:103: Wrote output to /tmp/p-it-runner-loadbalanc-a0080cef-3274804897/command-output/yarn-install.20231204-212632.a8cde.log
      program.go:1094: released yarn install lock
      program.go:2053: projdir: /tmp/p-it-runner-loadbalanc-a0080cef-3274804897
      program.go:1318: Initializing project (dir /tmp/p-it-runner-loadbalanc-a0080cef-3274804897; stack p-it-runner-loadbalanc-a0080cef)
      command.go:34: **** Invoke '/opt/hostedtoolcache/pulumi/3.95.0/x64/pulumi login --cloud-url file:///tmp/TestAccLoadbalancer3298498485/001' in '/tmp/p-it-runner-loadbalanc-a0080cef-3274804897'
      command.go:103: Wrote output to /tmp/p-it-runner-loadbalanc-a0080cef-3274804897/command-output/pulumi-login.20231204-212633.68f6a.log
      command.go:34: **** Invoke '/opt/hostedtoolcache/pulumi/3.95.0/x64/pulumi stack init p-it-runner-loadbalanc-a0080cef' in '/tmp/p-it-runner-loadbalanc-a0080cef-3274804897'
      command.go:103: Wrote output to /tmp/p-it-runner-loadbalanc-a0080cef-3274804897/command-output/pulumi-stack-init.20231204-212633.b6d20.log
[...]

PR #1569, opened 2023-12-12
Capture more upstream provider logs

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1536
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1556

More upstream provider logs are now captured by Pulumi logging.

This is a take 2 for this PR, incorporating learnings from 1556.


t0yv0:
python scripts/downstream_checks.py --hash ba3ef0bc9481818d1246cdc007ebed83e8451912                                        ~/code/pulumi-terraform-bridge
FAILED https://github.com/pulumi/pulumi-azuredevops/pull/235

azdo failure is unrelated I think.

VenelinMartinov (approved):
Very interesting discovery - mirroring the TF behaviour seems like the right thing here!

PR #1571, opened 2023-12-14
Support .Elem() for custom types

iwahbe:
This is necessary to support https://github.com/pulumi/pulumi-aws/issues/3110.

### DIFF
--- a/pf/internal/schemashim/attr_schema.go
+++ b/pf/internal/schemashim/attr_schema.go
@@ -15,11 +15,9 @@
 package schemashim
 
 import (
-	"fmt"
 	bridge "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 
 	pfattr "github.com/hashicorp/terraform-plugin-framework/attr"
-	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
 
 	"github.com/pulumi/pulumi-terraform-bridge/pf/internal/pfutils"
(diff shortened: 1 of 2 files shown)


iwahbe:
@t0yv0 I'm not sure the best place to insert a test here. Any recommendations?

pf/internal/schemashim/attr_schema.go
```diff
@@ -86,39 +84,25 @@ func (*attrSchema) StateFunc() shim.SchemaStateFunc {
...
-	case types.MapType:
-		schema = newTypeSchema(tt.ElemType, s.attr.Nested())
-	case types.ListType:
-		schema = newTypeSchema(tt.ElemType, s.attr.Nested())
-	case types.SetType:
-		schema = newTypeSchema(tt.ElemType, s.attr.Nested())
+	// t does not support any kind of element type.
 	default:
-		panic(fmt.Errorf("This Elem() case is not yet supported: %v", t))
+		return nil
```
t0yv0, line 104: This panic was very useful. In particular we got alerted about the missing case right away during tfgen and pointed to the right code to fix, instead of failing somewhere down the line with an obscure error. Are you sure we want to remove it?

pf/internal/schemashim/attr_schema.go
```diff
@@ -86,39 +84,25 @@ func (*attrSchema) StateFunc() shim.SchemaStateFunc {
...
 		return res
-	}
-	if tt, ok := t.(pfattr.TypeWithElementTypes); ok {
-		var res shim.Resource = newTuplePseudoResource(tt)
+	case pfattr.TypeWithElementTypes:
+		var res shim.Resource = newTuplePseudoResource(t)
 		return res
-	}
+	case pfattr.TypeWithElementType:
+		return shim.Schema(newTypeSchema(t.ElementType(), s.attr.Nested()))
```
t0yv0, line 100: This looks reasonable, sounds like it subsumes all the Map/List/Set cases as well as might handle custom types that implement pfattr.TypeWithElementType interface without necessarily being one of the above concrete types. Custom collections?

t0yv0:
> @t0yv0 I'm not sure the best place to insert a test here. Any recommendations?

Do any test files in this package work like custom_type_test.go or add a new one?

t0yv0:
Thanks for fixing this @iwahbe - https://github.com/pulumi/pulumi-aws/issues/3110 uncorking the build is important; I'd also suggest not blocking on it but verifying that this kind of attribute works all the way at runtime. I always have a little suspicion that just getting past tfgen does not necessarily get us to full support. So it's possible that the affected AWS resource does no deploy right. I would not block on it but good to backlog.

t0yv0 (approved):
LGTM with some comments, prefer to add the panic / assert  back.

pf/internal/schemashim/attr_schema.go
iwahbe: Previously, we would assert when:
1. `t` didn't implement `basetypes.ObjectTypable`.
2. `t` didn't implement `pfattr.TypeWithElementTypes`.
3. `t` did implement `pfattr.TypeWithElementType`
4. but was not a `types.MapType`, `types.ListType` or `types.SetType`.

 We now accept all types that implement `pfattr.TypeWithElementType`. What would the new assert criteria be?

pf/internal/schemashim/attr_schema.go
iwahbe: The change is necessitated to handle custom collections. This is the type that AWS introduced:

https://github.com/hashicorp/terraform-provider-aws/blob/3de2fd3ac7f5c9ddb7f64134fdc0dc5b7110fb79/internal/framework/types/list_nested_objectof.go#L17C1-L20

pf/internal/schemashim/attr_schema.go
t0yv0: Ah never mind, I was a bit hasty. I like to have the switches total. But it looks like it is total already. For scalars etc the right answer from Elem() is nil it looks like.

pf/internal/schemashim/attr_schema.go
t0yv0: Great fodder for a test case.

t0yv0:
Never mind the panic, thanks for pointing it out the switch statement looks correct.

PR #1590, opened 2023-12-20
Efficient schema traversal

iwahbe:
The goal is a general solution that abstracts traversing each SchemaInfo in a provider and applying some action. To prevent the provider from bogging down during startup, the API needs to do a full search during tfgen time, and then visit only impactful fields during the providers runtime.

This implementation will replace gcp's labels traversal, aws's tags traversal and will be used in akamai to unify types.

The implementation takes after https://github.com/pulumi/pulumi-terraform-bridge/pull/1544. Visitors return if they had an "effect". During `make tfgen`, every path is visited and a list of "effectful" `SchemaPath`s is built. During the provider's normal runtime, this list is deserialized and only these paths are visited.

### DIFF
--- a/pkg/tfbridge/walk.go
+++ b/pkg/tfbridge/walk.go
@@ -15,11 +15,17 @@
 package tfbridge
 
 import (
+	"errors"
+	"fmt"
+
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/util"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/walk"
+	md "github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata"
 )
 
 type SchemaPath = walk.SchemaPath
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/walk.go
```diff
@@ -168,3 +265,347 @@ func LookupSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *Schema
...
+		err := md.Set(prov.GetMetadata(), key, effects)
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.Join(errs...)
+}
+
+type traversalRecord struct {
```
t0yv0, line 441: There is some non-trivial TCO to keeping things serialized. Why is this code storing this? What does this data mean?

pkg/tfbridge/walk.go
```diff
@@ -114,8 +121,98 @@ func propertyPathToSchemaPathInner(
...
+// Translate a a bridged provider schema path into a Pulumi property path.
+//
+// The function hides the complexity of mapping Terraform names to Pulumi property names, joining Schema with user
+// overrides in SchemaInfos, and accounting for MaxItems=1 situations where Pulumi flattens collections to plain values.
+// and therefore SchemaPath values are longer than the PropertyPath values.
+//
+// SchemaPathToPropertyPath may return nil if there is no matching schema found. This may happen when drilling down to
+// values of unknown type, attributes not tracked in schema, or when there is a type mismatch between the path and the
+// schema.
+func SchemaPathToPropertyPath(
```
t0yv0, line 133: This semantically makes me uneasy just looking at the types, because PropertyPath represents paths in values while SchemaPath represents paths in types. What's the deal with map/list/set elements?   Skimming the implementation it looks like SchemaPaths pointing to list/map/set elements will translate to property paths with '*', for example:

```
			NewSchemaPath().GetAttr("x").Element().GetAttr("y") => x.*.y
```

This could make sense but perhaps call this `SchemaPathToPropertyPathPattern` and/or emphasize in the doc string?

pkg/tfbridge/walk.go
```diff
@@ -168,3 +265,347 @@ func LookupSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *Schema
 		return nil
 	}
 }
+
+// Drill down a path from a SchemaInfo object and find a matching SchemaInfo. If no
+// SchemaInfo exists, it will be created.
+func getSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *SchemaInfo {
```
t0yv0, line 271: https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/walk.go#L131 is this materially different from LookupSchemaInfoMapPath, maybe null handling? 

"If no SchemaInfo exists, it will be created."

Looks like that's the difference. Perhaps call it "getOrCreateSchemaInfoByPath"

pkg/tfbridge/walk.go
iwahbe: I'll add a doc comment explaining. The tldr is that `traversalRecord` is a list of resource/datasource/config schema paths that need to be visited during runtime.

I added this text to the PR description:
> The implementation takes after https://github.com/pulumi/pulumi-terraform-bridge/pull/1544. Visitors return if they had an "effect". During make tfgen, every path is visited and a list of "effectful" SchemaPaths is built. During the provider's normal runtime, this list is deserialized and only these paths are visited.

iwahbe:
I requested the review from t0yv0 after he reviewed but before I refreshed my screen (and saw the review). False alarm.

pkg/tfbridge/walk_test.go
```diff
@@ -95,12 +102,18 @@ func TestPropertyPathToSchemaPath(t *testing.T) {
 		},
 		{
 			name:     "list",
-			pp:       []any{"listStr"},
+			pp:       []any{"listStrs"},
```
t0yv0, line 105: Was there a bug here? Why is the test changing?

pkg/tfbridge/walk.go
```diff
@@ -168,3 +265,347 @@ func LookupSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *Schema
...
+
+func (r VisitDataSourceRoot) PulumiToken() tokens.Type { return r.token }
+func (r VisitDataSourceRoot) isVisitRoot()             {}
+
+type VisitConfigRoot struct{ token tokens.Type }
+
+func (r VisitConfigRoot) PulumiToken() tokens.Type { return r.token }
+func (r VisitConfigRoot) isVisitRoot()             {}
+
+type Visitor = func(VisitInfo, SchemaPath) (VisitResult, error)
```
t0yv0, line 350: Style golf, perhaps simplify to `func(VisitContext) error`, VisitContext can make other things handy like methods to retrieve "current" schema or info.

pkg/tfbridge/walk_test.go
iwahbe: From the commit message of https://github.com/pulumi/pulumi-terraform-bridge/pull/1590/commits/eca5d902fabd438d6224e0e0c96d4dfeb4c13e2a:

```
Note:

Some tests had PP names that didn't make sense with the given schema. I corrected them to
allow round-tripability:

`` `patch
-			pp:       []any{"objSet", 0},
+			pp:       []any{"objSets", 0},
`` `

I added a test with name overrides to confirm that it still worked as expected.
```

Basically, the property path `[]any{"listStr"}` would have never been generated by the bridge from it's schema, since `TerraformToPulumiName` would have pluralized `listStr` to `listStrs`. This worked in the context of the previous test, but lost information in the translation.

---

P.S. I really don't like that GH doesn't surface commit messages.

pkg/tfbridge/walk.go
```diff
@@ -168,3 +265,347 @@ func LookupSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *Schema
...
+// traversalId must be unique for each traversal within a `make tfgen`, but the same
+// between `make tfgen` and the provider runtime. A simple descriptive string like `"apply
+// labels"` works great here.
+func TraverseSchemas(prov *ProviderInfo, traversalID string, visitor Visitor) error {
+	prov.MetadataInfo.assertValid() // We must have metadata info for an efficient (sparse) traversal.
+
+	var errs []error
+
+	const key = "traverse-effects"
+	forEffect := currentRuntimeStage == runningProviderStage
```
t0yv0, line 368: This should be a parameter and not relying on the auto detection, to make the code more testable. Already been running into some trouble with this hidden auto-detector, i.e. benchmarks in pulumi-aws cannot be made to measure the runtime overhead because the auto-detector things that bencharking is happening at the runtime phase.

pkg/tfbridge/walk_test.go
t0yv0: Ah sorry I'll make a point of reading through those next time. Glad there was not a prod bug.

pkg/tfbridge/walk.go
iwahbe: Property paths support `*` as a glob element. I'll update the doc comment to call this out explicitly.

t0yv0:
So if I delete SchemaPathToPropertyPath this still compiles? Perhaps Efficient Schema traversal does not need it but you have an application in mind anyway? Would it be easy to have this as separate PR for posterity?

pkg/tfbridge/walk.go
t0yv0: Well yeah they probably should not, a separate type would be better, but some of the code that accepts PropertyPath actually accepts patterns. Sounds good.

iwahbe:
> So if I delete SchemaPathToPropertyPath this still compiles? Perhaps Efficient Schema traversal does not need it but you have an application in mind anyway? Would it be easy to have this as separate PR for posterity?

I was testing the API with https://github.com/pulumi/pulumi-gcp/pull/1474, which needs `SchemaPathToPropertyPath`. I'll pull it out into a separate PR.

t0yv0:
This is clever, perhaps I worry this is a little too clever. I like it overall, it's an interesting way to speedup runtime traversals. 

The chief reason not to do it would be that it might be an abstraction that doesn't fit all use cases like a glove and the alternative would be that individual traversals do their own thing each (and specially decide on the shape of data to store up in the runtime metadata blob or elsewhere).

The general pattern for passes is something akin to this, with Edits not necessarily being constrained to paths. 

```
type Pass[Edits Serializable] interface {
  PrecomputeEdits(*ProviderInfo) Edits
  ApplyEdits(*ProviderInfo, Edits) 
}
```

For AWS for example I've just been special-casing a lot to get past RAM/CPU costs: 
https://github.com/pulumi/pulumi-aws/pull/3172

And I'm not sure that AWS things can be shoehorned profitably into TraverseSchema. 

gcp's labels traversal is IMO not a strong example as it's overdone as the only use case is for top-level labels but the code does a deep traversal so something really simple and fast could be done instead. 

Really would be curious to see how it plays out in Akamai.

t0yv0:
https://github.com/pulumi/pulumi-gcp/pull/1474 scratch that, looks like there's 2-3 cases of non-top-level labels fix in GCP.

pkg/tfbridge/walk.go
```diff
@@ -168,3 +265,347 @@ func LookupSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *Schema
...
+}
+
+type VisitResult struct {
+	// If the visitor has done something.
+	//
+	// Visits will only be replayed during runtime startup if `HasEffect: true`.
+	HasEffect bool
+}
+
+func (v *VisitInfo) ShimSchema() shim.Schema                        { return v.shimSchema }
```
t0yv0, line 304: This is a confusing API, I think. I'm confused just trying to verify this :)

The EnclosingSchemaShimMap is not enclosing anything. It calls getShimSchemaMap which unpacks the resource fields at the current schema location.

iwahbe:
> And I'm not sure that AWS things can be shoehorned profitably into TraverseSchema.

I'm not 100% sure if the tags work can be fit into this framework (I think it can), but `SetAutonaming` can be fit into this framework (and should be... for all providers).

> Really would be curious to see how it plays out in Akamai.

I don't have a working solution for Akamai yet. The bridge is giving me naming issues still (unrelated to this PR). I'm playing with code like this (and seeing progress):

```go
	tfbridge.MustTraverseSchemas(&prov, "rules", func(ctx tfbridge.VisitInfo, path tfbridge.SchemaPath) (tfbridge.VisitResult, error) {
		_, ok := ctx.Root.(tfbridge.VisitDataSourceRoot)
		if !ok {
			return tfbridge.VisitResult{}, nil
		}

		setType := func(typ tokens.Type) {
			info := ctx.SchemaInfo()
			info.Type = typ
			if info.Elem == nil {
				info.Elem = &tfbridge.SchemaInfo{}
			}
			info.Elem.Type = typ
		}

		last, _ := path[len(path)-1].(walk.GetAttrStep)

		switch last.Name {
		case "cloudlet_policy":
			setType("akamai:index/cloudletPolicy:CloudletPolicy")
		case "cp_code":
			setType("akamai:index/cpCode:CpCode")
		case "behavior":
			setType("akamai:index/rulesBehavior:RulesBehavior")

		// TODO: Investigate
		// case "issuer_rdns":
		// case "subject_rdns":
		// case "value":
		// case "net_storage":
		default:
			return tfbridge.VisitResult{}, nil
		}
		return tfbridge.VisitResult{HasEffect: true}, nil
	})
```

t0yv0:
Yeah I think you're right, on second thought it can be made to work. I still worry about the understandability of the APIs exposed but we can't fix it all at once.. see we don't have a good set of types/vocabulary to talk about different schema contexts in this codebase like SchemaMap and SchemaInfo are way too overloaded and confusing. Is this a resource? Is this an object-typed field inside a resource? Having this spelled out would make it so much easier to explain. That said this should have pre-existed before TraverseSchema so not a reason to block on it..

Since this is not a terrible rush let's keep playing with it a bit more and try to finish the Akamai thing; I'll take another look on Tue (since holiday break is coming) and can take on trying to rewrite AWS with it.

t0yv0:
Second look, this looks good I can approve as-is (thanks for factoring out the other PR). One last round of API design. I 

```
type Visitor = func(VisitInfo) (VisitResult, error)
```

The visitor is invoked in these contexts:

- Function 
- Provider Config
- Resource
- Nested Property (rooted in either Function, Provider Config, or Resource)

Current design uses discriminated types for the kinds of roots, but makes Schema/SchemaMap accessors available for every context.  IF we're using discriminated types already I'd consider splitting this up so only Schema or SchemaMap accessor is available depending on what we're visiting and kind of push the user to correct usage (even if it's a little more verbose), such as:

- when visiting roots (Function, Resource, ProviderConfig) you probably want to work with SchemaMap not Schema
- when visiting nested properties you probably want to work with Schema, with the exception of Object-typed properties, which can be made explicit e.g.

```
if ctx.IsObjectType() {
   ctx.ObjectSchemaMap()
}
```

We're getting a bit blind to this as we've been working on the codebase a lot, curious for a fresh perspective like @VenelinMartinov thanks.

VenelinMartinov (commented):
Looks great, I had a few questions on the implementation!

@t0yv0's idea about exposing only the right properties to the visitor sounds quite sensible since any correct visitor would need to check which one to use in the current implementation.

As a person somewhat unfamiliar with the codebase it can be quite confusing what the difference between a Schema, SchemaInfo, SchemaMap, etc is.

pkg/tfbridge/walk.go
```diff
@@ -168,3 +284,394 @@ func LookupSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *Schema
...
+type VisitResourceRoot struct {
+	token tokens.Type
+
+	Info *ResourceInfo
+
+	TfToken string
+}
+
+func (r VisitResourceRoot) PulumiToken() tokens.Type { return r.token }
+func (r VisitResourceRoot) isVisitRoot()             {}
```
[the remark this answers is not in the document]
iwahbe: `isVisitRoot()` is here only to satisfy the interface, and yes this is a common go idiom. Because `isVisitRoot` is a private function, this means that only types in the `tfbridge` package can implement the `VisitRoot` interface, making `VisitRoot` a "closed" interface. 

I think the naming is idiomatic.

pkg/tfbridge/walk.go
```diff
@@ -168,3 +284,394 @@ func LookupSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *Schema
...
+	prov.MetadataInfo.assertValid() // We must have metadata info for an efficient (sparse) traversal.
+
+	var errs []error
+
+	var traverseOptions traversalOptions
+	for _, opt := range opts {
+		opt(&traverseOptions)
+	}
+
+	// forEffect indicates that the traversal is being run for side effects only, so
```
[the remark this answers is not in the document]
iwahbe: The caller can provide an override for `forEffect`, but the default is to run in `forEffect` mode when `currentRuntimeStage == runningProviderStage`. See:

https://github.com/pulumi/pulumi-terraform-bridge/blob/a25a2fee3c454d02da84a1c46239da4a1812b40a/pkg/tfbridge/walk.go#L408-L412

>  Did you include it here because the caller is in the provider code?

I'm not sure what you mean here. Can you rephrase?

pkg/tfbridge/walk.go
```diff
@@ -168,3 +284,394 @@ func LookupSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *Schema
...
+		var err error
+		effects, ok, err = md.Get[traversalRecord](prov.GetMetadata(), traversalID)
+		contract.Assertf(ok, "could not find traversalID %q in metadata, are you missing a `make tfgen`?", traversalID)
+		if err != nil {
+			return fmt.Errorf("unable to find effect list: %w", err)
+		}
+	}
+
+	rMap := prov.P.ResourcesMap()
+	for tfName, rInfo := range prov.Resources {
```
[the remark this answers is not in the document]
iwahbe: No. I see why it looks like that, but it isn't happing. I'll take another pass at simplifying the code here. I agree it's a little confusing.

iwahbe:
> One last round of API design. I
> 
> ```
> type Visitor = func(VisitInfo) (VisitResult, error)
> ```

I really appreciate your help with a clearer API design. I can tell the current design is confusing.

> The visitor is invoked in these contexts:
> 
> * Function
> * Provider Config
> * Resource
> * Nested Property (rooted in either Function, Provider Config, or Resource)

The visitor is only invoked on properties ("Nested Property" from your list, though not all properties are nested). Each property is anchored in a Function, Provider Config or Resource.

> Current design uses discriminated types for the kinds of roots, but makes Schema/SchemaMap accessors available for every context. 

`shim.Schema` and `tfbridge.SchemaInfo` are associated with each visited property. `shim.SchemaMap` and `map[string]*SchemaInfo` is available only via `EnclosingSchemaShimMap` and `EnclosingSchemaInfoMap`, respectively. Every property is enclosed in a `shim.SchemaMap` and `map[string]*SchemaInfo` by construction, so these are always valid.

For example, consider the resource:

```yaml
xyz_my_resource:
  field1:
    foo:
      type: string
    bar:
      type: int
  field2:
    type: string
```

The visitor will visit `field1`, `foo`, `bar`, `field2`. It will *not* visit `xyz_my_resource`, since that is not a property.

Every property is assumed[^1] to have an enclosing map by many bridge functions[^2]. The enclosing map of `foo` and `bar` is `{ foo: ..., bar: ...}` for both `shim.SchemaMap` and `map[string]*SchemaInfo`. The enclosing map of `field1` and `field2` is `{field1: ..., field2: ...}`.

[^1]: `.Elem` properties don't really have an enclosing map, but the rest of the bridge assumes that they do. 

[^2]: For example, `PropertyPathToSchemaPath` takes the enclosing map of the field it is translating.

> IF we're using discriminated types already I'd consider splitting this up so only Schema or SchemaMap accessor is available depending on what we're visiting and kind of push the user to correct usage (even if it's a little more verbose), such as:
>
> * when visiting roots (Function, Resource, ProviderConfig) you probably want to work with SchemaMap not Schema
> * when visiting nested properties you probably want to work with Schema, with the exception of Object-typed properties, which can be made explicit e.g.
> 
> ```
> if ctx.IsObjectType() {
>    ctx.ObjectSchemaMap()
> }
> ```

As a mentioned in the previous field, roots are never visited directly, only properties. The function visits everything with a valid `SchemaInfo / shim.Schema`. It sounds like that name is confusing. Should I change the name to `TraverseProperties`?

pkg/tfbridge/walk.go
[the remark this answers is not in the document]
VenelinMartinov: That makes perfect sense - I was trying to confirm that the reason for the default is that provider code calls this function. Hence you can't set the default at the calling place.

pkg/tfbridge/walk.go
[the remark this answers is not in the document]
VenelinMartinov: I'm still a bit confused about why you are iterating over all resources here in the `forEffect=true` case.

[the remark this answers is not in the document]
VenelinMartinov: ah, nvm just realised you have a condition on line 624

pkg/tfbridge/walk.go
```diff
@@ -281,3 +287,394 @@ func LookupSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *Schema
...
+	for tfName, dInfo := range prov.DataSources {
+		rShim, ok := dMap.GetOk(tfName)
+		if !ok {
+			continue
+		}
+
+		var err error
+		effect := ensureMap(&effects.DataSources)[tfName]
+
+		effect, err = traverseResourceOrDataSource(tfName, rShim, dInfo,
```
VenelinMartinov, line 456: your variable naming is slightly confusing - you are using `effect` and `effects` both in this function and in `traverseResourceOrDataSource` but in `traverseResourceOrDataSource` you return an `effects` and assign it to `effect`.

It'd be a bit clearer if you have 3 variable names and match up the ones which are the same across function boundaries.

VenelinMartinov: The type of `effects` also changes - it is `effects traversalRecord` here but `effects []SchemaPath` in `traverseSchemaMap`

VenelinMartinov (approved):
I didn't mean to block with my previous comments. Approving to unblock

pkg/tfbridge/walk.go
VenelinMartinov: perhaps something like `effectRecords` -> `effectPaths` -> `effect`

t0yv0:
> The visitor is only invoked on properties 

Ok yeah this is another consistent way to look at this I guess, thanks for explaining. So it visits properties only. So a couple of questions.

1. Would it ever visit list elements? These can have associated shim.Schema as well as possible SchemaInfo overrides. When visiting list elements, what will EnclosingSchemaShimMap be? Would it be nil or would it be the grandparent map? 

2. When visiting a property x on resource R of type object is this right: EnclosingSchemaShimMap is R's fields; SchemaShimMap is something with Elem() having type shim.Resource that you need to unpack if you want to unpack it; however you may just wait until x.z gets visited in another invocation, where EnclosingSchemaShimMap is going to be x's fields, and SchemaShimMap is going do be the x.z's type.

t0yv0:
TraverseProperties sounds good!

pkg/tfbridge/walk.go
t0yv0: Ah I think this works fine but I got confused because the implementation assumes this nasty trick that apparently I put in:

```
func wrapSchemaMap(sm shim.SchemaMap) shim.Schema {
	return (&schema.Schema{
		Type: shim.TypeMap,
		Elem: (&schema.Resource{Schema: sm}).Shim(),
	}).Shim()
}
```

This trick unifies SchemaMap and Schema types but it creates Schema values that don't "naturally" occur.  However this is part of the machinery here.. so the resources's properties for example can be unpacked via Elem().(*Resource) but only when it got allocated by this util function.

pkg/tfbridge/walk.go
```diff
@@ -281,3 +287,394 @@ func LookupSchemaInfoPath(schemaPath SchemaPath, schemaInfo *SchemaInfo) *Schema
...
+	//
+	// Visits will only be replayed during runtime startup if `HasEffect: true`.
+	HasEffect bool
+}
+
+func (v *VisitInfo) SchemaPath() SchemaPath                         { return v.schemaPath }
+func (v *VisitInfo) ShimSchema() shim.Schema                        { return v.shimSchema }
+func (v *VisitInfo) SchemaInfo() *SchemaInfo                        { return v.getSchemaInfo() }
+func (v *VisitInfo) EnclosingSchemaInfoMap() map[string]*SchemaInfo { return v.getSchemaInfoMap() }
+func (v *VisitInfo) EnclosingSchemaShimMap() shim.SchemaMap         { return v.getShimSchemaMap() }
```
t0yv0, line 332: Consider "EnclosingSchemaMap".

pkg/tfshim/walk/walk.go
```diff
@@ -114,6 +116,38 @@ func DecodeSchemaPath(path string) SchemaPath {
...
+		return err
+	}
+	*p = DecodeSchemaPath(s)
+	return nil
+}
+
+// Provide a stable ordering of SchemaPaths.
+//
+// Beyond determinism, there is no specific sort order.
+func SortSchemaPaths(paths []SchemaPath) {
```
t0yv0, line 145: You can avoid this if you represent a "set" as a map[string]struct{}.

t0yv0 (approved):
Thanks for explaining! Left some further questions but they're not blocking.

iwahbe:
> > The visitor is only invoked on properties
> 
> Ok yeah this is another consistent way to look at this I guess, thanks for explaining. So it visits properties only. So a couple of questions.
> 
> 1. Would it ever visit list elements? These can have associated shim.Schema as well as possible SchemaInfo overrides. When visiting list elements, what will EnclosingSchemaShimMap be? Would it be nil or would it be the grandparent map?

Yes (also map elements). It will be the result of `walk.LookupSchemaMapPath(p[:len(p)-1], schema).m.Elem().(shim.Resource)`, which should be `nil`.

> 2. When visiting a property x on resource R of type object is this right: EnclosingSchemaShimMap is R's fields; SchemaShimMap is something with Elem() having type shim.Resource that you need to unpack if you want to unpack it; however you may just wait until x.z gets visited in another invocation, where EnclosingSchemaShimMap is going to be x's fields, and SchemaShimMap is going do be the x.z's type.

I think that is correct.

This is the object you described:

```yaml
R:
  x:              # <-- EnclosingSchemaShimMap is `R`'s fields. If you want to see `x`'s fields from
                      # here, you need to do
                      # `VisitInfo.ShimSchema().Elem().(shim.Resource).Schema()`, which will give
                      # `{y: ...}`. If you just want to access `R.x.y`, you can wait.
    type: object
    fields:
      y:          # <-- EnclosingSchemaShimMap is `R.x`'s
                      # fields. `VisitInfo.ShimSchema().Elem().(shim.Resource).Schema()` will give
                      # `{z: {type: string}}`
        type: object
        fields:
          z:      # <-- EnclosingSchemaShimMap is `R.x.y`'s fields. `VisitInfo.ShimSchema().Type()
                      # == shim.TypeString`
            type: string
```

PR #1593, opened 2023-12-21
Decode schema faster on init and do not retain it

t0yv0:
In the latest iteration the PF bridge no longer needs to hold onto a PackageSpec reference. It does show in the heap profile though. This PR takes the opportunity to remove the packageSpec reference to reclaim this memory.

### DIFF
--- a/pf/tfbridge/provider.go
+++ b/pf/tfbridge/provider.go
@@ -15,17 +15,18 @@
 package tfbridge
 
 import (
+	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 
 	"github.com/blang/semver"
+	"github.com/json-iterator/go"
+
 	pfprovider "github.com/hashicorp/terraform-plugin-framework/provider"
 	"github.com/hashicorp/terraform-plugin-framework/providerserver"
 	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 
-	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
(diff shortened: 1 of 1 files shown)


pf/tfbridge/provider.go
```diff
@@ -192,7 +204,7 @@ func (p *provider) Close() error {
 
 // Pkg fetches this provider's package.
 func (p *provider) PkgWithContext(_ context.Context) tokens.Package {
-	return tokens.Package(p.packageSpec.Name)
+	return p.pkgName
```
iwahbe, line 207: I think we can make this simpler:

Looking through the generate chain, we see that `ProviderInfo.Name` is ultimately used as the package name:

`GenerateSchemaWithOptions`:

https://github.com/pulumi/pulumi-terraform-bridge/blob/0f7ccfdef31d3fddd4eb46c9770d460bbfbe81e3/pkg/tfgen/generate.go#L693-L714

`genPulumiSchema`:

https://github.com/pulumi/pulumi-terraform-bridge/blob/0f7ccfdef31d3fddd4eb46c9770d460bbfbe81e3/pkg/tfgen/generate_schema.go#L246-L250

`genPackageSpec`:

https://github.com/pulumi/pulumi-terraform-bridge/blob/0f7ccfdef31d3fddd4eb46c9770d460bbfbe81e3/pkg/tfgen/generate_schema.go#L258-L260

I think we can ignore the schema artifact here and just use:

```suggestion
	return tokens.Package(p.info.Name)
```

iwahbe (commented):
This is an improvement, and I believe that it lessens the memory consumption of bridged providers.

I believe that I have a simpler and more performant implementation. (in a comment) If you think my solution won't work, please let me know why and I'll approve this PR as is.

iwahbe (approved):

PR #1594, opened 2023-12-24
Add a SchemaPathToPropertyPath

iwahbe:
This function is currently unused in the bridge itself. It is expected that the function will be used by providers.

See https://github.com/pulumi/pulumi-gcp/pull/1474 for an example.

---

Some tests had PP names that didn't make sense with the given schema. I corrected them to allow round-tripability:

```patch
-			pp:       []any{"objSet", 0},
+			pp:       []any{"objSets", 0},
```

I added a test with name overrides to confirm that it still worked as expected.  Put simply, the property path []any{"listStr"} would have never been generated by the bridge from it's schema, since TerraformToPulumiName would have pluralized listStr to listStrs. This worked in the context of the previous test, but lost information in the translation.

### DIFF
--- a/pkg/tfbridge/walk.go
+++ b/pkg/tfbridge/walk.go
@@ -18,6 +18,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/util"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/walk"
 )
 
@@ -119,8 +120,7 @@ func propertyPathToSchemaPathInner(
 //
 // The function hides the complexity of mapping Terraform names to Pulumi property names,
 // joining Schema with user overrides in [SchemaInfo]s, and accounting for MaxItems=1
-// situations where Pulumi flattens collections to plain values.  and therefore SchemaPath
-// values are longer than the PropertyPath values.
+// situations where Pulumi flattens collections to plain values.
 //
 // [SchemaPathToPropertyPath] may return nil if there is no matching schema found. This
 // may happen when drilling down to values of unknown type, attributes not tracked in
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/walk.go
```diff
@@ -114,8 +115,117 @@ func propertyPathToSchemaPathInner(
 	return nil
 }
 
+// Translate a a bridged provider schema path into a Pulumi property path.
+//
+// The function hides the complexity of mapping Terraform names to Pulumi property names,
+// joining Schema with user overrides in [SchemaInfo]s, and accounting for MaxItems=1
+// situations where Pulumi flattens collections to plain values.  and therefore SchemaPath
```
t0yv0, line 122: " to plain values.  and therefore SchemaPath" - nit something's a bit odd with grammar there, do we need a "."?

pkg/tfbridge/walk.go
```diff
@@ -114,8 +115,117 @@ func propertyPathToSchemaPathInner(
...
+// values are longer than the PropertyPath values.
+//
+// [SchemaPathToPropertyPath] may return nil if there is no matching schema found. This
+// may happen when drilling down to values of unknown type, attributes not tracked in
+// schema, or when there is a type mismatch between the path and the schema.
+//
+// ## Element handling
+//
+// [SchemaPath]s can be either attributes or existential elements. For example:
```
t0yv0, line 131: What does "existential" mean here?

pkg/tfbridge/walk.go
```diff
@@ -114,8 +115,117 @@ func propertyPathToSchemaPathInner(
...
+//	NewSchemaPath().GetAttr("x").Element().GetAttr("y")
+//
+// [resource.PropertyPath]s have attributes or instantiated elements. For example:
+//
+//	x[3].y
+//
+// [SchemaPathToPropertyPath] translates all existential elements into the "*" (_any_)
+// element. For example:
+//
+//	NewSchemaPath().GetAttr("x").Element().GetAttr("y") => x["*"].y
```
t0yv0, line 142: Thank you for including this, makes the intent perfectly clear.

pkg/tfbridge/walk.go
```diff
@@ -114,8 +115,117 @@ func propertyPathToSchemaPathInner(
...
+
+	if len(schemaPath) == 0 {
+		return basePath
+	}
+
+	if schemaInfo == nil {
+		schemaInfo = &SchemaInfo{}
+	}
+
+	// Detect single-nested blocks (object types).
```
t0yv0, line 197: If you're looking for somewhere to put a helper method perhaps somewhere under tfshim would do. 

https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfshim/util/types.go#L24 is one place.

t0yv0 (approved):
Great addition, this will be super useful down the line. Thank you!

pkg/tfbridge/walk.go
iwahbe: I clarified in https://github.com/pulumi/pulumi-terraform-bridge/pull/1594/commits/7cabf06a35f94e56fdf3e76946fc0f8644230de1. In summary:

In a `SchemaPath`, `.Element()` represents the existence of some element access, but not a specific element access. I'm calling that an existential element. This is different from an element index in a `PropertyPath`: I'm calling `[3]` an instantiated element access, since it represents a specific index.

PR #1605, opened 2023-12-29
feat: Implement Provider Muxer

tmeckel:
This PR contains changes to Pulumi Terraform Bridge which allows the specification of provider mixins to add or replace functions or resources to a wrapped Terraform Provider without the need to change the code of upstream provider.


pkg/tfbridge/serve.go
```diff
@@ -26,19 +26,14 @@ import (
...
-	opts := opts{}
-	for _, applyOption := range options {
-		applyOption(&opts)
-	}
-
+func Serve(module string, version string, info ProviderInfo, pulumiSchema []byte) error {
 	// Create a new resource provider server and listen for and serve incoming connections.
 	return provider.Main(module, func(host *provider.HostClient) (pulumirpc.ResourceProviderServer, error) {
-		if len(opts.muxWith) > 0 {
+		if info.MuxWith != nil && len(info.MuxWith) > 0 {
```
iwahbe, line 32: We should probably encapsulate this whole block L32-L57 in its own function: `newMuxWithProvider` to make testing easier. We can then test the runtime behavior similar to this:

https://github.com/pulumi/pulumi-terraform-bridge/blob/2bb619413b2cca3d6acec58c76acb25dbb844fcb/pkg/tests/regress_923_test.go#L29

iwahbe (commented):
@tmeckel This is awesome! I'm really exited about getting this change in.

CC @blampe, @guineveresaenger This would provide a bridge supported way to stack a native Pulumi provider on top of a bridged provider... which currently describes pulumi-docker. I'd appreciate if you gave the API a once-over since you have experience in this area.

CC @t0yv0

pkg/tfbridge/serve.go
```diff
@@ -26,19 +26,14 @@ import (
...
 			// If we have multiple providers to serve, Mux them together.
 
 			var mapping muxer.DispatchTable
-			if m, found, err := metadata.Get[muxer.DispatchTable](info.GetMetadata(), "muxer"); err != nil {
+			if m, found, err := metadata.Get[muxer.DispatchTable](info.GetMetadata(), "mux"); err != nil {
```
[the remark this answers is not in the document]
tmeckel: Because the name differed between tfbridge and pf. I like to keep names equal for things which are equal.

image
image

pkg/tfgen/generate_schema.go
```diff
@@ -252,7 +255,38 @@ func genPulumiSchema(pack *pkg, name tokens.Package, version string,
...
+		}
+		muxSchemas := make([]pschema.PackageSpec, len(info.MuxWith)+1)
+		muxSchemas[0] = pulumiPackageSpec
+		for i, v := range info.MuxWith {
+			spec, err := v.GetSpec()
+			if err != nil {
+				return pschema.PackageSpec{}, err
+			}
+			// TODO: do we have to enforce name equality between the main
+			// provider and a mixin (spec.Name == pulumiPackageSpec.Name)?
```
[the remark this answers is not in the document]
tmeckel: @iwahbe I wasn't sure if we really want to enforce this. Actually it makes totally sense, on the other hand: why not mix and match different providers into one single provider.

x/muxer/mapping.go
```diff
@@ -46,9 +46,8 @@ func MergeSchemasAndComputeDispatchTable(schemas []schema.PackageSpec) (Dispatch
 	muxedSchema.Config = schema.ConfigSpec{}
 	muxedSchema.Provider = schema.ResourceSpec{}
 
-	for i, s := range schemas {
-		i, s := i, s
-		dispatchTable.layerSchema(muxedSchema, &s, i)
+	for i := range schemas {
+		dispatchTable.layerSchema(muxedSchema, &schemas[i], i)
```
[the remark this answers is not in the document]
tmeckel: @iwahbe actually it was a linting issue: `G601: Implicit memory aliasing in for loop.` Do we need a test for this?

https://husni.dev/beware-of-implicit-memory-aliasing-in-go-foor-loop/

pkg/tfgen/generate_schema.go
```diff
@@ -252,7 +255,38 @@ func genPulumiSchema(pack *pkg, name tokens.Package, version string,
 		info:           info,
 		renamesBuilder: renamesBuilder,
 	}
-	return g.genPackageSpec(pack)
+	pulumiPackageSpec, err := g.genPackageSpec(pack)
+	if err != nil {
+		return pschema.PackageSpec{}, err
+	}
+
+	if info.MuxWith != nil {
```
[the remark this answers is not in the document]
tmeckel: @iwahbe had to make the `assertValid` method public (capitalized) because it's in the `tfbridge` package.

pkg/tfgen/generate_schema.go
```diff
@@ -252,7 +255,38 @@ func genPulumiSchema(pack *pkg, name tokens.Package, version string,
...
+		if info.MetadataInfo == nil {
+			return pschema.PackageSpec{},
+				fmt.Errorf("ProviderInfo.MetadataInfo is required and cannot be nil for muxed providers")
+		}
```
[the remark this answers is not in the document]
tmeckel: @iwahbe had to make the assertValid method public (capitalized) because it's in the tfbridge package.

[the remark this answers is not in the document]
tmeckel: @iwahbe changed my mind :-D now using `info.GetMetadata()` which implicitly calls `assertValid()`.

pkg/tfbridge/serve.go
tmeckel: @iwahbe First: What signature is required for the new method `newMuxWithProvider`? Currently the code relies completely on closures. They must be unwound using parameters of the new function. From my point of view we need a combination of `Serve` and `provider.Main`. Not sure if being able to test this satisfies the effort. The parameter list might be get long and ugly.

Second: how/when will the method being called? After `if len(opts.muxWith) > 0 {` or in some other way?

tmeckel:
> CC @blampe, @guineveresaenger This would provide a bridge supported way to stack a native Pulumi provider on top of a bridged provider... which currently describes pulumi-docker. I'd appreciate if you gave the API a once-over since you have experience in this area.
> 
@blampe, @guineveresaenger I created a demo of the new muxing feature using the Azure Classic provider.

https://github.com/tmeckel/pulumi-azure/commit/91e0c68eb302b734007d14faff99b419923a85d1

x/muxer/mapping.go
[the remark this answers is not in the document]
iwahbe: Hmmm. I wonder why you are hitting this. Maybe a different version of the linter. In that case, I don't think we need a test.

pkg/tfbridge/serve.go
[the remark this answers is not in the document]
iwahbe: I don't think this segment was in use, so I think it is safe to change. As soon as this becomes load bearing, it becomes impossible to change.

pkg/tfbridge/serve.go
iwahbe: I played around with this vis a vis testing. I came up with [this](https://github.com/tmeckel/pulumi-terraform-bridge/pull/1).

tmeckel: @iwahbe merged

tmeckel: I've to see with what I can come up with.

pkg/tfbridge/main.go
```diff
@@ -69,20 +67,7 @@ func Main(pkg string, version string, prov ProviderInfo, pulumiSchema []byte, op
...
 }
-
-type Option func(*opts)
-
-type opts struct {
-	muxWith []func(*provider.HostClient) (pulumirpc.ResourceProviderServer, error)
-}
-
-// This is an experimental API.
-func MuxWith(s func(*provider.HostClient) (pulumirpc.ResourceProviderServer, error)) Option {
```
t0yv0, line 84: This is not used anywhere? Should we go through a //deprecated step first before removing completely?

x/muxer/provider.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/pkg/v3/resource/provider"
+	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
+)
+
+// Provider defines an interface which must be implemented by providers
+// that shall be used as mixins of a wrapped Terraform provider
+type Provider interface {
```
t0yv0, line 11: This interface belongs in the main package perhaps, since it's part of defining ProviderInfo and sounds like we're committing to it, WDYT?

t0yv0:
This looks reasonable to me and cleans up the API a fair bit, thank you!

pkg/tfbridge/provider.go
```diff
@@ -168,6 +183,34 @@ func NewProvider(ctx context.Context, host *provider.HostClient, module string,
...
+		return nil, err
+	} else if found {
+		mapping = m
+	} else {
+		return nil, fmt.Errorf("missing pre-computed muxer mapping")
+	}
+
+	servers := []muxer.Endpoint{{
+		Server: func(host *provider.HostClient) (pulumirpc.ResourceProviderServer, error) {
+			return newProvider(context.Background(), host, module, version, info.P, info, pulumiSchema), nil
```
t0yv0, line 200: Document why context.Background() instead of `ctx` from the parameter?

t0yv0 (approved):

pkg/tfbridge/provider.go
tmeckel: @t0yv0 I merged a PR of @iwahbe -> https://github.com/tmeckel/pulumi-terraform-bridge/pull/1 I think that's an oversight of him, but perhaps there's a reason for this.

pkg/tfbridge/main.go
iwahbe: I can't [find any usage](https://github.com/search?q=tfbridge.MuxWith&type=code) and I'm not 100% sure that it worked in the first place (untested). The API is marked as experimental. I'm comfortable just removing it.

x/muxer/provider.go
tmeckel: @t0yv0 @iwahbe because the interface is only used when defining a mixin for a wrapped TF provider, I think it's reasonable to locate it in the `muxer` package. However if it shall be moved to the main package it must be renamed.

pkg/tfbridge/provider.go
t0yv0: I think before the refactor the outer code didn't have the context parameter. I think it does not matter *yet* but it's best to propagate `ctx` rather than initiate a new context chain - that should always be unusual and documented with a comment. Things like implicit loggers are passed via context sometimes in this codebase and this can get interesting.

t0yv0:
panic: TODO [recovered] in the tests - a bit more work left it looks like.

tmeckel:
> panic: TODO [recovered] in the tests - a bit more work left it looks like.

Obviously 🤣 Working on it! Provider testing, the next new dungeon in the Pulumi Bridge that I've to dig myself into.

pkg/tfbridge/provider.go
tmeckel: @t0yv0 I'd prefer to propagate the ctx down as well. As I said I think it's a copy-paste fault.

tmeckel: @iwahbe @t0yv0 propagating down `ctx` now

pkg/tfbridge/serve.go
tmeckel: @t0yv0 @iwahbe initial muxing tests implemented

blampe:
> > CC @blampe, @guineveresaenger This would provide a bridge supported way to stack a native Pulumi provider on top of a bridged provider... which currently describes pulumi-docker. I'd appreciate if you gave the API a once-over since you have experience in this area.
> 
> @blampe, @guineveresaenger I created a demo of the new muxing feature using the Azure Classic provider.
> 
> [tmeckel/pulumi-azure@91e0c68](https://github.com/tmeckel/pulumi-azure/commit/91e0c68eb302b734007d14faff99b419923a85d1)

Thanks for this @tmeckel! I haven't looked at the PR, only this example provider. Only one question -- why do we need this wrapper struct, shouldn't the inferred `gop.Provider` be sufficient for muxing?
```go
 type Provider struct {
 	prov gop.Provider
 }

 func NewProvider() *Provider {
 	inst := Provider{}
 	inst.prov = infer.Provider(infer.Options{
 		Functions: []infer.InferredFunction{
 			infer.Function[*core.GetClientToken, core.GetClientTokenArgs, core.GetClientTokenResult](),
 			infer.Function[*core.GetResources, core.GetResourcesArgs, core.GetResourcesResult](),
 		},
 		ModuleMap: map[tokens.ModuleName]tokens.ModuleName{
 			"mux": "index",
 		},
 	})
 	inst.prov.Configure = configure
 	return &inst
 }
```

More generally, the muxing in the docker provider isn't bad at all but if this is something we expect to do more regularly then it make sense to add some sugar like this.

t0yv0:
> the muxing in the docker provider isn't bad 

I would challenge that and count up P1s and deep bugs that bottom out at that muxing. More importantly testing the muxing facilities in the docker provider feels a bit misplaced so it's likely relatively shallow. I think it makes sense to treat it more like a "framework" thing that we need to maintain and test outside of a specific provider.

tmeckel:
> Thanks for this @tmeckel! I haven't looked at the PR, only this example provider. Only one question -- why do we need this wrapper struct, shouldn't the inferred `gop.Provider` be sufficient for muxing?
> 
> ```go
>  type Provider struct {
>  	prov gop.Provider
>  }
> 
>  func NewProvider() *Provider {
>  	inst := Provider{}
>  	inst.prov = infer.Provider(infer.Options{
>  		Functions: []infer.InferredFunction{
>  			infer.Function[*core.GetClientToken, core.GetClientTokenArgs, core.GetClientTokenResult](),
>  			infer.Function[*core.GetResources, core.GetResourcesArgs, core.GetResourcesResult](),
>  		},
>  		ModuleMap: map[tokens.ModuleName]tokens.ModuleName{
>  			"mux": "index",
>  		},
>  	})
>  	inst.prov.Configure = configure
>  	return &inst
>  }
> ```
> 
> More generally, the muxing in the docker provider isn't bad at all but if this is something we expect to do more regularly then it make sense to add some sugar like this.

Thanks for having a look at the muxing API @blampe! What you called a _wrapper struct_ is an interface to abstract away the implementation of a provider that shall be used as a mixin. 

Both `tfgen` and the `tfbridge` use the interface in two scenarios.

`tfgen` uses this interface to create the mixed schema and the pre-calculated dispatch table which contains a mapping between a Pulumi token and the (final) implementing provider of a token. This is done automatically and when I understand your code example correctly, a provider maintainer has to do this manually with your strategy. Not very convenient from my point of view. The `muxer.Provider` interface can be simply added to arbitrary provider to enable it to be used as a mixin, regardless what implementation has been used i.e. via the (new) Pulumi GO Provider SDK or via the classic way of implementing `UnimplementedResourceProviderServer`.

`tfbride` uses the `muxer.Provider` interface to initialize (start) the configured provider instances during startup. Using an interface here gives the provider maintainer the chance to perform any type of stuff before returning an instance of the provider. If we would initialize the provider directly in `tfbridge` this possibility isn't available.

Finally I'm a big fan of using interfaces when it comes to tight components together and thus I implemented it that way. 😀

pkg/tests/mux_with_test.go
```diff
@@ -0,0 +1,295 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
+	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+// func todo[T any]() T {
```
t0yv0, line 41: not needed looks like?

pkg/tests/mux_with_test.go
```diff
@@ -0,0 +1,295 @@
...
+		nil,          /* hostClient */
+		info.Name,    /* module */
+		info.Version, /* version */
+		info.P,       /* tf */
+		info,         /* info */
+		data,         /* pulumiSchema */
+	)
+	return
+}
+func TestMuxWithProvider(t *testing.T) {
```
t0yv0, line 122: Empty line before func?

tmeckel: @t0yv0 shall I apply `gofumpt` instead of addin an empty line?

pkg/tests/mux_with_test.go
tmeckel: Sloppy oversight... 👎🏼

pkg/tests/mux_with_test.go
t0yv0: Ah it does that? We never updated this codebase to gofumpt but it sounds like a good idea. Ignore for now I'll talk to folks.

tmeckel: @t0yv0 perhaps the bridge should then adopt the `.golangci.yml` of the `pulumi/pulumi` repo.

t0yv0: Yeah great idea. Let me file it up.

tmeckel: @t0yv0 don't be that excited! 😁 I tested this yesterday -> plenty of linting errors!

blampe:
> Finally I'm a big fan of using interfaces when it comes to tight components together and thus I implemented it that way. 😀

Of course. To rephrase my question, doesn't it make more sense for `pulumi-go-provider` to implement this interface instead of requiring the user to roll their own implementation?

tmeckel:
> > Finally I'm a big fan of using interfaces when it comes to tight components together and thus I implemented it that way. 😀
> 
> Of course. To rephrase my question, doesn't it make more sense for `pulumi-go-provider` to implement this interface instead of requiring the user to roll their own implementation?

@blampe ahhh ... totally misunderstood you. Sorry! Well from my point of view it should be an explicit decision to enable a Provider for mixin. If the GO Provider would implicitly implement the interface a provider maintainer can't make such a decision. As a compromise the GO provider could provider some sort of `WithMixinSupport` or the like. So that a provider maintainer only must add this method to get the implementation added.

pkg/tfgen/generate_schema.go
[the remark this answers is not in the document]
iwahbe: Thinking about it a bit more, I think that this will already be enforced by schema validation. `pulumiPackageSpec.Name` is not enforced (and doesn't need to be), but if the package component of tokens don't match up, schema validation will fail.

x/muxer/provider.go
iwahbe: @tmeckel I agree with @t0yv0 that this should be located in the `tfbridge` package. I have 2 reasons:
1. `x/muxer` is an implementation detail of the bridge, and shouldn't need to be referenced by external users.
2. I think interfaces should be defined where they are consumed or provided, and `x/muxer.Provider` is neither.

pkg/tests/mux_with_test.go
iwahbe: Let's fix this before merging.

iwahbe (approved):
I left comments for 3 small changes. After that I think we are ready to merge!

iwahbe:
> > > Finally I'm a big fan of using interfaces when it comes to tight components together and thus I implemented it that way. 😀
> > 
> > 
> > Of course. To rephrase my question, doesn't it make more sense for `pulumi-go-provider` to implement this interface instead of requiring the user to roll their own implementation?
> 
> @blampe ahhh ... totally misunderstood you. Sorry! Well from my point of view it should be an explicit decision to enable a Provider for mixin. If the GO Provider would implicitly implement the interface a provider maintainer can't make such a decision. As a compromise the GO provider could provider some sort of `WithMixinSupport` or the like. So that a provider maintainer only must add this method to get the implementation added.

I played around with implementing `muxer.Provider` for a generic `pulumi-go-provider` provider. It makes me think we should change the interface a bit. The obvious implementation looks like this:

```go
func (p *Provider) GetSpec(ctx context.Context, name, version string) (schema.PackageSpec, error) {
	return GetSchema(ctx, name, version, p.WithDefaults())
}
func (p *Provider) GetInstance(_ context.Context, name, version string, host *pprovider.HostClient) (rpc.ResourceProviderServer, error) {
	return newProvider(name, version, p.WithDefaults())(host)
}
```

I propose changing the `muxer.Provider` interface to:

```go
type Provider interface {
	GetSpec(ctx context.Context, name, version string) (schema.PackageSpec, error)
	GetInstance(ctx context.Context, name, version string, host *pprovider.HostClient) (rpc.ResourceProviderServer, error)
}
```

Passing a `context.Context` is generally a good idea, and provides a helpful side-channel for passing information. It makes the interface much more future-proof.

Passing `name, version string` avoids us needing to copy in the `name` and `version` another time.

x/muxer/provider.go
tmeckel: @iwahbe @t0yv0 I think the interface should then be renamed like `MuxProvider`, `ProviderMux`, `MixinProvider`. Right?

tmeckel: @iwahbe @t0yv0 I moved the intreface to the `pkg/tfbridge` package and renamed it to `ProviderMixin`, but that really stutters. We should find a common sense on mix, mux , muxing, mixing! 😁 

image

iwahbe: The field is called `MuxWith`, and under the hood it uses a `muxer`, so I think I would go with `mux` instead of `mixin`. Since I think this interface will only be used (at least externally to the bridge) with this field, I would call it `MuxWithProvider`.

Naming is hard ([`AugmentShimWithDisjointPF`](https://github.com/pulumi/pulumi-terraform-bridge/blob/ba7cc1d3083f782df65129423fa72e8cdf7eaac6/pf/internal/muxer/muxer.go#L64)), so I would love to hear @t0yv0's opinion on this.

t0yv0: Naming is hard. The problem with "mixin" word is that in Pulumi this means something already like we have hand-written node code like https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/s3/s3Mixins.ts that mixes into the generated SDK. So that term is taken. 

The word "mux" is entering our lexicon from TF prior art (https://github.com/hashicorp/terraform-plugin-mux). between "mix" and "mux" I'd lean to "mux".

tmeckel: @t0yv0 Okay, well then I'd propose `MuxWith []MuxProvider` or `MuxWith []ProviderMux`. I prefer the first alternative, because it makes clear for what purpose the interface has and when it'll be used then.

CC @iwahbe

iwahbe: I also prefer `MuxWith []MuxProvider`. That sounds good to me.

pkg/tfgen/generate_schema.go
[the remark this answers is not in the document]
tmeckel: Done

x/muxer/provider.go
tmeckel: @iwahbe @t0yv0 renamed to `MuxProvider`

pkg/tests/mux_with_test.go
tmeckel: Done

pkg/tests/mux_with_test.go
tmeckel: Done

tmeckel:
@t0yv0 @iwahbe hopefully the last update before merging

pkg/tfbridge/provider.go
```diff
@@ -66,6 +69,16 @@ type Provider struct {
 	memStats        memStatCollector
 }
 
+// ProviderMixin defines an interface which must be implemented by providers
```
t0yv0, line 72: Self-reference wasn't renamed in the doc comment.

t0yv0 (approved):

pkg/tfbridge/provider.go
tmeckel: Fixed

iwahbe (approved):

guineveresaenger:
@tmeckel happy new year and sorry I missed this over the holidays - thank you for this work, I'm super excited to take it for a spin. 💟

PR #1610, opened 2024-01-02
Continue configuring in the normal way

t0yv0:
@blampe points out in https://github.com/pulumi/pulumi-go-provider/pull/172 it's might not be necessary to marshal with configencoding once we get past CheckConfig.  Interesting idea. This works fine for a simple program on custom AWS build:

```
name: aws-yaml-configs
runtime: yaml
description: A minimal AWS Pulumi YAML program
outputs:
  # Export the name of the bucket
  bucketName: ${my-bucket.id}
resources:
  provider:
    type: pulumi:providers:aws
    properties:
      defaultTags:
        tags:
          foo: bar
    options:
      version: 6.17.0
  my-bucket:
    type: aws:s3:Bucket
    options:
      provider: ${provider}
``` 

Let me think of further testing. This can improve granularity of secret tracking.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -369,7 +369,13 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
 	secretNews := MarkSchemaSecrets(ctx, p.config, p.info.Config, resource.NewObjectProperty(news)).ObjectValue()
 
 	// In case news was modified by pre-configure callbacks, marshal it again to send out the modified value.
-	newsStruct, err := configEnc.MarshalProperties(secretNews)
+	newsStruct, err := plugin.MarshalProperties(secretNews, plugin.MarshalOptions{
+		Label:        "config",
+		KeepUnknowns: true,
+		SkipNulls:    true,
+		KeepSecrets:  true,
+		RejectAssets: true,
+	})
 	if err != nil {
 		return nil, err
 	}


pf/tests/provider_checkconfig_test.go
```diff
@@ -667,7 +674,7 @@ func TestPreConfigureCallback(t *testing.T) {
 		  "response": {
 		    "inputs": {
 		      "version": "6.54.0",
-	              "configValue": "04da6b54-80e4-46f7-96ec-b56ff0331ba9"
+	              "configValue": "dd056dcd-154b-4c76-9bd3-c8f88648b5ff"
```
t0yv0, line 677: This is now UnknownObjectValue = "dd056dcd-154b-4c76-9bd3-c8f88648b5ff" ... ever so slightly weird.

pf/tests/provider_checkconfig_test.go
```diff
@@ -439,13 +440,19 @@ func TestCheckConfig(t *testing.T) {
...
-	              "batching": {
-	                "4dabf18193072939515e22adb298388d": "1b47061264138c4ac30d75fd1eb44270",
-	                "value": "{\"enableBatching\":true,\"sendAfter\":\"1s\"}"
-	              },
-	              "scopes": "[\"a\",\"b\"]",
-	              "version": "6.54.0"
-	            }
+                  "batching": {
+                    "enableBatching": true,
+                    "sendAfter": {
```
t0yv0, line 445: This is great - improved granularity of secret marking.

blampe (approved):

Issue #1626, opened 2024-01-18
Talos provider fails to build on latest bridge with type conflict


t0yv0:
### What happened?

From @ringods :

 I updated my Talos package to use the latest tf-bridge commit. I want to start testing the muxer that Thomas Meckel contributed. My go.mod changes:
    github.com/pulumi/pulumi-terraform-bridge/pf v0.23.1-0.20240111163851-bda33f369ae3
	github.com/pulumi/pulumi-terraform-bridge/v3 v3.70.1-0.20240111163851-bda33f369ae3

But my provider doesn't build anymore:
error: failed to create Pulumi schema: talos:machine/generated/MachineSecretsSecrets:MachineSecretsSecrets: conflicting type definitions

The error is introduced in https://github.com/pulumi/pulumi-terraform-bridge/pull/1550 but does not seem to give sufficient guidance how to resolve the type collision. 

### Example

See above

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
Confirmed with a minified repro https://github.com/pulumi/pulumi-terraform-bridge/pull/1550 broke simple ExtraTypes examples that unfortunately were untested in this codebase. Let me look at reverting it.

t0yv0:
Looks like per https://github.com/pulumi/pulumi-terraform-bridge/pull/1628#issuecomment-1899167921 there might be an additional error in the new version of OCI that might have a similar but slightly different root cause CC @iwahbe

iwahbe:
@t0yv0 Can you post the minified repro?

t0yv0:
It's in a PR.

iwahbe:
This was fixed by reverting https://github.com/pulumi/pulumi-terraform-bridge/pull/1628.

PR #1627, opened 2024-01-18
Regress 1626

t0yv0:
See #1626 - this is a minified repro .


iwahbe (approved):
Let's merge this as a regression test. I'm going to re-add add https://github.com/pulumi/pulumi-terraform-bridge/pull/1550, and we want to make sure that we don't regress again.

t0yv0:
Sure sounds good!

PR #1628, opened 2024-01-18
Revert "Allow the `Type` override to effect downstream types"

t0yv0:
Reverts pulumi/pulumi-terraform-bridge#1550

Discovered that ExtraTypes support is broken by this change per https://github.com/pulumi/pulumi-terraform-bridge/issues/1626

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -426,10 +426,6 @@ type SchemaInfo struct {
 	CSharpName string
 
 	// a type to override the default; "" uses the default.
-	//
-	// If the type overridden is an object type, `Type: newName` is interpreted as a
-	// move operation, pointing the property to a type called `newName` and creating
-	// `newName` with the schema described by this type.
 	Type tokens.Type
 
 	// alternative types that can be used instead of the override.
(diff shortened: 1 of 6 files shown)


iwahbe (approved):
I agree with the revert to unblock talos and oci. That said, `ExtraTypes` isn't the only thing wrong here. pulumi-oci is blocked on the same [error message](https://github.com/pulumi/pulumi-oci/actions/runs/7550497575/job/20556231046), but it doesn't specify `ExtraTypes`.

t0yv0:
Can you please capture everything relevant in the tracker. Appreciated.

iwahbe:
> Can you please capture everything relevant in the tracker. Appreciated.
> 
> As far as I can see OCI built on [pulumi/pulumi-oci#327](https://github.com/pulumi/pulumi-oci/pull/327) without an issue.

The current OCI version builds correctly, but the [new OCI version](https://github.com/pulumi/pulumi-oci/issues/325) [does not](https://github.com/pulumi/pulumi-oci/actions/runs/7550497575).


Screenshot 2024-01-18 at 12 36 10 PM

t0yv0:
OK I captured this in 1626 for now, we can split later on if it turns out a different issue. Thanks! Friendly reminder we need to track things..

PR #1639, opened 2024-01-23
Refactor IgnoreChanges

t0yv0:
The intent here is to refactor the code without any observable changes to make possible further extensions such as https://github.com/pulumi/pulumi-terraform-bridge/pull/1614 .

Specifically, before this PR:

```go

type Provider interface {
        ...
	Diff(t string, s InstanceState, c ResourceConfig) (InstanceDiff, error)
}

type InstanceDiff interface {
        ...
	IgnoreChanges(ignored map[string]bool)
}

```

That is the interface is to compute a diff first, and then side-effect to ignore changes. 

After this PR we have this instead:

```go
type Provider interface {
        ...
	Diff(t string, s InstanceState, c ResourceConfig, opts DiffOptions) (InstanceDiff, error)
}
type IgnoreChanges = func() map[string]struct{}
type DiffOptions struct {
  IgnoreChanges IgnoreChanges
}
```

So the algorithm to ignore changes is specified upfront when constructing the diff. This seems to be useful to un-constrain the implementation of shim.Provider a little bit, specifically in #1614 that is trying to change the internal representation of InstanceDiff implementation in such a way that it is inconvenient to process IgnoreChanges after the InstanceDiff is constructed already.

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -252,12 +252,12 @@ func newIgnoreChanges(
 		return nil
 	}
 	return func() map[string]struct{} {
-		return doIgnoreChanges(ctx, tfs, ps, olds, news, ignoredPaths)
+		return computeIgnoreChanges(ctx, tfs, ps, olds, news, ignoredPaths)
 	}
 }
 
 // Computes the ignored key set.
-func doIgnoreChanges(
+func computeIgnoreChanges(
 	ctx context.Context,
 	tfs shim.SchemaMap,
 	ps map[string]*SchemaInfo,
@@ -284,6 +284,25 @@ func doIgnoreChanges(
 		visitPropertyValue(ctx, en, string(k), v, etf, eps, shimutil.IsOfTypeMap(etf), visitor)
 	}
 	return ignoredKeySet
+}
+
+func doIgnoreChanges(
+	ctx context.Context,
+	tfs shim.SchemaMap,
+	ps map[string]*SchemaInfo,
+	olds, news resource.PropertyMap,
+	ignoredPaths []string,
+	diff shim.InstanceDiff,
+) {
+	if diff == nil {
+		return
+	}
+	ignored := computeIgnoreChanges(ctx, tfs, ps, olds, news, ignoredPaths)
+	m := map[string]bool{}
+	for k := range ignored {
+		m[k] = true
+	}
+	diff.IgnoreChanges(m)
 }
 
 // makeDetailedDiff converts the given state (olds), config (news), and InstanceDiff to a Pulumi property diff.
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -241,22 +241,37 @@ func makePropertyDiff(ctx context.Context, name, path string, v resource.Propert
 	visitPropertyValue(ctx, name, path, v, tfs, ps, rawNames, visitor)
 }
 
-func doIgnoreChanges(ctx context.Context, tfs shim.SchemaMap, ps map[string]*SchemaInfo,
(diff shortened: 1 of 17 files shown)


t0yv0:
Opening for early feedback. We don't have to merge yet I guess until I have an end to end demonstration that 1614 is viable on top of this (I will rebase). But this seems to me like it could help.

pkg/tfshim/shim.go
```diff
@@ -42,6 +42,9 @@ type InstanceDiff interface {
 	Destroy() bool
 	RequiresNew() bool
 
+	// Deprecated. Instead of constructing an InstanceDiff object and then calling IgnoreChanges
+	// on it, please pass IgnoreChanges option to Diff so that the ignores are processed as part
+	// of constructing InstanceDiff.
 	IgnoreChanges(ignored map[string]bool)
```
iwahbe, line 48: Is there a reason we are not just removing This? We are already making a breaking change to all consumers of `shim/*` (changing the method signature of `Diff` and `DiffWithContext`.

---

If we decide to keep, let's use `:` instead of `.`, since that triggers the `staticcheck` linter.

```suggestion
	// Deprecated: Instead of constructing an InstanceDiff object and then calling IgnoreChanges
	// on it, please pass IgnoreChanges option to Diff so that the ignores are processed as part
	// of constructing InstanceDiff.
	IgnoreChanges(ignored map[string]bool)
```

pkg/tfshim/shim.go
```diff
@@ -213,14 +216,49 @@ type ProviderWithContext interface {
...
+	IgnoreChanges IgnoreChanges
+}
+
+func NewDiffOptions(opts ...DiffOption) DiffOptions {
+	options := DiffOptions{}
+	for _, o := range opts {
+		o(&options)
+	}
+	return options
+}
```
iwahbe, line 241: Nit: I think that this pattern is beneficial when you want composable, nested, and opaque settings. In this case, both sides of the contract are public: `DiffOption` and `DiffOptions`. I'm not sure what the benefit is here over just having the caller pass `DiffOptions` directly.

pkg/tfbridge/diff.go
```diff
@@ -268,8 +283,26 @@ func doIgnoreChanges(ctx context.Context, tfs shim.SchemaMap, ps map[string]*Sch
 		en, etf, eps := getInfoFromPulumiName(k, tfs, ps, false)
 		visitPropertyValue(ctx, en, string(k), v, etf, eps, shimutil.IsOfTypeMap(etf), visitor)
 	}
+	return ignoredKeySet
+}
 
-	tfDiff.IgnoreChanges(ignoredKeySet)
+func doIgnoreChanges(
```
iwahbe, line 289: I couldn't find where this function is actually called (except in tests). Can we remove it or move it into a `*_test.go` file?

iwahbe (commented):
I left a couple of comments, but I'm over all in favor of a refactor like this.

pkg/tfshim/shim.go
t0yv0: Ah right, diffOptions shouldn't be public probably, I was in a bit of hurry.

pkg/tfshim/shim.go
t0yv0: Well adding functional options doesn't break callers just implementers but yeah. I'm mostly thinking of not causing issues for Cosmic but perhaps that's a moot concern. If we can take breaking changes over these internals that's good.

pkg/tfbridge/diff.go
t0yv0: Yeah moving to _test is good. I left it around to not rewrite the tests.

t0yv0:
@iwahbe PTAL, I simplified this a bit and removed old way of doing things instead of deprecating it.

iwahbe (approved):
LGTM

Issue #1541, opened 2023-11-27
Cleanup ProviderWithContext


t0yv0:
In https://github.com/pulumi/pulumi-terraform-bridge/pull/1537 we introduced a careful refactor that adds shim.ProviderWithContext interface to assist in context.Context object propagation and get rid of context.TODO().  This was done to make the PR minimally scoped and avoid a technically breaking change, however the desired end state is to actually take the slightly breaking change and make shim.Provider interfaces require context.Context where needed. There is a pretty good chance that this change although technically breaking will not have a big effect on consuming repositories as shim.Provider is used mostly internally.

PR #1642, opened 2024-01-25
Breaking: require context.Context in shim.Provider

t0yv0:
This change should not break bridged provider users or authors, but is technically a small breaking change in the pkg/tfshim module.

The shim.Provider interface methods now requires a context.Context as the first argument.

The shim.ProviderWithContext compatibility interface is now removed. Use shim.Provider.

The shim.Provider NewDestroyDiff method now additionally accepts the resource token:

        // Create a Destroy diff for a resource identified by the TF token t.
        NewDestroyDiff(ctx context.Context, t string) InstanceDiff

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1541

### DIFF
--- a/pf/internal/schemashim/provider.go
+++ b/pf/internal/schemashim/provider.go
@@ -59,67 +59,82 @@ func (p *SchemaOnlyProvider) DataSourcesMap() shim.ResourceMap {
 	return &schemaOnlyDataSourceMap{dataSources}
 }
 
-func (p *SchemaOnlyProvider) Validate(c shim.ResourceConfig) ([]string, []error) {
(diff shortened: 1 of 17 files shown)


t0yv0:
Decided to try this before https://github.com/pulumi/pulumi-terraform-bridge/pull/1639 - if we're OK with small internally-breaking changes we should do this cleanup first.

iwahbe (approved):
Let's do it.

I don't expect that external users will be working with the internals of `shim.Provider`, so I think this is a safe change to make.

It's probably worth adding a doc comment to `package shim` saying that this package is for internal use, and may experience breaking changes. That has always been the case, but it would be helpful to make it explicit.

PR #1643, opened 2024-01-25
Package comment noting that tfshim is considered internal

t0yv0:
Add a package comment explaining the purpose of the tfshim package and our policy on accepting breaking changes in it without bumping the bridge major version.


iwahbe (approved):
Thanks!

PR #1645, opened 2024-01-25
Fix master

t0yv0:
Recent commit merged poorly and the tests are no longer working. After the deprecation and move to ithub.com/pulumi/providertest/replay a tests needs to be adjusted.


iwahbe (approved):
Thanks for taking this fix.

PR #1650, opened 2024-01-26
Get extra nested docs from schema map

### LINKED ISSUE #1045: Bug: tfgen does not include descriptions for nested config fields
### What happened?

A [user reported missing examples for the Docker provider configuration](https://github.com/pulumi/pulumi-docker/issues/550) and [I noticed that even the descriptions were missing](https://www.pulumi.com/registry/packages/docker/api-docs/provider/#providerregistryauth).

After failing to create an upstream patch against an `index.md` file, I realized that unlike for Resources/Data Sources, for Provider.Config, we read descriptions (and only descriptions) from the [TF schema directly](https://github.com/kreuzwerker/terraform-provider-docker/blob/master/internal/provider/provider.go#L69).

The `Description` fields exist for the nested resource; we are just not consuming them.

### Expected Behavior

I expected nested schema descriptions to be read into the provider schema correctly.

### Steps to reproduce

1. Patch `upstream/internal/provider/provider.go` on the Docker provider - edit the description for `ca_material`, and edit the one for `registry_auth:address`
2. With the new patch in place, run `make tfgen`
3. Observe `caMaterial` having an updated description, while `registryAuth.Adddress` has no description at all.

### Output of `pulumi about`

n/a

### Additional context

_No response_

### Contributing

guineveresaenger:
This pull request aims to improve accuracy for nested property docs descriptions.

It uses the SchemaMap to look up and read `Description` fields [such as this one](https://github.com/kreuzwerker/terraform-provider-docker/blob/master/internal/provider/provider.go#L111) in the case where trying to obtain documentation form `entityDocs` does not yield any results.

The original impetus for this change was to address #1045, which is very specific to Config() fields. The Provider configuration docs are special-cased in the bridge and we do not have access to any `entityDocs`. Instead, we were reading in the TF schema's Description() field into `rawdocs` when generating a top-level config variable. This did not persist into any nested fields on Configuration variables, hence this fix. But it turns out that we can catch quite a few more missing descriptions if we extend implementation across all nested fields (see stats below).

As a side note - the Terraform Plugin Framework implements `Description()` in a way that respects and prioritizes descriptions that come from Markdown docs; for `sdkv2` resources we still need to use, and prefer, our parsed-from-markdown `entityDocs` descriptions. It would be interesting to see if we get improved docs for TFPF resources by not reading descriptions from `entityDocs`.
It was also necessary to return an empty string from TFPF `typeSchema` so as not to cause a panic during schemagen on TFPF using providers. Open to improvements/suggestions here!

This PR aims to be fully additive - we are not changing the way in which docs are generated; we only fill in missing docs.
Nevertheless, this PR does also seem to improve the correctness of nested docs where some description fields have been matched to the wrong input field.

Please see a [new Docker schema generated against these changes](https://github.com/pulumi/pulumi-docker/blob/nested-configs-sample-schema/provider/cmd/pulumi-resource-docker/schema.json)
Quick stats:
```
pulumi/pulumi-docker🦉 git diff master --stat -- provider/cmd
 provider/cmd/pulumi-resource-docker/schema.json | 111 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------------------------
 1 file changed, 68 insertions(+), 43 deletions(-)
 ```

Here is [an AWS schema](https://github.com/pulumi/pulumi-aws/blob/testing-docsgen/provider/cmd/pulumi-resource-aws/schema.json) - you will want to pull that branch and look at the git diff to `master` locally.
Quick stats:
```
pulumi/pulumi-aws🦉 git diff master --stat -- provider/cmd 
 provider/cmd/pulumi-resource-aws/schema.json | 1810 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------------------
 1 file changed, 1206 insertions(+), 604 deletions(-)
 ```

This PR fixes #1045.

### DIFF
--- a/internal/testprovider/schema_nested_descriptions.go
+++ b/internal/testprovider/schema_nested_descriptions.go
@@ -66,8 +66,7 @@ func resourceNestedDescriptionsSchema() map[string]*schema.Schema {
 						Description: "Version of the ruleset to deploy.",
 					},
 					"action_parameters": {
-						Type: schema.TypeList,
-						// MaxItems:    1,
+						Type:        schema.TypeList,
 						Optional:    true,
 						Description: "List of parameters that configure the behavior of the ruleset rule action.",
 						Elem: &schema.Resource{
--- a/pf/internal/schemashim/object_type_test.go
+++ b/pf/internal/schemashim/object_type_test.go
@@ -155,7 +155,6 @@ func assertHasSimpleObjectAttributes(t *testing.T, r shim.Resource) {
 	assert.True(t, r.Schema().Get("r").Required(), "r is required")
 	assert.True(t, r.Schema().Get("co").Computed() && r.Schema().Get("co").Optional(), "co is computed and optional")
 	assert.Equal(t, r.Schema().Get("desc").Description(), "I am a description")
-
 }
 
 func assertIsObjectType(t *testing.T, shimmed shim.Schema) {
--- a/pkg/tfgen/internal/testprovider/nesteddescriptions.go
+++ b/pkg/tfgen/internal/testprovider/nesteddescriptions.go
@@ -15,30 +15,12 @@
 package testprovider
 
 import (
-	"unicode"
-
+	testproviderdata "github.com/pulumi/pulumi-terraform-bridge/v3/internal/testprovider"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	shimv2 "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/sdk-v2"
-	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
-
-	testproviderdata "github.com/pulumi/pulumi-terraform-bridge/v3/internal/testprovider"
 )
 
 func ProviderNestedDescriptions() tfbridge.ProviderInfo {
-
-	member := func(mod string, mem string) tokens.ModuleMember {
-		return tokens.ModuleMember("cloudflare" + ":" + mod + ":" + mem)
-	}
-
-	typ := func(mod string, typ string) tokens.Type {
-		return tokens.Type(member(mod, typ))
-	}
-
-	resource := func(mod string, res string) tokens.Type {
-		fn := string(unicode.ToLower(rune(res[0]))) + res[1:]
-		return typ(mod+"/"+fn, res)
-	}
-
 	return tfbridge.ProviderInfo{
 		P:           shimv2.NewProvider(testproviderdata.ProviderNestedDescriptions()),
 		Name:        "cloudflare",
@@ -49,7 +31,7 @@ func ProviderNestedDescriptions() tfbridge.ProviderInfo {
 		Repository:  "https://github.com/pulumi/pulumi-cloudflare",
 		Resources: map[string]*tfbridge.ResourceInfo{
 			"cloudflare_ruleset": {
-				Tok: resource("index", "Ruleset"),
+				Tok: tfbridge.MakeResource("cloudflare", "index", "Ruleset"),
 				Fields: map[string]*tfbridge.SchemaInfo{
 					"rules": {
 						Elem: &tfbridge.SchemaInfo{
--- a/pkg/tfgen/internal/testprovider/defaultinfo.go
+++ b/pkg/tfgen/internal/testprovider/defaultinfo.go
@@ -15,29 +15,12 @@
 package testprovider
 
 import (
-	"unicode"
-
+	testproviderdata "github.com/pulumi/pulumi-terraform-bridge/v3/internal/testprovider"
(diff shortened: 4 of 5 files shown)


iwahbe:
I can see the docker diff in GH: https://github.com/pulumi/pulumi-docker/compare/master...nested-configs-sample-schema.

I believe that the corrections come from (downstream from the change):

https://github.com/pulumi/pulumi-terraform-bridge/blob/6dc3163b8e1d04f452c7eb851b6ef0b27c7062e0/pkg/tfgen/generate_schema.go#L567-L573

pkg/tfgen/generate.go
```diff
@@ -456,6 +460,10 @@ func (g *Generator) makeObjectPropertyType(typePath paths.TypePath,
 		// This seems wrong, so we ignore the second return value here for now.
 		doc, _ := getNestedDescriptionFromParsedDocs(entityDocs, objPath.join(key))
 
+		// If we have no result from entityDocs, we look up the TF schema Description.
+		if doc == "" {
```
iwahbe, line 464: I would love to see a test that includes elided text from a `.Description()`, and then shows that our elision and replace logic applies to `.Description()` based descriptions. I'm pretty sure that neither currently apply, since their application are tied into parsing `entityDocs`.

Untangling the replace logic will be painful, so I don't believe it is worth blocking this PR. I think we need to make sure that [`reformatText`](https://github.com/pulumi/pulumi-terraform-bridge/blob/c85c6b160505d6ababf9d898204e88c7df8292c4/pkg/tfgen/docs.go#L1967-L1969) is called on `.Description()` based properties.

Looking at https://github.com/pulumi/pulumi-docker/compare/master...nested-configs-sample-schema, we can actually see several instances where normally elided text was added.

iwahbe (commented):
This is a big improvement, and I'm looking forward to getting this merged. I think we need to address one additional category of errors before merging.

pf/internal/schemashim/type_schema.go
```diff
@@ -98,8 +98,8 @@ func (*typeSchema) DefaultValue() (interface{}, error) {
 	return nil, bridge.ErrSchemaDefaultValue
 }
 
-func (*typeSchema) Description() string {
-	panic("Description() should not be called during schema generation")
+func (s *typeSchema) Description() string {
```
t0yv0, line 101: Nit: "s" not needed.

pkg/tfgen/generate.go
t0yv0: Yeah looks like editing the test bridge *indirectly* covers the functionality, I can't recall do we have an assert somewhere that regenerating test bridge matches the checked in schema? We might have that..  So technically this is covered - but just to help our future selves I'd also love to see a simple dedicated test that asserts the desired description behavior.

guineveresaenger: I created a dedicated test provider that ensures that these nested descriptions are modified as expected. The granular behavior of `reformatText` is already under test of course.

pkg/tfgen/internal/testprovider/nesteddescriptions.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	typ := func(mod string, typ string) tokens.Type {
+		return tokens.Type(member(mod, typ))
+	}
+
+	resource := func(mod string, res string) tokens.Type {
+		fn := string(unicode.ToLower(rune(res[0]))) + res[1:]
+		return typ(mod+"/"+fn, res)
+	}
```
iwahbe, line 40: Nit: There is a `tfbridge.MakeResource(pkg, mod, res)` function which does this already.

internal/testprovider/schema_nested_descriptions.go
```diff
@@ -0,0 +1,93 @@
...
+						Description: "Unique rule identifier.",
+					},
+					"version": {
+						Type:        schema.TypeString,
+						Computed:    true,
+						Description: "Version of the ruleset to deploy.",
+					},
+					"action_parameters": {
+						Type: schema.TypeList,
+						// MaxItems:    1,
```
iwahbe, line 70:
```suggestion
```

pf/internal/schemashim/object_type_test.go
```diff
@@ -143,6 +154,8 @@ func assertHasSimpleObjectAttributes(t *testing.T, r shim.Resource) {
 	assert.True(t, r.Schema().Get("c").Computed(), "c is computed")
 	assert.True(t, r.Schema().Get("r").Required(), "r is required")
 	assert.True(t, r.Schema().Get("co").Computed() && r.Schema().Get("co").Optional(), "co is computed and optional")
+	assert.Equal(t, r.Schema().Get("desc").Description(), "I am a description")
+
```
iwahbe, line 158:
```suggestion
```

iwahbe (approved):
A couple minuscule formatting nits. Otherwise LGTM!

PR #1656, opened 2024-01-29
Test schemashim against the interface

t0yv0:


t0yv0:
https://github.com/pulumi/pulumi-terraform-bridge/issues/1657 looks like the issue is by design.

PR #1659, opened 2024-01-30
[PF] Call `Read` after `ImportState` when importing a resource

### LINKED ISSUE #793: [Plugin Framework] Add a test for Read handling a not-found resource
## Hello!

## Issue details

Read provider method is called for `pulumi refresh` and `pulumi import`, and one particular interesting error case is what happens when the resource being sought is not found in the cloud anymore. The task is to add tests for the PF implementation locking in reasonable behavior for the not-found Read.

### LINKED ISSUE #1655: [PF] Pulumi's `Read` does not invoke TF's `Read` during an import
### What happened?

When the PF implements Pulumi's `Read`, it calls TF's ImportState and then returns. According to TF's documentation on the resource lifecycle, it should also call `Read`:

According to [upstream's documentation](https://developer.hashicorp.com/terraform/plugin/framework/resources/read):

>Read (refresh) is part of the basic Terraform lifecycle for managing resources. During the [terraform apply](https://developer.hashicorp.com/terraform/cli/commands/apply), [terraform plan](https://developer.hashicorp.com/terraform/cli/commands/plan), and [terraform refresh](https://developer.hashicorp.com/terraform/cli/commands/refresh) commands, Terraform calls the provider [ReadResource](https://developer.hashicorp.com/terraform/plugin/framework/internals/rpcs#readresource-rpc) RPC, in which the framework calls the resource.Resource interface Read method. The Read method is also executed after [resource import](https://developer.hashicorp.com/terraform/plugin/framework/resources/import).

### Example

The PR that motivated this discovery was https://github.com/pulumi/pulumi-aws/issues/3308. That issue saw an `import` on a missing resource panic or error out instead of giving a "missing resource" failure.

Root causing the problem, the TF Import succeeded (as a no-op), and then the bridge returned without calling TF's Read. Pulumi, seeing the Import succeed, called `Check` on invalid data, causing the TF provider to panic.

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

iwahbe:
This PR makes two changes to PF's import semantics:
1. It calls `ReadResource` after `ImportResourceState`.
2. It checks for an empty state (TF's signal that a `Read` found no resource) and returns appropriately.

Fixes #1655 
Fixes https://github.com/pulumi/pulumi-aws/issues/3308
Fixes #793

### DIFF
--- a/pf/internal/convert/convert.go
+++ b/pf/internal/convert/convert.go
@@ -70,6 +70,9 @@ func DecodePropertyMap(dec Decoder, v tftypes.Value) (resource.PropertyMap, erro
 	if err != nil {
 		return nil, err
 	}
+	if pv.IsNull() {
+		return nil, nil
+	}
 	if !pv.IsObject() {
 		return nil, fmt.Errorf("Expected an Object, got: %v", pv.String())
 	}
--- a/pf/internal/convert/convert.go
+++ b/pf/internal/convert/convert.go
@@ -70,15 +70,6 @@ func DecodePropertyMap(dec Decoder, v tftypes.Value) (resource.PropertyMap, erro
 	if err != nil {
 		return nil, err
 	}
-
-	// Pulumi's property value representation doesn't do a good job distinguishing
-	// between the empty property map (PropertyMap{}) and a nil value
-	// (resource.NullProperty).
-	//
-	// We allow converting a null property into a nil map.
-	if pv.IsNull() {
-		return nil, nil
-	}
 	if !pv.IsObject() {
 		return nil, fmt.Errorf("Expected an Object, got: %v", pv.String())
 	}
--- a/pf/internal/convert/convert.go
+++ b/pf/internal/convert/convert.go
@@ -70,6 +70,12 @@ func DecodePropertyMap(dec Decoder, v tftypes.Value) (resource.PropertyMap, erro
 	if err != nil {
 		return nil, err
 	}
+
+	// Pulumi's property value representation doesn't do a good job distinguishing
+	// between the empty property map (PropertyMap{}) and a nil value
+	// (resource.NullProperty).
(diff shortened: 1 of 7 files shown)


pf/internal/convert/convert.go
```diff
@@ -70,6 +70,7 @@ func DecodePropertyMap(dec Decoder, v tftypes.Value) (resource.PropertyMap, erro
 	if err != nil {
 		return nil, err
 	}
+
```
t0yv0, line 73: Needless diff?

pf/tests/provider_read_test.go
```diff
@@ -188,22 +188,46 @@ func TestImportingResourcesWithBlocks(t *testing.T) {
...
             }
           }
-	}`
+        }`
+	testutils.Replay(t, server, testCase)
+}
+
+// Check that importing a resource that does not exist returns an empty property bag and
+// no ID.
+func TestImportingMissingResources(t *testing.T) {
```
t0yv0, line 216: Good!

pf/internal/convert/convert.go
iwahbe: I had a change there that turned out to be unnecessary. Fixed.

t0yv0:
This is great! I was just appending the same tests (in spirit) to https://github.com/pulumi/pulumi-terraform-bridge/pull/1614 - to cover import found/not-found and refresh found/not found. It turns out that in SDKV2 bridge the pulumi Read method is performing tf.Import and then unconditionally falling through to tf.ReadResource. So yes the status quo is to always do the Read. What I find peculiar is that the import logic may indicate not-found in which pulumi Read short-circuits. This may not be necessary here though but some food for thought.

t0yv0 (approved):
🚀

PR #1663, opened 2024-01-31
Improve error messages for Invalid or unknown key

t0yv0:
Per https://github.com/pulumi/pulumi/issues/3546 it is possible that when a user specifies `aws:foo` config in a Pulumi program she is not intending it to be picked up by the aws provider. This change improves the error message emitted in this case to guide the user better.


pkg/tfbridge/provider_test.go
```diff
@@ -1118,8 +1149,8 @@ func TestCheckConfig(t *testing.T) {
 		})
 		require.NoError(t, err)
 		require.Equal(t, 1, len(resp.Failures))
-		require.Equal(t, "could not validate provider configuration: "+
-			"Invalid or unknown key. Examine values at 'explicitprovider.requiredprop'.",
+		require.Equal(t, "could not validate provider configuration: Invalid or unknown key. "+
```
t0yv0, line 1152: Accidental edit, disregard, this stays the same. The explicit provider error should be very rare as it should be a compilation error typically in TypeScript.

t0yv0:
CC @zbuchheit

pkg/tfbridge/provider_test.go
```diff
@@ -1142,9 +1173,9 @@ func TestCheckConfig(t *testing.T) {
 		require.NoError(t, err)
 		require.NoError(t, err)
 		require.Equal(t, 1, len(resp.Failures))
-		require.Equal(t, "could not validate provider configuration: "+
-			"Invalid or unknown key. Check `pulumi config get testprovider:cofnigValue`. "+
-			"Did you mean `testprovider:configValue`?",
+		require.Equal(t, "`testprovider:cofnigValue` is not a valid configuration key for the testprovider provider. "+
```
zbuchheit, line 1176:
```suggestion
		require.Equal(t, "`testprovider:configValue` is not a valid configuration key for the testprovider provider. "+
```

pkg/tfbridge/provider_test.go
```diff
@@ -1142,9 +1173,9 @@ func TestCheckConfig(t *testing.T) {
...
+			"Did you mean `testprovider:configValue`? "+
+			"If the key is not intended for the provider, please choose a different namespace from `testprovider:`.",
```
zbuchheit, line 1178:
```suggestion
			"If the referenced key is not intended for the provider, please choose a different namespace from `testprovider:`.",
```
nit: makes it more specific

iwahbe (approved):
Looks like a clear improvement.

dirien:
@t0yv0 and @iwahbe, I think we have an issue as this PR triggered an error in https://github.com/dirien/pulumi-vultr/issues/423

iwahbe:
@dirien Can you open an issue in this repo with the code that used to work, and now doesn't?

dirien:
done @iwahbe: https://github.com/pulumi/pulumi-terraform-bridge/issues/2582

Issue #1595, opened 2023-12-26
Refresh pollutes state for programs that use Assets and Archives


t0yv0:
### What happened?

When using assets and archives, as common for lambas in AWS for example, refresh wipes out the asset/archive information from the state and replaces it with machine-local references. This is inherently dangerous if multiple boxes are used to perform `pulumi` operations against a shared statefile. 

### Example

The motivating example is a highly upvoted issue in pulumi-aws:

https://github.com/pulumi/pulumi/issues/6235

https://github.com/pulumi/pulumi-aws/issues/3185 even more minimal repro is here. 

### Output of `pulumi about`

See above.

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
Seeing lots of incidences in azure tests as well. Any time azure:storage/blob:Blob is used, in particular when used with overlays that automatically provision functions like this one:

```typescript
export const subscription = topic.onEvent("test", async (context, arg) => {
```

iwahbe:
I've added this bug to the planning sheet for next iteration.

PR #1671, opened 2024-02-02
Fixes refresh of resources with unchanged assets

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1595, https://github.com/pulumi/pulumi/issues/6235

Before this change, refreshing an unchanged resource which used an Asset or an Archive resulted in polluting Pulumi state with a machine-local filename.

The simplest example this is tested on is:

```typescript
    const exampleBucketObject = new aws.s3.BucketObject("exampleBucketObject", {
        key: "someobject",
        bucket: bucket.id,
        source: new pulumi.asset.FileAsset(inFile),
    });
```

After this change this example refreshes correctly with no changes detected.

Unfortunately this change by itself is insufficient to guarantee that these resources refresh correctly when Pulumi-tracked state is out of date with respect to the cloud state, and further work is required to make this correct for individual resources in each provider. For the example above, the upstream provider for BucketObject does not implement fetching contents of the object upon Read into a temporary file. Therefore refreshing the object in this situation updates the "etag" property but does not guide the user to the need of reconciling the FileAsset with the updated cloud state.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -1123,7 +1123,7 @@ func (p *Provider) Read(ctx context.Context, req *pulumirpc.ReadRequest) (*pulum
 		}
 	}
 
-	config, _, err := MakeTerraformConfig(ctx, p, oldInputs, res.TF.Schema(), res.Schema.Fields)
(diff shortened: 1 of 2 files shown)


VenelinMartinov (approved):
💯

Issue #1572, opened 2023-12-15
Double-counted examples stats under PULUMI_CONVERT=1


t0yv0:
Noticing that there's something off with the example stats when rolling out PULUMI_CONVERT in pulumi-azuread.

https://github.com/pulumi/pulumi-azuread/pull/484

Stats indicate more examples rendered but I could not find them in the SDKs. I think what happened is that the same example  gets double-counted in a weird way, I was spot-checking one example showing under `#/resources/azuread:index/user:User` with PULUMI_CONVERT=0 and as two examples User#0 and User#1 under PULUMI_CONVERT=1.

In the latter case, the two examples have identical OriginalHCL.. the first one has no conversion results and the second copy does have them. So it seems that this is benign for the end-users but does distort the stats. 

The files were obtained by:

```
go run ~/code/pulumi-terraform-bridge/unstable/scripts/exconverter/main.go --dir "$PWD/cache"
``` 

The two examples in detail:

```
jq '.|select(.ExampleName=="#/resources/azuread:index/user:User#0")' cache/experimental/byExample.json                                                                                                   ~/code/pulumi-azuread
{
  "ExampleName": "#/resources/azuread:index/user:User#0",
  "OriginalHCL": "resource \"azuread_user\" \"example\" {\n  user_principal_name = \"jdoe@hashicorp.com\"\n  display_name        = \"J. Doe\"\n  mail_nickname       = \"jdoe\"\n  password            = \"SecretP@sswd99!\"\n}",
  "ConversionResults": {}
}
t0yv0@Antons-MacBook-Pro> jq '.|select(.ExampleName=="#/resources/azuread:index/user:User#1")' cache/experimental/byExample.json                                                                                                       ~/code/pulumi-azuread
{
  "ExampleName": "#/resources/azuread:index/user:User#1",
  "OriginalHCL": "resource \"azuread_user\" \"example\" {\n  user_principal_name = \"jdoe@hashicorp.com\"\n  display_name        = \"J. Doe\"\n  mail_nickname       = \"jdoe\"\n  password            = \"SecretP@sswd99!\"\n}",
  "ConversionResults": {
    "csharp": {
      "FailureSeverity": 0,
      "FailureInfo": "",
[...]

PR #1676, opened 2024-02-05
Fix example double-counting under PULUMI_CONVERT=1

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1572

Before the change examples were accidentally double-counted under PULUMI_CONVERT=1 skewing stats. This is now fixed. 

Also the interface to the coverage tracker assumed that the methods are called in a particular order. This confused the PULUMI_CONVERT because it shuffles the ordering of discovering and post-processing the examples. This is now made more explicit by passing *Example pointers in the interface to the converter.

### DIFF
--- a/pkg/tfgen/examples_coverage_tracker.go
+++ b/pkg/tfgen/examples_coverage_tracker.go
@@ -124,13 +124,12 @@ func (ct *CoverageTracker) getOrCreateExample(pageName string, hcl string) *Exam
 		example := Example{hcl, make(map[string]*LanguageConversionResult)}
 		existingPage.Examples = append(existingPage.Examples, example)
 		return &example
-	} else {
-		// Initializing a page for this example.
-		example := Example{hcl, make(map[string]*LanguageConversionResult)}
-		examples := []Example{example}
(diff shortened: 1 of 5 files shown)


pkg/tfgen/examples_coverage_tracker.go
```diff
@@ -92,36 +90,56 @@ const (
...
+			return &e
+		}
 	}
-	ct.currentPageName = pageName
+	return nil
+}
 
+// Similar to getExample, but instead of returning nil when not found, creates an appropriate page
+// and Example object and registers it.
+func (ct *CoverageTracker) getOrCreateExample(pageName string, hcl string) *Example {
```
guineveresaenger, line 115: I really like this rename.

guineveresaenger (commented):
Had a discussion about the `useCoverageTracker`/`getExample` logic and could refine a bit

t0yv0:
Addressed feedback @guineveresaenger PTAL.

blampe (approved):

PR #1684, opened 2024-02-14
Raise the default log level of TF from WARN to ERROR

iwahbe:
Fixes https://github.com/pulumi/pulumi-aws/issues/3389

### DIFF
--- a/unstable/logging/logging.go
+++ b/unstable/logging/logging.go
@@ -207,10 +207,17 @@ func setupRootLoggers(ctx context.Context, output io.Writer) context.Context {
 }
 
 // Choose the default level carefully: logs at this level or higher (more severe) will be shown to the user of Pulumi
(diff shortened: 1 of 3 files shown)


unstable/logging/logging.go
```diff
@@ -270,26 +275,7 @@ func (w *logSinkWriter) Write(p []byte) (n int, err error) {
...
-	case "TRACE":
-		return hclog.Trace
-	case "DEBUG":
-		return hclog.Debug
-	case "OFF":
-		return hclog.Off
-	default:
-		return hclog.NoLevel
-	}
+	return hclog.LevelFromString(os.Getenv(tfLogEnvVar))
```
iwahbe, line 278: This is a simple refactor to use `hclog`'s own switch statement. It doesn't change any behavior.

guineveresaenger (approved):

PR #1685, opened 2024-02-15
Disable HTML encoding of Pulumi Schemas

iwahbe:
This encoding results in .NET examples such as:

```
    /// return await Deployment.RunAsync(() =&gt;
```

These render correctly in the registry, but it makes the in-code examples invalid. Hugo (and  markdown in general) already handles character escaping, so it should not be replicated in Pulumi Schemas.


t0yv0:
This feels like it's papering over another issue somewhere. The JSON encoding option SetEscapeHTML does not affect the content of the information stored in JSON, it simply offers a choice between two semantically identical ways to represent the same content in JSON. If this ends up having a meaningful difference, then there is something that's reading JSON improperly.

iwahbe:
> This feels like it's papering over another issue somewhere. The JSON encoding option SetEscapeHTML does not affect the content of the information stored in JSON, it simply offers a choice between two semantically identical ways to represent the same content in JSON. If this ends up having a meaningful difference, then there is something that's reading JSON improperly.

Hmmm. I think you're right. Quick testing shows that json escapes `<>` to unicode sequences `\uXXXX` where as we are seeing `&gt;`. I think that is a different subsystem doing the encoding, but I'm not sure which one.

PR #1696, opened 2024-02-22
Fix unexplained detailed diffs

t0yv0:
Fixes https://github.com/pulumi/pulumi-aws/issues/3439

makeDetailedDiff does not report collection changes. It instead hopes that elements of the collection will produce diff entries. When changes involve collections all the way down and there are no element entries, makeDetailedDiff produces no records. 

However we have some workarounds now that produce a DIFF_SOME even when there are no entries per
https://github.com/pulumi/pulumi-terraform-bridge/issues/1501

```
	if p.info.XSkipDetailedDiffForChanges && len(diff.Attributes()) > 0 {
```

The change in this PR embellishes this case with collection-level diffs so there is at least something to show.

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -156,9 +156,18 @@ func visitPropertyValue(
 	}
 }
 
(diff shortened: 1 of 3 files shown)


iwahbe (approved):
LGTM.

PR #1697, opened 2024-02-23
Revert "Make sure no nil lists are passed to TF"

t0yv0:
Reverts pulumi/pulumi-terraform-bridge#1688


guineveresaenger (approved):
This will allow us to release the bridge today. We will re-revert and adjust providers.

edit: a summary of resulting failures is [on Slack](https://pulumi.slack.com/archives/C02FXTZEZ6W/p1708718289918489)

PR #1698, opened 2024-02-23
Remove rename tables and tfgen namecheck

t0yv0:
TF to Pulumi name projections were suspect at some point in the evolution of this project, and Renames and namecheck were used to verify that name translations are consistent and decided at tfgen time. We have had no failures in the namecheck for many months, and we currently trust the PulumiToTerraformName and inverse functions, and this functionality can be retired.


iwahbe (approved):
LGTM

Will this remove the `"renames"` section from existing providers automatically?

If not, we might want to follow up with a PR that will explicitly remove it.

PR #1701, opened 2024-02-23
DO NOT MERGE Use prebuilt schema in tfgen

t0yv0:
Due to a historical accident that failed to cache schema generation bridged providers currently keep rebuilding the schema for every target language. That is `pulumi-tfgen-azure python --out sdk/python/` will regenerate the schema from scratch, including converting examples, when `pulumi-tfgen-azure schema --out provider/cmd/pulumi-resource-azure` has already done the work of generating schema.json. This change introduces a new parameter that skips this, this command proceeds almost instantaneously now with the extra parameter as in:

```
pulumi-tfgen-azure python --out sdk/python/ --schema provider/cmd/pulumi-resource-azure/schema.json
```

This will need a small ci-mgmt follow up to take effect but should be worth it. Especially since PULUMI_CONVERT=1 rollout slows down schema generation due to interop with a separate converter process.


guineveresaenger (approved):

t0yv0:
So reformatText and fixupPropertyReference specifically depend on the target language (Go, C#) so
the descriptions in the schema differ accordingly from the ones in the SDKs. I also suspect
markdownLink.ReplaceAllStringFunc to be responsible for cross-rendering something like this found in
the source:

```
---
page_title: "Linode: linode_vpc"
description: |-
  Manages a Linode VPC.
---

# linode\_vpc

Manages a Linode VPC.

Please refer to [linode_vpc_subnet](vpc_subnet.html.markdown) to manage the subnets under a Linode VPC.
```

These look like simple regex passes, could they be retrofitted to work on top of PackageSpec instead
of entityDocs? What information is possibly missing from rendering SDK descriptions? Directionally we'll need to be making the flow explicit if we are to make this code maintainable. 

For the immediate purposes though I really find it objectionable that example rendering happens repeatedly, and burning repeat CPU as well as emitting repeat warnings times language. Using schema.json as a vehicle to share example renderings is not the only way forward, perhaps a shared example rendering cache could satisfy the same aim.

iwahbe:
> For the immediate purposes though I really find it objectionable that example rendering happens repeatedly, and burning repeat CPU as well as emitting repeat warnings times language. Using schema.json as a vehicle to share example renderings is not the only way forward, perhaps a shared example rendering cache could satisfy the same aim.

I think it makes a lot of sense to share a cache between schema generation and SDK generation. I would be in favor of a cache created by `make tfgen` and then consumed by each non-schema generate call. We could do something as dumb as a set of files where the name is a hash of the input and the contents is a the cached output.

t0yv0:
Closing in favor of https://github.com/pulumi/pulumi-terraform-bridge/pull/1717

PR #1717, opened 2024-02-28
Improve examples cache

t0yv0:
Fleshing out the examples cache a bit. With the introduction of PULUMI_CONVERT=1 the build times are making large providers like AWS very unpleasant to work with. In particular examples are being re-rendered repeatedly across `make tfgen` and `make build_sdks` as well as across multiple passes through the provider. This PR improves pre-existing examples cache a bit with the following goals in mind:

- a lot more information that might be relevant to examples generation is captured into the cache key
- so that changes in tooling versions that might affect example rendering should invalidate the cache
- cache directories now have a metadata file explaining the details of the cache key
- PULUMI_CONVERT=1 style conversions now support accelerating through the cache
- this is slightly tricky as PULUMI_CONVERT=1 bulk-detects HCLs but now needs to be aware of the cache to skip them
- cache hits should not affect example coverage tracker metrics


pkg/tfgen/docs.go
```diff
@@ -1606,7 +1568,10 @@ func (g *Generator) convertHCLToString(e *Example, hclCode, path, languageName s
 	var diags hcl.Diagnostics
 	var err error
 
-	if cliConverterEnabled() {
+	cache := g.getOrCreateExamplesCache()
+	if r, ok := cache.Lookup(hclCode, languageName); ok {
+		convertedHcl = r
```
iwahbe, line 1573: Are we not returning here because we want 	`g.coverageTracker.languageConversionSuccess(e, languageName, convertedHcl)` to trigger later down?

iwahbe (approved):
Looks good to me. 

I'd be comfortable disabling `gosec`. I'm struggling to think of a case where it flagged an error for us.

pkg/tfgen/docs.go
t0yv0: That's right. Looks like this results in a redundant cache.Store call. I can fix that up.

PR #1744, opened 2024-03-08
Revert schema path changes

iwahbe:
Reverts #1741, #1710 

This means that GCP will not be able to upgrade until these commits are fixed. 

Let's get the bridge out in the meantime.


pkg/tfgen/docs.go
```diff
@@ -1271,9 +1271,9 @@ func (g *Generator) convertExamples(docs string, path examplePath) (result strin
 	}
 
 	// This function is very expensive for large providers. Permit experimental disk-based caching if the user
-	// specifies the PULUMI_CONVERT_EXAMPES_CACHE_DIR environment variable, pointing to a folder for the cache.
+	// specifies the PULUMI_CONVERT_EXAMPLES_CACHE_DIR environment variable, pointing to a folder for the cache.
```
t0yv0, line 1274: Can we keep this correction? Also good example why atomic PRs are good, one PR per concern.

pkg/tfbridge/provider.go
```diff
@@ -780,7 +780,9 @@ func (p *Provider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (*pul
 	rescfg := MakeTerraformConfigFromInputs(ctx, p.tf, inputs)
 	warns, errs := p.tf.ValidateResource(ctx, tfname, rescfg)
 	for _, warn := range warns {
-		GetLogger(ctx).Warn(fmt.Sprintf("%v verification warning: %v", urn, warn))
+		if err = p.host.Log(ctx, diag.Warning, urn, fmt.Sprintf("%v verification warning: %v", urn, warn)); err != nil {
```
t0yv0, line 783: This is another weird change that needs its own PR. I missed the in review. The `return err` from a logging interface  feels a bit dangerous. We shouldn't be failing the provider when for some reason the gRPC machinery for sending warnings logs has malfunctioned.  Perhaps this should be reflected in the the Log() methods not returning `error`.

pkg/tfgen/docs.go
iwahbe: We are keeping that correction (commit is https://github.com/pulumi/pulumi-terraform-bridge/pull/1744/commits/490e7f5eff00a69a6c407eaed2976087379f99c2).

t0yv0 (approved):
Always should be OK to revert. Let's follow up with a deep dive next week on what happened here. The changes looked really sound.

pkg/tfgen/docs.go
iwahbe: The previous correction was accidentally reverted by https://github.com/pulumi/pulumi-terraform-bridge/pull/1689.

PR #1748, opened 2024-03-11
Notes on set handling

t0yv0:


iwahbe:
@t0yv0 Is this something that you think is worth cleaning up and merging? Should it be appended to https://github.com/pulumi/pulumi-terraform-bridge/blob/5c13cc4d81e56baa7498b6a3f8b37f07a9a3dc7d/docs/guides/sets.md?

Issue #1651, opened 2024-01-26
Incomplete schema timeouts handling


t0yv0:
### What happened?

When working on https://github.com/pulumi/pulumi-terraform-bridge/pull/1648 noting that Schema-defined (provider-defined) resource operation timeouts are not propagated consistently into all lifecycle methods. They are propagated when Create resources is called, but not on Update and Delete. User-defined timeout overrides are always propagated. This omission may mean that the bridged providers may behave sub-optimally when upstream providers customize the timeouts.

### Example

N/A yet - need to dig up and example of a resource that actually specifies custom timeouts in the schema.

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
QQ: grep providers to see if the timeouts are actually specified/sometimes/ a lot? For which resources?

PR #1749, opened 2024-03-11
Fix custom timeout propagation for Update

t0yv0:
Fixes a regression responsible for the https://github.com/pulumi/pulumi-aws/issues/3442 P1.

When refactoring timeout handling in https://github.com/pulumi/pulumi-terraform-bridge/pull/1648 the behavior of custom timeouts for Update and Delete was inadvertently broken, but passed the test suite. 

The regression was introduced in v3.72.0 of the bridge and affects all SDKv2 based resources. Custom timeouts were not respected for Update and Delete but were respected for Create. 

The fix is very self-contained but this PR also backfills test coverage with a matrix of tests for custom timeout propagation and respecting both schema-based and user-based timeouts.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -1684,7 +1684,7 @@ func transformFromState(
 func newTimeoutOverrides(key shim.TimeoutKey, maybeTimeoutSeconds float64) map[shim.TimeoutKey]time.Duration {
 	timeoutOverrides := map[shim.TimeoutKey]time.Duration{}
 	if maybeTimeoutSeconds != 0 {
-		timeoutOverrides[shim.TimeoutCreate] = time.Duration(maybeTimeoutSeconds * float64(time.Second))
+		timeoutOverrides[key] = time.Duration(maybeTimeoutSeconds * float64(time.Second))
 	}
 	return timeoutOverrides
 }
--- a/pkg/tfbridge/provider_test.go
+++ b/pkg/tfbridge/provider_test.go
@@ -3260,7 +3260,7 @@ func TestSchemaFuncsNotCalledDuringRuntime(t *testing.T) {
 	})
 }
 
-// TOOD[pulumi/pulumi#15636] if/when Pulumi supports customizing Read timeouts these could be added here.
+// TODO[pulumi/pulumi#15636] if/when Pulumi supports customizing Read timeouts these could be added here.
 func TestCustomTimeouts(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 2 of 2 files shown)


pkg/tfbridge/provider_test.go
```diff
@@ -3259,3 +3259,181 @@ func TestSchemaFuncsNotCalledDuringRuntime(t *testing.T) {
...
+	skips := func(tc testCase) string {
+		if tc.userSpecifiedTimeout != nil {
+			return ""
+		}
+		if tc.schemaTimeout == nil {
+			return ""
+		}
+		switch tc.cud {
+		case "Update", "Delete":
+			return "TODO[pulumi/pulumi-terraform-bridge#1651] - schema-specified timeouts are not " +
```
t0yv0, line 3420: This suspicion is now confirmed for #1651  - custom Delete or Update timeouts from the upstream provider schema are not currently respected. Any chance that's by design? If we want it respected I can follow up with a quick PR here.

pkg/tfbridge/provider.go
```diff
@@ -1684,7 +1684,7 @@ func transformFromState(
 func newTimeoutOverrides(key shim.TimeoutKey, maybeTimeoutSeconds float64) map[shim.TimeoutKey]time.Duration {
 	timeoutOverrides := map[shim.TimeoutKey]time.Duration{}
 	if maybeTimeoutSeconds != 0 {
-		timeoutOverrides[shim.TimeoutCreate] = time.Duration(maybeTimeoutSeconds * float64(time.Second))
+		timeoutOverrides[key] = time.Duration(maybeTimeoutSeconds * float64(time.Second))
```
iwahbe, line 1687: Shouldn't we have a lint for unused function args? That would have caught this.

pkg/tfbridge/provider_test.go
iwahbe: Unless we can find a reason why they are not respected, let's respect them.

This is less Chesterton's fence then a pile of scattered sticks in a field.

iwahbe (approved):

pkg/tfbridge/provider.go
iwahbe: 😢 
https://github.com/pulumi/pulumi-terraform-bridge/blob/22fdb29205f4415e371a3dc12165334fe59fc50d/.golangci.yml#L26

iwahbe: I opened https://github.com/pulumi/pulumi-terraform-bridge/issues/1750.

t0yv0: Nice.

PR #1765, opened 2024-03-18
fix: fixes are not applied to property deprecation messages

corymhall:
We have logic in `fixupPropertyReference` to replace terraform resource references with pulumi references (i.e. `aws_s3_bucket` => `aws.s3.BucketV2`).

This logic is applied to the `Description` property, but it is not applied to the separate `DeprecationMessage` property. These two properties exist separately in the pulumi resource schema and are combined when creating the website docs and the SDK docs.

fixes #3028

### DIFF
--- a/pkg/tfgen/internal/testprovider/miniaws.go
+++ b/pkg/tfgen/internal/testprovider/miniaws.go
@@ -1,0 +1,72 @@
+// Copyright 2016-2022, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


pkg/tfgen/internal/testprovider/miniaws.go
```diff
@@ -0,0 +1,75 @@
...
+		DocRules: &tfbridge.DocRuleInfo{EditRules: func(defaults []tfbridge.DocsEdit) []tfbridge.DocsEdit {
+			return []tfbridge.DocsEdit{
+				{
+					Path: "*",
+					Edit: func(_ string, content []byte) ([]byte, error) {
+						content = bytes.ReplaceAll(
+							content,
+							// This replacement is done in the aws provider
+							// here. This replacement is necessary because the
+							// bridge will drop any docs that contain the work
```
t0yv0, line 45: Contain the **word** - typo I think.

t0yv0 (approved):
🚢

Issue #1776, opened 2024-03-21
TFGen starts to generate (MISSING) into Description


t0yv0:
Repro:

```bash
cd pf
GOWORK=off make build.testproviders
git status # should not produce a diff but does
```

With the bug present, `RandomPassword` special field has a description with (MISSING) in it:

```
"description": "Include special characters in the result. These are `!@#$%!\u0026(MISSING)*()-_=+[]{}\u003c\u003e:?`. Default value is `true`.\n"
```

Versions:

```org
| v3.65.0                                  | no repro   |
| 2d6a5c06d6e2fa666c4a35ebe7d51c7fcb187bfd | reproduces |
```

t0yv0:
Customer reports hitting the same with

```
Got another one on tf-bridge: if your TF schema has a field description that contains valid fmt.Sprintf formatting directives, they get all jacked up at some point in the tfgen process.
For example, we have a field that looks something like this:
keys.Mapname: {
	Description: "Config Map key. Format: %SOMETHING%:%SOMETHING%:uniqueKey. Eg: some.test.domain:8080:uniqueKey. blah blah",
	Type:        schema.TypeString,
	Required:    true,
},
After upgrading to the latest version of tfbridge (from v3.73.0), we get a diff like this in our Pulumi schema:
                 "mapname": {
                     "type": "string",
-                    "description": "Config Map key. Format: %SOMETHING%:%SOMETHING%:uniqueKey. Eg: some.test.domain:8080:uniqueKey. blah blah\n"
+                    "description": "Config Map key. Format: %!S(MISSING)OMETHING%!:(MISSING)%!S(MISSING)OMETHING%!:(MISSING)uniqueKey. Eg: some.test.domain:8080:uniqueKey. blah blah\n"
                 }
```

PR #1777, opened 2024-03-21
Test RandomBytes PF support

t0yv0:
Small PR to cover support for resources without a proper ID like RandomBytes. To do this:

- upgraded localized pulumi-random provider to the version we actually use
- wrote a quick test
- discovered #1776 
- worked around it for now by manually undoing its schema edits


VenelinMartinov (approved):
lgtm

PR #1780, opened 2024-03-21
Fixes issues with name overrides in Plugin Framework invokes

t0yv0:
This change fixes an issue where Plugin Framework invokes did not respect SchemaInfo overrides, in particular there could be issues with incorrect translation of incoming PropertyMap data when SchemaInfo specified a different name for a data source property.

On top of [t0yv0/improve-convert-module](https://github.com/pulumi/pulumi-terraform-bridge/pull/1779).

### DIFF
--- a/pf/internal/convert/encoding_test.go
+++ b/pf/internal/convert/encoding_test.go
@@ -295,6 +295,43 @@ func TestDataSourceDecoder(t *testing.T) {
 				},
 			}),
 		},
+		{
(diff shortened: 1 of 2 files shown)


VenelinMartinov (approved):
Did you spot this randomly in the code or was this motivated by a provider issue?

t0yv0:
Spotted when adding coverage.

PR #1793, opened 2024-03-26
Make the cache key more precise

t0yv0:
Make the examples cache key more precise:

- do not look at provider version
- do not look at the entirety of go.mod, only check key package versions
- include MarshallableProviderInfo in the cache
- if go.work is used, still include that as before


t0yv0:
I think I'm getting decent results with this on AWS in https://github.com/pulumi/pulumi-aws/pull/3737

iwahbe (approved):
LGTM

PR #1798, opened 2024-03-27
Read in descriptions for nested properties

### LINKED ISSUE #1774: Some property descriptions missing from nested type fields
We have descriptions missing in some nested type fields. I'm not sure the reasoning and can't really find a pattern for when this occurs that's obvious, but here is an example on the lambda page. 

https://www.pulumi.com/registry/packages/aws/api-docs/lambda/function/#functionsnapstart (optimizationStatus missing)
https://www.pulumi.com/registry/packages/aws/api-docs/lambda/function/#functionvpcconfig (vpcId missing)

guineveresaenger:
Uses a nested path lookup for properties whose descriptions come to us via entityDocs.Attributes.

Allows for the regex to match against `.` in nested output names.

Fixes #1774.

~This is a draft so far as there's a few issues here.~

update: Please see a build of the AWS provider docs, for a [schema built with these changes](https://github.com/pulumi/pulumi-aws/compare/guin/fix-misaligned-attribute-content?expand=1) where specific missing fields are now correctly populated:

[MQ Broker Instance before](https://www.pulumi.com/registry/packages/aws/api-docs/mq/broker/#brokerinstance)
[MQ Broker Instance after](http://registry--origin-pr-4222-df3bcdc8.s3-website.us-west-2.amazonaws.com/registry/packages/aws/api-docs/mq/broker/#endpoints_nodejs)

[Lambda Function Snapstart Before](https://www.pulumi.com/registry/packages/aws/api-docs/lambda/function/#functionsnapstart)
[Lambda Function Snapstart After](http://registry--origin-pr-4222-df3bcdc8.s3-website.us-west-2.amazonaws.com/registry/packages/aws/api-docs/lambda/function/#functionsnapstart)

[Elasticsearch DomainVPCOptions Before](https://www.pulumi.com/registry/packages/aws/api-docs/elasticsearch/domain/#domainvpcoptions)
[Elasticsearch DomainVPCOptions After](http://registry--origin-pr-4222-df3bcdc8.s3-website.us-west-2.amazonaws.com/registry/packages/aws/api-docs/elasticsearch/domain/#domainvpcoptions)

** Notes**

1. I discovered that _a majority_ of nested property descriptions (at least in AWS) come from what is described as a "fallback" in the code comment, which I edited to add that information.

2. This fixes this bug where extra bullet points in a property description were attached to the previously-parsed entry. 

**Before:**

image
Screenshot 2024-03-26 at 5 33 34 PM

**After** (localhost):

Screenshot 2024-03-26 at 5 33 22 PM
Screenshot 2024-03-26 at 5 33 00 PM

~3. However, this may need some more work to properly map multi-nested documentation [as seen in `mq_broker.instances`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/mq_broker#instances) which [still results in missing descriptions on our end](https://www.pulumi.com/registry/packages/aws/api-docs/mq/broker/#brokerinstance)~

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -44,7 +44,6 @@ import (
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tf2pulumi/convert"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
-	"github.com/ryboe/q"
 )
 
 const (
@@ -278,25 +277,11 @@ func getDocsForResource(g *Generator, source DocsSource, kind DocKind,
 		docInfo = info.GetDocs()
 	}
 
-	//if rawname == "aws_lambda_function" && kind == ResourceDocs {
-	//
-	//	q.Q(rawname)
-	//	q.Q(kind)
-	//	q.Q(source)
-	//	q.Q(docInfo)
-	//
-	//}
-
 	var docFile *DocFile
 	var err error
 	switch kind {
 	case ResourceDocs:
 		docFile, err = source.getResource(rawname, docInfo)
-		//if rawname == "aws_lambda_function" {
-		//	q.Q(docFile.FileName)
-		//	q.Q(string(docFile.Content))
-		//}
-
 	case DataSourceDocs:
 		docFile, err = source.getDatasource(rawname, docInfo)
 	default:
@@ -367,10 +352,6 @@ func getDocsForResource(g *Generator, source DocsSource, kind DocKind,
 
 			overlayArgsToArgs(sourceDocs, &doc)
 		}
-	}
-	if rawname == "aws_lambda_function" && kind == ResourceDocs {
-		q.Q("doc.Attributes")
-		q.Q(doc.Attributes)
 	}
 
 	return doc, nil
@@ -556,7 +537,6 @@ func (p *tfMarkdownParser) parseSupplementaryExamples() (string, error) {
 }
 
 func (p *tfMarkdownParser) parse(tfMarkdown []byte) (entityDocs, error) {
-
 	p.ret = entityDocs{
 		Arguments:  make(map[docsPath]*argumentDocs),
 		Attributes: make(map[string]string),
@@ -604,7 +584,6 @@ func (p *tfMarkdownParser) parse(tfMarkdown []byte) (entityDocs, error) {
 		if err := p.parseSection(section); err != nil {
 			return entityDocs{}, err
 		}
-
 	}
 
 	// Get links.
@@ -990,7 +969,6 @@ func parseAttributesReferenceSection(subsection []string, ret *entityDocs) {
 
 func (p *tfMarkdownParser) parseImports(subsection []string) {
 	var token string
-
 	if p.info != nil && p.info.GetTok() != "" {
 		token = p.info.GetTok().String()
 	}
@@ -1019,13 +997,14 @@ func (p *tfMarkdownParser) parseImports(subsection []string) {
 			}
 		}
 	}
+
 	if i, ok := tryParseV2Imports(token, subsection); ok {
 		p.ret.Import = i
 		return
 	}
+
 	var importDocString string
 	for _, section := range subsection {
-
 		if strings.Contains(section, "**NOTE:") || strings.Contains(section, "**Please Note:") ||
 			strings.Contains(section, "**Note:**") {
 			// This is a Terraform import specific comment that we don't need to parse or include in our docs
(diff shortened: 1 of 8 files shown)


pkg/tfgen/docs.go
```diff
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+
```
iwahbe, line 23:
```suggestion
```

pkg/tfgen/docs.go
```diff
@@ -1023,6 +1025,7 @@ func (p *tfMarkdownParser) parseImports(subsection []string) {
 		section = strings.Replace(section, "```", "", -1)
 
 		if strings.Contains(section, "terraform import") {
+
```
iwahbe, line 1028:
```suggestion
```

pkg/tfgen/docs.go
```diff
@@ -440,7 +442,7 @@ var (
 	)
 
 	attributeBulletRegexp = regexp.MustCompile(
-		"^\\s*[*+-]\\s*`([a-zA-z0-9_]*)`\\s*[:–-]?\\s*(.*)",
+		"^\\s*[*+-]\\s*`([a-zA-z0-9._]*)`\\s*[:–-]?\\s*(.*)",
```
iwahbe, line 445: I think we are at the point where we need to break up this regex into multiple constants so we can name them:

```suggestion
		bulletPointRegexp + "`" + pathNameRegexp + `"` + "\\s*[:–-]?\\s*(.*)", // I'm not sure what the last section is.
```

iwahbe (commented):
This looks directionally good.

pkg/tfgen/docs.go
```diff
@@ -965,6 +970,10 @@ func parseAttributesReferenceSection(subsection []string, ret *entityDocs) {
 	}
 }
 
+func flattenListAttributeKey(attribute string) string {
+	return strings.ReplaceAll(attribute, ".0", "")
+}
```
iwahbe, line 975: Where do these lists come from? Are they always `.0`, never `.N` or `.1`?

pkg/tfgen/docs_test.go
```diff
@@ -672,6 +672,43 @@ func TestParseAttributesReferenceSection(t *testing.T) {
...
+		Attributes: make(map[string]string),
+	}
+	parseAttributesReferenceSection([]string{
+		"The following attributes are exported:",
+		"",
+		"* `id` - The ID of the Droplet",
+		"* `urn` - The uniform resource name of the Droplet",
+		"* `name`- The name of the Droplet",
+		"* `region` - The region of the Droplet",
+		"* `region.zone` - The zone of the Droplet region",
```
iwahbe, line 687: Can we add a test for when `region.zone` comes before `region`?

pkg/tfgen/generate.go
```diff
@@ -1874,13 +1873,20 @@ func getNestedDescriptionFromParsedDocs(entityDocs entityDocs, path docsPath) (s
...
 		// In case #1 above, we are generating an incorrect schema because the upstream schema is incorrect, and we would
 		// arguably be better off not having any description in our docs. In case #2 above, this is fairly risky fallback
 		// behavior with may result in incorrect docs, per pulumi-terraform-bridge#550.
 		//
 		// We should work to minimize the number of times this fallback behavior is triggered (and possibly eliminate it
 		// altogether) due to the difficulty in determining whether the correct description is actually found.
-		return attribute, true
+		if description, ok := entityDocs.Attributes[string(attrPath)]; ok {
+			return description, true
+		}
```
iwahbe, line 1888: I don't see where this code section is under test. Can you add a test here?

pkg/tfgen/docs.go
guineveresaenger: This is an example of a [TF schema.TypeList](https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/mq/broker.go#L175) which I believe only ever uses the `.0` to access the list elements. I'm admittedly a bit fuzzy on this so I'm happy to be corrected here.
I also searched the p-aws schema as well as upstream/website for any indications of !=`.0` and came up with nothing. I know that's not proof and if you want to make this digit-agnostic, we can.

pkg/tfgen/docs_test.go
guineveresaenger: Added!

pkg/tfgen/generate.go
guineveresaenger: Added.

pkg/tfgen/docs.go
iwahbe: I trust your check. This is the type of information that would make an excellent comment on line 974.

iwahbe (approved):

Issue #1328, opened 2023-08-04
Type-check incoming values against the Pulumi Package Schema


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Bridged providers receive resource.PropertyValue user data from the engine encoding resource or function arguments that might not have the expected shape, in the sense that it mismatches the type published in Pulumi Package Schema. In these cases the transformation to the Terraform-expected domain may result in data of unexpected shape passed down to the TF internals.  This leads to panics.

This is especially likely to happen when using languages that do not guarantee type-correctness or where it can be bypassed, such as TypeScript and Python.

Link to related issues:

- https://github.com/pulumi/pulumi-terraform-bridge/issues/787
- https://github.com/pulumi/pulumi-terraform-bridge/issues/1138

The suggested enhancement is to check the incoming data against the schema as soon as the provider receives it, and provide a nicely formatted error message to the user, closer to Pulumi terms, of how to fix the program.

Ideally this feature is built as a layer independent from the bridge as nothing here assumes that a provider is a bridged provider; it could be implemented as at a GRPC transforming layer or as middleware.

### Affected area/feature

```[tasklist]
### Tasks
- [ ] https://github.com/pulumi/pulumi-terraform-bridge/issues/1891
- [ ] https://github.com/pulumi/pulumi-terraform-bridge/issues/1892
```

t0yv0:
Another GCP issue where the error message could have been better with a type-check pass: pulumi/pulumi-gcp#1377

t0yv0:
https://github.com/pulumi/pulumi-aws/issues/3092 could have gotten a much more lucid error message here.

PR #1800, opened 2024-03-27
feat: type check values based on pulumi schema

corymhall:
This adds an initial type checking implementation to `pkg/tfbridge`. The actually type checker is pretty generic and only has a dependency on `pulumi/pulumi` so it would be possible to pull this out into a separate library as a general type checker.

**Feature Support**
- [X] required properties
- [X] property names
- [X] `string` types
- [X] `number` types
- [X] `object` types
- [X] `array` types
- [X] `null` types
- [X] `bool` types
- [X] `output` types
- [X] `secret` types
- [ ] `asset` types
- [ ] `archive` types
- [ ] `resourceReference` types
- [ ] String Enums
- [ ] Number Enums - *It knows about enums, but it does not validate enum values since enums are not exhaustive*

Along with the type checker it will also build an error message that contains the path to the property.

**Todo**
- [ ] Add to `pf/tfbridge` (can be done in a later pr)
- [x] Add tests that hit the new type checker at the provider level
- [x] How do we roll this out?
      - By default this will only warn users, but can override with an env var

re #1328


pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,471 @@
...
+	cc := *c
+	cc.paths = append(c.paths, fmt.Sprintf("%d", idx))
+	cc.isList = true
+	return cc
+}
+
+// toPath renders the property path as a string
+func (c *pathBuilder) toPath() string {
+	return strings.Join(c.paths, ".")
+}
```
iwahbe, line 42: Can we use [`resource.PropertyPath`](https://github.com/pulumi/pulumi/blob/28f098a660aa698c83a77058da28f7867707d0a0/sdk/go/common/resource/properties_path.go#L14-L16) here?

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,471 @@
...
+// validateTypeSpec is the main function for validating a PropertyValue against the pulumi schema.
+// It is a recursive function that will validate nested types and arrays. It will return a list of
+// type failures if any are found
+func (v *PulumiInputValidator) validateTypeSpec(
+	propertyKey string,
+	propertyValue resource.PropertyValue,
+	typeSpec pschema.TypeSpec,
+	pathBuilder pathBuilder,
+	requiredProps []string,
+) *[]TypeFailure {
```
iwahbe, line 140: Why do we need `*[]TypeFailure`? Go slices are already nullable.
```suggestion
) []TypeFailure {
```

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,471 @@
...
+	// don't type check
+	// - resource references
+	// - assets
+	// - archives
+	if propertyValue.IsResourceReference() || propertyValue.IsAsset() || propertyValue.IsArchive() {
+		return nil
+	}
+
+	// handle non-required properties that are null
+	if propertyValue.TypeString() == "null" {
```
iwahbe, line 154:
```suggestion
	if propertyValue.IsNull() {
```

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,471 @@
...
+				},
+			}
+		}
+	}
+
+	failures := []TypeFailure{}
+	pathBuilder.add(propertyKey)
+
+	propertyValueType := propertyValue.TypeString()
+	if propertyValue.IsOutput() || propertyValue.IsComputed() || propertyValue.IsSecret() {
```
iwahbe, line 169: computed property values don't really understand their underlying type. I know they say they do, but they don't.

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,471 @@
...
+				Reason: fmt.Sprintf(
+					"expected %s type, got %s type",
+					possible.toString(),
+					propertyValue.TypeString(),
+				),
+			},
+		}
+	}
+
+	switch propertyValueType {
```
t0yv0, line 192: Matching based on `propertyValue.TypeString()` is not ideal since that feels like the code becomes sensitive to how these values are presented to the user for debugging. If you unpack how TypeString itself is implemented it's doing matching over its node kind, I think that'd read more naturally here as well:

https://github.com/pulumi/pulumi/blob/master/sdk/go/common/resource/properties.go#L549

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,471 @@
...
+// type defined in a couple different ways. This will check the different ways
+// and return the correct type spec If the array supports multiple types (i.e.
+// `oneOf`) then it will find the correct one and return it. This
+// can be used to provide a more helpful error message
+func (v *PulumiInputValidator) getArrayTypeSpec(
+	spec *pschema.TypeSpec,
+	propertyValue resource.PropertyValue,
+) (
+	typeSpec *pschema.TypeSpec,
+	requiredProps []string,
```
t0yv0, line 79: Required props always nil it seems like, could reflect in signature.

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,471 @@
...
+		return nil
+	}
+
+	// handle non-required properties that are null
+	if propertyValue.TypeString() == "null" {
+		if slices.Contains(requiredProps, propertyKey) {
+			return &[]TypeFailure{
+				{
+					ResourcePath: pathBuilder.toPath(),
+					Reason:       fmt.Sprintf("property %s is required", propertyKey),
```
t0yv0, line 159: It seems that this check is redundant with `!slices.Contains(stableKeys, resource.PropertyKey(requiredProp))` done when checking objects? Or am I missing something?

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,471 @@
...
+	}
+
+	if len(failures) > 0 {
+		return &failures
+	}
+
+	return nil
+}
+
+func (v *PulumiInputValidator) validateInputType(
```
t0yv0, line 310: Does this validate resources only or Invoke (what TF calls DataSource) inputs as well? I suspect only the former, in which case it would be great to highlight that with a better name and comment.

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,471 @@
...
+) (*pschema.TypeSpec, []string, bool) {
+
+	allSpecs := []string{}
+	for _, spec := range specs {
+		allSpecs = append(allSpecs, spec.Type)
+		switch inputValue.TypeString() {
+		case "[]":
+			if spec.Type == "array" {
+				if spec.Items != nil {
+					return spec.Items, []string{}, true
```
t0yv0, line 375: This jumps from an array type like `Array<string>` to an element type like `string`, which feels very strange and unrelated to the idea of unrolling OneOf. The entire function I'd like to understand what it's trying to do, starting from the signature as it does not seem to use inputValue.

t0yv0: I was trying to brush up on OneOf semantic, here's a good piece of code that interprets it during "binding" to Type.

```https://github.com/pulumi/pulumi/blob/master/pkg/codegen/schema/bind.go#L842```

This is actually interesting as it leads to UnionType https://github.com/pulumi/pulumi/blob/master/pkg/codegen/schema/schema.go#L185

It's interesting -- this code might be easier to write over the bound Type objects than raw TypeSpec

Also thinking that the schema is optimized more for code generation (primary purpose) than type checking we're attempting here, so interpreting UnionType.DefaultType is a little tricky.

IN the first pass it's not unreasonable to punt on union types entirely.

If we were to attempt to type check them, I imagine something could be tried along the lines of:

```
var ut UnionType
var foundCleanMatch
for i, possibleType := range ut.ElementTypes {
   tcRes = tc(value, possibleType)
   if !tcRes.Error() {
      foundCleanMatch = true
   }
}
if foundCleanMatch {
   return result.NoError()
}
return result.Format("Unexpected value %v, expecting one of %s", value, formatOptions(..))
```

This can also get tedious to get right from usability pov as deeply nested union errors can get very verbose.. But we're looking for something to try out here for the first time so no need to get it perfect.

File: pkg/tfbridge/validate_input_types.go [hunk 10]
```diff
@@ -0,0 +1,471 @@
...
+				return &spec, []string{}, true
+			}
+
+		}
+	}
+	return nil, allSpecs, false
+}
+
+// The input type (i.e. PropertyValue.TypeString() do not match 100% to the pschema.TypeSpec.Type)
+func typeEqual(schemaType, inputType string) bool {
```
t0yv0, line 396: This is a bit difficult to keep giving the right answer if anything changes in the underlying types, curious if we could get by without it at all.

pkg/tfbridge/validate_input_types_test.go
```diff
@@ -0,0 +1,2122 @@
...
+	"testing"
+
+	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestValidateInputTypesV2 tests the validateInputTypesV2 function
+func TestValidateInputType_objects(t *testing.T) {
+	testCases := []struct {
```
t0yv0, line 14: Did we check MapType? Something like `Map<string,string>` or `Map<string,someObject>`. Skimming quickly not seeing it but perhaps need to look again.

pkg/tfbridge/validate_input_types.go
corymhall: Should we just exclude `IsComputed` then?

pkg/tfbridge/validate_input_types.go
corymhall: Sure I can switch to doing it that way.

pkg/tfbridge/validate_input_types.go
corymhall: It doesn't look like `PropertyPath` includes any utilities for building a path, it looks like it can create a `PropertyPath` from a string and can edit `PropertyValue`s.

pkg/tfbridge/validate_input_types.go
corymhall: The property may be a top level property that is not an object property.

pkg/tfbridge/provider.go
```diff
@@ -751,6 +755,44 @@ func (p *Provider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (*pul
...
 
+	// for now we are just going to log warnings if there are failures.
+	// over time we may want to turn these into actual errors
+	_, validateShouldError := os.LookupEnv("PULUMI_ERROR_TYPE_CHECKER")
+	schemaMap, schemaInfos := res.TF.Schema(), res.Schema.GetFields()
+	if p.pulumiSchema != nil {
+		var schema pschema.PackageSpec
+		if err := json.Unmarshal(p.pulumiSchema, &schema); err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
corymhall: Done!

pkg/tfbridge/validate_input_types.go
corymhall: Done!

pkg/tfbridge/validate_input_types.go
corymhall: I refactored this to use more of the `PropertyValue` type, but I don't think we can do without this since it's not a 1:1 mapping between the schema types and the `PropertyValue` types.

pkg/tfbridge/validate_input_types_test.go
corymhall: That should be what all of these tests are. `object` types.

pkg/tfbridge/validate_input_types.go
corymhall: I added a comment to this function to try and better explain what it is doing.


 For example, given the following schema:
```
	"prop": {
		TypeSpec: pschema.TypeSpec{
			OneOf: []pschema.TypeSpec{
				{
					Type: "array",
					Items: &pschema.TypeSpec{
						Type: "object",
						// using ref to test specific object keys
						Ref: "#/types/pkg:index/type:ObjectStringType",
					},
				},
				{
					Type: "string",
				},
			},
		},
	}
```

 and the following input value:
```
	"prop": []map[string]interface{}{{"key": "value"}}
```
 In this case the actual type that we want to compare "prop" to is the `ObjectStringType`
 so this function will return the `Items` `TypeSpec` for comparison.

pkg/tfbridge/validate_input_types.go
iwahbe: `PropertyPath` is defined as:

```go
type PropertyPath []interface{}
```

That means that `func (c *pathBuilder) add(path string)` is equivalent to `c = append(c, path)` and `func (c *pathBuilder) addListIndex(idx int)` is equivalent to `c = append(c, idx)`.

pkg/tfbridge/validate_input_types.go
iwahbe: Yes.

pkg/tfbridge/validate_input_types.go
t0yv0: Ah yes that's right.. That's good then. There's another weird way to handle it is wrapping the top-level property-set schema for a resource in an object although it isn't, but it can be confusing. Sounds good then.

pkg/tfbridge/validate_input_types.go
t0yv0: This doesn't make sense to me tbh. Let's dig in here on a call.

pkg/tfbridge/validate_input_types.go
corymhall: Ok I think I was able to refactor it to work with `PropertyPath`!

pkg/tfbridge/validate_input_types.go
corymhall: based on our conversation I think I was able to refactor this to what we discussed. Let me know what you think.

pkg/tfbridge/provider.go
```diff
@@ -765,6 +770,42 @@ func (p *Provider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (*pul
 		return nil, err
 	}
 
+	logger := GetLogger(ctx)
+	// for now we are just going to log warnings if there are failures.
+	// over time we may want to turn these into actual errors
+	_, validateShouldError := os.LookupEnv("PULUMI_ERROR_TYPE_CHECKER")
+	schemaMap, schemaInfos := res.TF.Schema(), res.Schema.GetFields()
+	if p.pulumiSchema != nil {
+		schema := sync.OnceValue(func() *pschema.PackageSpec {
```
t0yv0, line 779: Hm this is not going to do any caching I think since the state created by sync.OnceValue(f) is not stored in p.Provider therefore it's going to be re-created and GC-d on every call.

t0yv0:
Round of suggestions here: https://github.com/pulumi/pulumi-terraform-bridge/pull/1878 - happy to take some time to talk through that tomorrow.

pkg/tfbridge/provider.go
corymhall: @t0yv0 I moved this sync.OnceValue function to the `Provider` object. I _think_ this does what we are wanting, but I'm not sure if we really need the `OnceValue` function at all. Is there a downside to unmarshaling it inside `newProvider` and just storing it on the `Provider` object?

corymhall:
> Round of suggestions here: #1878 - happy to take some time to talk through that tomorrow.

@t0yv0 I've rebased against your PR and added in some of the TODOs. Let me know if you think. 

I took a stab at the `OneOf`, but we can also just remove it and defer it to later.

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,361 @@
...
+			),
+		}}
+	case "object":
+		// This is not really an object but a map type with some element type, which is assumed to be string if
+		// unspecified. Check accordingly. This should be very similar to the "array" case.
+		//
+		// elementTypeSpec := typeSpec.AdditionalProperties
+		if propertyValue.IsObject() {
+			if typeSpec.AdditionalProperties == nil {
+				// Unknown item type so nothing more to check
```
t0yv0, line 285: The binder actually assumes this would be a stringly typed map[string]string. Could go either way here.

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,361 @@
...
+					objType.Required,
+					objType.Properties,
+					propertyPath,
+				)
+			}
+			objectValue := propertyValue.ObjectValue()
+			failures := []TypeFailure{}
+			for _, propertyKey := range objectValue.StableKeys() {
+				pb := append(propertyPath, string(propertyKey))
+				failure := v.validatePropertyValue(objectValue[propertyKey], *typeSpec.AdditionalProperties, pb)
```
t0yv0, line 307: This looks right.

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,361 @@
...
+		// unspecified. Check accordingly. This should be very similar to the "array" case.
+		//
+		// elementTypeSpec := typeSpec.AdditionalProperties
+		if propertyValue.IsObject() {
+			if typeSpec.AdditionalProperties == nil {
+				// Unknown item type so nothing more to check
+				return nil
+			}
+			// if there is a ref to another type, get that and then validate
+			if typeSpec.AdditionalProperties.Ref != "" {
```
t0yv0, line 289: You don't need this if clause at all, I think. What's going to happen is that v.validatePropertyValue on *typeSpec.AdditionalProperties is going to take full care of this.

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,361 @@
...
+		}
+
+		failures := []TypeFailure{}
+		for _, spec := range typeSpec.OneOf {
+			failure := v.validatePropertyValue(propertyValue, spec, propertyPath)
+			if len(failure) == 0 {
+				// we have found the correct type
+				return nil
+			}
+			failures = append(failures, failure...)
```
t0yv0, line 212: Merging all these failures will be unreadable. Perhaps just ignore them all for now and emit one generic failure that "matched none of the union type members"

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,361 @@
...
+		}
+
+		// don't validate types with Discriminator (yet)
+		if typeSpec.Discriminator != nil {
+			return nil
+		}
+
+		// default type
+		if typeSpec.Type != "" {
+			typeSpec.OneOf = append(typeSpec.OneOf, pschema.TypeSpec{Type: typeSpec.Type})
```
t0yv0, line 202: I always worry this mutation would leak outside the scope of this func, but looks like it doesn't perhaps because we have our own struct copy.

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,361 @@
...
+
+		return v.validatePropertyMap(
+			propertyValue.ObjectValue(),
+			objType.Required,
+			objType.Properties,
+			propertyPath,
+		)
+	}
+
+	if typeSpec.OneOf != nil {
```
t0yv0, line 184: This looks right

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,361 @@
...
+func (v *PulumiInputValidator) validatePropertyMap(
+	propertyMap resource.PropertyMap,
+	required []string,
+	propertyTypes map[string]pschema.PropertySpec,
+	propertyPath resource.PropertyPath,
+) []TypeFailure {
+	stableKeys := propertyMap.StableKeys()
+	failures := []TypeFailure{}
+
+	// handle required properties
```
t0yv0, line 62: While this looks right, I'm still worried here we actually allow some cases where "required" properties are actually filled out by the bridge post-check and therefore this will be too strict. I can try constructing some examples.

pkg/tfbridge/validate_input_types.go
corymhall: Are you saying that if `Type: "object", AdditionalProperties: nil` then the assumption is that the type is `Type: "object", AdditionalProperties: { Type: "string" }`?

pkg/tfbridge/validate_input_types.go
t0yv0: Yeah I'd be more comfortable starting with that.

pkg/tfbridge/validate_input_types.go
t0yv0: Yes! Last I checked that seemed to be literally what the binder code was saying.  It is not easy to tell though if that's intentional or "let's assume some type here". I guess I prefer your version here for now - no type specified so we won't check.

pkg/tfbridge/provider.go
t0yv0: >  Is there a downside to unmarshaling it inside newProvider and just storing it on the Provider object?

No downside, that'd be the more natural thing to do.

pkg/tfbridge/validate_input_types.go
corymhall: I took a stab at trying to return a better single error message, let me know what you think.

pkg/tfbridge/validate_input_types.go
corymhall: Done!

pkg/tfbridge/validate_input_types.go
flostadler: I don't quite understand what this is doing. Why is `typeSpec.Type` appended to `OneOf`?

corymhall: I took that logic from https://github.com/pulumi/pulumi/blob/477a54b3de081a4927df3038c7b2266d599d5f3c/pkg/codegen/schema/bind.go#L849-L854

Not sure exactly why, but it looks like a "default" type can be stored in the `Type` field.

pkg/tfbridge/provider.go
```diff
@@ -770,6 +782,36 @@ func (p *Provider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (*pul
 		return nil, err
 	}
 
+	logger := GetLogger(ctx)
+	// for now we are just going to log warnings if there are failures.
+	// over time we may want to turn these into actual errors
+	_, validateShouldError := os.LookupEnv("PULUMI_ERROR_TYPE_CHECKER")
```
iwahbe, line 788: `PULUMI_ERROR_TYPE_CHECKER=false` should not be equivalent to `PULUMI_ERROR_TYPE_CHECKER=true`.
```suggestion
	validateShouldError := "github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil".IsTruthy(os.Getenv("PULUMI_ERROR_TYPE_CHECKER"))
```

pkg/tfbridge/validate_input_types_test.go
```diff
@@ -0,0 +1,1883 @@
...
+		{
+			name:     "oneof_object_multi_type_failure_object",
+			typeRef:  "ObjectMultiType",
+			typeName: "object",
+			input: resource.NewObjectProperty(resource.NewPropertyMapFromMap(map[string]interface{}{
+				"prop": []interface{}{"foo"},
+			})),
+			failures: []TypeFailure{
+				{
+					Reason:       "expected string type, got [] type",
```
iwahbe, line 268: If we can't resolve the inner type here, can we call this a list type?
```suggestion
					Reason:       "expected string type, got list type",
```

iwahbe (commented):
This looks pretty good to me. Before we roll out a change like this, can you create a branch that has `PULUMI_ERROR_TYPE_CHECKER` default to on, then run all of our downstream tests. If our downstream tests are valid and the type checker is valid, then they should all pass.

pkg/tfbridge/validate_input_types.go
t0yv0: This is a strange encoding I don't quite fully understand how it's interpreted either, would need to dig in.

Looking for examples in the wild.

https://github.com/pulumi/pulumi-aws/blob/upstream-v5.46.0/provider/cmd/pulumi-resource-aws/schema.json#L126765

Looks like for all examples I could find, these are enums that can also be strings. So the encoding is specifying that .type is string ("default"), but then also providing string as one of the alternatives, and the enum type as the other alternative.

pkg/tfbridge/validate_input_types.go
t0yv0: I see it's under test, checking up on it, hm..

pkg/tfbridge/validate_input_types_test.go
corymhall: To generate this I'm just using `PropertyValue.TypeString()` and it unfortunately uses `[]` as the value for array.

https://github.com/pulumi/pulumi/blob/ff8b280cc962b66dcc0c7756575b4469a6c0cf7d/sdk/go/common/resource/properties.go#L558-L560

I can definitely customize this to return `list` instead if that is preferred.

pkg/tfbridge/validate_input_types.go
t0yv0: Removing this block makes TestValidateInputType_arrays/object_double_nested_object_type_success4 fail, I've instrumented that test case and I think it's actually not asserting the right behavior, with this block removed I think the validator correctly fails on that test case. Step by step:

```
=== RUN   TestValidateInputType_arrays/object_double_nested_object_type_success

# validating against array of objects
validatePropertyValue {[{map[prop:{map[objectStringProp:{foo}]}]}]} object_double_nested_object_type_success4 {
  "type": "array",
  "items": {
    "$ref": "#/types/pkg:index/type:ObjectDoubleNestedObjectType"
  }
}

# validating first element against an object
validatePropertyValue {map[prop:{map[objectStringProp:{foo}]}]} object_double_nested_object_type_success4[0] {
  "$ref": "#/types/pkg:index/type:ObjectDoubleNestedObjectType"
}

# down to prop attribute of the object, validating against a map of objects
validatePropertyValue {map[objectStringProp:{foo}]} object_double_nested_object_type_success4[0].prop {
  "type": "object",
  "additionalProperties": {
    "type": "object",
    "$ref": "#/types/pkg:index/type:ObjectStringType"
  }
}

# down objectStringProp map key, validating "foo" string against an object type
validatePropertyValue {foo} object_double_nested_object_type_success4[0].prop.objectStringProp {
  "type": "object",
  "$ref": "#/types/pkg:index/type:ObjectStringType"
}

# this should fail as it does 
    validate_input_types_test.go:1552: 0 failures, got 1: [{expected object type, got string type object_double_nested_object_type_success4[0].prop.objectStringProp}]
--- FAIL: TestValidateInputType_arrays (0.00s)
    --- FAIL: TestValidateInputType_arrays/object_double_nested_object_type_success4 (0.00s)
```

t0yv0: E.g. for the double nested object type, TypeScript would have to supply something like this:

```
{"prop":  {"key1": {"objectStringProp": "A"}, "key2": {"objectStringProp": "B"}}}
```

t0yv0: We can always check by generating a TypeScript SDK for a schema that looks like that.

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,337 @@
...
+		failures := []TypeFailure{}
+		for _, spec := range typeSpec.OneOf {
+			failure := v.validatePropertyValue(propertyValue, spec, propertyPath)
+			if len(failure) == 0 {
+				// we have found the correct type
+				return nil
+			}
+			failures = append(failures, failure...)
+		}
+		// try to find the best failure message
```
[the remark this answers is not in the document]
corymhall: ok i'll remove the `OneOf` checking and leave a TODO.

pkg/tfbridge/validate_input_types_test.go
iwahbe: If it's coming from upstream, we can leave it for now.

pkg/tfbridge/provider.go
corymhall: Updated!

pkg/tfbridge/validate_input_types.go
```diff
@@ -0,0 +1,337 @@
...
+// recursive function that will validate nested types and arrays. Returns a list of type failures if any are found.
+func (v *PulumiInputValidator) validatePropertyMap(
+	propertyMap resource.PropertyMap,
+	propertyTypes map[string]pschema.PropertySpec,
+	propertyPath resource.PropertyPath,
+) []TypeFailure {
+	stableKeys := propertyMap.StableKeys()
+	failures := []TypeFailure{}
+
+	// TODO: handle required properties. Deferring for now
```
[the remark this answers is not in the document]
corymhall: Done!

pkg/tfbridge/validate_input_types.go
corymhall: Ok I remove that block

t0yv0 (approved):
This LGTM to me if we can pass a version of this that panics instead of warning and have all downstream tests pass with that (sorry for being so paranoid). Code looks good but we've historically had some trouble with this codebase behaving safely.

t0yv0:
Thanks @corymhall for all the work here.

flostadler (approved):

corymhall:
> This LGTM to me if we can pass a version of this that panics instead of warning and have all downstream tests pass with that (sorry for being so paranoid). Code looks good but we've historically had some trouble with this codebase behaving safely.

Ran this branch through the downstream tests and nothing failed
https://github.com/pulumi/pulumi-terraform-bridge/tree/corymhall/better-type-checking-enabled-test

PR #1812, opened 2024-03-29
Correctly handle MaxItems: 1 for extractSchemaInputs

iwahbe:
Fixes https://github.com/pulumi/pulumi-cloudflare/issues/554

This corrects MaxItems: 1 handling for `extractSchemaInputs`. As part of that, I have separated the handling of `map<string,elem>` and `object` types to their own function. This vastly simplifies the code.

Tests are incoming shortly.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -1652,6 +1652,32 @@ func isDefaultOrZeroValue(tfs shim.Schema, ps *SchemaInfo, v resource.PropertyVa
 func extractSchemaInputs(state resource.PropertyValue, tfs shim.Schema, ps *SchemaInfo,
 	rawNames bool) resource.PropertyValue {
 
+	if ps == nil {
+		ps = &SchemaInfo{}
+	}
+
+	for IsMaxItemsOne(tfs, ps) {
+		ps = ps.Elem
+		if ps == nil {
+			ps = &SchemaInfo{}
+		}
+
+		switch e := tfs.Elem().(type) {
+		case shim.Schema:
+			tfs = e
+		case shim.Resource:
+			if state.IsObject() {
+				return resource.NewProperty(
+					extractSchemaInputsObject(state.ObjectValue(), e.Schema(), ps.Fields),
+				)
+			} else {
+				// The shape of state and tfs have de-synchronized, so we
+				// just return state as is.
+				return state
+			}
+		}
+	}
+
 	switch {
 	case state.IsArray():
 		etfs, eps := elemSchemas(tfs, ps)
@@ -1663,34 +1689,25 @@ func extractSchemaInputs(state resource.PropertyValue, tfs shim.Schema, ps *Sche
 		}
 		return resource.NewArrayProperty(v)
 	case state.IsObject():
-		var tfflds shim.SchemaMap
-		if tfs != nil {
-			if res, isres := tfs.Elem().(shim.Resource); isres {
-				tfflds = res.Schema()
-			}
-		}
-		var psflds map[string]*SchemaInfo
-		if ps != nil {
-			psflds = ps.Fields
-		}
-		isMap := tfflds == nil
-
 		obj := state.ObjectValue()
-		v := make(map[resource.PropertyKey]resource.PropertyValue, len(obj))
-		for k, e := range obj {
-			_, etfs, eps := getInfoFromPulumiName(k, tfflds, psflds, rawNames || shimutil.IsOfTypeMap(tfs))
-			if isInput := isMap || etfs != nil && (etfs.Optional() || etfs.Required()); !isInput {
-				glog.V(9).Infof("skipping '%v' (not an input)", k)
-				continue
-			}
-
-			ev := extractSchemaInputs(e, etfs, eps, rawNames || shimutil.IsOfTypeMap(tfs))
-			if mustSet := isMap || etfs != nil && (etfs.Required() || !isDefaultOrZeroValue(etfs, eps, ev)); !mustSet {
-				glog.V(9).Infof("skipping '%v' (not required + default or zero value)", k)
-				continue
-			}
-			v[k] = ev
+		if tfflds, ok := shimutil.CastToTypeObject(tfs); ok {
+			return resource.NewProperty(
+				extractSchemaInputsObject(obj, tfflds, ps.Fields),
+			)
 		}
+
+		// state does not represent an object, so it represents a map.
+		//
+		// That means that state's keys don't have special semantics.
+		v := make(map[resource.PropertyKey]resource.PropertyValue, len(obj))
+		etfs, eps := elemSchemas(tfs, ps)
+		for k, e := range obj {
+			v[k] = extractSchemaInputs(e, etfs, eps, rawNames)
+		}
+
+		// To match previous behavior, we insert the default key for Map types.
+		//
+		// TODO: We should probably remove the extraneous defaultsKey here.
 		v[defaultsKey] = resource.NewArrayProperty([]resource.PropertyValue{})
 		return resource.NewObjectProperty(v)
 	default:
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/schema.go
```diff
@@ -1663,59 +1690,75 @@ func extractSchemaInputs(state resource.PropertyValue, tfs shim.Schema, ps *Sche
...
+) resource.PropertyMap {
+	v := make(map[resource.PropertyKey]resource.PropertyValue, len(state))
+	for k, e := range state {
+		_, etfs, eps := getInfoFromPulumiName(k, tfs, ps, false)
+		typeKnown := tfs != nil && etfs != nil
+
+		if !typeKnown || !(etfs.Optional() || etfs.Required()) {
+			glog.V(9).Infof("skipping '%v' (not an input)", k)
+			continue
+		}
```
iwahbe, line 1731: This matches prior logic, but I'm surprised that we ignore (drop) unknown properties.

pkg/tfbridge/schema.go
```diff
@@ -1663,59 +1690,75 @@ func extractSchemaInputs(state resource.PropertyValue, tfs shim.Schema, ps *Sche
...
-				continue
-			}
-			v[k] = ev
+			v[k] = extractSchemaInputs(e, etfs, eps, rawNames)
 		}
+
+		// To match previous behavior, we insert the default key for Map types.
+		//
+		// TODO: We should probably remove the extraneous defaultsKey here.
 		v[defaultsKey] = resource.NewArrayProperty([]resource.PropertyValue{})
```
iwahbe, line 1712: Before this PR, we didn't clearly distinguish between Map and Object very well, so we inserted the defaults list in all cases. To avoid changing a bunch of existing tests, I'm leaving the old behavior as is for now. 

Map's can't have field defaults, so I'm pretty sure that this is wrong. Fixing is a followup PR job.

pkg/tfbridge/schema.go
```diff
@@ -1652,6 +1652,33 @@ func isDefaultOrZeroValue(tfs shim.Schema, ps *SchemaInfo, v resource.PropertyVa
 func extractSchemaInputs(state resource.PropertyValue, tfs shim.Schema, ps *SchemaInfo,
 	rawNames bool) resource.PropertyValue {
 
+	if ps == nil {
+		ps = &SchemaInfo{}
+	}
+
+	// If ps is MaxItemsOne, then it will not be present in the
```
t0yv0, line 1659: Dangling sentence.

pkg/tfbridge/schema.go
```diff
@@ -1652,6 +1652,33 @@ func isDefaultOrZeroValue(tfs shim.Schema, ps *SchemaInfo, v resource.PropertyVa
...
+		switch e := tfs.Elem().(type) {
+		case shim.Schema:
+			tfs = e
+		case shim.Resource:
+			if state.IsObject() {
+				return resource.NewProperty(
+					extractSchemaInputsObject(state.ObjectValue(), e.Schema(), ps.Fields),
+				)
+			} else {
+				// The shape of state and tfs have de-synchronized, so we
```
t0yv0, line 1675: Sure about this? This is tolerating schema mismatch? Tough choices.

pkg/tfbridge/schema.go
```diff
@@ -1652,6 +1652,33 @@ func isDefaultOrZeroValue(tfs shim.Schema, ps *SchemaInfo, v resource.PropertyVa
...
+		if ps == nil {
+			ps = &SchemaInfo{}
+		}
+
+		switch e := tfs.Elem().(type) {
+		case shim.Schema:
+			tfs = e
+		case shim.Resource:
+			if state.IsObject() {
+				return resource.NewProperty(
```
t0yv0, line 1671: should this be NewObjectProperty to be more type-precise (nit)

pkg/tfbridge/schema.go
t0yv0: Yeah make a comment to highlight that it's surprising? I don't have a good intuition on why the input extraction should ignore known properties, could be useful eventually to understand. Looks like some "mismatching schema" toleration features leaking here tbh.

t0yv0:
Thanks for wrangling this, looking forward to the tests. Not the easiest code to work on.

pkg/tfbridge/schema.go
iwahbe: We were always tolerating schema mismatches, this just calls it out explicitly.

t0yv0: Got it. Hmpf.

pkg/tfbridge/schema.go
iwahbe: I added a comment to the code.

pkg/tfbridge/schema.go
iwahbe: Can you clarify what you mean by "more type-precise". `NewObjectProperty` is more explicit, but I don't see why we shouldn't use type inference. [`NewProperty`](https://github.com/pulumi/pulumi/blob/28f098a660aa698c83a77058da28f7867707d0a0/sdk/go/common/resource/properties.go#L237-L240) is fully type safe:

```go
// PropertyValueType enumerates the actual types that may be stored in a PropertyValue.
//
//nolint:lll
type PropertyValueType interface {
	bool | float64 | string | *asset.Asset | *archive.Archive | Computed | Output | *Secret | ResourceReference | []PropertyValue | PropertyMap
}

// NewProperty creates a new PropertyValue.
func NewProperty[T PropertyValueType](v T) PropertyValue {
	return PropertyValue{v}
}
```

t0yv0: Just out of habit, I'd use least general primitives, in case the underlying type switches to string, that's a bug. But it's purely theoretical so disregard the nit.

pkg/tfbridge/schema_test.go
```diff
@@ -2588,6 +2589,200 @@ func TestExtractDefaultIntegerInputs(t *testing.T) {
...
+
+	provider := func(info *ResourceInfo) *Provider {
+		if info == nil {
+			info = &ResourceInfo{}
+		}
+		if info.Tok == "" {
+			info.Tok = tokens.NewTypeToken("module", "importableResource")
+		}
+
+		listOfObj := func(maxItems int) *schemav1.Schema {
```
t0yv0, line 2602: prefer schemav2 for new tests perhaps? or is cloudflare on schemav1

pkg/tfbridge/schema_test.go
```diff
@@ -2588,6 +2589,200 @@ func TestExtractDefaultIntegerInputs(t *testing.T) {
 	assert.Equal(t, expected, ins)
 }
 
+func TestExtractSchemaInputsNestedMaxItemsOne(t *testing.T) {
```
t0yv0, line 2592: Looks good! Trust these test cases discriminate the bug.

iwahbe: I checked. Both fail in different ways on the old code.

pkg/tfbridge/schema.go
```diff
@@ -1652,6 +1652,32 @@ func isDefaultOrZeroValue(tfs shim.Schema, ps *SchemaInfo, v resource.PropertyVa
 func extractSchemaInputs(state resource.PropertyValue, tfs shim.Schema, ps *SchemaInfo,
 	rawNames bool) resource.PropertyValue {
 
+	if ps == nil {
+		ps = &SchemaInfo{}
+	}
+
+	for IsMaxItemsOne(tfs, ps) {
```
t0yv0, line 1659: It's fantastic to get quick fix out! IN the longer term maybe worth looking into abstracting out the traversal using path-based or some such methods, so the codebase deals with these idiosyncrasies in one place and gets it right once and for all.

pkg/tfbridge/schema_test.go
iwahbe: Cloudflare is on v2. This test doesn't really care about the underlying provider framework, but I can switch to v2.

PR #1813, opened 2024-03-29
Ensure that asset translation does not clobber the files

t0yv0:
Adds a test to regress against the situation that caused the AWS P1 in https://github.com/pulumi/pulumi-aws/issues/3622


VenelinMartinov (approved):

t0yv0:
Good to know. Yes we can merge next week, just a regression test. It's fixed.

PR #1836, opened 2024-04-03
Don't interpolate provider description fields

iwahbe:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1776

- 028b4898a583a2ae54a411e800de26c6caad447e adds a CI check to prevent this kind of issue from re-accuring.


VenelinMartinov (approved):
Approving to unblock.

It'd be clearer if you separated out the refactoring changes from your meaningful changes.
I don't understand which part is the actual fix since most of the changes look like you are propagating errors?

Have you verified that the existing tests catch the error and we won't regress here?

guineveresaenger:
This LGTM but I would've like to see a test showing the (MISSING) behavior.

to answer the reviewer question: wrapping the input into `%s` every time prevents other raw upstream format strings from being written to the buffer.

pkg/tfgen/docs.go
```diff
@@ -1467,27 +1470,26 @@ func (g *Generator) convertExamplesInner(
...
-						fprintf(output, endPulumiCodeChooser)
+						fprintf("%s\n%s\n%s",
+							startPulumiCodeChooser, convertedBlock, endPulumiCodeChooser)
 					}
 				}
 			} else {
 				// Take already-valid code blocks as-is.
 				if hasHeader {
-					fprintf(output, docs[tfBlock.headerStart:tfBlock.start])
+					fprintf("%s", docs[tfBlock.headerStart:tfBlock.start])
```
iwahbe, line 1482: This is probably the actual fix.

iwahbe:
> It'd be clearer if you separated out the refactoring changes from your meaningful changes. I don't understand which part is the actual fix since most of the changes look like you are propagating errors?

I did a `fmt.*f` cleanliness check over the whole file. Anytime you call a function like `fmt.Sprintf(someString, arg1, arg2)` you are interpolating `arg1` and `arg2` into `someString`. If you know what `someString` is, then thats fine. If `someString` is partially user defined, then it's a problem. My expectation is the actual fix is [here](https://github.com/pulumi/pulumi-terraform-bridge/pull/1836/files#r1550167607), but it could have been several places.

> Have you verified that the existing tests catch the error and we won't regress here?

Yes. That https://github.com/pulumi/pulumi-terraform-bridge/pull/1836/commits/d7e1645d5c4838a6a05673db193450ee977cac59 fails CI ([check](https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/8542392449/job/23403902465)) shows the area is under test.

PR #1841, opened 2024-04-04
Add a helper function for delegating to another field with ComputeID

iwahbe:
ComputeID is used in 2 places by Pulumi:

- In `pulumi-random` to delegate "id" to  "base64"
- In `pulumi-artifactory` to delegate "id" to "key"

The function to do this is non-trivial.

This PR adds a helper method `tfbridge.DelegateIDField` to perform this function.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -385,8 +385,13 @@ type ResourceInfo struct {
 	// resources. It is called in Create(preview=false) and Read provider methods.
 	//
 	// This option is currently only supported for Plugin Framework based resources.
-	ComputeID func(ctx context.Context, state resource.PropertyMap) (resource.ID, error)
+	//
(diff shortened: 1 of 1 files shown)


pkg/tfbridge/info.go
```diff
@@ -1374,3 +1379,36 @@ type SkipExamplesArgs struct {
...
+
+		c := fmt.Sprintf(". This is an error in %s resource provider, please report at "+
+			"%s.", providerName, repoURL)
+
+		fieldValue, ok := state[field]
+		if !ok {
+			return "", fmt.Errorf("Could not find required property '%s' in state"+c, field)
+		}
+
+		// ComputeID is only called during when preview=false, so we don't need to
```
t0yv0, line 1394: This could be an assert.

t0yv0 (approved):
LGTM after it's covered by a test - for good form. Thanks.

Issue #1839, opened 2024-04-03
AWS build warning: unexpected HCL snippet in Convert (x118)


t0yv0:
### What happened?

```
   1 warning: failed to convert HCL for #/types/aws:autoscaling/GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements:GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements/acceleratorManufacturers to csharp: <nil>: unexpected HCL snippet in Convert; 
   1 warning: failed to convert HCL for #/types/aws:autoscaling/GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements:GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements/acceleratorManufacturers to go: <nil>: unexpected HCL snippet in Convert; 
   1 warning: failed to convert HCL for #/types/aws:autoscaling/GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements:GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements/acceleratorManufacturers to java: <nil>: unexpected HCL snippet in Convert; 
   1 warning: failed to convert HCL for #/types/aws:autoscaling/GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements:GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements/acceleratorManufacturers to python: <nil>: unexpected HCL snippet in Convert; 
   1 warning: failed to convert HCL for #/types/aws:autoscaling/GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements:GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements/acceleratorManufacturers to typescript: <nil>: unexpected HCL snippet in Convert; 
   1 warning: failed to convert HCL for #/types/aws:autoscaling/GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements:GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements/acceleratorManufacturers to yaml: <nil>: unexpected HCL snippet in Convert; 
[...]

PR #1842, opened 2024-04-04
Do not fail conversion of non-HCL fenced markdown snippets

t0yv0:
Fix https://github.com/pulumi/pulumi-terraform-bridge/issues/1839

The documentation pipeline was looking for Markdown code-fenced blocks. In the case where they specify terraform or hcl
as the language, it is clear that they should be converted. However the unmarked blocks may be carrying code that needs
to be converted or else they can be carrying text that needs to be displayed with a fixed font.

This change introduces a regex-based guess that allows non-Terraform code snippets to pass through as-is.

Fixes the following AWS build warnings (107-118 total, getting numbers a bit mixed up):

```
   1 warning: failed to convert HCL for #/types/aws:autoscaling/GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements:GroupMixedInstancesPolicyLaunchTemplateOverrideInstanceRequirements/acceleratorManufacturers to csharp: <nil>: unexpected HCL snippet in Convert;
```

### DIFF
--- a/pkg/tfgen/convert_cli.go
+++ b/pkg/tfgen/convert_cli.go
@@ -173,6 +173,8 @@ func (cc *cliConverter) FinishConvertingExamples(p pschema.PackageSpec) pschema.
 	return result
 }
 
+const cliConverterErrUnexpectedHCLSnippet = "unexpected HCL snippet in Convert"
+
 // During FinishConvertingExamples pass, generator calls back into this function to continue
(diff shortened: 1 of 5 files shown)


iwahbe (approved):
Looks reasonable to me.

guineveresaenger (commented):
Nice refinement! Thank you.

PR #1843, opened 2024-04-04
Reduce tfgen failing example warning verbosity

t0yv0:
Before this change we emitted a lot of duplicated warnings for examples that failed to convert. This change removes the
duplication and aggregates similar errors so that when every language conversion fails with the same error, which is
common when the failure pertains to the HCL->PCL pass of the conversion, we get one warning only not many.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1793,8 +1793,8 @@ func (g *Generator) convertHCL(e *Example, hcl, path string, languages []string)
 			err = multierror.Append(err, fmt.Errorf("[%s] %w", ls, convertErr))
 		}
 
-		g.warn(fmt.Sprintf("unable to convert HCL example for Pulumi entity '%s': %v. The example will be dropped "+
-			"from any generated docs or SDKs.", path, err))
+		g.warn("unable to convert HCL example for Pulumi entity '%s'. The example will be dropped "+
+			"from any generated docs or SDKs: %v", path, err)
 
 		return "", err
 	}
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1757,59 +1757,57 @@ func (g *Generator) convertHCL(e *Example, hcl, path string, languages []string)
 	}
 
 	result.WriteString(hclConversionsToString(hclConversions))
-	if len(failedLangs) == 0 {
+
+	switch {
+	// Success
+	case len(failedLangs) == 0:
+		return result.String(), nil
+	// Complete failure - every language conversion failed; error
+	case len(failedLangs) == len(languages):
+		err := g.warnUnableToConvertHCLExample(path, failedLangs)
+		return "", err
+	// Partial failure - not returning an error but still emit the warning
+	default:
+		err := g.warnUnableToConvertHCLExample(path, failedLangs)
+		contract.IgnoreError(err)
 		return result.String(), nil
 	}
+}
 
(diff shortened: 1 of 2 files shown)


pkg/tfgen/docs.go
```diff
@@ -1767,6 +1764,34 @@ func (g *Generator) convertHCL(e *Example, hcl, path string, languages []string)
...
+			m := convertErr.Error()
+			seen[m] = struct{}{}
+
+			langs := []string{}
+			for l := range languagesByErrMsg[m] {
+				langs = append(langs, l)
+			}
+			sort.Strings(langs)
+			ls := strings.Join(langs, ", ")
+			err = multierror.Append(err, fmt.Errorf("[%s] %w", ls, convertErr))
```
guineveresaenger, line 1792: at first glance, this looks like we're not just skipping duplicate errors but also consolidating different errors?

pkg/tfgen/docs.go
```diff
@@ -1767,6 +1764,34 @@ func (g *Generator) convertHCL(e *Example, hcl, path string, languages []string)
 	isCompleteFailure := len(failedLangs) == len(languages)
 
 	if isCompleteFailure {
+		languagesByErrMsg := map[string]map[string]struct{}{}
+		for lang, convertErr := range failedLangs {
+			m := convertErr.Error()
```
guineveresaenger, line 1769: As this is used further down, can we name `m` `errMsg`?

pkg/tfgen/docs.go
```diff
@@ -1767,6 +1764,34 @@ func (g *Generator) convertHCL(e *Example, hcl, path string, languages []string)
...
+			if _, ok := languagesByErrMsg[m]; !ok {
+				languagesByErrMsg[m] = map[string]struct{}{}
+			}
+			languagesByErrMsg[m][lang] = struct{}{}
+		}
+
+		var err error
+
+		seen := map[string]struct{}{}
```
guineveresaenger, line 1778: Why does this map need to be mapping to a struct?

guineveresaenger (commented):
I had a couple questions, and I would like to see a few comments explaining our approach to consolidating the errors but this looks good to me.

pkg/tfgen/docs.go
t0yv0: This is a classic go representation of Sets. Unfortunately there is no Set in stdlib like in Python.

pkg/tfgen/docs.go
t0yv0: Good point, done.

pkg/tfgen/docs.go
t0yv0: Right.

pkg/tfgen/docs.go
```diff
@@ -1767,6 +1764,34 @@ func (g *Generator) convertHCL(e *Example, hcl, path string, languages []string)
 	isCompleteFailure := len(failedLangs) == len(languages)
 
 	if isCompleteFailure {
+		languagesByErrMsg := map[string]map[string]struct{}{}
```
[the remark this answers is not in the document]
t0yv0: Done

t0yv0:
Missed handling the partial failure case that's still super duplicative. Adding a few more commits here shortly.

t0yv0:
Done now

guineveresaenger (approved):
LGTM

PR #1848, opened 2024-04-09
Add env vars to speed up bisecting tfgen

t0yv0:
Schema generation takes a long time for large providers like pulumi-aws and adding these helper env vars to limit it to
one problematic resource or data source can be a helpful tool to speed up bisecting over bridge ranges.


VenelinMartinov (approved):

iwahbe (approved):
This is great for quick iteration. 

Can you add a reference to this in `docs` so this is findable for others? Maybe `docs/dev#env vars`?

t0yv0:
Sure

PR #1851, opened 2024-04-10
Remove `rawNames` from `makeObjectTerraformInputs`

### LINKED ISSUE #1830: Refactor `extractInputs` to remove `rawNames`
`rawNames` makes the implementation of `extractInputs` confusing and is incorrect. We should clean up the area and narrow down the desired behavior of `rawNames`, document the desired behavior and put it under test.

I know recent PRs have changed behavior here for `Read` and tests did not flag the change.

iwahbe:
It never makes sense to pass `rawNames=true` to `makeObjectTerraformInputs` since that implies we have an object type with properties, but that we should not do property name translations. This is never the case.

This PR introduces a subtle change in behavior when walking values of unknown type:

Before this PR, we assumed that an untyped `resource.PropertyMap` was an object (name translation is performed) but all nested `resource.PropertyMap` values were then assumed to be a map (name translation is not performed).

After this PR, we treat all untyped `resource.PropertyMap` values as maps. IMO this makes more sense: the rule is we don't apply name translation unless we know a value is an object (and can thus generate naming tables for it).

I don't expect that this change will break users since user's should never hit untyped values. If they do, neither before or after will work consistently.

Addressing the test change:

"object_property_value" was used to test object names but never given a type. In light of the above change, we needed to make it typed to get the expected name translation.

---

Fixes #1830 

This PR is best reviewed on a commit by commit basis. Each commit passes tests and includes a description of what it does (if anything). The commits build, with each commit able to remove `rawNames` from one additional place.

### DIFF
--- a/pkg/tfbridge/diff.go
+++ b/pkg/tfbridge/diff.go
@@ -24,7 +24,6 @@ import (
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
-	shimutil "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/util"
 )
 
 // containsComputedValues returns true if the given property value is or contains a computed value.
@@ -70,7 +69,7 @@ type propertyVisitor func(attributeKey, propertyPath string, value resource.Prop
 // check to see if the InstanceDiff has an entry for that path.
 func visitPropertyValue(
 	ctx context.Context, name, path string, v resource.PropertyValue, tfs shim.Schema,
-	ps *SchemaInfo, rawNames bool, visitor propertyVisitor) {
+	ps *SchemaInfo, visitor propertyVisitor) {
 
 	if IsMaxItemsOne(tfs, ps) {
 		if v.IsNull() {
@@ -127,21 +126,33 @@ func visitPropertyValue(
 			}
 
 			en := name + "." + ti
-			visitPropertyValue(ctx, en, ep, e, etfs, eps, rawNames, visitor)
+			visitPropertyValue(ctx, en, ep, e, etfs, eps, visitor)
 		}
 	case v.IsObject():
-		var tfflds shim.SchemaMap
 		if tfs != nil {
-			if res, isres := tfs.Elem().(shim.Resource); isres {
-				tfflds = res.Schema()
+			if res, ok := tfs.Elem().(shim.Resource); ok {
+				tfflds := res.Schema()
+				var psflds map[string]*SchemaInfo
+				if ps != nil {
+					psflds = ps.Fields
+				}
+
+				for k, e := range v.ObjectValue() {
+					var elementPath string
+					if strings.ContainsAny(string(k), `."[]`) {
+						elementPath = fmt.Sprintf(`%s.["%s"]`, path, strings.ReplaceAll(string(k), `"`, `\"`))
+					} else {
+						elementPath = fmt.Sprintf("%s.%s", path, k)
+					}
+
+					en, etf, eps := getInfoFromPulumiName(k, tfflds, psflds, false)
+					visitPropertyValue(ctx, name+"."+en, elementPath, e, etf, eps, visitor)
+				}
+				return
 			}
 		}
-		var psflds map[string]*SchemaInfo
-		if ps != nil {
-			psflds = ps.Fields
-		}
 
-		rawElementNames := rawNames || shimutil.IsOfTypeMap(tfs)
+		etfs, eps := elemSchemas(tfs, ps)
 		for k, e := range v.ObjectValue() {
 			var elementPath string
 			if strings.ContainsAny(string(k), `."[]`) {
@@ -150,8 +161,7 @@ func visitPropertyValue(
 				elementPath = fmt.Sprintf("%s.%s", path, k)
 			}
 
-			en, etf, eps := getInfoFromPulumiName(k, tfflds, psflds, rawElementNames)
-			visitPropertyValue(ctx, name+"."+en, elementPath, e, etf, eps, rawElementNames, visitor)
+			visitPropertyValue(ctx, name+"."+string(k), elementPath, e, etfs, eps, visitor)
 		}
 	}
 }
@@ -166,7 +176,7 @@ func makePropertyDiff(
 	forceDiff *bool,
 	tfs shim.Schema,
 	ps *SchemaInfo,
-	finalize, rawNames bool,
+	finalize bool,
 ) {
 
 	visitor := func(name, path string, v resource.PropertyValue) bool {
@@ -257,7 +267,7 @@ func makePropertyDiff(
 		return false
 	}
 
-	visitPropertyValue(ctx, name, path, v, tfs, ps, rawNames, visitor)
+	visitPropertyValue(ctx, name, path, v, tfs, ps, visitor)
 }
 
 func newIgnoreChanges(
@@ -296,11 +306,11 @@ func computeIgnoreChanges(
 	}
 	for k, v := range olds {
 		en, etf, eps := getInfoFromPulumiName(k, tfs, ps, false)
-		visitPropertyValue(ctx, en, string(k), v, etf, eps, shimutil.IsOfTypeMap(etf), visitor)
+		visitPropertyValue(ctx, en, string(k), v, etf, eps, visitor)
 	}
 	for k, v := range news {
 		en, etf, eps := getInfoFromPulumiName(k, tfs, ps, false)
-		visitPropertyValue(ctx, en, string(k), v, etf, eps, shimutil.IsOfTypeMap(etf), visitor)
+		visitPropertyValue(ctx, en, string(k), v, etf, eps, visitor)
 	}
 	return ignoredKeySet
 }
@@ -353,17 +363,17 @@ func makeDetailedDiffExtra(
 	for k, v := range olds {
 		en, etf, eps := getInfoFromPulumiName(k, tfs, ps, false)
 		makePropertyDiff(ctx, en, string(k), v, tfDiff, diff, collectionDiffs, forceDiff,
-			etf, eps, false, shimutil.IsOfTypeMap(etf))
+			etf, eps, false)
 	}
 	for k, v := range news {
 		en, etf, eps := getInfoFromPulumiName(k, tfs, ps, false)
 		makePropertyDiff(ctx, en, string(k), v, tfDiff, diff, collectionDiffs, forceDiff,
-			etf, eps, false, shimutil.IsOfTypeMap(etf))
+			etf, eps, false)
 	}
 	for k, v := range olds {
 		en, etf, eps := getInfoFromPulumiName(k, tfs, ps, false)
 		makePropertyDiff(ctx, en, string(k), v, tfDiff, diff, collectionDiffs, forceDiff,
-			etf, eps, true, shimutil.IsOfTypeMap(etf))
+			etf, eps, true)
 	}
 
 	changes := pulumirpc.DiffResponse_DIFF_NONE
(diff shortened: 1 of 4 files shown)


t0yv0:
> After this PR, we treat all untyped resource.PropertyMap values as maps. IMO this makes more sense: the rule is we don't apply name translation unless we know a value is an object (and can thus generate naming tables for it).

I think that indeed makes a lot more sense. But it might break existing programs out there. I don't know how often we hit the "lost the type" case, I suspect it's a lot less than a year ago where we had bugs in matching the right type to the right input. So we can move forward with this but recommend calling out explicitly in release notes.

pkg/tfbridge/schema.go
```diff
@@ -351,7 +351,6 @@ func (ctx *conversionContext) makeTerraformInput(
 	old, v resource.PropertyValue,
 	tfs shim.Schema,
 	ps *SchemaInfo,
-	rawNames bool,
```
t0yv0, line 354: Reducing the parameter spaces is always heart-warming.

pkg/tfbridge/schema.go
```diff
@@ -479,51 +478,40 @@ func (ctx *conversionContext) makeTerraformInput(
...
 			if ps != nil {
 				psflds = ps.Fields
 			}
-			var err error
-			input, err = ctx.makeObjectTerraformInputs(oldObject, v.ObjectValue(),
-				tfflds, psflds, rawNames)
+			obj, err := ctx.makeObjectTerraformInputs(oldObject, v.ObjectValue(), tfflds, psflds)
 			if err != nil {
-				return nil, err
+				return obj, err
```
t0yv0, line 498: Should this really be under `if err!=nil`?

iwahbe: Fixed.

pkg/tfbridge/diff.go
```diff
@@ -127,21 +126,33 @@ func visitPropertyValue(
...
+				tfflds := res.Schema()
+				var psflds map[string]*SchemaInfo
+				if ps != nil {
+					psflds = ps.Fields
+				}
+
+				for k, e := range v.ObjectValue() {
+					var elementPath string
+					if strings.ContainsAny(string(k), `."[]`) {
+						elementPath = fmt.Sprintf(`%s.["%s"]`, path, strings.ReplaceAll(string(k), `"`, `\"`))
```
t0yv0, line 143: Umm.. How is elementPath used? Should this be a type distinct from string that encapsulates stepping down like this?

iwahbe: I'm pretty sure it should be a `resource.PropertyPath`. I didn't change it because I didn't want to expand the scope of this PR any more. I'll stack another PR on top of this to refactor `elementPath`.

t0yv0: Hm yeah.

t0yv0:
I appreciate cleaning up rawNames as this was very difficult for me to grok, but I believe the only purpose of that indeed was to track through the recursion whether we are dealing with an object or a map, or else we lost the schema. 

I am very hesitant with touching this code without adding a ton of coverage first. This is a true minefield that can cause a lot of downstream pain. I'm trying to read existing tests now to see if I can spot anything or anything we could add.

One evidence is that the PR had a mismatched `if err` handling but did that get caught by the tests? 

One confusing property of how this function is written is how it's handling MaxItemsOne=1 interaction with the types. I think it *might* just work but it's not obvious to me.. I like code really obvious, like in F# and Rust.... the reason it works is MIO wraps the value in an array which then forces the schema lookup to do elemSchemas which forces the recursion down to Elem, which might just be what we need here.

Let me read the tests again and do some case analysis.

iwahbe:
> I appreciate cleaning up rawNames as this was very difficult for me to grok, but I believe the only purpose of that indeed was to track through the recursion whether we are dealing with an object or a map, or else we lost the schema.

That is the purpose, but `rawNames` as used is not fit for purpose. If we have a type then we can use that to determine if we are looking at a map or an object, and if we have no type then we have lost the schema.

> I am very hesitant with touching this code without adding a ton of coverage first. This is a true minefield that can cause a lot of downstream pain. I'm trying to read existing tests now to see if I can spot anything or anything we could add.

I'm pretty sure I fixed a couple of lurking bugs around nested objects. I'll try to add some tests for them.

> One evidence is that the PR had a mismatched `if err` handling but did that get caught by the tests?

That was a cosmetic error:

```patch
			obj, err = ctx.makeObjectTerraformInputs(oldObject, v.ObjectValue(),
				tfflds, psflds)
			if err != nil {
+				return nil, err
-				return obj, err

			}
```

Since `ctx.makeObjectTerraformInputs` always returns `nil, err` if `err != nil`, no test could have caught it.

> One confusing property of how this function is written is how it's handling MaxItemsOne=1 interaction with the types. I think it _might_ just work but it's not obvious to me.. I like code really obvious, like in F# and Rust.... the reason it works is MIO wraps the value in an array which then forces the schema lookup to do elemSchemas which forces the recursion down to Elem, which might just be what we need here.
> 
> Let me read the tests again and do some case analysis.

pkg/tfbridge/schema_test.go
```diff
@@ -583,10 +591,18 @@ func TestTerraformOutputsWithSecretsSupported(t *testing.T) {
 					},
 				},
 				f.NewSchemaMap(map[string]*schema.Schema{
-					// Type mapPropertyValue as a map so that keys aren't mangled in the usual way.
 					"float_property_value":     {Type: shim.TypeFloat},
 					"my_string_property_value": {Type: shim.TypeString},
-					"map_property_value":       {Type: shim.TypeMap},
+					"object_property_value": {
```
t0yv0, line 596:
```go
func TestTerraformInputSchemas(t *testing.T) {
	sch := map[string]*schemav2.Schema{
		// "schema_map": {Type: schemav2.TypeMap},  --> this is invalid
		"schema_map": {
			Type:     schemav2.TypeMap,
			Optional: true,
		},
		"unknown_array_value2": {
			Optional: true,
			Type:     schemav2.TypeList,
			MinItems: 1,
			Elem: (&schema.Resource{
				Schema: schemaMap(map[string]*schema.Schema{
					"required_property": {Type: shim.TypeString, Required: true},
					"conflicts_a":       {Type: shim.TypeString, ConflictsWith: []string{"conflicts_b"}},
					"conflicts_b":       {Type: shim.TypeString, ConflictsWith: []string{"conflicts_a"}},
				}),
			}).Shim(),
		},
		// INVALID: Type must be specified
		//
		// "object_property_value": {
		// 	Elem: &schemav2.Resource{
		// 		Schema: map[string]*schemav2.Schema{
		// 			"property_a": {Type: schemav2.TypeString, Required: true},
		// 			"property_b": {Type: schemav2.TypeBool, Optional: true},
		// 		},
		// 	},
		// },
	}
	err := schemav2.InternalMap(sch).InternalValidate(nil)
	require.NoError(t, err)
}

```

t0yv0: Let's try to add test cases that are representable in TF under SDKv2 and therefore realistic. I'm checking all prior test cases against InternalValidate, not all of them pass, sometimes it's little things like Optional: true.

t0yv0:
Consider rebasing on https://github.com/pulumi/pulumi-terraform-bridge/pull/1853 and first having commits that add test cases, and then having the change in behavior match the change in tests, that'd be awesome (as well as separate changes that refactor from changes that change behavior)!

This may be separate but having re-read these test cases, they don't really cover that many variations which is a pity. Having this covered would be interesting:

- [x] single-nested block is covered
- [ ] list-nested block
- [ ] set-nested block
- [ ] map attribue
- [ ] set attribute
- [ ] list attribute

I'm curious what happens with nils and unknowns in blocks in particular. This is coming up for cross-tests as well so maybe will be adding cases as part of that. A lot of things look quite suspect.

iwahbe:
> * [x]  single-nested block is covered
> * [ ]  list-nested block
> * [ ]  set-nested block
> * [ ]  map attribue
> * [ ]  set attribute
> * [ ]  list attribute

I'm going to open a PR to cover this list and get existing behavior under test before I merge here.

pkg/tfbridge/schema_test.go
iwahbe: This has now been addressed by https://github.com/pulumi/pulumi-terraform-bridge/pull/1853 and https://github.com/pulumi/pulumi-terraform-bridge/pull/1855, ensuring that all modified tests are `Validate`able. This PR has rebased on both.

iwahbe:
@t0yv0 After the escapade that ended in https://github.com/hashicorp/terraform-plugin-sdk/issues/62, I believe that only 2 commits introduce semantic changes, both of which update existing tests, demonstrating their behavior. The rest are pure refactors. Each commit has a commit message that details if it is a pure refactor or what behavioral change is expected. Given the tests in https://github.com/pulumi/pulumi-terraform-bridge/pull/1853,  https://github.com/pulumi/pulumi-terraform-bridge/pull/1855 and https://github.com/pulumi/pulumi-terraform-bridge/pull/1851/commits/7df97a6d2031756f820c061208a2719c4261c1dd (which doesn't change with the rest of this PR), I think we have enough coverage to merge.

pkg/tfbridge/schema_test.go
```diff
@@ -396,10 +396,31 @@ func TestTerraformOutputsWithSecretsSupported(t *testing.T) {
 		},
 		{
 			name: "object_property_value",
+
+			// NOTE: This input does not match the schema's type. The tfValue *should* be wrapped
+			// in a []any{}. [MakeTerraformOutputs] handles this case, but it shouldn't need to
+			// and tf should never return it.
```
t0yv0, line 402: Maybe it handles it because it's trying to do MaxItems=1 mismatch toleration or something. Hard to know.

pkg/tfbridge/schema_test.go
```diff
@@ -396,10 +396,31 @@ func TestTerraformOutputsWithSecretsSupported(t *testing.T) {
...
 			tfValue: map[string]interface{}{
 				"property_a": "a",
 				"property_b": true,
 			},
+			tfType: &schema.Schema{
```
t0yv0, line 407: Hmm, strange test case it didn't have tfType before so it was trying to parse untyped?

pkg/tfbridge/schema_test.go
```diff
@@ -3129,6 +3129,32 @@ func Test_makeTerraformInputsNoDefaults(t *testing.T) {
 			//nolint:lll
 			expect: autogold.Expect(map[string]interface{}{"nested_resource": []interface{}{map[string]interface{}{"configuration": map[string]interface{}{"configurationValue": true}}}}),
 		},
+		{
```
t0yv0, line 3132: Nice.

pkg/tfbridge/schema_test.go
```diff
@@ -3082,7 +3098,7 @@ func Test_makeTerraformInputsNoDefaults(t *testing.T) {
 				},
 			}),
 			//nolint:lll
-			expect: autogold.Expect(map[string]interface{}{"property_a": "a", "property_b": true, "property_c": map[string]interface{}{"nested_property_a": true}}),
+			expect: autogold.Expect(map[string]interface{}{"property_a": "a", "property_b": true, "property_c": map[string]interface{}{"nestedPropertyA": true}}),
```
t0yv0, line 3101: This is an observable change, and the tests is showing it. map_property_value indicates a Map with unknown element types, which is supported in TF, and the change is that the keys are now passed as-is instead of inflected to Pulumi name notation, which makes perfect sense.

t0yv0 (approved):
Thanks for wrangling difficult code! 

This is almost perfect form for legacy refactoring, one thing I'd do is add tests at the bottom of the commit stack and make changes on top, that'd be perfect form. 

Anyhow I think the change in behavior over TypeMap without a type looks reasonable. The new behavior is much more intuitive. 

We have not quite reached consensus that we need to freeze development on this codebase and only make changes that are motivated directly by customer-visible scenarios that cannot be satisfied without the change, so I'm on balance for shipping this, but at some point I believe we'll need to consider "freeze" the bridge in the above manner since in foundational codebases that are used everywhere every change is fraught with breaking something.

PR #1853, opened 2024-04-10
Refactor MakeTerraformInput tests

t0yv0:
As part of hardening the bridge, revisit tests for MakeTerraformInput.

It turns out that a lot of these tests used schema examples that do not pass InternalValidate that bridged providers should be running at build-time, and therefore these tests do not represent functionality actually used in production.

To assist making sure we can write good tests, shim layer is extended to expose SchemaMap.Validate().

Where possible, the test cases were corrected to pass the validation.

One exception is map_with_resource_elem test case that is discarded as this is rejected by the validation. Instead, a test case for regular map attributes is added.

Several test cases are suspect and may be regression-testing incorrect behavior, NOTE is added for those.

Finally, test cases are rearranged into table-driven tests to make them easier to maintain.


VenelinMartinov (approved):
looks good, thanks for fixing

PR #1855, opened 2024-04-11
Refactor TestTerraformOutputsWithSecretsSupported

iwahbe:
Followup to https://github.com/pulumi/pulumi-terraform-bridge/pull/1853.

Since https://github.com/pulumi/pulumi-terraform-bridge/pull/1851 also edits these tests, I thought they should also have (and require) correct types.

### DIFF
--- a/pkg/tfbridge/schema_test.go
+++ b/pkg/tfbridge/schema_test.go
@@ -330,170 +330,307 @@ type MyString string
 // TestTerraformOutputsWithSecretsSupported verifies that we translate Terraform outputs into Pulumi outputs and
 // treating sensitive outputs as secrets
 func TestTerraformOutputsWithSecretsSupported(t *testing.T) {
-	ctx := context.Background()
-	for _, f := range factories {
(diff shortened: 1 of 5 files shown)


pkg/tfbridge/schema_test.go
```diff
@@ -330,170 +330,289 @@ type MyString string
...
+			},
+			tfType: &schema.Schema{
+				Type:     shim.TypeList,
+				MaxItems: 1,
+				Optional: true,
+				Elem: (&schema.Resource{
+					Schema: schemaMap(map[string]*schema.Schema{
+						"secret_value": {
+							Type:      shim.TypeString,
+							Sensitive: true,
```
t0yv0, line 558: Schema-based secrets, nice!

t0yv0 (approved):
These are strange test cases as not all of them test secrets. But I really appreciate the new rewritten form as that makes each testcase more localized and easily readable. We can extend this table as needed when in doubt about secrets handling. LGTM.

VenelinMartinov (approved):
yeah, I agree much more readable, thanks!

PR #1877, opened 2024-04-16
feat: add resource urn to context in Check call

corymhall:
This PR adds the resource URN to the context in the `Check` method, along with methods to set and get the URN from context.

The motivation for this change was driven by trying to implement the suggested fix in [this
comment](https://github.com/pulumi/pulumi-aws/issues/3788#issuecomment-2051881787) in the aws provider. We want to be able to keep some global state of which resources have been seen by `Check`, but realized that we don't actually get any identifying information in the `PreCheckCallback` function.

An alternative approach would be to update the `PreCheckCallback` signature to also contain the URN, but that would be a breaking change.

re https://github.com/pulumi/pulumi-aws/issues/3788

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -34,6 +34,26 @@ import (
 	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/logging"
 )
 
+type urnCtxKeyType string
+
+const urnCtxKey urnCtxKeyType = "urn"
+
+// WithUrn returns a copy of ctx with the resource URN value
+func WithUrn(ctx context.Context, urn resource.URN) context.Context {
+	return context.WithValue(ctx, urnCtxKey, urn)
+}
+
+// GetUrn gets a resource URN from context
+//
+//	urn, ok := GetUrn(ctx)
+//	if !ok {
+//	    // urn was not found
+//	}
+func GetUrn(ctx context.Context) (resource.URN, bool) {
+	urn, ok := ctx.Value(urnCtxKey).(resource.URN)
+	return urn, ok
+}
+
 const (
 	MPL20LicenseType      TFProviderLicense = "MPL 2.0"
 	MITLicenseType        TFProviderLicense = "MIT"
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/provider.go
```diff
@@ -726,6 +726,7 @@ func (p *Provider) Configure(ctx context.Context,
 func (p *Provider) Check(ctx context.Context, req *pulumirpc.CheckRequest) (*pulumirpc.CheckResponse, error) {
 	ctx = p.loggingContext(ctx, resource.URN(req.GetUrn()))
 	urn := resource.URN(req.GetUrn())
+	ctx = WithUrn(ctx, urn)
```
t0yv0, line 729: I'm amenable to this but the question becomes why only in Check? Perhaps you could retrofit `p.loggingContext` to do this; it's already installed in all the methods pretty much.

pkg/tfbridge/info.go
```diff
@@ -34,6 +34,26 @@ import (
...
+// GetUrn gets a resource URN from context
+//
+//	urn, ok := GetUrn(ctx)
+//	if !ok {
+//	    // urn was not found
+//	}
+func GetUrn(ctx context.Context) (resource.URN, bool) {
+	urn, ok := ctx.Value(urnCtxKey).(resource.URN)
+	return urn, ok
+}
```
iwahbe, line 55: It feels like for any call here, there either should or should not be a result for `GetUrn`, and which one should be knowable statically. Given that, I would recommend an API that doesn't support a failure mode:

```suggestion
// GetUrn gets a resource URN from context.
//
//	urn := GetUrn(ctx)
//
// GetUrn is available on any context associated with a resource. 
// Calling GetUrn on a context associated with an Invoke will panic.
func GetUrn(ctx context.Context) resource.URN {
	urn, ok := ctx.Value(urnCtxKey).(resource.URN)
	return urn, ok
}
```

EronWright:
An alternative to consider is to add a `PostCheckCallback`, so that the provider gets the checked inputs that represent the planned state.  It could be that the planned state is more useful than the pre-check state. I could imagine that defaulting or transformations happen during check.  Food for thought.

t0yv0:
Check does not return planed state in bridged providers, alas.

iwahbe (commented):
As implemented, I don't think this will work for PF based providers. If I'm wrong and it does, then we need to add a quick test to lock in that behavior. If it doesn't, we will need to add support (and a quick test to lock in behavior).

The equivalent method to `loggingContext` in PF land is `initLogging`:

https://github.com/pulumi/pulumi-terraform-bridge/blob/de2933ba29e365331516d1888cf4ee75a060c5a6/pf/tfbridge/logging.go#L25-L28

corymhall:
> The equivalent method to `loggingContext` in PF land is `initLogging`:

Done!

iwahbe (approved):
LGTM

PR #1878, opened 2024-04-16
Suggestions for better type checking branch

t0yv0:
Results of me digging into code review, please consider these alt structure for the code 🙏  

@corymhall 

CC @iwahbe


pkg/tfbridge/validate_input_types.go
```diff
@@ -35,215 +35,249 @@ func NewInputValidator(urn resource.URN, schema pschema.PackageSpec) *PulumiInpu
...
-
-	if spec.AdditionalProperties != nil {
-		if spec.AdditionalProperties.Ref != "" {
-			complexSpec = v.getType(spec.AdditionalProperties.Ref)
-			// if we don't find the reference then just continue instead of failing
-			if complexSpec != nil {
-				requiredProps = complexSpec.Required
+// validatePropertyValue is the main function for validating a PropertyMap against the Pulumi Schema. It is a
+// recursive function that will validate nested types and arrays. Returns a list of type failures if any are found.
+func (v *PulumiInputValidator) validatePropertyMap(
```
t0yv0, line 40: Both resource inputs and object types can fit this signature, makes it a little easier to reuse.

pkg/tfbridge/validate_input_types.go
```diff
@@ -35,215 +35,249 @@ func NewInputValidator(urn resource.URN, schema pschema.PackageSpec) *PulumiInpu
...
+	required []string,
+	propertyTypes map[string]pschema.PropertySpec,
+	propertyPath resource.PropertyPath,
+) []TypeFailure {
+	stableKeys := propertyMap.StableKeys()
+	failures := []TypeFailure{}
+
+	// handle required properties
+	//
+	// TODO: what if properties have defaults specified, should this still be an error?
```
t0yv0, line 51: Actually important to dig into this TODO, I think we might be throwing extraneous warnings otherwise, how things stand, where bridge would actually auto-fill the default post-Check. Pretty sure that's the case.. so perhaps just not validate these.

pkg/tfbridge/validate_input_types.go
```diff
@@ -35,215 +35,249 @@ func NewInputValidator(urn resource.URN, schema pschema.PackageSpec) *PulumiInpu
...
 
-	// perform some initial easy type matching. fail fast.
-	possible, ok := v.matchType(propertyValue, typeSpec)
-	if !ok {
-		return []TypeFailure{
-			{
+	// now we are going to switch on the desired type
+	//
+	// a good reference here for the semantics of the TypeSpec is
+	// https://github.com/pulumi/pulumi/blob/master/pkg/codegen/schema/bind.go#L881
```
t0yv0, line 133: This is the key reference for me to read to copy how it does case analysis to untangle the representation of TypeSpec. Including here in a comment.

pkg/tfbridge/validate_input_types.go
```diff
@@ -35,215 +35,249 @@ func NewInputValidator(urn resource.URN, schema pschema.PackageSpec) *PulumiInpu
...
+	// this code is not using the binder functionality as that would require a plugin loader to resolve references,
+	// which is not desirable for the provider runtime
+	//
+	// possibly this could be done in the future with a way to still use the bind code while resolving references to
+	// unchecked types as using the bound representation would make the code cleaner
+	//
+	// for now the code simply follows bindSpec
+
+	if typeSpec.Ref != "" {
+		objType := v.getType(typeSpec.Ref)
```
t0yv0, line 144: Per getType comment this is very low-power compared to what the PackageSpec supports. It only finds local object type definitions.

pkg/tfbridge/validate_input_types.go
```diff
@@ -35,215 +35,249 @@ func NewInputValidator(urn resource.URN, schema pschema.PackageSpec) *PulumiInpu
...
-				}
-			} else if len(typeSpec.OneOf) > 0 {
-				failure := v.validateOneOf("", arrayValue, typeSpec.OneOf, pb)
-				if failure != nil {
-					failures = append(failures, failure...)
-				}
-			}
-		}
+	if typeSpec.OneOf != nil {
+		// TODO type check against union types.
```
t0yv0, line 169: This is important case for maps and need to be filled in, should be trivial just omitting for lack of time.

t0yv0, line 169: I think this can be left for later. Fairly complex and not as frequently used.

pkg/tfbridge/validate_input_types.go
```diff
@@ -35,215 +35,249 @@ func NewInputValidator(urn resource.URN, schema pschema.PackageSpec) *PulumiInpu
...
+			Reason: fmt.Sprintf(
+				"expected an array, got %s type",
+				propertyValue.TypeString(),
+			),
+		}}
+	case "object":
+		// This is not really an object but a map type with some element type, which is assumed to be string if
+		// unspecified. Check accordingly. This should be very similar to the "array" case.
+		//
+		// elementTypeSpec := typeSpec.AdditionalProperties
```
t0yv0, line 244: This is the only case that needs to handle AdditionalProperties; This needs to be filled in I'm just leaving a TODO for lack of time. IT should be very straightforward along the array lines.

pkg/tfbridge/validate_input_types.go
```diff
@@ -35,215 +35,249 @@ func NewInputValidator(urn resource.URN, schema pschema.PackageSpec) *PulumiInpu
...
-func (v *PulumiInputValidator) getType(typeRef string) *pschema.ComplexTypeSpec {
+// getType gets a type definition from a schema reference. Currently it only supports types from the same schema that
+// are object types. It does not support enum types, foreign type references, special references such as
+// "pulumi.json#/Archive", references to resources or providers or anything else.
+func (v *PulumiInputValidator) getType(typeRef string) *pschema.ObjectTypeSpec {
 	if strings.HasPrefix(typeRef, "#/types/") {
 		ref := strings.TrimPrefix(typeRef, "#/types/")
 		if typeSpec, ok := v.schema.Types[ref]; ok {
-			return &typeSpec
+			// Exclude enum types here.
```
t0yv0, line 259: We could handle enums but it's more work, defer for later.

pkg/tfbridge/validate_input_types.go
```diff
@@ -35,215 +35,249 @@ func NewInputValidator(urn resource.URN, schema pschema.PackageSpec) *PulumiInpu
...
 
 // ValidateInputs will validate a set of inputs against the pulumi schema. It will
 // return a list of type failures if any are found
-func (v *PulumiInputValidator) ValidateInputs(inputs resource.PropertyMap) *[]TypeFailure {
-	failures := []TypeFailure{}
-	for key, value := range inputs {
-		if failure := v.validateResourceInputType(string(key), value, resource.PropertyPath{}); failure != nil {
-			failures = append(failures, *failure...)
-		}
+func (v *PulumiInputValidator) ValidateInputs(resourceToken tokens.Type, inputs resource.PropertyMap) *[]TypeFailure {
```
t0yv0, line 270: This really needs a resourceToken to find the right schema I think.

PR #1880, opened 2024-04-17
Enable bisecting failing conversions

t0yv0:
In the rare case when the bulk example converter panics, the experience of isolating the failing example and reporting a bug does not spark joy. This little change introduces a bisect algo that tries to narrow down an example-set to the one failing example, and also a better error reporting mechanism that prints a repro.sh in a temp folder where the maintainer can inspect further.

### DIFF
--- a/pkg/tfgen/convert_cli.go
+++ b/pkg/tfgen/convert_cli.go
@@ -22,7 +22,9 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
LGTM

A test would be great, but since this doesn't go out to pulumi users I'm fine without.

t0yv0:
The happy path is under test and the error path I got to use to diagnose https://github.com/pulumi/pulumi-digitalocean/issues/683 issue.

PR #1882, opened 2024-04-18
Remove docs matching hack and add more correct strategies for docs discovery.

### LINKED ISSUE #1789: Some property descriptions map to the wrong property
### What happened?

When we generate docs, we scrape upstream property descriptions and apply them to the corresponding pulumi property. We don't always get the mapping right.

Sometimes, especially when there are multiple properties with the same name (but different nesting), we match the wrong description with a property.

### Example

This issue motivated me to open a tracking issue in the bridge, but I have seen this happen multiple times.

- https://github.com/pulumi/pulumi-cloudflare/issues/661

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

### LINKED ISSUE #1875: Docs: When parameters are nested under other parameters, docs discovery does not work correctly
Example: [AWS CloudWatch Event Connection Header](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_event_connection#header) supports three subparameters, which are indented bullet points. 

Docs discovery does not currently have the ability to find and populate these fields correctly, relying instead on matching similar related field descriptions, often incorrectly, as [rendered on Pulumi in this case](https://www.pulumi.com/registry/packages/aws/api-docs/cloudwatch/eventconnection/#eventconnectionauthparametersoauthoauthhttpparametersheader).

Screenshot 2024-04-15 at 10 21 27 AM

### LINKED ISSUE #1876: Docs: Support subparameters for multiple headers
When upstream docs are of the format

`parameter_one` and `parameter_two` support the following arguments:
- arg1
- arg2
- arg3

then we do not correctly associate all of the subarguments with _both_ headers, resulting in the [need to match similar parameters with their descriptions](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfgen/generate.go#L1849), which is inaccurate and results in incorrect docs elsewhere.

We should introduce logic to match documentation fields correctly for this use case.

guineveresaenger:
This pull request is a draft.

It includes changes to docs discovery that ensure we are matching entityDocs to the correct property field descriptions.

- adds number pattern to regex so we can match "S3"
- uses full parent path as docs path so we can properly map different "key" field descriptions to the correct parent field
- adds new regex to nested header discovery
- allows for dot-spaced keys in entityDocs for more precise lookup

Fixes #1789

Fixes #1876

Fixes #1875

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -597,7 +597,6 @@ func (p *tfMarkdownParser) parse(tfMarkdown []byte) (entityDocs, error) {
 	footerLinks := getFooterLinks(markdown)
 
 	doc, _ := cleanupDoc(p.rawname, p.sink, p.infoCtx, p.ret, footerLinks)
-
 	return doc, nil
 }
 
@@ -822,14 +821,16 @@ func (p *tfMarkdownParser) parseSchemaWithNestedSections(subsection []string) {
 
 // parseArgFromMarkdownLine takes a line of Markdown and attempts to parse it for a Terraform argument and its
 // description
-func parseArgFromMarkdownLine(line string) (string, string, bool) {
+func parseArgFromMarkdownLine(line string) (string, string, bool, bool) {
 	matches := argumentBulletRegexp.FindStringSubmatch(line)
-
+	indentedBullet := false
 	if len(matches) > 4 {
-		return matches[1], matches[4], true
+		if strings.HasPrefix(matches[0], " ") {
+			indentedBullet = true
+		}
+		return matches[1], matches[4], true, indentedBullet
 	}
-
-	return "", "", false
+	return "", "", false, indentedBullet
 }
 
 var genericNestedRegexp = regexp.MustCompile("supports? the following:")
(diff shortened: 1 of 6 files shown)


pkg/tfgen/docs.go
```diff
@@ -822,82 +821,166 @@ func (p *tfMarkdownParser) parseSchemaWithNestedSections(subsection []string) {
 
 // parseArgFromMarkdownLine takes a line of Markdown and attempts to parse it for a Terraform argument and its
 // description
-func parseArgFromMarkdownLine(line string) (string, string, bool) {
+func parseArgFromMarkdownLine(line string) (string, string, bool, bool) {
```
iwahbe, line 824: I have no idea what this returns. Can we add a type here?
```suggestion
func parseArgFromMarkdownLine(line string) (struct{ bullet: string, text: string, indentedBullet: bool}, bool) {
```

I'm not sure the labels that I suggested are correct (which is kinda the point), but I think we need clarity here.

pkg/tfgen/docs.go
```diff
@@ -822,82 +821,166 @@ func (p *tfMarkdownParser) parseSchemaWithNestedSections(subsection []string) {
...
+	blockNames := tokenInBackticks.FindAllString(match, -1)
+	for _, blockName := range blockNames {
+		if blockName != "" {
+			blockName = strings.Trim(blockName, "`")
+			blockName = strings.ToLower(blockName)
+			blockName = strings.Replace(blockName, " ", "_", -1)
+			blockName = strings.TrimSuffix(blockName, "[]")
+			if subNest != "" {
+				// For the format ""The `grpc_route`, `http_route` and `http2_route` 's `action` object
+				//supports the following:" the result should be grpc_route.action
```
iwahbe, line 909: Nit: Formatting

```suggestion
				// For the format:
				//
				//	The `grpc_route`, `http_route` and `http2_route` 's `action` object supports the following: 
				//
				// the result should be grpc_route.action
```

pkg/tfgen/docs.go
```diff
@@ -822,82 +821,166 @@ func (p *tfMarkdownParser) parseSchemaWithNestedSections(subsection []string) {
...
+				blockName = blockName + "." + subNest
+			}
+			nestedBlockNames = append(nestedBlockNames, blockName)
+		}
+	}
+	return nestedBlockNames
 }
 
-// getNestedBlockName take a line of a Terraform docs Markdown page and returns the name of the nested block it
+func splitMatchOnColon(match string) []string {
```
iwahbe, line 918: I'm not sure I understand what this function does? Can we add an example in the doc comment and maybe look at changing the name. To me, `splitMatchOnColon` implies `strings.Split(match, ":")`.

iwahbe (approved):
Looks good to me.

We are waiting on an OK from the AWS folks to merge, right?

guineveresaenger:
Yeah. I opened https://github.com/pulumi/pulumi-aws/pull/3846 which only holds the schema diff when generated with d8ccf4c.

If we decide that that diff is too lossy for AWS, we have two options:

1. add a flag here - by contrast, Azure, GCP, and Cloudflare docs all look like they are very happy with these changes so we should definitely roll this out elsewhere.m
2. implement https://github.com/pulumi/pulumi-terraform-bridge/issues/1886, see if that helps significantly for AWS and go through another round of checking with AWS squad. 

Of course, we can also do 1 first and then implement 2 and remove the flag if deemed sufficiently better; if not, we could explore the option to flag resource-by-resource if top-used resources are negatively impacted.

guineveresaenger:
The AWS folks say 🚢 since it this change removes a lot of incorrect docs, a net improvement for them. Thanks @t0yv0 for taking a look!

PR #1887, opened 2024-04-19
Exclude code choosers in SDKs

iwahbe:
Right now, we generate code blocks for the schema that look like this:

    
    ```typescript
    // TS Code
    ```
    ```python
    # Python Code
    ```
    ...
    

That works great in the registry. For SDKs, we generate code that looks like this:

```python
  # <!--Start PulumiCodeChooser -->
  # ```python
  # # Python Code
  # ```
  # <!--End PulumiCodeChooser -->
```

This is not good, since our users see our PulumiCodeChooser directives directly.

This PR changes the bridge's behavior to only emit the PulumiCodeChooser directives when we are targeting the schema.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1466,7 +1466,9 @@ func (g *Generator) convertExamplesInner(
 
 					if err != nil {
 						// We do not write this section, ever.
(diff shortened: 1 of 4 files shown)


guineveresaenger (approved):
LGTM! Thanks for this improvement.

PR #1896, opened 2024-04-22
iwahbe/test large int as string roundtrip

iwahbe:
- **Refactor TestOverridingTFSchema for clarity.**
  This commit is a pure refactor of the tests, and does not add or remove any tests.
  

- **Support override int with string**
  This conversion was previously handled by by terraform as a best-effort approach. We
  should handle it in-house for 2 reasons:
  1. To provider errors at the pulumi level.
  2. To ensure that providers receive the same inputs as from TF, instead of relying on TF
  to convert on the wrong type.

### DIFF
--- a/pkg/tfbridge/schema.go
+++ b/pkg/tfbridge/schema.go
@@ -393,23 +393,33 @@ func (ctx *conversionContext) makeTerraformInput(
 		v = nv
 	}
 
+	if tfs == nil {
+		tfs = (&schema.Schema{}).Shim()
+	}
+
+	pathErr := func(v any, err error) (any, error) {
+		if err == nil {
+			return v, nil
+		}
+
+		return v, fmt.Errorf("%s: %w", name, err)
+	}
+
 	switch {
 	case v.IsNull():
 		return nil, nil
 	case v.IsBool():
-		if tfs != nil && tfs.Type() == shim.TypeString {
+		switch tfs.Type() {
+		case shim.TypeString:
 			if v.BoolValue() {
 				return "true", nil
 			}
 			return "false", nil
+		default:
+			return v.BoolValue(), nil
 		}
-		return v.BoolValue(), nil
 	case v.IsNumber():
-		var typ shim.ValueType
-		if tfs != nil {
-			typ = tfs.Type()
-		}
-		switch typ {
+		switch tfs.Type() {
 		case shim.TypeFloat:
 			return v.NumberValue(), nil
 		case shim.TypeString:
@@ -418,7 +428,12 @@ func (ctx *conversionContext) makeTerraformInput(
 			return int(v.NumberValue()), nil
 		}
 	case v.IsString():
-		return v.StringValue(), nil
+		switch tfs.Type() {
+		case shim.TypeInt:
+			return pathErr(strconv.ParseInt(v.StringValue(), 10, 64))
+		default:
+			return v.StringValue(), nil
+		}
 	case v.IsArray():
 		var oldArr []resource.PropertyValue
 		if old.IsArray() {
@@ -1100,6 +1115,10 @@ func MakeTerraformOutput(
 			v = list
 		}
 
+		if ps == nil {
+			ps = &SchemaInfo{}
+		}
+
 		// We use reflection instead of a type switch so that we can support mapping values whose underlying type is
 		// supported into a Pulumi value, even if they stored as a wrapper type (such as a strongly-typed enum).
 		//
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/schema_test.go
```diff
@@ -1241,92 +1241,165 @@ func TestInvalidAsset(t *testing.T) {
...
-	}
-
-	tfSchema := shimv1.NewSchemaMap(map[string]*schemav1.Schema{
-		"pulumi_override_tf_string_to_boolean":    {Type: schemav1.TypeString},
-		"pulumi_override_tf_string_to_bool":       {Type: schemav1.TypeString},
-		"pulumi_empty_tf_override":                {Type: schemav1.TypeString},
-		"pulumi_override_tf_string_to_int":        {Type: schemav1.TypeString},
-		"pulumi_override_tf_string_to_integer":    {Type: schemav1.TypeString},
-		"tf_empty_string_to_pulumi_bool_override": {Type: schemav1.TypeString},
+	// We need to assert that when both the Pulumi type (String) and the Terraoform
```
guineveresaenger, line 1251:
```suggestion
	// We need to assert that when both the Pulumi type (String) and the Terraform
```

pkg/tfbridge/schema_test.go
```diff
@@ -1241,92 +1241,165 @@ func TestInvalidAsset(t *testing.T) {
...
+			tfOutput: resource.NewNullProperty(),
+		},
+		{
+			name: "tf_int_to_pulumi_string",
+
+			tfSchema: &schemav1.Schema{Type: schemav1.TypeInt},
+			info:     &SchemaInfo{Type: "string"},
+
+			tfInput:  largeNumber,
+			tfOutput: resource.NewProperty(strconv.FormatInt(largeNumber, 10)),
```
guineveresaenger, line 1335: So when Terraform gives us a large number, we transform it into a string? But if the number is _small_, we transform to an int64, but one of a certain size?

I'm confused because both `largeNumber` as well as other tf.Output values all look like int64. I presume this is a property of shim.TypeInt?

pkg/tfbridge/schema_test.go
```diff
@@ -1241,92 +1241,165 @@ func TestInvalidAsset(t *testing.T) {
 }
 
 func TestOverridingTFSchema(t *testing.T) {
-	ctx := context.Background()
+	t.Parallel()
+
+	const (
+		k                 = "prop"
```
guineveresaenger, line 1247: I'd love for this to be "testProp" hard coded everywhere instead of abstracted here - I find that this var is just far enough away from most of the tests to be not really clear as to why it's called "k".

pkg/tfbridge/schema.go
```diff
@@ -393,23 +393,33 @@ func (ctx *conversionContext) makeTerraformInput(
 		v = nv
 	}
 
+	if tfs == nil {
+		tfs = (&schema.Schema{}).Shim()
+	}
+
+	pathErr := func(v any, err error) (any, error) {
```
guineveresaenger, line 400: I'm not sure why this is called `pathErr` - or why we need to have it declared separately. If the intention is to reuse it elsewhere, then maybe we can make it a standalone function, but a slightly more awkward name like "valueOrError" or "resourcePathOrError" might help here; as is it reads like we're declaring a kind of error on line 433.

guineveresaenger (approved):
Had a question around the integer size. Do we not need to adjust code in MakeTerraformOutput?

pkg/tfbridge/schema_test.go
iwahbe: I'm happy to call it testProp and move it closer to where it is used.

pkg/tfbridge/schema_test.go
iwahbe: > So when Terraform gives us a large number, we transform it into a string? But if the number is small, we transform to an int64, but one of a certain size?

We never do anything based on the size of the number given. If the TF type is an Int (`int64` in Go) and the Pulumi type is `"string"` we ensure that the engine sees a string and that the upstream provider sees an `int64`. The test here is to make sure we do the conversion without going through a datatype that cannot express all of `int64` (like `float64`).

> I'm confused because both `largeNumber` as well as other tf.Output values all look like int64. I presume this is a property of shim.TypeInt?

`strconv.FormatInt` is of type `func(n int64, base int) string`, so `tfOutput` holds a string `resource.PropertyValue`.

iwahbe:
> Had a question around the integer size. Do we not need to adjust code in MakeTerraformOutput?

We do: 

https://github.com/pulumi/pulumi-terraform-bridge/pull/1896/files#diff-5f10eecfa13490de36f3e3e71c19a2233316ebe835e2971f9a3e7d83e3b20250R1133-R1138

pkg/tfbridge/schema.go
iwahbe: It was called `pathErr` because it wrapped the returned error in path information. To improve clarity, I changed the name to `wrapError` and reused it where appropriate.

PR #1906, opened 2024-04-27
Fix panic in ReadDataDiff

### LINKED ISSUE #1809: Data Source Read receives a nil RawConfig resulting in panics
### What happened?

From https://github.com/pulumi/pulumi-aws/issues/3752 @mikhailshilkov root-cause establishes that TF works but bridged providers panic. 

### Example

See above.

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

guineveresaenger:
Previously, we passed a `nil` object as the terraform InstanceState.
Now, we pass a stuctured RawConfig object to avoid a panic when GetRawConfig is called in the ReadWithoutTimeout data source method.
Fixes #1809.

### DIFF
--- a/pkg/tfshim/sdk-v2/provider.go
+++ b/pkg/tfshim/sdk-v2/provider.go
@@ -160,7 +160,7 @@ func (p v2Provider) ReadDataDiff(
 	t string,
 	c shim.ResourceConfig,
 ) (shim.InstanceDiff, error) {
-	r, ok := p.tf.DataSourcesMap[t]
+	resource, ok := p.tf.DataSourcesMap[t]
 	if !ok {
 		return nil, fmt.Errorf("unknown resource %v", t)
 	}
@@ -170,12 +170,12 @@ func (p v2Provider) ReadDataDiff(
 		return nil, err
 	}
 	config := configFromShim(c)
-	rawConfig := makeResourceRawConfig(providerOpts.diffStrategy, config, r)
+	rawConfig := makeResourceRawConfig(providerOpts.diffStrategy, config, resource)
 
-	instanceState := terraform.InstanceState{
-		RawConfig: rawConfig,
+	diff, err := resource.Diff(ctx, nil, config, p.tf.Meta())
+	if diff != nil {
+		diff.RawConfig = rawConfig
 	}
-	diff, err := r.Diff(ctx, &instanceState, configFromShim(c), p.tf.Meta())
 	return diffToShim(diff), err
 }
 
--- a/pkg/tfshim/sdk-v2/provider.go
+++ b/pkg/tfshim/sdk-v2/provider.go
@@ -3,7 +3,6 @@ package sdkv2
 import (
 	"context"
 	"fmt"
-
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/logging"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
@@ -165,7 +164,18 @@ func (p v2Provider) ReadDataDiff(
 	if !ok {
 		return nil, fmt.Errorf("unknown resource %v", t)
 	}
-	diff, err := r.Diff(ctx, nil, configFromShim(c), p.tf.Meta())
+
+	providerOpts, err := getProviderOptions(p.opts)
+	if err != nil {
+		return nil, err
+	}
+	config := configFromShim(c)
+	rawConfig := makeResourceRawConfig(providerOpts.diffStrategy, config, r)
+
+	instanceState := terraform.InstanceState{
+		RawConfig: rawConfig,
+	}
+	diff, err := r.Diff(ctx, &instanceState, configFromShim(c), p.tf.Meta())
 	return diffToShim(diff), err
 }
 
--- a/pkg/tests/invoke_raw_config_test.go
+++ b/pkg/tests/invoke_raw_config_test.go
@@ -1,0 +1,98 @@
+// Copyright 2016-2023, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


pkg/tfshim/sdk-v2/provider.go
```diff
@@ -165,7 +164,18 @@ func (p v2Provider) ReadDataDiff(
...
-	diff, err := r.Diff(ctx, nil, configFromShim(c), p.tf.Meta())
+
+	providerOpts, err := getProviderOptions(p.opts)
+	if err != nil {
+		return nil, err
+	}
+	config := configFromShim(c)
+	rawConfig := makeResourceRawConfig(providerOpts.diffStrategy, config, r)
+
+	instanceState := terraform.InstanceState{
```
t0yv0, line 175: Slightly worrisome that this is not a complete InstanceState, so this is not a complete fix. How does the reference gRPC server in TF SDKv2 handle invokes? 

```
type InstanceState struct {
	// A unique ID for this resource. This is opaque to Terraform
	// and is only meant as a lookup mechanism for the providers.
	ID string `json:"id"`

	// Attributes are basic information about the resource. Any keys here
	// are accessible in variable format within Terraform configurations:
	// ${resourcetype.name.attribute}.
	Attributes map[string]string `json:"attributes"`

	// Ephemeral is used to store any state associated with this instance
	// that is necessary for the Terraform run to complete, but is not
	// persisted to a state file.
	Ephemeral EphemeralState `json:"-"`

	// Meta is a simple K/V map that is persisted to the State but otherwise
	// ignored by Terraform core. It's meant to be used for accounting by
	// external client code. The value here must only contain Go primitives
	// and collections.
	Meta map[string]interface{} `json:"meta"`

	// Tainted is used to mark a resource for recreation.
	Tainted bool `json:"tainted"`

	mu sync.Mutex
}
```

pkg/tfshim/sdk-v2/provider.go
```diff
@@ -165,7 +164,18 @@ func (p v2Provider) ReadDataDiff(
...
 		return nil, fmt.Errorf("unknown resource %v", t)
 	}
-	diff, err := r.Diff(ctx, nil, configFromShim(c), p.tf.Meta())
+
+	providerOpts, err := getProviderOptions(p.opts)
+	if err != nil {
+		return nil, err
+	}
+	config := configFromShim(c)
+	rawConfig := makeResourceRawConfig(providerOpts.diffStrategy, config, r)
```
t0yv0, line 173: Perhaps commit to ```recoverAndCoerceCtyValue(resource, config.Raw)``` for new code? It's stricter, it doesn't produce plausible but invalid garbage. I understand there's risks with handling an error from that, possibly need to start with rolling out under warning handling of the error?

t0yv0:
This looks like a decent improvement but I wonder if it could be good to reference plans to actually finish Invoke support here to be fully compatible with TF?

pkg/tfshim/sdk-v2/provider.go
iwahbe: It's worth looking at. This seems like it's strictly better, so if emulating upstream perfectly is difficult, then I think we can get the incremental improvement in and then iterate.

t0yv0: Yeah I agree!

t0yv0: https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/schema/grpc_provider.go#L1209 is how TF does it.

t0yv0: Notably

```
	diff, err := res.Diff(ctx, nil, config, s.provider.Meta())
	if err != nil {
		resp.Diagnostics = convert.AppendProtoDiag(ctx, resp.Diagnostics, err)
		return resp, nil
	}

	// Not setting RawConfig here is okay, as ResourceData.GetRawConfig()
	// will return a NullVal of the schema if there is no InstanceDiff.
	if diff != nil {
		diff.RawConfig = configVal
	}
```

t0yv0: It passes Nil to diff and then it patches RawConfig. I think it might be worth matching this up, it can only take an hour or so.

pkg/tfshim/sdk-v2/provider.go
```diff
@@ -161,11 +160,22 @@ func (p v2Provider) ReadDataDiff(
...
-	diff, err := r.Diff(ctx, nil, configFromShim(c), p.tf.Meta())
+
+	providerOpts, err := getProviderOptions(p.opts)
+	if err != nil {
+		return nil, err
+	}
+	config := configFromShim(c)
+	rawConfig := makeResourceRawConfig(providerOpts.diffStrategy, config, resource)
+
+	diff, err := resource.Diff(ctx, nil, config, p.tf.Meta())
```
t0yv0, line 175: Nice ty!

pkg/tfshim/sdk-v2/provider.go
```diff
@@ -161,11 +160,22 @@ func (p v2Provider) ReadDataDiff(
 	t string,
 	c shim.ResourceConfig,
 ) (shim.InstanceDiff, error) {
-	r, ok := p.tf.DataSourcesMap[t]
+	resource, ok := p.tf.DataSourcesMap[t]
```
t0yv0, line 163: Call this var dataSource perhaps.

pkg/tfshim/sdk-v2/provider.go
```diff
@@ -161,11 +160,22 @@ func (p v2Provider) ReadDataDiff(
...
 		return nil, fmt.Errorf("unknown resource %v", t)
 	}
-	diff, err := r.Diff(ctx, nil, configFromShim(c), p.tf.Meta())
+
+	providerOpts, err := getProviderOptions(p.opts)
+	if err != nil {
+		return nil, err
+	}
+	config := configFromShim(c)
+	rawConfig := makeResourceRawConfig(providerOpts.diffStrategy, config, resource)
```
t0yv0, line 173: I'll follow up removing this diffStrategy in a separate PR.

t0yv0 (approved):
LGTM now thanks!

pkg/tfshim/sdk-v2/provider.go
guineveresaenger: gotcha - we can then just bring the error fallback/warning logic to this method instead.

Issue #1919, opened 2024-05-01
Pulumi doesn't handle cases when state was removed with TF `RemoveResource`


andrii-glukhyi:
### What happened?

Terraform  recommend for  resources deleted out of bound do following call `RemoveResource` method that should clean up state on the fly. 
```
  // Treat HTTP 404 Not Found status as a signal to recreate resource
    // and return early
    if httpResp.StatusCode == http.StatusNotFound {
        resp.State.RemoveResource(ctx)

        return
    }
```
More [here](https://developer.hashicorp.com/terraform/plugin/framework/resources/read)

Pulumi cannot handle this case and just throws error on refresh. 

### Example

On refresh it throws error:
```
     Type                         Name                        Plan        Info
     pulumi:pulumi:Stack                aaa-test              1 error
     ├─ pulumi:providers:custom           provider
 ~   ├─ custom:index:Resource          v2                          update      [diff: +csId,size,state,storageType]
 ~   ├─ custom:index:Resource          v1                          update      [diff: +csId,size,state,storageType]
 ~   └─ custom:index:Resource2          a-resource           refresh     1 error

Diagnostics:
  custom:index:Resource (a-resource):
    error: Preview failed: Expected an Object, got: {<nil>}

  pulumi:pulumi:Stack (aaa-test):
    error: preview failed
```

Where terraform logs looks like following';
```
Note: Objects have changed outside of Terraform

Terraform detected the following changes made outside of Terraform since the last "terraform apply" which may have affected
this plan:

  # custom.resource has been deleted
  - resource "custom" "test" {
      - id                         = "12345" -> null
        name                       = "123455"
        tags                       = {}
        # (26 unchanged attributes hidden)
    }


Unless you have made equivalent changes to your configuration, or ignored the relevant attributes using ignore_changes, the
following plan may include actions to undo or respond to these changes.

─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the
following symbols:
  + create

Terraform will perform the following actions:

  # custom.test will be created
  + resource "custom" "test" {
      + id.                           = (known after apply)
      + actions                   = (known after apply)
```

### Output of `pulumi about`

```
CLI
Go Version   go1.22.0
Go Compiler  gc

Plugins
NAME    VERSION
custom     1.4.1
nodejs  unknown

Host
OS       darwin
Version  14.4.1
Arch     x86_64

[...]

PR #1920, opened 2024-05-01
Instrument DecodePropertyMap failures to assist root-causing refresh issues in Plugin Framework

t0yv0:
Investigating https://github.com/pulumi/pulumi-terraform-bridge/issues/1919

DecodePropertyMap called from Read fails because it decoded something to a null but a PropertyMap is expected. It appears it should never happen because Read will not call DecodePropertyMap with a tftypes.Value=nil but instead short-circuit the "resource not found path".

`resp.State.RemoveResource(ctx)` likewise seems to work correctly.

Testing this further, still unable to reproduce the scenario in test. Instead, opting for adding some more debugging instrumentation around the line that emits the error, and doing some conservative code tightening around this code path.

### DIFF
--- a/pf/tfbridge/provider_read.go
+++ b/pf/tfbridge/provider_read.go
@@ -151,7 +151,7 @@ func (p *provider) readResource(
 		return plugin.ReadResult{}, nil
 	}
 
-	readStateMap, err := readState.ToPropertyMap(rh)
+	readStateMap, err := readState.ToPropertyMap(ctx, rh)
 	if err != nil {
 		return plugin.ReadResult{}, fmt.Errorf("converting to property map: %w", err)
 	}
@@ -225,7 +225,7 @@ func (p *provider) importResource(
 		return plugin.ReadResult{}, err
 	}
 
-	readStateMap, err := readState.ToPropertyMap(rh)
+	readStateMap, err := readState.ToPropertyMap(ctx, rh)
 	if err != nil {
 		return plugin.ReadResult{}, err
 	}
--- a/pf/tfbridge/provider_read.go
+++ b/pf/tfbridge/provider_read.go
@@ -141,19 +141,14 @@ func (p *provider) readResource(
 		return plugin.ReadResult{}, nil
 	}
 
-	// TF interpretes a null new state as an indication that the resource does not
-	// exist in the cloud provider.
-	newStateNull, err := resp.NewState.IsNull()
-	if err != nil {
-		return plugin.ReadResult{}, fmt.Errorf("checking null state: %w", err)
-	}
-	if newStateNull {
-		return plugin.ReadResult{}, nil
-	}
-
 	readState, err := parseResourceStateFromTF(ctx, rh, resp.NewState, resp.Private)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


pf/tfbridge/provider_read.go
```diff
@@ -141,22 +141,17 @@ func (p *provider) readResource(
 		return plugin.ReadResult{}, nil
 	}
 
-	// TF interpretes a null new state as an indication that the resource does not
-	// exist in the cloud provider.
-	newStateNull, err := resp.NewState.IsNull()
```
t0yv0, line 146: Previously IsNull check was done on the DynamicValue domain, and after the change it's done on the tftypes.Value domain. There's a very slim chance the issue was there, namely that there were values that return IsNull() = false on DynamicValue domain that then translate to Null values on tftypes.Value domain.

pf/tfbridge/provider_read.go
```diff
@@ -141,22 +141,17 @@ func (p *provider) readResource(
...
-		return plugin.ReadResult{}, nil
-	}
-
 	readState, err := parseResourceStateFromTF(ctx, rh, resp.NewState, resp.Private)
 	if err != nil {
 		return plugin.ReadResult{}, fmt.Errorf("parsing resource state: %w", err)
 	}
 
-	readStateMap, err := readState.ToPropertyMap(rh)
+	// TF interprets a null new state as an indication that the resource does not exist in the cloud provider.
```
t0yv0, line 149: I still believe this is how TF works.

pf/tfbridge/resource_state.go
```diff
@@ -115,15 +115,30 @@ func parseResourceStateFromTF(
...
+	resourceSchemaVersion int64,
+	state *tfprotov6.DynamicValue,
+	private []byte,
+) (*upgradedResourceState, error) {
+	rs := &resourceState{
+		TFSchemaVersion: resourceSchemaVersion,
 		Private:         private,
-	}}, nil
+	}
+	if state == nil {
```
t0yv0, line 132: Tightening to also handle physical nulls, probably unnecessary but just in case.

pkg/convert/convert.go
```diff
@@ -102,13 +103,18 @@ func EncodePropertyMap(enc Encoder, pmap resource.PropertyMap) (tftypes.Value, e
...
 
-func DecodePropertyMap(dec Decoder, v tftypes.Value) (resource.PropertyMap, error) {
+func DecodePropertyMap(ctx context.Context, dec Decoder, v tftypes.Value) (resource.PropertyMap, error) {
 	pv, err := dec.toPropertyValue(v)
 	if err != nil {
 		return nil, err
 	}
 	if !pv.IsObject() {
-		return nil, fmt.Errorf("Expected an Object, got: %v", pv.String())
+		details := fmt.Sprintf(`DecodePropertyMap expected the decoder to return an Object
```
t0yv0, line 112: Emitting a little more information in the debug log. Not returning it as an err in case it is printed on stdout since the values may be sensitive.

VenelinMartinov (approved):
Seems sensible, sounds like it should handle the reported error.

PR #1948, opened 2024-05-09
Remove spurious ConflictsWith warnings on import

t0yv0:
Toward https://github.com/pulumi/pulumi-terraform-bridge/issues/1225 - this fixes the special case of ConflictsWith warnings. This fixes spurious warnings on `pulumi import`, popular bugs such as:

- https://github.com/pulumi/pulumi-aws/issues/2318 
- https://github.com/pulumi/pulumi-aws/issues/3670
- https://github.com/pulumi/pulumi-gitlab/issues/293
- https://github.com/pulumi/pulumi-gcp/issues/844
- https://github.com/pulumi/pulumi-linode/issues/373

TF does not guarantee Read results to be compatible with calling Check on, in particular Read can return results that run afoul of ConflictsWith constraint. This change compensates by arbitrarily dropping data from the Read result until it passes ConflictsWith checks.

This affects `pulumi refresh` as well as I think it should although I have not seen "in the wild" cases where refresh is affected, since it typically will not copy these properties to the input bag unless they're present in old inputs, which are usually correct wrt to ConflictsWith.

### DIFF
--- a/pkg/tfbridge/deconflict_test.go
+++ b/pkg/tfbridge/deconflict_test.go
@@ -16,7 +16,6 @@ package tfbridge
 
 import (
 	"context"
-	"fmt"
 	"testing"
 
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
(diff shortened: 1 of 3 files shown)


pkg/tfbridge/deconflict_test.go
```diff
@@ -0,0 +1,82 @@
...
+			ConflictsWith: []string{"availability_zone_id"},
+		}).Shim(),
+		"availability_zone_id": (&schema.Schema{
+			Type:          shim.TypeString,
+			Optional:      true,
+			Computed:      true,
+			ForceNew:      true,
+			ConflictsWith: []string{"availability_zone"},
+		}).Shim(),
+	}
```
iwahbe, line 48: Can you add a test for different ConflictsWith topologies?

At the very least:

A conflicts with B (where B does not conflict with A)

A conflicts with B and C, B conflicts with A and C, C conflicts with A and B

iwahbe (approved):
I'd like some more testing on different conflicts with topologies, but otherwise I think this looks good.

pkg/tfbridge/deconflict_test.go
t0yv0: Yeah good point, adding a few.

PR #1958, opened 2024-05-10
Fix spurious ForcesProviderReplace activation

t0yv0:
The ForceProviderReplace setting introduces cascading replaces when a certain Provider-level configuration setting changes. For example, changing the region in the AWS Provider is able to recreate all the affected resources in the new desired region with this setting. There was however a practical problem fixed with this PR that caused accidental re-creation of resources in the same region.

Specifically, some of the resources states provisioned by older versions of Pulumi CLI may not have oldInputs sufficiently populated in the call to DetailedDiff. This caused ForceProviderReplace logic to assume that the region is changing from "" to "us-east-1" where it did not, in fact, change. To compensate for this, the logic no longer initiates cascading replace when the oldInputs map does not contain the value.

See also:

- pulumi/pulumi-aws#3826
- pulumi/pulumi-aws#3674

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -675,6 +675,12 @@ func (p *configDiffer) DiffConfig(
 		keyPath, err := resource.ParsePropertyPath(key)
 		contract.AssertNoErrorf(err, "Unexpected failed parse of PropertyPath %q", key)
 		if p.forcesProviderReplace(keyPath) {
+			// NOTE: for states provisioned on the older versions of Pulumi CLI oldInputs will have no entry
+			// for the changing property. Causing cascading replacements in this case is undesirable, since
+			// it is not a real change. Err on the side of not replacing (pulumi/pulumi-aws#3826).
+			if _, ok := keyPath.Get(resource.NewObjectProperty(oldInputs)); !ok {
+				continue
+			}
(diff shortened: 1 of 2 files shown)


t0yv0:
This seems the most direct approach to the problem but feels a little odd, possibly does not generalize very well and will need to be revisited. If we know of other use cases or approaches let's consider here.

iwahbe (approved):
I'm fine with this change given the current design.

The need here does make me question the original design. Instead of setting a flag (`bool`), we should have allowed a diff function 
```go
ForceProviderReplace func(ctx context.Context, olds, news resource.PropertyValue) (bool, error)
```

VenelinMartinov (commented):
This looks like something we'd feature flag in the future. I doubt all provider parameters are not really changes when they move from unspecified to specified.

Should we make it opt-in (so needs a change now) or opt out (safe to merge like this, add a feature flag later)?

EDIT: never mind, this is already flagged, we can add another in the future if we find examples of the reverse behaviour. Happy with this.

VenelinMartinov (approved):

t0yv0:
Yes a callback would be the way to go. This currently has only 1 use case (AWS region) unfortunately. Let's leave as-is but if revisiting go with the callback and deprecate the flag.

PR #1973, opened 2024-05-14
Dynamic Bridged Providers: Schema Generation

iwahbe:
Initial PR for dynamically bridged providers.

Currently, support is limited to Parameterize and GetSchema calls. We are getting very close to where it would be helpful for multiple people to work on the project, so I'd like to get ready to merge soon.

Fixes https://github.com/pulumi/home/issues/3409
Fixes https://github.com/pulumi/home/issues/3410


dynamic/internal/shim/loader.go
```diff
@@ -0,0 +1,263 @@
...
+// providers.Interface against it.
+func runProvider(ctx context.Context, meta *providercache.CachedProvider) (Provider, error) {
+	execFile, err := meta.ExecutableFile()
+	if err != nil {
+		return nil, err
+	}
+
+	config := &plugin.ClientConfig{
+		HandshakeConfig:  tfplugin.Handshake,
+		Logger:           logging.NewProviderLogger(""),
```
VenelinMartinov, line 221: Is this correct? Is logging handled in syncStdout/err below?

VenelinMartinov (commented):
There's quite a bit of code here, it'd be helpful if you added a short description of which bits are eventful.

From previous discussions IIRC, the stuff under internal/shim is directly translating the GRPC calls, right?

I'd appreciate some comments on the new pf code.

dynamic/internal/shim/loader.go
```diff
@@ -0,0 +1,263 @@
...
+		return dir, nil
+	}
+
+	pulumiHome := os.Getenv("PULUMI_HOME")
+	if pulumiHome == "" {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return "", err
+		}
+		pulumiHome = filepath.Join(home, ".pulumi")
```
t0yv0, line 121: pulumi/pulumi may have some linkable helper code to resolve pulumi home.

dynamic/internal/shim/loader.go
```diff
@@ -0,0 +1,263 @@
...
+) (Provider, error) {
+	cacheDir, err := getPluginCache()
+	if err != nil {
+		return nil, err
+	}
+
+	systemCache := providercache.NewDir(cacheDir)
+
+	// Look through existing packages to see if the package we want is already downloaded.
+	for k, packages := range systemCache.AllAvailablePackages() {
```
t0yv0, line 157: O(N)?

dynamic/internal/shim/loader.go
```diff
@@ -0,0 +1,263 @@
...
+
+	// store the client so that the plugin can kill the child process
+	protoVer := client.NegotiatedVersion()
+	switch protoVer {
+	case 5:
+		p := raw.(*tfplugin.GRPCProvider)
+		p.PluginClient = client
+		p.Addr = meta.Provider
+
+		v6, err := tf5to6server.UpgradeServer(ctx, func() tfprotov5.ProviderServer {
```
t0yv0, line 250: slog a message here we're using a shim to approximate v5 to v6 protocol perhaps.

dynamic/internal/shim/loader.go
```diff
@@ -0,0 +1,263 @@
...
+			return v5shim.New(tfplugin5.NewProviderClient(rpcClient.(*plugin.GRPCClient).Conn))
+		})
+		if err != nil {
+			return nil, err
+		}
+		return provider{v6, meta.Provider.Type, meta.Version.String(), rpcClient.Close}, nil
+	case 6:
+		p := tfplugin6.NewProviderClient(rpcClient.(*plugin.GRPCClient).Conn)
```
t0yv0, line 258: Do all the downcasts always succeed or worth adding nil checks?

dynamic/internal/shim/protov5/translate/tfplugin5.go
```diff
@@ -0,0 +1,674 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package translate
+
+import (
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"github.com/opentofu/opentofu/internal/tfplugin5"
```
t0yv0, line 20: Ah two models for the same data, how sad. 

Are you sure you need these translations?

E.g.  tfprotov5.GetMetadataRequest -> tfplugin5.GetMetadata_Request

These are all protobuf-level models I think. So you could turn around through proto:

tfprotov5.GetMetadataRequest -> proto-bytes -> tfplugin5.GetMetadata_Request

dynamic/main.go
```diff
@@ -0,0 +1,126 @@
...
+		XGetSchema: func(ctx context.Context, req plugin.GetSchemaRequest) ([]byte, error) {
+
+			info := defaultInfo
+			if tfServer != nil {
+				info = providerInfo(ctx, tfServer)
+			}
+
+			packageSchema, err := tfgen.GenerateSchemaWithOptions(tfgen.GenerateSchemaOptions{
+				ProviderInfo: info,
+				DiagnosticsSink: diag.DefaultSink(io.Discard, os.Stderr, diag.FormatOptions{
```
t0yv0, line 62: Can you actually log to stderr from providers safely?

Should stdout be redirected somewhere other than discard?

dynamic/main.go
```diff
@@ -0,0 +1,126 @@
...
+		DisplayName:  "Any Terraform Provider",
+		P:            proto.Empty(),
+		Name:         name,
+		Version:      version,
+		Description:  "A Pulumi provider for dynamically bridging terraform provdiers.",
+		MetadataInfo: &tfbridge.MetadataInfo{Path: "", Data: tfbridge.ProviderMetadata(nil)},
+	}
+
+	metadata := pfbridge.ProviderMetadata{
+		XGetSchema: func(ctx context.Context, req plugin.GetSchemaRequest) ([]byte, error) {
```
t0yv0, line 53: Is this called only once, someone handling caching?

dynamic/provider_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+func grpcTestServer(ctx context.Context, t *testing.T) pulumirpc.ResourceProviderServer {
+	defaultInfo, metadata, close := initialSetup()
+	t.Cleanup(func() { assert.NoError(t, close()) })
+	s, err := pfbridge.NewProviderServer(ctx, nil, defaultInfo, metadata)
+	require.NoError(t, err)
+	return s
+}
+
+func TestSchemaGeneration(t *testing.T) {
```
t0yv0, line 30: I would be curious to compare the schema generated by this method vs the one generated by static TFGen for something large like AWS, not to check it in but to grasp the discrepancies. I suspected historically that there is some information we're not able to recover from the gRPC protocol possibly including

- whether a property is optional or not
- whether a proeprty is MaxItems=1 or not

This would be great to understand and document, as obviously a gRPC schema loading approach is preferable, if the limitations aren't there or can be worked around we could consider further leaning into it.

pkg/tfgen/generate.go
```diff
@@ -856,6 +864,7 @@ func NewGenerator(opts GeneratorOptions) (*Generator, error) {
 		skipExamples:     opts.SkipExamples,
 		coverageTracker:  opts.CoverageTracker,
 		editRules:        getEditRules(info.DocRules),
+		noDocsRepo:       opts.XInMemoryDocs,
```
t0yv0, line 867: why not name it xInMemoryDocs for symmetry?

pf/tfbridge/provider_get_mapping.go
```diff
@@ -54,7 +53,7 @@ func (p *provider) GetMappingWithContext(ctx context.Context, key, provider stri
 
 func (p *provider) marshalProviderInfo(ctx context.Context) *tfbridge.MarshallableProviderInfo {
 	var providerInfoCopy tfbridge.ProviderInfo = p.info
-	providerInfoCopy.P = schemashim.ShimSchemaOnlyProvider(ctx, p.tfProvider)
+	providerInfoCopy.P = p.info.P
```
t0yv0, line 56: Is this safe? p.info.P always in place?

pf/tfbridge/provider.go
```diff
@@ -95,45 +98,70 @@ func ShimProviderWithContext(ctx context.Context, p pfprovider.Provider) shim.Pr
...
-		return nil, fmt.Errorf("Fatal failure gathering resource metadata: %w", err)
-	}
+	var (
+		server6            tfprotov6.ProviderServer
+		resources          pfutils.Resources
+		datasources        pfutils.DataSources
+		providerConfigType *tftypes.Object
+	)
+	switch p := info.P.(type) {
+	case *schemashim.SchemaOnlyProvider:
```
t0yv0, line 109: There is some code smell here with the case analysis, perhaps the capability this code is seeking is missing an interface.

pf/proto/internal/gather/attr.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+
+	"github.com/pulumi/pulumi-terraform-bridge/pf/internal/pfutils"
+)
+
+type _attr struct{ s *tfprotov6.SchemaAttribute }
+
+var _ = pfutils.Attr(_attr{})
```
t0yv0, line 26: I'm wondering if there is any way to avoid layering these adapters on top of pfutils adapters.

pfutils.Attr is a workaround as it doc comment states

There is some additional highly suspect reflective code in:

https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pf/internal/pfutils/reflector.go#L66

A "clean" adapter set would go from the Protocol V6 Schema encoding straight into shim.Schema. 

Is that not practical?

pf/proto/internal/gather/type.go
```diff
@@ -0,0 +1,71 @@
...
+package gather
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+type _type struct{ s tftypes.Type }
```
t0yv0, line 24: This looks rather suspect why is it needed.

t0yv0:
This could be a few PRs for easier review maybe, we can check the plugin cache and resolver first perhaps? 

I'm thinking if pfutils could be killed with fire instead of reused here, perhaps pf/provider over-assumes schema related assumptions from here:

```
	resources     pfutils.Resources
	datasources   pfutils.DataSources
```

Looking quickly what it really seems to want is these things:

```
       rh.schema.ResourceSchemaVersion(),
       rh.schema.Type().TerraformType(ctx) // tftypes.Object
```

t0yv0:
Looks like this need can be satisfied by this instead:

https://github.com/hashicorp/terraform-plugin-go/blob/main/tfprotov6/schema.go#L105

```
func (s *Schema) ValueType() tftypes.Type {
```

So stepping back, there are two sets of requirements:

- schema generation requires a subset of the interfaces defined in shim/Provider - which we could split to make that formal, yay for smaller interfaces
 
- pf/tfbridge runtime could use a bit of cleanup to require less, it really just needs tfprotov6.ProviderServer-like handle and schema-wise it needs tftypes.Type for conversion and looks like it needs some bridge schema info as well

If this resonates.. cleaner=more robust usually.

dynamic/internal/shim/loader.go
iwahbe: We don't need to set a prefix, so it looks correct to me. We might want to iterate on how logging works here later, but this seems to work for now.

dynamic/internal/shim/loader.go
iwahbe: It does:

https://github.com/pulumi/pulumi/blob/7c0489fafcdc5c393f06ced3e0e8b2da413bcbec/sdk/go/common/workspace/paths.go#L264-L266

Thanks for suggesting it.

dynamic/internal/shim/loader.go
iwahbe: The downcasts always succeed as far as I can tell.

dynamic/internal/shim/protov5/translate/tfplugin5.go
iwahbe: They both represent the same thing, but they aren't both protobuf models. tfprotov5 is just slightly higher level then tfplugin5.

For example, `tfplugin5.ServerCapabilities` vs `*tfprotov5.ServerCapabilities`:

```go
// ServerCapabilities allows providers to communicate extra information
// regarding supported protocol features. This is used to indicate
// availability of certain forward-compatible changes which may be optional
// in a major protocol version, but cannot be tested for directly.
type ServerCapabilities struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// The plan_destroy capability signals that a provider expects a call
	// to PlanResourceChange when a resource is going to be destroyed.
	PlanDestroy bool `protobuf:"varint,1,opt,name=plan_destroy,json=planDestroy,proto3" json:"plan_destroy,omitempty"`
	// The get_provider_schema_optional capability indicates that this
	// provider does not require calling GetProviderSchema to operate
	// normally, and the caller can used a cached copy of the provider's
	// schema.
	GetProviderSchemaOptional bool `protobuf:"varint,2,opt,name=get_provider_schema_optional,json=getProviderSchemaOptional,proto3" json:"get_provider_schema_optional,omitempty"`
	// The move_resource_state capability signals that a provider supports the
	// MoveResourceState RPC.
	MoveResourceState bool `protobuf:"varint,3,opt,name=move_resource_state,json=moveResourceState,proto3" json:"move_resource_state,omitempty"`
}
```

```go
// ServerCapabilities allows providers to communicate optionally supported
// protocol features, such as forward-compatible Terraform behavior changes.
//
// This information is used in GetProviderSchemaResponse as capabilities are
// static features which must be known upfront in the provider server.
type ServerCapabilities struct {
	// GetProviderSchemaOptional signals that this provider does not require
	// having the GetProviderSchema RPC called first to operate normally. This
	// means the caller can use a cached copy of the provider's schema instead.
	GetProviderSchemaOptional bool

	// MoveResourceState signals that a provider supports the MoveResourceState
	// RPC.
	MoveResourceState bool

	// PlanDestroy signals that a provider expects a call to
	// PlanResourceChange when a resource is going to be destroyed. This is
	// opt-in to prevent unexpected errors or panics since the
	// ProposedNewState in PlanResourceChangeRequest will be a null value.
	PlanDestroy bool
}
```

I thought of trying to automate the mapping with reflect, but there are a couple of places where you need genuine type conversions:

https://github.com/pulumi/pulumi-terraform-bridge/blob/114bc771e674f48f61b071bffec0456cc7ff214f/dynamic/internal/shim/protov5/translate/tfplugin5.go#L190-L202

dynamic/main.go
iwahbe: I couldn't when I generated the schema promptly, but I can now that this is correctly part of Parameterize.

dynamic/provider_test.go
iwahbe: The v6 protocol (and I think the v5 protocol) has support for both optional properties and MaxItems properties. It's definitely worth doing a direct comparison to see the difference.

iwahbe: I opened an issue to track this.

pf/proto/internal/gather/type.go
iwahbe: Why? I needed a type that would bridge `tftypes.Type` back to the higher level `attr.Type`. `_type` projects a `tftype.Type` into the richer typespace of `attr.Type`.

iwahbe: It's needed because `pfutils.Attr` thinks in `attr.Type`, instead of the lower level `tftypes.Type`.

dynamic/provider_test.go
t0yv0: Thanks a ton! It might be a much cleaner and more robust way to do things than what's in place now.

pf/proto/internal/gather/type.go
t0yv0: That's exactly what's suspect to me, I was surprised to see pfutils used at all here, and looks like that's what's pulling in attr.Type.

t0yv0:
There's no blockers here at first glance, do you want me to try to work it out again and approve or is there a chance to split it out into smaller PRs for easier review? Appreciated.

dynamic/main.go
iwahbe: I'm not aware of any part of the pulumi provider lifecycle where it is called more then once. The underlying TF provider is cached, but the getSchema call is not.

I'm not sure I want to cache the provider schema multiple times (even in different forms), since it's a big object to keep around.

pkg/tfgen/generate.go
iwahbe: Because we also set `noDocsRepo` when we fail to find a repo during normal docs generation.

pf/tfbridge/provider_get_mapping.go
iwahbe: `p.info.P` is always in place (it's where we got `p.tfProvider` from). I'm assuming that the code looked like this since the initial bridge PF implementation passed the provider seperatily and did leave `p.info.P` as `nil`.

pf/tfbridge/provider.go
iwahbe: We could convert to an interface.

```go
type PFProviderServer interface {
	Server()      tfprotov6.ProviderServer
	Resources()   pfutils.Resources
	DataSources() pfutils.DataSources
	Config()      *tftypes.Object
}
```

I think that would help clean up some existing code that does the same thing. Good idea.

pf/proto/internal/gather/attr.go
iwahbe: > A "clean" adapter set would go from the Protocol V6 Schema encoding straight into shim.Schema.
>
>Is that not practical?

I think it's practical, but it would force this PR to reach much farther into `/pf` then I'd like. I'd rather merge with the current `pfutils` setup and then refactor afterwards.

iwahbe:
I'll take a look at refactoring into more reviewable chunks today.

pf/tfbridge/provider.go
t0yv0: Bonus points if pfutils.* is not there or is simplified. It might be profitable to revisit and retroactively review the design of that.

dynamic/main.go
t0yv0: It's great then.

dynamic/README.md
```diff
@@ -0,0 +1,97 @@
+<!-- -*- fill-column: 110 -*- -->
+# Dynamic Bridged Provider
```
thomas11, line 2: Awesome README, thank you for documenting this.

thomas11 (commented):
The Autogold tests are great but unit tests would be good, too.

pf/proto/internal/gather/type.go
iwahbe: Refactoring out `pfutils` allowed me to remove this.

pf/proto/internal/gather/attr.go
iwahbe: `pfutils` has been removed here.

pf/proto/block.go
```diff
@@ -0,0 +1,134 @@
...
+func (m blockSchema) DefaultFunc() shim.SchemaDefaultFunc { return nil }
+func (m blockSchema) DefaultValue() (interface{}, error)  { return nil, nil }
+func (m blockSchema) Description() string                 { return m.block.Block.Description }
+func (m blockSchema) Computed() bool                      { return false }
+func (m blockSchema) ForceNew() bool                      { return false }
+func (m blockSchema) StateFunc() shim.SchemaStateFunc     { return nil }
+func (m blockSchema) Elem() interface{}                   { return blockResource{block: m.block.Block} }
+func (m blockSchema) MaxItems() int                       { return int(m.block.MaxItems) }
+func (m blockSchema) MinItems() int                       { return int(m.block.MinItems) }
+func (m blockSchema) ConflictsWith() []string             { return nil }
```
VenelinMartinov, line 125: I guess these are the bits which are inaccessible through GRPC? Are default inaccessible here too?

pf/proto/block.go
```diff
@@ -0,0 +1,134 @@
...
+type blockSchema struct{ block tfprotov6.SchemaNestedBlock }
+
+func (m blockSchema) Type() shim.ValueType                { return shim.TypeMap }
+func (m blockSchema) Optional() bool                      { return false }
+func (m blockSchema) Required() bool                      { return false }
+func (m blockSchema) Default() interface{}                { return nil }
+func (m blockSchema) DefaultFunc() shim.SchemaDefaultFunc { return nil }
+func (m blockSchema) DefaultValue() (interface{}, error)  { return nil, nil }
+func (m blockSchema) Description() string                 { return m.block.Block.Description }
+func (m blockSchema) Computed() bool                      { return false }
```
VenelinMartinov, line 119: Is this meant to always return false?

pf/proto/protov6.go
```diff
@@ -0,0 +1,151 @@
...
+
+func filter[T any](m []T, keep func(T) bool) []T {
+	for i := len(m) - 1; i >= 0; i-- {
+		if keep(m[i]) {
+			continue
+		}
+
+		// If i is not the last element in the slice, put the last element in the
+		// slice into i.
+		if i < len(m)-1 {
```
VenelinMartinov, line 142: If this happens would we check the last element in the new position with `keep`?

I'd appreciate a test on this, not obviously correct.

pf/proto/attribute.go
```diff
@@ -0,0 +1,117 @@
...
+
+type attribute struct{ attr tfprotov6.SchemaAttribute }
+
+// Simple schema options
+
+func (a attribute) Optional() bool      { return a.attr.Optional }
+func (a attribute) Required() bool      { return a.attr.Required }
+func (a attribute) Description() string { return a.attr.Description }
+func (a attribute) Computed() bool      { return a.attr.Computed }
+func (a attribute) ForceNew() bool      { return false }
```
t0yv0, line 36: Deserve a comment why it's ok to return false?

pf/proto/attribute.go
```diff
@@ -0,0 +1,117 @@
...
+
+		return object{obj: tfprotov6.SchemaObject{
+			Attributes: attrs,
+			Nesting:    tfprotov6.SchemaObjectNestingModeMap,
+		}}
+	}
+
+	switch t := t.(type) {
+	case tftypes.Set:
+		return attribute{tfprotov6.SchemaAttribute{Type: t.ElementType}}
```
t0yv0, line 92: This case reads weird but may behave OK semi-accidentally. Suppose you have an attribute that corresponds to a Map<string,number>. So the element type is number. However. The element type `number` is not really an attribute in TF parlance. It cannot have Optional/Computed/Required and so on, I think. This could be made more clear by not having this represented as `attribute` struct. Or at least make a comment that this is a pseudo-attribute we are creating here.

pf/proto/block.go
```diff
@@ -0,0 +1,134 @@
...
+		return a.Name != key
+	})
+	m.block.BlockTypes = filter(m.block.BlockTypes, func(b *tfprotov6.SchemaNestedBlock) bool {
+		return b.TypeName != key
+	})
+}
+
+func (m blockMap) Validate() error { return nil }
+
+type blockSchema struct{ block tfprotov6.SchemaNestedBlock }
```
t0yv0, line 110: Not convinced that SchemaNestedBlockNestingMode cases are all handled, will appreciate an elaboration or a pointer to some tests. In particular SchemaNestedBlockNestingMode=single is weird as it expresses object types directly. It is expressible and does appear for Plugin Framework based resources.

pf/proto/block.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (m blockSchema) Type() shim.ValueType                { return shim.TypeMap }
```
t0yv0, line 112: That doesn't feel right either. List-nested block will need to have TypeList.

pf/proto/block.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	for _, b := range m.block.BlockTypes {
+		if !each(b.TypeName, blockSchema{*b}) {
+			return
+		}
+	}
+}
+
+func (m blockMap) Set(key string, value shim.Schema) {
```
t0yv0, line 84: Curious if we can panic here and be OK with an immutable model. Does it get used?

pf/proto/block.go
t0yv0: This one is true, only attributes can be Computed, the notion does not exist for blocks, and I think we represent this as false for the bridge.

VenelinMartinov: Yeah, that makes a lot of sense.

pf/proto/attribute.go
```diff
@@ -0,0 +1,117 @@
...
+	default:
+		panic(fmt.Sprintf("UNKNOWN TYPE of %#v", t)) // TODO: Remove for release
+		// return shim.TypeInvalid
+	}
+}
+
+func (a attribute) Elem() interface{} {
+	t := a.attr.ValueType()
+	switch {
+	case a.attr.NestedType != nil:
```
t0yv0, line 70:
```	// NestedType indicates that this is a NestedBlock-style object masquerading
	// as an attribute. This field conflicts with Type.
	NestedType *SchemaObject
```

Oof. Would be interesting to see a test for this.

pf/proto/block.go
t0yv0: Yeah no luck with detecting these over gRPC. We can just detect validation failures.

pf/proto/object.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+var (
+	_ = shim.Resource(object{})
+	_ = shim.SchemaMap(attrMap{})
+)
+
+type object struct {
+	pseudoResource
+	obj tfprotov6.SchemaObject
```
t0yv0, line 31: SchemaObjectNestingMode handled? Looks pretty suspect.

t0yv0:
This pretty clean! Excellent. I'm good with merging and then trying to address some refinements I've pointed out as follow-up.

dynamic/info.go
```diff
@@ -0,0 +1,103 @@
...
+
+func providerInfo(ctx context.Context, p shim.Provider) tfbridge.ProviderInfo {
+	prov := tfbridge.ProviderInfo{
+		P:           proto.New(ctx, p),
+		Name:        p.Name(),
+		Version:     p.Version(),
+		Description: "A Pulumi provider dynamically bridged from " + p.Name() + ".",
+
+		// To avoid bogging down schema generation speed, we skip all examples.
+		SkipExamples: func(tfbridge.SkipExamplesArgs) bool { return true },
```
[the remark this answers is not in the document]
iwahbe: The schema is used for generating SDKs. There is no `make tfgen` step for dynamic providers.

pf/proto/block.go
iwahbe: Yep.

dynamic/main.go
```diff
@@ -0,0 +1,125 @@
...
+
+			info := defaultInfo
+			if tfServer != nil {
+				info = providerInfo(ctx, tfServer)
+			}
+
+			packageSchema, err := tfgen.GenerateSchemaWithOptions(tfgen.GenerateSchemaOptions{
+				ProviderInfo: info,
+				DiagnosticsSink: diag.DefaultSink(os.Stdout, os.Stderr, diag.FormatOptions{
```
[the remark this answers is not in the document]
iwahbe: > Should it go to stdout/stderr? 

I'm not sure where else to put it. I believe that stdout/stderr are simply passed through, so this seams a reasonable default.

> Not going to interfere with port printing? Thanks.

No, since `XGetSchema` is only called to respond to the `GetSchema` gRPC call which means that port mapping has already taken place.

pf/proto/attribute.go
iwahbe: I'm happy to add one. It's because this information isn't exposed by terraform at the gRPC level.

pf/proto/attribute.go
iwahbe: That is a good idea. I'll do that.

dynamic/main.go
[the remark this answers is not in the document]
t0yv0: Its' reasonable I'm just wondering if stdout will look like this:

```
random tfgen output
12345
```

The wrapping Pulumi CLI needs to read the port number .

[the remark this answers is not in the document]
t0yv0: You could also directly send it to the host logging system I guess. Without indirecting through stdout stderr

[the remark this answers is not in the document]
t0yv0: > No, since XGetSchema is only called to respond to the GetSchema gRPC call which means that port mapping has already taken place.

I see. So perhaps OK then.

pf/proto/attribute.go
t0yv0: Trying to remember if we use it for docs-only or use it to drive behavior?

iwahbe: This is the only place where it impacts behavior:

https://github.com/pulumi/pulumi-terraform-bridge/blob/6d24ffc501f1b9757b85548b92f5e837e2e30994/pkg/tfbridge/provider.go#L1053-L1063

I think that's just used for display logic.

pf/proto/protov6.go
iwahbe: We wouldn't need to, since it would have already been checked. 

Doesn't matter though, the standard library has `slices.DeleteFunc`, which is what we should have used. I'm removing `filter` in place of `slices.DeleteFunc`.

pf/proto/attribute.go
t0yv0: I think stables something deprecate-able. So all good then.

pf/proto/block.go
iwahbe: They are not handled. I didn't look closely enough at the nesting. I'll fix it.

pf/proto/object.go
iwahbe: It is not handled in [506589f](https://github.com/pulumi/pulumi-terraform-bridge/pull/1973/commits/506589f50ae024783592c84b70987a2915c27071).

iwahbe: This should be fixed in [deed5e7](https://github.com/pulumi/pulumi-terraform-bridge/pull/1973/commits/deed5e7ce50778d2a540ff5ef833f357f38c915a).

pf/proto/block.go
iwahbe: I'd rather not panic. If we don't need these methods, let's remove them from `shim.SchemaMap`. I'll explore that in a seperate PR.

pf/proto/attribute.go
```diff
@@ -0,0 +1,103 @@
...
+		//
+		// [tfprotov6.SchemaObjectNestingModeMap]: `a` represents a Map<Object>, so it's `.Type()`
+		// should be [shim.TypeMap] and it's `.Elem()` should be a [shim.Schema] whose `.Type()` is
+		// [shim.TypeMap] and whose `.Elem()` is a [shim.Resource].
+
+		switch obj.Nesting {
+		case tfprotov6.SchemaObjectNestingModeMap:
+			// We are careful to only assign to variables on [attribute.Elem]'s stack. We *do not*
+			// mutate the caller.
+			obj.Nesting = tfprotov6.SchemaObjectNestingModeSingle
```
t0yv0, line 72: This looks tricky though possibly correct. Wouldn't mind longer boring code here.

t0yv0 (approved):
Thanks for addressing the feedback, as far as I can tell this should be good now. I think we'll still need to add tests around proto->shim.Schema before shipping to cover these strange attr/block combinations but can come as a follow up PR.

iwahbe:
> Thanks for addressing the feedback, as far as I can tell this should be good now. I think we'll still need to add tests around proto->shim.Schema before shipping to cover these strange attr/block combinations but can come as a follow up PR.

100%

Issue #1904, opened 2024-04-25
Refresh reorders set elements causing spurious diffs


t0yv0:
### What happened?

When servicing `pulumi refresh` the provider only gets to influence the result in a Read call, but the diff will be computed outside of the provider by the Pulumi CLI and the engine within; the engine does not understand sets and therefore deals with a list projection. The inputs are ordered arbitrarily according to program order, but the state read from the cloud might be (should be?) ordered by TF-canonical order (using Set function that gives an identity to each element).  This makes Pulumi display spurious reordering diffs the are not visible in TF.

What we could do here:

- protocol change that lets bridged provider participate in computing diffs for refresh; make that set-aware

- without a protocol change we can do a pass over Read results that attempts to reorder set element projection to match the order from inputs as much as possible, that is intentionally minimize the spurious diffs within the current model; this is doable quickly I suggest we stat here

### Example

A good example for Plugin Framework based resources is provided in https://github.com/pulumi/pulumi-aws/issues/3303

I'm searching for more examples for SDKv2 based resources. 

### Output of `pulumi about`

See above.

### Additional context

N/A

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

iwahbe:
Thanks for raising the issue. We will prioritize next iteration.

t0yv0:
Slightly more involved proposal that could also fix for this: https://github.com/pulumi/pulumi/issues/16072 - in this proposal the state out of Read would be threaded through the program again, so the usual Diff machinery would kick in that handles sets already.

t0yv0:
This might change significantly with https://github.com/pulumi/pulumi/pull/16146 that plumbs refresh results through Diff, though I've not verifed.

t0yv0:
@mikhailshilkov reminds of an issue https://github.com/pulumi/pulumi-gcp/issues/1935 affecting import (not just refresh). Possibly same fix if Read could canonicalize the ordering of sets.

guineveresaenger:
https://github.com/pulumi/pulumi-gcp/issues/1935 was fixed via https://github.com/pulumi/pulumi/pull/16024.

t0yv0:
[...]

PR #1974, opened 2024-05-15
[Breaking] Remove UnknownValue from shim.Schema

iwahbe:
It looks like nothing uses this code, so we don't need to keep it around. While this is technically a breaking change, I don't expect that anyone would use this method.

### DIFF
--- a/pf/internal/schemashim/attr_schema.go
+++ b/pf/internal/schemashim/attr_schema.go
@@ -138,10 +138,6 @@ func (s *attrSchema) Sensitive() bool {
 	return s.attr.IsSensitive()
 }
 
-func (*attrSchema) UnknownValue() interface{} {
-	panic("UnknownValue() should not be called during schema generation")
-}
-
 func (*attrSchema) SetElement(config interface{}) (interface{}, error) {
 	panic("SetElement() should not be called during schema generation")
 }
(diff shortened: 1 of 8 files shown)


t0yv0:
No concerns of this change breaking anyone, but there is some knowledge encoded here that's worth calling out, at the very least in https://github.com/pulumi/pulumi-terraform-bridge/issues/1904

Similar code to this one has moved (or been copied) to MakeTerraformInputs and is causing problems.

Fundamentally my theory of what this is about is that list/set-nested blocks in TF can never accept null or unknown value but in Pulumi the projection of these blocks does indeed have to handle receiving nulls and unknowns. It appears that it has attempted doing so by replacing them with plausible empty values.

Is this common knowledge or should I take time to document/track a bit more? If this is understood I'm LGTM on merging the simplification.

t0yv0 (approved):

iwahbe:
I think that as long as the code is present in `MakeTerraformInputs` (where it is actually used), then I'm happy with it as is.

> Fundamentally my theory of what this is about is that list/set-nested blocks in TF can never accept null or unknown value but in Pulumi the projection of these blocks does indeed have to handle receiving nulls and unknowns. It appears that it has attempted doing so by replacing them with plausible empty values.

This is definitely what the code is supposed to do. The code is just never invoked.

> Is this common knowledge or should I take time to document/track a bit more? If this is understood I'm LGTM on merging the simplification.

I don't think it's common knowledge but it is present in the part of the code base where it is needed: `MakeTerraformInputs`.

PR #2005, opened 2024-05-21
Unify upgradeResourceState between `provider` and `provider2`

iwahbe:
This PR extracts the non provider2 specific part of `planResourceChangeImpl.upgradeState` into it's own function: `upgradeResourceStateGRPC`. This part is a pure refactor. 

The PR then re-implements `upgradeResourceState` in terms of `upgradeResourceStateGRPC`.

The driving motivation for this change is to reduce complexity and allow both upgraders to handle large JSON numbers.

Replaces https://github.com/pulumi/pulumi-terraform-bridge/pull/1735

### DIFF
--- a/pkg/tfshim/sdk-v2/provider.go
+++ b/pkg/tfshim/sdk-v2/provider.go
@@ -124,7 +124,7 @@ func (p v2Provider) Apply(
 	if !ok {
 		return nil, fmt.Errorf("unknown resource %v", t)
 	}
-	state, err := upgradeResourceState(ctx, p.tf, r, stateFromShim(s))
+	state, err := upgradeResourceState(ctx, t, p.tf, r, stateFromShim(s))
 	if err != nil {
 		return nil, fmt.Errorf("failed to upgrade resource state: %w", err)
 	}
@@ -148,7 +148,7 @@ func (p v2Provider) Refresh(
 		return nil, fmt.Errorf("unknown resource %v", t)
 	}
 
-	state, err := upgradeResourceState(ctx, p.tf, r, stateFromShim(s))
+	state, err := upgradeResourceState(ctx, t, p.tf, r, stateFromShim(s))
 	if err != nil {
 		return nil, fmt.Errorf("failed to upgrade resource state: %w", err)
 	}
--- a/pkg/tfshim/sdk-v2/upgrade_state.go
+++ b/pkg/tfshim/sdk-v2/upgrade_state.go
@@ -92,11 +92,6 @@ func upgradeResourceState(ctx context.Context, t string, p *schema.Provider, res
 		return nil, nil
 	}
 
-	version, hasVersion, err := extractSchemaVersion(instanceState.Meta)
-	if err != nil {
-		return nil, err
-	}
-
 	rawState := instanceState.RawState
 
 	// If RawState is not set but attributes is, we need to hydrate RawState
@@ -105,6 +100,12 @@ func upgradeResourceState(ctx context.Context, t string, p *schema.Provider, res
 		// We default to assuming that the old state has the same shape as the new
 		// resource.
 		typ := res.CoreConfigSchema().ImpliedType()
+
+		// Find the version, if present, to deserialize instanceState into.
+		version, hasVersion, err := extractSchemaVersion(instanceState.Meta)
+		if err != nil {
+			return nil, err
+		}
 
 		// If we have a version, we use the schema shape that matches the version
 		// specified.
--- a/pkg/tfshim/sdk-v2/upgrade_state.go
+++ b/pkg/tfshim/sdk-v2/upgrade_state.go
@@ -6,91 +6,147 @@ import (
 	"strconv"
 
 	"github.com/hashicorp/go-cty/cty"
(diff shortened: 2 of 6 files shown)


pkg/tfshim/sdk-v2/upgrade_state.go
```diff
@@ -6,92 +6,148 @@ import (
...
+		newMeta[k] = v
 	}
+	newMeta["schema_version"] = strconv.Itoa(res.SchemaVersion)
 
-	// Next, migrate the JSON state up to the current version.
-	json, err = schema.UpgradeJSONState(ctx, version, json, res, p.Meta())
+	return newState, newMeta, nil
+}
+
+func extractSchemaVersion(meta map[string]any) (int64, bool, error) {
```
iwahbe, line 71: This needed to be extracted out since we use it in multiple places:
- We use it in `upgradeResourceStateGRPC` to send the current schema version to the provider server.
- We use it in `upgradeResourceState` to find the correct type to hydrate the instanceState.Attributes into.

pkg/tfshim/sdk-v2/upgrade_state.go
```diff
@@ -103,8 +160,13 @@ func findID(v cty.Value) (string, bool) {
 	if !ok {
 		return "", false
 	}
-	if !id.Type().Equals(cty.String) {
+	if !id.Type().Equals(cty.String) || id.IsNull() {
```
t0yv0, line 163: Interesting, I thought nulls in cty are typed, perhaps this accepts null that are not typed String.

iwahbe: Null's in cty are typed, but `id.AsString()` where `id.IsNull()` panics. `findID` already returned `"", false` when it couldn't identify an accessible string ID. This just covers one more corner case.

pkg/tfshim/sdk-v2/upgrade_state.go
```diff
@@ -6,92 +6,149 @@ import (
...
-	// Next, migrate the JSON state up to the current version.
-	json, err = schema.UpgradeJSONState(ctx, version, json, res, p.Meta())
-	if err != nil {
-		return nil, err
+	newMeta := make(map[string]interface{}, len(meta))
+	// copy old meta into new meta
+	for k, v := range meta {
+		newMeta[k] = v
 	}
+	newMeta["schema_version"] = strconv.Itoa(res.SchemaVersion)
```
t0yv0, line 66: Omit if 0.

pkg/tfshim/sdk-v2/provider.go
```diff
@@ -148,7 +148,7 @@ func (p v2Provider) Refresh(
 		return nil, fmt.Errorf("unknown resource %v", t)
 	}
 
-	state, err := upgradeResourceState(ctx, p.tf, r, stateFromShim(s))
+	state, err := upgradeResourceState(ctx, t, p.tf, r, stateFromShim(s))
```
t0yv0, line 151: Should this use GRPC version? Possibly later?

pkg/tfshim/sdk-v2/upgrade_state.go
```diff
@@ -6,92 +6,149 @@ import (
...
 	}
-	newState.RawConfig = instanceState.RawConfig
 
-	// Copy the original ID and meta to the new state and stamp in the new version.
+	newState.RawConfig = instanceState.RawConfig
+	newState.RawState = v
+	newState.Meta = newMeta
 	newState.ID = instanceState.ID
 
 	// If state upgraders have modified the ID, respect the modification.
```
t0yv0, line 148: I vaguely recall working on the fix for this, would be interesting to see if the gRPC version just handles it right.

t0yv0 (approved):

pkg/tfshim/sdk-v2/upgrade_state.go
t0yv0: Ah that makes perfect sense. Same as the expanded form:


```
	if !id.Type().Equals(cty.String) {
		return "", false
	}
        if id.IsNull() {
  		return "", false
        }
	return id.AsString(), true
```

pkg/tfshim/sdk-v2/provider.go
iwahbe: It does. `upgradeResourceState` shims from the representation used by `v2Provider` to `cty.Value`, invokes `upgradeResourceStateGRPC` and then shims back.

pkg/tfshim/sdk-v2/upgrade_state.go
iwahbe: It doesn't, since the gRPC version returns a `(cty.Value, map[string]any, error)`. This is still needed to bridge back to `*terraform.InstanceState`.

t0yv0: Brilliant. We can connect this to https://github.com/pulumi/pulumi-terraform-bridge/issues/1895 - I believe if we do this right and transition to PlanResourceChange cty.Value is all that is needed.

PR #2029, opened 2024-05-24
Only validate IDs for PF resources

iwahbe:
This PR is best reviewed commit by commit:
- eb9cc3c38c47fdfbeb15f056fced769eedcaa354 restricts ID validation to PF based resources. SDK based resources all have an implicit ID field and so don't need this check.
- 5754d071ba41edf01b283fae82143b5cf7ce36a8 moves the various provider validation functions into 1 main function: `checkProvider`. This ensures that every entry point can just call `checkProvider`. Previously we had two validation functions and we didn't always call both of them.
- 9641d85d4f16ee9e205b3aaabc6f8c17c038cbc2 refactors `checkProvider` and it's sub-functions into their own internal package. We do this to prevent directly calling the non-primary function.
- df23bf25511ca4a638453c6bb8ce8c83da416d23 adds some tests to the check in general, and the new behavior introduced in eb9cc3c38c47fdfbeb15f056fced769eedcaa354 in particular.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2030

### DIFF
--- a/pf/internal/check/check_test.go
+++ b/pf/internal/check/check_test.go
@@ -1,0 +1,125 @@
+// Copyright 2016-2024, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package check
+
+import (
+	"bytes"
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/provider"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	sdkschema "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	sdkv2 "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/sdk-v2"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
+	property "github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+
+	pfbridge "github.com/pulumi/pulumi-terraform-bridge/pf/tfbridge"
+)
+
+func TestMissingIDProperty(t *testing.T) {
+	stderr, err := test(t, tfbridge.ProviderInfo{
+		P: pfbridge.ShimProvider(testProvider{}),
+	})
+
+	assert.Equal(t, "error: Resource test_res has a problem: no \"id\" attribute. "+
+		"To map this resource consider specifying ResourceInfo.ComputeID\n", stderr)
+
+	assert.ErrorContains(t, err, "There were 1 unresolved ID mapping errors")
+}
+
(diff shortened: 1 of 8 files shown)


pf/internal/check/checks.go
```diff
@@ -12,20 +12,44 @@
...
-	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
 )
 
-func checkIDProperties(sink diag.Sink, info tfbridge.ProviderInfo) error {
+// Validate that info is valid as either a PF provider or a PF & SDK based provider.
+//
+// This function should be called in the generate step, but before schema generation (so
+// as to error as soon as possible).
+func Provider(sink diag.Sink, info tfbridge.ProviderInfo) error {
+	isPFResource := func(string) bool { return true }
```
guineveresaenger, line 33: Can we add a comment here that explains why we set the default to `true`? 
I presume "any provider that is _not_ muxed by default only contains PF resources so this always returns true for those".

pf/internal/check/checks.go
```diff
@@ -12,20 +12,44 @@
...
+		checkIDProperties(sink, info, isPFResource),
+		notSupported(sink, info, isPFResource, isPFDataSource),
+	)
+}
+
+func checkIDProperties(sink diag.Sink, info tfbridge.ProviderInfo, doCheck func(tfToken string) bool) error {
 	errors := 0
 
 	info.P.ResourcesMap().Range(func(rname string, resource shim.Resource) bool {
+		if !doCheck(rname) {
```
guineveresaenger, line 50: I think this would be more clear if it were still called `isPfResource`. Ideally we would only call `checkIDProperties` at all if the resource in question is a PF resource, but since that's not how the (muxed) resources map is set up, I think a name that says "if this _isn't_ a PF resource, we always return true" is more clear. The decision to do the check is inherent to the PF-y-ness.

guineveresaenger (approved):
Couple of nits around comeents/naming.

It's not entirely clear to me that this needs to be in an `internal` package. Is it not sufficient to export `Provider` (and `Provider` only) if we have a new `check` package?

iwahbe:
> It's not entirely clear to me that this needs to be in an `internal` package.

I'm not sure what you mean. Are you suggesting that it shouldn't be a package or that it shouldn't be internal?

>  Is it not sufficient to export `Provider` (and `Provider` only) if we have a new `check` package?

We do export `Provider` (and `Provider` only), but we only export them for `tfgen`. We don't want end users to be calling these functions. We don't have any backwards compatibility guarantees on names or signatures.

pf/internal/check/checks.go
iwahbe: Added in 83fd9b4eb5b338c66ba9699b2cbdb12a31245315.

pf/internal/check/checks.go
iwahbe: I added a name change and a comment in 83fd9b4eb5b338c66ba9699b2cbdb12a31245315.

guineveresaenger:
I see, you are saying end users should not be calling `check.Provider()` at all either.

I don't have a particularly strong feeling here; I would've left it at `tfgen.checkProvider` but usage wise `internal` makes sense too. Feel free to resolve.

iwahbe:
Waiting on API token limits to get to a green CI and merge.

pf/internal/check/check_test.go
```diff
@@ -0,0 +1,125 @@
...
+	err := Provider(sink, info)
+
+	// We should not write diags to stdout
+	assert.Empty(t, stdout.String())
+
+	return stderr.String(), err
+}
+
+type (
+	testProvider struct{ provider.Provider }
```
t0yv0, line 94: Lots of tests need this. It looks like the Plugin Framework has this functionality internally (unfortunately) so we cannot just use that, but we have this instead which is very similar and can be considered useful here:

https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pf/tests/internal/providerbuilder/build_provider.go#L14

Just a helper that allows one to mostly fill out callbacks instead of defining brand new types.

t0yv0 (approved):
The changes LGTM.

pf/internal/check/check_test.go
iwahbe: I didn't know that existed. I'd love to use it, except that it's internal and pf/test is it's own module. I would need to move some files around to make this work, and its not worth it for this PR IMO.

t0yv0: Yeah we'd need to move it first. Just FYI next time.

PR #2044, opened 2024-05-29
Bump pulumi/pulumi to v3.117.0

iwahbe:
The weekly job ran, but it looks like it failed silently: https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/9208221284.


t0yv0:
Something still looks not quite right in the test suite.

iwahbe:
Superseeded by https://github.com/pulumi/pulumi-terraform-bridge/pull/2045.

PR #2050, opened 2024-05-31
PlanResourceChange: do not call state upgraders during Create

t0yv0:
When testing rolling out PlanResourceChange to the entirety of AWS, basic tests such as TestAccCloudWatch  started failing with a panic from the provider upgrade machinery.  The issue is that provider upgrades should not be called on Create. This is now fixed.

### DIFF
--- a/pkg/tests/cross-tests/adapt.go
+++ b/pkg/tests/cross-tests/adapt.go
@@ -103,6 +103,8 @@ func (ta *typeAdapter) NewValue(value any) tftypes.Value {
 		case map[string]any:
 			values := map[string]tftypes.Value{}
 			for k, el := range v {
+				_, ok := aT[k]
+				contract.Assertf(ok, "no type for attribute %q", k)
 				values[k] = fromType(aT[k]).NewValue(el)
 			}
 			return tftypes.NewValue(t, values)
@@ -112,6 +114,7 @@ func (ta *typeAdapter) NewValue(value any) tftypes.Value {
 }
 
 func fromType(t tftypes.Type) *typeAdapter {
+	contract.Assertf(t != nil, "type cannot be nil")
 	return &typeAdapter{t}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/tests/cross-tests/tf_driver.go
```diff
@@ -120,6 +120,9 @@ func (d *tfDriver) coalesce(t T, x any) *tftypes.Value {
 		return nil
 	}
 	objectType := convert.InferObjectType(sdkv2.NewSchemaMap(d.res.Schema), nil)
+	for k := range objectType.AttributeTypes {
+		objectType.OptionalAttributes[k] = struct{}{}
```
t0yv0, line 124: To be able to express missing values, rather than nils, in cross-tests, just assume all attributes are optional. I think this should work out.

pkg/tfshim/sdk-v2/provider2.go
```diff
@@ -328,6 +328,11 @@ func (p *planResourceChangeImpl) upgradeState(
 	res := p.tf.ResourcesMap[t]
 	state := p.unpackInstanceState(t, s)
 
+	// In the case of Create, prior state is encoded as Null and should not be subject to upgrades.
+	if state.stateValue.IsNull() {
```
t0yv0, line 332: The fix. Don't upgrade nil state.

pkg/tests/cross-tests/adapt.go
```diff
@@ -103,6 +103,8 @@ func (ta *typeAdapter) NewValue(value any) tftypes.Value {
 		case map[string]any:
 			values := map[string]tftypes.Value{}
 			for k, el := range v {
+				_, ok := aT[k]
+				contract.Assertf(ok, "no type for attribute %q", k)
```
t0yv0, line 107: The stack traces are not helpful otherwise.

VenelinMartinov (approved):
Great find, thank you for fixing!

PR #2068, opened 2024-06-06
Emit upstream Installation instructions

guineveresaenger:
This pull request is the scaffold implementation of generating installation docs for bridged providers from upstream source.

In this change, a new "language" called `registry-docs` is added to the bridge. (Note that "language" in this case is a bit of a misnomer - it is the `tfgen` CLI's input for determining which generating tactic to use - this is why `schema` is a language as well).

This change reads the upstream `index.md` file and emits it as a Pulumi [package installation instruction file](https://www.pulumi.com/docs/using-pulumi/pulumi-packages/how-to-author/#overview-installation--configuration) to a top-level folder called `docs/` (the current registry standard). We also emit a minimal `_index.md` file. When these files exist in a provider, the registry will use those instead of the hard coded files existing in the registry currently.

This pull request includes a fair bit of TODOs in the docs parser - currently only the Hugo front matter is implemented, to show the scaffold. Transformations for this document to be Pulumi-friendly will follow in a separate PR.

This change does not affect any current use of the bridge - to flip it on, we'd add another Make target to a provider's Makefile, e.g. `./bin/pulumi-tfgen-openstack registry-docs --out docs/`. I'd like us to consider adding this change despite the TODOs, so we can keep the changes somewhat atomic and reduce merge conflicts.

Closes https://github.com/pulumi/pulumi-terraform-bridge/issues/2082.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -2362,12 +2362,11 @@ func guessIsHCL(code string) bool {
 	return guessIsHCLPattern.MatchString(code)
 }
 
-func plainDocsParser(docFile *DocFile, pkgName string, g *Generator) ([]byte, error) {
+func plainDocsParser(docFile *DocFile, g *Generator) ([]byte, error) {
 	// Replace upstream front matter with Pulumi registry's
-	contentStr, err := replaceUpstreamFrontMatter(string(docFile.Content), pkgName)
-	if err != nil {
-		return nil, err
-	}
+	contentStr, title := getBodyAndTitle(string(docFile.Content))
+	contentStr = writeFrontMatter(title) + contentStr
+
 	//TODO: See https://github.com/pulumi/pulumi-terraform-bridge/issues/2078
 	// - translate code blocks with code choosers
 	// - apply default edit rules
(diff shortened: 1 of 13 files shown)


pkg/tfgen/docs.go
```diff
@@ -2358,3 +2361,53 @@ var (
 func guessIsHCL(code string) bool {
 	return guessIsHCLPattern.MatchString(code)
 }
+
+func plainDocsParser(docFile *DocFile, pkgName string, g *Generator) ([]byte, error) {
+	// Replace upstream front matter with Pulumi registry's
```
iwahbe, line 2366: Typo?
```suggestion
	// Replace upstream front matter within Pulumi's registry
```

pkg/tfgen/docs.go
```diff
@@ -2358,3 +2361,53 @@ var (
...
+		return nil, err
+	}
+
+	// TODO: translate code blocks
+
+	//TODO: apply default edit rules
+
+	//TODO: reformat text
+
+	//TODO: Light translation / possible eliding for certain headers such as "Arguments Reference"
```
iwahbe, line 2378: Do you want to merge with these TODOs in place? If so they will need an associated issue.

pkg/tfgen/docs.go
```diff
@@ -2358,3 +2361,53 @@ var (
...
+	// or "Configuration block"
+	return []byte(contentStr), nil
+}
+
+func replaceUpstreamFrontMatter(content, pkgName string) (string, error) {
+	// Capitalize the package name
+	capitalize := cases.Title(language.English)
+	pkgName = capitalize.String(pkgName)
```
iwahbe, line 2386: Shouldn't this be the `DisplayName` field on the schema?

pkg/tfgen/generate.go
```diff
@@ -88,12 +88,13 @@ type Generator struct {
...
-	CSharp Language = "dotnet"
-	Schema Language = "schema"
-	PCL    Language = "pulumi"
+	Golang       Language = "go"
+	NodeJS       Language = "nodejs"
+	Python       Language = "python"
+	CSharp       Language = "dotnet"
+	Schema       Language = "schema"
+	PCL          Language = "pulumi"
+	RegistryDocs Language = "registry-docs"
```
iwahbe, line 97: We've talked a lot about why it makes sense for `RegistryDocs` to be a "language". Can you add a comment here summarizing why this makes sense?

pkg/tfgen/source.go
```diff
@@ -236,6 +249,13 @@ func getDocsPath(repo string, kind DocKind) ([]string, error) {
 
 	var paths []string
 
+	if kind == InstallationDocs {
+		// ${repo}/docs/
```
iwahbe, line 253: Frontmatter isn't present for repos that use `website`?

guineveresaenger: I'm not sure I understand. Repos that use `website` have front matter: https://raw.githubusercontent.com/hashicorp/terraform-provider-aws/main/website/docs/index.html.markdown

pkg/tfgen/docs.go
guineveresaenger: would `Replace upstream front matter with Pulumi registry's front matter` be more clear? we're just subbing in what our registry needs here.

iwahbe: Yes. Much more clear.

pkg/tfgen/source.go
iwahbe: This path is only `${repo}/docs/`. I'm saying we probably want to accommodate the `${repo}/website/docs/` path (as is used by your example).

pkg/tfgen/docs.go
guineveresaenger: It should. But we do not in fact enforce this field on our providers. I'll add some logic to use DisplayName preferentially.

pkg/tfgen/docs.go
guineveresaenger: Filed https://github.com/pulumi/pulumi-terraform-bridge/issues/2078 to track this work.

iwahbe (approved):

PR #2073, opened 2024-06-07
Experiments with walking sdkv2 schema

t0yv0:


t0yv0:
@VenelinMartinov I'm getting to PASS TestCollectionsRefreshClean. 

Couple of things.

1. had to implement Create; this gets clean refresh to pass on baseline
2. the nested-block rewrite lost the important idea that we need to normalize to config value and not to null value; if config is ok we should not normalize it
3. further testing uncovered that it's not the config value that matters but the old state value in fact we should be looking at for comparison
4. looks like not only normalizing empty collections to null but vice versa is needed

t0yv0:
Closing in favor of #2065 where @VenelinMartinov is taking it to completion.

PR #2074, opened 2024-06-07
Add an edit rule: "Made by <elided>" to "Made by Pulumi"

iwahbe:
This applies to resources like
https://www.pulumi.com/registry/packages/scm/api-docs/addressobject/. Currently, the example shows:

```ts
import * as scm from "@pulumi/scm";

const example = new scm.AddressObject("example", {
    folder: "Shared",
    name: "example",
    description: "Made by Terraform",
    ipNetmask: "10.2.3.4",
});
```

It will now show:

```ts
import * as scm from "@pulumi/scm";

const example = new scm.AddressObject("example", {
    folder: "Shared",
    name: "example",
    description: "Made by Pulumi",
    ipNetmask: "10.2.3.4",
});
```


pkg/tfgen/docs.go
```diff
@@ -147,98 +147,6 @@ const (
 	DataSourceDocs DocKind = "data-sources"
 )
 
-// Create a regexp based replace rule that is bounded by non-ascii letter text.
```
iwahbe, line 150: I moved these wholesale to `edit_rules.go`. They are the same in the new file.

pkg/tfgen/docs.go
```diff
@@ -147,98 +147,6 @@ const (
...
-	}
-	return contents, nil
-}
-
-// Get the replace rule set for a DocRuleInfo.
-//
-// getEditRules is only called once during `tfgen`, so we move the cost of compiling
-// regexes into getEditRules, avoiding a marginal startup time penalty.
-func getEditRules(info *tfbridge.DocRuleInfo) editRules {
-	defaults := []tfbridge.DocsEdit{
```
iwahbe, line 225: The default list was refactored into it's own function: `defaultEditRules` for clarity, but otherwise this function remains unchanged.

pkg/tfgen/docs_test.go
```diff
@@ -1647,30 +1647,33 @@ func TestParseTFMarkdown(t *testing.T) {
...
-			}
-
-			tc.providerInfo.DocRules = &tfbridge.DocRuleInfo{
-				EditRules: func(defaults []tfbridge.DocsEdit) []tfbridge.DocsEdit {
-					return append([]tfbridge.DocsEdit{rule}, defaults...)
-				},
-			}
-		}),
+		test("custom-replaces",
+			editRule(func(path string, content []byte) ([]byte, error) {
```
iwahbe, line 1671: This is a pure refactor. I did this when I thought I was going to need to extend this test, but I believe that it remains an improvement.

pkg/tfgen/edit_rules.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+)
+
+func defaultEditRules() editRules {
+	return editRules{
+		// Replace content such as "`terraform plan`" with "`pulumi preview`"
+		boundedReplace("[tT]erraform [pP]lan", "pulumi preview"),
+		// Replace content such as " Terraform Apply." with " pulumi up."
+		boundedReplace("[tT]erraform [aA]pply", "pulumi up"),
+		reReplace(`"([mM])ade (by|with) [tT]erraform"`, `"Made $2 Pulumi"`),
```
iwahbe, line 32: Adding this rule is the actual change.

pkg/tfgen/test_data/signalfx-log-timeline/expected.json
```diff
@@ -1,5 +1,5 @@
...
+  "Description": "You can add logs data to your Observability Cloud dashboards without turning your logs into metrics first.\n\nA log timeline chart displays timeline visualization in a dashboard and shows you in detail what is happening and why.\n\n## Example\n\n```tf\nresource \"signalfx_log_timeline\" \"my_log_timeline\" {\n  name        = \"Sample Log Timeline\"\n  description = \"Lorem ipsum dolor sit amet, laudem tibique iracundia at mea. Nam posse dolores ex, nec cu adhuc putent honestatis\"\n\n  program_text = \u003c\u003c-EOF\n  logs(filter=field('message') == 'Transaction processed' and field('service.name') == 'paymentservice').publish()\n  EOF\n\n  time_range = 900\n\n}\n```",
```
iwahbe, line 2: This was unrelated to my change, and serializes the same (so tests pass with both representations).

pkg/tfgen/edit_rules.go
guineveresaenger: Do we feel confident that capitalizing the replacement will make sense in all situations?

guineveresaenger (approved):
Really appreciate the restructuring into a new file here.

pkg/tfgen/edit_rules.go
iwahbe: I think so. It's only for quoted strings, so I think this will only happen in description fields where it would make sense.

PR #2075, opened 2024-06-07
Standardize a pf/tfbridge compatible provider on pf.ShimProvider

iwahbe:
This is a pure refactor and will not have any end user consequences.

This was broken off from https://github.com/pulumi/pulumi-terraform-bridge/pull/1973.


VenelinMartinov (approved):
lgtm

PR #2080, opened 2024-06-11
Dynamic Bridged Providers: Plugin Aquisition

iwahbe:
Broken off from https://github.com/pulumi/pulumi-terraform-bridge/pull/1973 for easier review.

Fixes https://github.com/pulumi/home/issues/3409


dynamic/internal/shim/loader.go
```diff
@@ -0,0 +1,251 @@
...
+	tfaddr "github.com/opentofu/registry-address"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+)
+
+// TODO:
+//
+// - If TF is setup and has already cached a provider, we should try to use that
```
t0yv0, line 49: This one is tricky I see the speed benefits but creating an interference with TF's own cache sounds like something to be avoided perhaps, especially things like attempting to write to it.

dynamic/internal/shim/loader.go
```diff
@@ -0,0 +1,251 @@
...
+// provider.
+//
+// - If not, but TF is setup we should probably contribute to the existing cache.
+//
+// - If TF is not setup, we should cache within PULUMI_HOME to avoid creating new dirs.
+const envPluginCache = "TF_PLUGIN_CACHE_DIR"
```
t0yv0, line 55: Should this be prefixed with PULUMI?

dynamic/internal/shim/loader.go
```diff
@@ -0,0 +1,251 @@
...
+	switch protoVer {
+	case 5:
+		p := raw.(*tfplugin.GRPCProvider)
+		p.PluginClient = client
+		p.Addr = meta.Provider
+
+		slog.Info("Found v5 provider.. upgrading to v6")
+
+		v6, err := tf5to6server.UpgradeServer(ctx, func() tfprotov5.ProviderServer {
+			return v5shim.New(tfplugin5.NewProviderClient(rpcClient.(*plugin.GRPCClient).Conn))
```
t0yv0, line 239: This is interesting. Indeed looks like protobuf doesn't make this translation easy though it should be entirely mechanical.

t0yv0 (approved):

dynamic/internal/shim/loader.go
t0yv0: I see, this is actually one level up from raw protos. And something like this https://github.com/hashicorp/terraform-plugin-go/blob/v0.23.0/tfprotov5/internal/fromproto/provider.go#L31 is internal unfortunately.

dynamic/internal/shim/loader.go
iwahbe: It should be prefixed with Pulumi. Thanks for pointing that out.

dynamic/internal/shim/loader.go
iwahbe: Followup issue: https://github.com/pulumi/pulumi-terraform-bridge/issues/2090

Issue #2088, opened 2024-06-12
Panic in muxer when attaching with PULUMI_DEBUG_PROVIDERS


guineveresaenger:
### What happened?

Following [debug provider steps](https://github.com/pulumi/platform-providers-team/blob/main/playbooks/How-To/Debug%20providers.md) in p-aws, upon starting a program using PULUMI_DEBUG_PROVIDERS and the exposed port, the process crashes.

### Example

```
Got a connection, launched process /Users/guin/go/src/github.com/pulumi/pulumi-aws/bin/pulumi-resource-aws (pid = 97357).
55429
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1059f7714]

goroutine 55 [running]:
github.com/pulumi/pulumi/pkg/v3/resource/provider.(*HostClient).Close(0x0)
        /Users/guin/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.119.0/resource/provider/host.go:62 +0x24
github.com/pulumi/pulumi-terraform-bridge/x/muxer.(*muxer).Attach(0x140038df3e0, {0x11c57f200, 0x14007632c30}, 0x1400764e240)
        /Users/guin/go/src/github.com/pulumi/pulumi-terraform-bridge/x/muxer/muxer.go:438 +0x250
github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Attach_Handler.func1({0x11c57f200, 0x14007632c30}, {0x11b376520, 0x1400764e240})
        /Users/guin/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.119.0/proto/go/provider_grpc.pb.go:777 +0xbc
github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc.OpenTracingServerInterceptor.func1({0x11c57f200, 0x14007632c30}, {0x11b376520, 0x1400764e240}, 0x14007636180, 0x1400250a360)
        /Users/guin/go/pkg/mod/github.com/grpc-ecosystem/grpc-opentracing@v0.0.0-20180507213350-8e809c8a8645/go/otgrpc/server.go:57 +0x574
[...]

Issue #1293, opened 2023-07-20
Can't attach to processes using the muxer


danielrbradley:
When using `PULUMI_DEBUG_PROVIDERS` with a bridged provider using the muxer, it panics on attach.

The `host` field on `m` is nil in the following method. I would assume this is due to the host not being defined if we've started the provider manually instead of via the engine.

I would assume that this is likely fixed by only calling `Close` if `m.host` is not nil.

Relevant code:
```golang
func (m *muxer) Attach(ctx context.Context, req *rpc.PluginAttach) (*emptypb.Empty, error) {
	host, err := provider.NewHostClient(req.GetAddress())
	if err != nil {
		return nil, err
	}
	if err := m.host.Close(); err != nil {
		return nil, err
	}
	// Here we override the underlying host. This should replace the host instance of
	// each subsidiary provider.
	*m.host = *host
	return &emptypb.Empty{}, nil
}
```

Stacktrace from attaching to the AWS provider:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x10a329460]

goroutine 41 [running]:
github.com/pulumi/pulumi/pkg/v3/resource/provider.(*HostClient).Close(...)
        /Users/dan/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.75.0/resource/provider/host.go:62
github.com/pulumi/pulumi-terraform-bridge/x/muxer.(*muxer).Attach(0x14000cb94f0, {0x10e3f58cc?, 0x40?}, 0x11054d300?)
        /Users/dan/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/x/muxer@v0.0.4/muxer.go:401 +0x50
github.com/pulumi/pulumi/sdk/v3/proto/go._ResourceProvider_Attach_Handler.func1({0x113173b68, 0x14000774c00}, {0x111f90da0?, 0x1400077c240})
        /Users/dan/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.75.0/proto/go/provider_grpc.pb.go:699 +0x74
github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc.OpenTracingServerInterceptor.func1({0x113173b68, 0x140007749c0}, {0x111f90da0, 0x1400077c240}, 0x1400205a180, 0x14000764330)
[...]

PR #1299, opened 2023-07-24
Fix panic on muxer debug attach

danielrbradley:
Only attempt to close the existing host if it exists

Fixes #1293

PR #2091, opened 2024-06-13
Get new host client directly when attaching a muxed provider

guineveresaenger:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2088.

### DIFF
--- a/x/muxer/muxer.go
+++ b/x/muxer/muxer.go
@@ -16,9 +16,10 @@ package muxer
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	pbempty "github.com/golang/protobuf/ptypes/empty"
 	"io"
+	"reflect"
 	"sync"
 
 	"github.com/golang/glog"
@@ -424,12 +425,23 @@ func (m *muxer) GetPluginInfo(ctx context.Context, e *emptypb.Empty) (*rpc.Plugi
 }
 
 func (m *muxer) Attach(ctx context.Context, req *rpc.PluginAttach) (*emptypb.Empty, error) {
-	host, err := provider.NewHostClient(req.GetAddress())
-	if err != nil {
-		return nil, err
+	attach := make([]func() error, len(m.servers))
+	for i, s := range m.servers {
+		s := s
+		attach[i] = func() error {
+			_, err := s.Attach(ctx, req)
+			return err
+		}
 	}
-	m.host = host
-	return &pbempty.Empty{}, nil
+
+	var closeErr error
+	if !reflect.ValueOf(m.host).IsNil() {
+		closeErr = m.host.Close()
+	}
+
+	var err error
+	m.host, err = provider.NewHostClient(req.GetAddress())
+	return &emptypb.Empty{}, errors.Join(append(asyncJoin(attach), err, closeErr)...)
 }
 
 type getMappingArgs struct {
--- a/x/muxer/muxer.go
+++ b/x/muxer/muxer.go
@@ -435,6 +435,8 @@ func (m *muxer) Attach(ctx context.Context, req *rpc.PluginAttach) (*emptypb.Emp
 	}
 
 	var closeErr error
+	// Because in Go, an interface type is not nil even when its underlying value is nil, the nil check here
+	//must test the underlying type.
 	if !reflect.ValueOf(m.host).IsNil() {
 		closeErr = m.host.Close()
 	}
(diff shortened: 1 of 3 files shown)


iwahbe (approved):
It would be really great to add a test here, but I won't require it since this is us (not user) facing and because there isn't a test already.

x/muxer/muxer.go
```diff
@@ -424,24 +424,12 @@ func (m *muxer) GetPluginInfo(ctx context.Context, e *emptypb.Empty) (*rpc.Plugi
 }
 
 func (m *muxer) Attach(ctx context.Context, req *rpc.PluginAttach) (*emptypb.Empty, error) {
-	attach := make([]func() error, len(m.servers))
-	for i, s := range m.servers {
-		s := s
-		attach[i] = func() error {
-			_, err := s.Attach(ctx, req)
```
t0yv0, line 431: Why was this wrong for sub-providers to attach?

guineveresaenger: It never actually attached a host to the muxer, as far as I could tell.

t0yv0:
TBH I don't quite understand the fix and I think this still a borderline call whether this is a user-facing panic or not, I think it is since https://www.pulumi.com/docs/using-pulumi/pulumi-packages/debugging-provider-packages/ documents the feature for the user.

This issue was claimed to be fixed:

https://github.com/pulumi/pulumi-terraform-bridge/issues/1293
https://github.com/pulumi/pulumi-terraform-bridge/pull/1299

I recall referencing something in AWS claiming that the fix is coming. Since it's regressed, I'd like to request a test, and an RCA we can all understand. CC @mjeffryes for a second opinion.

t0yv0:
Ok the ticket also mentions https://github.com/pulumi/pulumi-terraform-bridge/pull/1716

That included a test that was skipped until https://github.com/pulumi/pulumi/pull/15526 which is now in.

Can this test be un-skipped now, looks like I failed to follow up on that since 15526 went in? Was this test not covering the feature properly or is it in fact failing once un-skipped?

t0yv0 (changes requested):
I was curious so it turns out that somewhere around here:

https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pf/tfbridge/main.go#L117 

The pulumi/pkg code allocates an unclosable host that would close upon panic.

The muxer Attach method tries to close the previously allocated host (this one) before re-setting it, which now panics.

The skipped test succeeds but alas testbridge that its' written against is not a muxed provider so it does not cover AWS test case.

```go
func TestAttach(t *testing.T) {
	source := filepath.Join("..", "testdata", "basicprogram")
	bin, err := filepath.Abs(filepath.Join("..", "bin"))
	require.NoError(t, err)
	pt := pulumitest.NewPulumiTest(t, source,
		opttest.AttachProviderBinary("testbridge", bin),
		opttest.SkipInstall())
	pr := pt.Preview()
	t.Logf("%v", pr.StdOut+pr.StdErr)
}

```

x/muxer/muxer.go
```diff
@@ -424,24 +424,12 @@ func (m *muxer) GetPluginInfo(ctx context.Context, e *emptypb.Empty) (*rpc.Plugi
...
+	host, err := provider.NewHostClient(req.GetAddress())
+	if err != nil {
+		return nil, err
 	}
-
-	var err error
-	m.host, err = provider.NewHostClient(req.GetAddress())
-
-	return &emptypb.Empty{}, errors.Join(append(asyncJoin(attach), err, closeErr)...)
+	m.host = host
```
t0yv0, line 431: Since this host client is now not propagated to the constituent providers they will not be able to work correctly (send logs to the host for example).

x/muxer/muxer.go
guineveresaenger: update: it i=turns out this is a parallel functionality that doesn't affect m.host

x/muxer/muxer.go
guineveresaenger: Yup, that was a red herring; see above

t0yv0:
It's better! Could we please re-enable TestAttach test and add one for muxed providers at integration level? Much appreciated.

danielrbradley:
Would love a brief explanation on why and how this fixes the issue. My go-foo is lacking here.

It's a beautiful one line fix .. but I have no idea _why_ it would fix it 🙃

t0yv0:
http://golang50shad.es/index.html#nil_in_nil_in_vals I think. It truly is non-intuitive.

t0yv0 (approved):

PR #2093, opened 2024-06-13
feat: type check config values

corymhall:
This PR extends the typechecker to also type check config values in `CheckConfig`. The motivation behind this is
https://github.com/pulumi/pulumi-aws/issues/4004. In that issue the user provides a key that is not in the allowed list and gets an opaque error message. In Terraform the user gets a more descriptive error message, so our behavior here is worse.

The big difference between the previous PR that is typechecking resource inputs is in that case we were preventing panics whereas in this case the primary motivation is to give _better_ error messages. I am still only writing a warning for now which gives a message to the user like this:

```console
  pulumi:providers:aws (provider):
    warning: Unexpected type at field "endpoints[0]":
               an unexpected argument "abcd" was provided
    error: pulumi:providers:aws resource 'chall-provider' has a problem: could not validate provider configuration: Invalid or unknown key. Examine values at 'chall-provider.endpoints'.
```

fixes https://github.com/pulumi/pulumi-aws/issues/4004

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -523,8 +523,8 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
 		}
 	}
 
-	if err := p.typeCheckConfig(ctx, urn, news); err != nil {
-		return err, nil
+	if check := p.typeCheckConfig(ctx, urn, news); check != nil {
+		return check, nil
 	}
 
 	checkFailures := validateProviderConfig(ctx, urn, p, config)
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -523,6 +523,39 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
 		}
 	}
 
+	if err := p.typeCheckConfig(ctx, urn, news); err != nil {
+		return err, nil
+	}
+
+	checkFailures := validateProviderConfig(ctx, urn, p, config)
+	if len(checkFailures) > 0 {
+		return &pulumirpc.CheckResponse{
+			Failures: checkFailures,
+		}, nil
+	}
+
+	// Ensure properties marked secret in the schema have secret values.
+	secretNews := MarkSchemaSecrets(ctx, p.config, p.info.Config, resource.NewObjectProperty(news)).ObjectValue()
+
+	// In case news was modified by pre-configure callbacks, marshal it again to send out the modified value.
+	newsStruct, err := configEnc.MarshalProperties(secretNews)
+	if err != nil {
+		return nil, err
+	}
+
+	return &pulumirpc.CheckResponse{
+		Inputs: newsStruct,
+	}, nil
+}
+
+func (p *Provider) typeCheckConfig(
+	ctx context.Context,
+	urn resource.URN,
+	news resource.PropertyMap,
+) *pulumirpc.CheckResponse {
+	span, _ := opentracing.StartSpanFromContext(ctx, "sdkv2.typeCheckConfig")
+	defer span.Finish()
+
 	logger := GetLogger(ctx)
 	// for now we are just going to log warnings if there are failures.
 	// over time we may want to turn these into actual errors
@@ -552,7 +585,7 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
 			if len(failures) > 0 {
 				return &pulumirpc.CheckResponse{
 					Failures: failures,
-				}, nil
+				}
 			}
 			logger.Warn("Type checking is still experimental. If you believe that a warning is incorrect,\n" +
 				"please let us know by creating an " +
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/provider.go
```diff
@@ -523,6 +523,45 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
 		}
 	}
 
+	logger := GetLogger(ctx)
```
t0yv0, line 526: nit: maybe a separate func here, I prefer smaller definitions.

t0yv0:
This is an interesting approach that might be viable! The concern as always is regressions, would it start rejecting programs that worked fine before? 

Also I am curious, how does TF present this error? For other TF validation errors we had a translation layer that picked up the TF error and converted it to a Pulumi error, is there a chance that that could work here as well as an alternative approach for this specific issue, or is TF just not emitting a well labelled error?

corymhall:
> This is an interesting approach that might be viable! The concern as always is regressions, would it start rejecting programs that worked fine before?

It shouldn't! We are starting off with a warning message so we should be ok.

> Also I am curious, how does TF present this error? For other TF validation errors we had a translation layer that picked up the TF error and converted it to a Pulumi error, is there a chance that that could work here as well as an alternative approach for this specific issue, or is TF just not emitting a well labelled error?

I'm not sure if Terraform has some other process for converting their error messages to user facing error messages because I couldn't trace the error message that you get from the CLI to what we are converting in the bridge. For example, this is the error you get in Terraform.

```console
╷
│ Error: Unsupported argument
│
│   on main.tf line 14, in provider "aws":
│   14:     workspacesweb = "http://localhost:4655"
│
│ An argument named "workspacesweb" is not expected here.
╵
```

> beyond this one particular error in AWS, though I can't quickly find example tickets pertaining to this.

Yeah I couldn't either, but this is the only good way I could think of to actually handle this type of error message.

pkg/tfbridge/provider.go
```diff
@@ -523,6 +523,45 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
...
+	// for now we are just going to log warnings if there are failures.
+	// over time we may want to turn these into actual errors
+	validateShouldError := cmdutil.IsTruthy(os.Getenv("PULUMI_ERROR_CONFIG_TYPE_CHECKER"))
+	schemaMap := p.config
+	schemaInfos := p.info.GetConfig()
+	if p.pulumiSchemaSpec != nil {
+		iv := NewInputValidator(urn, *p.pulumiSchemaSpec, true)
```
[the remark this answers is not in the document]
corymhall: Adding tracing to the function.

pkg/tfbridge/provider.go
corymhall: done

pkg/tfbridge/validate_input_types.go
```diff
@@ -51,6 +56,12 @@ func (v *PulumiInputValidator) validatePropertyMap(
 		objectValue := propertyMap[objectKey]
 		objType, knownType := propertyTypes[string(objectKey)]
 		if !knownType {
+			if v.validateUnknownTypes {
+				return []TypeFailure{{
+					Reason:       fmt.Sprintf("an unexpected argument %q was provided", string(objectKey)),
+					ResourcePath: propertyPath.String(),
+				}}
```
iwahbe, line 63: We should present as many errors as possible, not just the first one that comes up.
```suggestion
				failures = append(failures, TypeFailure{
					Reason:       fmt.Sprintf("an unexpected argument %q was provided", string(objectKey)),
					ResourcePath: propertyPath.String(),
				}
```

pkg/tfbridge/validate_input_types.go
```diff
@@ -274,3 +285,26 @@ func (v *PulumiInputValidator) ValidateInputs(resourceToken tokens.Type, inputs
...
+	failures := v.validatePropertyMap(
+		inputsToValidate,
+		v.schema.Config.Variables,
+		resource.PropertyPath{},
+	)
+	if len(failures) > 0 {
+		return failures
+	}
+
+	return nil
```
iwahbe, line 309: Distinguishing between `[]TypeFailure{}` and `[]TypeFailure(nil)` is an anti-pattern.

```suggestion
	return v.validatePropertyMap(
		inputsToValidate,
		v.schema.Config.Variables,
		resource.PropertyPath{},
	)
```

In general, we should strive to behave the same when given an empty slice or a `nil` slice.

pkg/tfbridge/validate_input_types_test.go
```diff
@@ -1639,3 +1639,95 @@ func TestValidateInputType_toplevel(t *testing.T) {
...
+
+			v := &PulumiInputValidator{
+				urn:                  urn,
+				schema:               pspec,
+				validateUnknownTypes: true,
+			}
+			failures := v.ValidateConfig(resource.PropertyMap{
+				resource.PropertyKey(tc.inputName): tc.input,
+			})
+			if failures != nil && len(failures) != len(tc.failures) {
```
iwahbe, line 1721: `len` works fine on `nil` slices.
```suggestion
			if len(failures) != len(tc.failures) {
```

pkg/tfbridge/validate_input_types_test.go
```diff
@@ -1639,3 +1639,95 @@ func TestValidateInputType_toplevel(t *testing.T) {
...
+				t.Fatalf("%d failures, got %d: %v", len(tc.failures), len(failures), failures)
+			}
+			if len(tc.failures) > 0 {
+				if failures == nil {
+					t.Fatalf("expected failures, got none")
+				} else {
+					assert.Equal(t, tc.failures, failures)
+				}
+			}
```
iwahbe, line 1730: Why do we have so much validation logic on failures to assert that `tc.failures == failures`? Can't we just use `assert.Equal` to handle all of that?
```suggestion
			assert.Equal(t, tc.failures, failures)
```

pkg/tfbridge/provider.go
```diff
@@ -523,6 +523,10 @@ func (p *Provider) CheckConfig(ctx context.Context, req *pulumirpc.CheckRequest)
 		}
 	}
 
+	if err := p.typeCheckConfig(ctx, urn, news); err != nil {
+		return err, nil
+	}
```
iwahbe, line 528: Let's call this `check` instead of `err`. By Go convention, `err` implies the type `error`.
```suggestion
	if check := p.typeCheckConfig(ctx, urn, news); check != nil {
		return check, nil
	}
```

iwahbe (approved):
This looks good to me.

PR #2107, opened 2024-06-18
Dynamic providers test diff

iwahbe:
Adds some tests for diff functionality on dynamic providers.

### DIFF
--- a/pf/internal/pfutils/proposed_new.go
+++ b/pf/internal/pfutils/proposed_new.go
@@ -79,13 +79,6 @@ func ProposedNew(
 			return configValue, nil
 		case pathToRoot:
 			if configValue != nil && !configValue.IsNull() && configValue.IsKnown() {
-				// Here priorStateValue must be unknown. If it was an object, Reconcile would not be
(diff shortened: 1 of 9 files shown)


pf/internal/pfutils/proposed_new.go
```diff
@@ -79,13 +79,6 @@ func ProposedNew(
 			return configValue, nil
 		case pathToRoot:
 			if configValue != nil && !configValue.IsNull() && configValue.IsKnown() {
-				// Here priorStateValue must be unknown. If it was an object, Reconcile would not be
-				// called. If it was null, it got substituted via newObjectWithDefaults.
-				//
-				// priorStateValue may be null if it is a scalar from a
-				// null value. For example, during a create.
-				contract.Assertf(!priorStateValue.IsKnown() || priorStateValue.IsNull(),
-					"priorStateValue must be unknown or empty")
```
iwahbe, line 88: It's not clear why this assert was present in the first place, and `ProposedNew` seems to be fine hitting this case.

guineveresaenger (approved):

Issue #1988, opened 2024-05-17
Panic when working with `types.Dynamic`


dirien:
### What happened?

The newest version of the azapi provider introduced this TF datatype called:

```yaml
Body                  types.Dynamic  `tfsdk:"body"`
```

This leads during the bridging to this error panic:

```bash
panic: fatal: A failure has occurred: impossible: sch.Type() should be one of TypeMap, TypeList, TypeSet at this point path: resource[key="azapi_data_plane_resource",token="azapi:index/dataPlaneResource:DataPlaneResource"].outputs.body, type: 

goroutine 1 [running]:
github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.failfast(...)
        /Users/dirien/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.114.0/go/common/util/contract/failfast.go:23
github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.Failf({0x101c39532?, 0x6e?}, {0x14000c4f008?, 0x14000693b84?, 0x9d?})
        /Users/dirien/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.114.0/go/common/util/contract/fail.go:32 +0xbc
github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen.(*Generator).makePropertyType(0x1400074a008, {0x10f9cb9c8, 0x140007c3f50}, {0x101b4d79d, 0x4}, {0x10f9effe0, 0x14000834480}, 0x0, 0x1, {{0x14000ac2000, ...}, ...})
        /Users/dirien/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.82.0/pkg/tfgen/generate.go:427 +0x4b8
github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen.(*Generator).propertyVariable(0x1400074a008, {0x10f9cb908, 0x140008e9db0}, {0x101b4d79d, 0x4}, {0x10f9ddf50, 0x140008e9dc0}, 0x14000bb0000, {0x140009bab60, 0x61}, ...)
        /Users/dirien/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.82.0/pkg/tfgen/generate.go:1687 +0x290
[...]

PR #2110, opened 2024-06-21
Error for dynamic types in makePropertyType

guineveresaenger:
Rather than panic, bubble up an error from makePropertyType when user attempts to make a dynamic property type.

Handles panic reported in #1988. Note that it does not implement the missing logic.

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -353,7 +353,7 @@ func (g *Generator) Sink() diag.Sink {
 
 func (g *Generator) makePropertyType(typePath paths.TypePath,
 	objectName string, sch shim.Schema, info *tfbridge.SchemaInfo, out bool,
-	entityDocs entityDocs) *propertyType {
+	entityDocs entityDocs) (*propertyType, error) {
 
 	t := &propertyType{}
 
@@ -368,7 +368,7 @@ func (g *Generator) makePropertyType(typePath paths.TypePath,
 
 	if sch == nil {
 		contract.Assertf(info != nil, "missing info when sch is nil on type: "+typePath.String())
-		return t
+		return t, nil
 	}
 
 	// We should carry across any of the deprecation messages, to Pulumi, as per Terraform schema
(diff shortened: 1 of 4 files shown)


pkg/tfgen/generate_test.go
```diff
@@ -164,7 +164,8 @@ func Test_makePropertyType(t *testing.T) {
 	}).Shim()
 
 	t.Run("String", func(t *testing.T) {
-		p := g.makePropertyType(path, "obj", strType, nil, false, entityDocs{})
+		p, err := g.makePropertyType(path, "obj", strType, nil, false, entityDocs{})
+		assert.NoError(t, err)
```
iwahbe, line 168: Nit: If we use an assert here and `makePropertyType` returns an error, it will show an error for L168 and L169. We don't need to show the error for L169, so we should use `require`, which stops the test on error.
```suggestion
		require.NoError(t, err)
```

pkg/tfgen/generate.go
```diff
@@ -433,13 +441,15 @@ func (g *Generator) makePropertyType(typePath paths.TypePath,
 		t.kind = kindList
 	case shim.TypeSet:
 		t.kind = kindSet
+	case shim.TypeDynamic:
+		return nil, errors.New("Error in schema generation: Dynamic types are not implemented")
 	default:
 		contract.Failf(
 			"impossible: sch.Type() should be one of TypeMap, TypeList, TypeSet at this point path: %s, type: %s",
```
iwahbe, line 448:
```suggestion
			"impossible: sch.Type() should be one of TypeMap, TypeList, TypeSet or TypeDynamic at this point path: %s, type: %s",
```

iwahbe (changes requested):
Let's make sure we add a test here. I don't think the change as implemented will actually fix https://github.com/pulumi/pulumi-terraform-bridge/issues/1988, since PF doesn't ever generate a `shim.TypeDynamic`. We need to add that logic too.

iwahbe (approved):
This looks good to me.

PR #2115, opened 2024-06-22
Add InputTypes flag to Python codegen information

julienp:
This flag was added to Pulumi Python codegen in
https://github.com/pulumi/pulumi/pull/15957 and enables the generation
of TypedDict based input types. This is currently opt in and defaults to
`classes`, leaving codegen unchanged. A future version of Python codegen
will flip this to `classes-and-dicts` to generate TypedDict types side
by side with Args classes.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -175,6 +175,15 @@ type OverlayInfo = info.Overlay
 // JavaScriptInfo contains optional overlay information for Python code-generation.
 type JavaScriptInfo = info.JavaScript
 
+const (
+	// Use the default type generation from pulumi/pulumi.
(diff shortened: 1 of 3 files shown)


iwahbe (commented):
Could we do an enum here?

```go
type PythonInputType string
const (
	PythonInputTypeDefault = ""
        PythonInputTypeClasses = "classes"
        PythonInputTypeClassesAndDicts = "classes-and-dicts"
)

julienp:
> Could we do an enum here?

Definitely, added here https://github.com/pulumi/pulumi-terraform-bridge/pull/2115/commits/109b0584329fd0448e33505bb43f2d8dc8f7d3b5

VenelinMartinov (approved):
lgtm

PR #2134, opened 2024-06-26
WIP: Detect recursive type unrolling

t0yv0:


pkg/tfgen/internal/unrec/recursion_detector.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+// Detects unrolled recursion.
+//
+// TF does not support recursive type definitions but Pulumi does. The idea here is to take a projection of TF recursion
+// unrolling trick and fold it back.
+//
+// Suppose ancestor(T1, T2) if T2 describes a sub-property of T1.
+//
+// Suppose T1 <= T2 if T1 has the subset of properties of T2 with types that are themselves Tx <= Ty.
```
t0yv0, line 31: This seems to work really well for the AWS WAFv2 example at least.

pkg/tfgen/internal/unrec/testdata/TestSimplify.golden
```diff
@@ -0,0 +1,117 @@
...
+        "fieldToMatch": {
+          "$ref": "#/types/myprov:index/WebAclStatementAndStatementStatementXssMatchStatementFieldToMatch:WebAclStatementAndStatementStatementXssMatchStatementFieldToMatch"
+        }
+      },
+      "type": "object"
+    },
+    "myprov:index/WebAclStatementAndStatementStatementXssMatchStatementFieldToMatch:WebAclStatementAndStatementStatementXssMatchStatementFieldToMatch": {
+      "properties": {
+        "method": {
+          "$ref": "#/types/myprov:index/WebAclStatementAndStatement:WebAclStatementAndStatement"
```
t0yv0, line 68: This isn't right, because it used to be an empty object. Need to correct handling of empty object types that trivially satisfy `A <= B`.

pkg/tfgen/internal/unrec/gen.go
```diff
@@ -0,0 +1,189 @@
...
+					Type:     schema.TypeString,
+					Optional: true,
+				},
+				"scope_down_statement": scopeDownStatementSchema(level - 1),
+			},
+		},
+	}
+}
+
+func scopeDownStatementSchema(level int) *schema.Schema {
```
t0yv0, line 91: This looks too similar to statement(), might need to add some variety there

pkg/tfgen/internal/unrec/comparer.go
```diff
@@ -0,0 +1,308 @@
...
+	if !cmp.EqualPropertySpecs(&a.PropertySpec, &b.PropertySpec) {
+		return false
+	}
+	return true
+}
+
+func (cmp *generalizedComparer) EqualObjectTypeSpecs(a, b pschema.ObjectTypeSpec) bool {
+	if a.Type != b.Type {
+		return false
+	}
```
iwahbe, line 127: Why don't we delegate to `cmp.EqualTypeSpecs` here?
```suggestion
	if !cmp.EqualTypeSpecs(a.Type, b.Type) {
		return false
	}
```

pkg/tfgen/internal/unrec/equality_classes.go
```diff
@@ -0,0 +1,68 @@
...
+			if sameEqualityClasses(eq, ac, yc) {
+				acc[i] = append(append([]T{}, ac...), yc...)
+				matched = true
+				break
+			}
+		}
+		if !matched {
+			acc = append(acc, yc)
+		}
+	}
```
iwahbe, line 66: I think this would be clearer with a label:
```suggestion
outer:
	for _, yc := range ys {
		for i, ac := range acc {
			if sameEqualityClasses(eq, ac, yc) {
				acc[i] = append(append([]T{}, ac...), yc...)
				continue outer
			}
		}
		acc = append(acc, yc)
	}
```

pkg/tfgen/internal/unrec/example_schema_test.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"encoding/json"
+	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+)
+
+//go:generate go run gen.go
```
iwahbe, line 26: This feels like a complicated way to do this for a test. Could we get the same effect by implementing `TestMain`?

pkg/tfgen/internal/unrec/package_spec_util.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+)
+
+func parseLocalRef(rawRef string) (tokens.Type, bool) {
+	if !strings.HasPrefix(rawRef, "#/types/") {
+		return "", false
+	}
+	cleanRef := strings.TrimPrefix(rawRef, "#/types/")
+	return tokens.Type(cleanRef), true
```
iwahbe, line 32: Is there a reason not to validate here?
```suggestion
	typ, err := tokens.ParseTypeToken(cleanRef)
	contract.AssertNoErrorf(err, "invalid schema")
	return typ, true
```

pkg/tfgen/internal/unrec/package_spec_util.go
```diff
@@ -0,0 +1,221 @@
...
+	for k := range refs {
+		if best == "" || len(k) < len(best) {
+			best = k
+		}
+	}
+	contract.Assertf(best != "", "expected to find at least one element")
+	return best
+}
+
+func findGenericTypeReferences(x any) (typeRefs, error) {
```
iwahbe, line 96: It feels like this could be implemented via `type_visitor.go` and it would be much easer to reason about.

pkg/tfgen/internal/unrec/recursion_detector.go
```diff
@@ -0,0 +1,85 @@
...
+
+func newRecursionDetector(schema *pschema.PackageSpec) *recursionDetector {
+	return &recursionDetector{
+		schema:                 schema,
+		detectedRecursiveTypes: map[tokens.Type]struct{}{},
+		cmp:                    &comparer{schema: schema},
+	}
+}
+
+func (rd *recursionDetector) Detect(roots []tokens.Type) []tokens.Type {
```
iwahbe, line 53: Is `roots` a resource token or a type token?

iwahbe (commented):
I'm really exited about this. I'd love to see what the diff looks like for a couple of our smaller providers (does it detect recursion, and if so do we agree).

pkg/tfgen/internal/unrec/example_schema_test.go
t0yv0: This avoids build-depending on tfgen. Later on tfgen can use this package then. Avoid the issue with recursive packages. If that's not going to be an issue I can clean up.

iwahbe: That's reasonable. Then a comment explaining ☝️ would be great.

pkg/tfgen/internal/unrec/recursion_detector.go
t0yv0: Always a type token.

pkg/tfgen/internal/unrec/comparer.go
t0yv0: a.Type is a string like "object":

```
	// Type must be "object" if this is an object type, or the underlying type for an enum.
	Type string `json:"type,omitempty" yaml:"type,omitempty"`
```

but

```
EqualTypeSpecs(a, b *pschema.TypeSpec) bool
```

I think it should stay as written.

PR #2143, opened 2024-06-27
tweaks for clarity

guineveresaenger:
This PR is just a suggestion - it wasn't very clear to me what "parameterization" means in the context of dynamic providers. It felt like it was referring to a very specific concept, but it didn't really go into what that was. It feels like we could wait to talk about it until we do have that doc and concept laid out.

Feel free to ignore - these would've been my comments on #2137.


iwahbe (approved):
It's a bit cleaner. We can take it.

PR #2147, opened 2024-06-27
Remove extra schema information for dynamic providers

iwahbe:


dynamic/main.go
```diff
@@ -44,6 +45,11 @@ func initialSetup() (tfbridge.ProviderInfo, pfbridge.ProviderMetadata, func() er
 		Version:      version.Version(),
 		Description:  "Use any Terraform provider with Pulumi",
 		MetadataInfo: &tfbridge.MetadataInfo{Path: "", Data: tfbridge.ProviderMetadata(nil)},
+		SchemaPostProcessor: func(spec *schema.PackageSpec) {
+			spec.Attribution = ""
+			spec.Provider = schema.ResourceSpec{}
```
t0yv0, line 50: I don't quite recall what these configure, perhaps a quick comment why this is not relevant for the dynamic provider case?

t0yv0 (approved):

PR #2151, opened 2024-06-29
feat: optimization of nested description parsing

### LINKED ISSUE #1886: Refine description matching when a description is unique
As part of #1789, which requires us to remove the [docs matching hack](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfgen/generate.go#L1849), we can gain extra correct descriptions for providers with difficult-to-parse documentation (such as AWS) by still matching docs path leaves to key path leaves, but only as long as those leaf keys have unique names in the resource.

guineveresaenger:
This pull request adds additional docs parsing capability for type descriptions.

Prior to this change, we match the type object key, with the the best-attempt docs paths for nested schemas from the docs parser Arguments, via GetNestedDescriptionsFromParsedDocs. If we do not find any Description via that, then we attempt to obtain a Description from the schema Description field.

With this change, when we have an empty Description, we look again at the parsed entityDocs, and **if** the leaf of the type's docsPath is unique within the resource, then we match that Description to the Type.

We do this matching _after_ the schema description lookup to scope this optimization as tightly as possible - when, for instance, a field is entirely missing from the entityDocs.Attributes, we don't want to accidentally be matching a type description with the same docsPath leaf. The chance of this happening is greatly reduced when checking all other, reliable, sources first.

On AWS, this adds ~1400 descriptions to type fields:

```
git diff master --stat -- provider/cmd/pulumi-resource-aws/schema.json 
 provider/cmd/pulumi-resource-aws/schema.json | 4272 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---------------------------------
 1 file changed, 2809 insertions(+), 1463 deletions(-)
 ```

Fixes #1886.

### DIFF
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -1971,6 +1971,35 @@ func getNestedDescriptionFromParsedDocs(entityDocs entityDocs, path docsPath) (s
 		p = p.withOutRoot()
 	}
 
+	//TODO: Compare leaf path descriptions and set those IFF they are unique: https://github.com/pulumi/pulumi-terraform-bridge/issues/1886
+
+	//uniqueLeaves := make([]docsPath, 0, len(entityDocs.Arguments/2)
+	leaf := path.leaf()
+
+	// Get all leaves
+	var allLeaves []string
+	uniqueLeaves := make(map[string]bool)
+	for arg := range entityDocs.Arguments {
+		allLeaves = append(allLeaves, arg.leaf())
+		_, exist := uniqueLeaves[arg.leaf()]
+		if !exist {
+			uniqueLeaves[arg.leaf()] = true
+		}
+	}
+
+	//``
+
+	// Now we have to check if our current leaf is in our unique map.
+	// If yes, then we can grab our description and assign it.
+
+	if _, exist := uniqueLeaves[leaf]; exist {
+		if val, ok := entityDocs.Arguments[path]; ok {
+			return val.description, false
+		}
+	}
+
+	// Determine if they're unique
+
 	// To maintain old behavior, we also check if the last segment of `path` matches
 	// with some other last segment of any other entity.
 	//
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -1949,42 +1949,29 @@ func emitFile(fs afero.Fs, relPath string, contents []byte) error {
 	return err
 }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: pkg/tfgen/generate.go [hunk 1]
```diff
@@ -1943,6 +1949,44 @@ func emitFile(fs afero.Fs, relPath string, contents []byte) error {
...
+	leaf := path.leaf()
+
+	uniqueLeaves := getUniqueLeafPaths(arguments)
+	if _, exist := uniqueLeaves[leaf]; exist {
+		for argKey, argDoc := range arguments {
+			if argKey.leaf() == leaf {
+				return argDoc.description
+			}
+		}
+	}
```
iwahbe, line 1962: This feels a little funny to me. We want to check `path.leaf()` is unique in `arguments`. The way we do that here is to:
1. Build a list of unique arguments.
2. Check if `path.leaf()` is in that list.
3. If it is, iterate through `arguments` again to find the description.

I think we could do this in a faster and simpler way:
```suggestion
	leaf := path.leaf()
	var foundArgumentDocs *argumentDocs
	for argKey, argDoc := range arguments {
		if argKey.leaf() != leaf {
			continue
		}
		if foundArgumentDocs != nil {
			// path.leaf() is not a unique leaf in arguments
			return ""
		}
		foundArgumentDocs = argDoc
	}
	// Could not find an argument suffixed by path.leaf()
	if foundArgumentDocs == nil {
		return ""
	}
	return foundArgumentDocs.description
```

We then wouldn't need `getUniqueLeafPaths` at all.

pkg/tfgen/generate.go
```diff
@@ -1943,6 +1949,44 @@ func emitFile(fs afero.Fs, relPath string, contents []byte) error {
...
+		if exist {
+			leaves[arg.leaf()]++
+			continue
+		}
+		leaves[arg.leaf()] = 1
+	}
+
+	// Only return leaves with a count of exactly 1.
+	// For quick lookup, this is a map where the key is our leaf, and the value a boolean, which we disregard.
+	uniqueLeavesMap := make(map[string]bool)
```
iwahbe, line 1981: If we disregard the boolean, then we should return a simpler type.
```suggestion
	// For quick lookup, this is a map where the key is our leaf, and the value a boolean, which we disregard.
	uniqueLeavesMap := make(map[string]struct{})
```

iwahbe (approved):
The change looks great. I left comments on how the implementation could be simplified.

PR #2184, opened 2024-07-15
Remove usage of plugin framework types in favor of interface types

### LINKED ISSUE #2119: Remove instances of of `types.*` from PF
We should remove all instances of `types.*` from the PF in favor of `basetypes.*`. `types` is the plugin frameworks pre-made version of a `List`, `Set`, `Map`, `Object`, ... while `basetypes.*` describes the interface that `types.*` & *custom types* are implemented against. To fully support custom types, we must remove all instances of `types.*`.

Related to https://github.com/pulumi/pulumi-terraform-bridge/issues/2020

guineveresaenger:
This is mostly a migration/refactor from `github.com/hashicorp/terraform-plugin-framework/types/` to `github.com/hashicorp/terraform-plugin-framework/types/basetypes` library. The main functional change is in type_schema.go's Elem() function.

Resolves #2119.

### DIFF
--- a/pf/internal/schemashim/convert_type.go
+++ b/pf/internal/schemashim/convert_type.go
@@ -19,7 +19,7 @@ import (
 	"fmt"
 
 	pfattr "github.com/hashicorp/terraform-plugin-framework/attr"
-	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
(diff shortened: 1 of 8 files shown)


iwahbe (approved):
We probably want to keep the `types.*` version in `dynamic/test/pfprovider` and `internal/testprovider`, since that is what users will write. Otherwise LGTM

PR #2187, opened 2024-07-16
Adjust documentation for Elem()

t0yv0:
Toward fixing #1828

Document current findings and add tests for PF representations.

### DIFF
--- a/pf/tests/repr_test.go
+++ b/pf/tests/repr_test.go
@@ -1,0 +1,177 @@
+// Copyright 2016-2024, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


pf/tests/repr_test.go
```diff
@@ -0,0 +1,211 @@
...
+								},
+							},
+						},
+					},
+				}},
+			},
+			autogold.Expect(`{
+  "resources": {
+    "_": {
+      "object_attribute": {
```
t0yv0, line 177: Interestingly enough this may look like a single-nested block to shim.Schema; it doesn't carry the distinction is it an attr or block in shim.Schema. May be allright for now but can cause some trouble in the corner cases perhaps.

pkg/tfshim/shim.go
iwahbe, line 1: It's probably worth calling out that `repr_test.go` can be used to show how various systems are represented in `shim`.

iwahbe (approved):

Issue #2149, opened 2024-06-28
Parameterized providers: support SDK parameterization


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Depends on https://github.com/pulumi/pulumi/pull/16491.

We need to update [`parseParameterizeParameters`](https://github.com/pulumi/pulumi-terraform-bridge/blob/1994ac18f985a74ed6c212414e749c8e879d20c0/dynamic/info.go#L71-L72) to accept `plugin.ParameterizeValue` based parameters.

I expect that we can just unmarshal the value into [`paramaterizeArgs`](https://github.com/pulumi/pulumi-terraform-bridge/blob/1994ac18f985a74ed6c212414e749c8e879d20c0/dynamic/info.go#L65).

Similarly, we will need to augment our [`XGetSchema` function](https://github.com/pulumi/pulumi-terraform-bridge/blob/1994ac18f985a74ed6c212414e749c8e879d20c0/dynamic/main.go#L51) to embed the appropriate parameters into the schema generated for a TF provider.[^1]

[^1]: We should embed `terraform-provider` as the base name, not `terraform-bridge`.

PR #2202, opened 2024-07-19
Fill in Parameterization Value

iwahbe:
This commit adds support for the new `.Parameterization` field on the schema. As part of
doing so, I have moved parsing parameterization into it's own package.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2149

### DIFF
--- a/dynamic/parameterize/value.go
+++ b/dynamic/parameterize/value.go
@@ -1,0 +1,106 @@
+// Copyright 2016-2024, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


dynamic/parameterize/value.go
```diff
@@ -0,0 +1,106 @@
...
+func ParseValue(b []byte) (Value, error) {
+	var value Value
+	err := value.Unmarshal(b)
+	return value, err
+}
+
+// IntoArgs converts a [Value] into an [Args].
+//
+// We can do this because [Value] is a fully resolved [Args], and so it is always possible to
+// go from [Value] to [Args].
```
mjeffryes, line 95: nit. I assume the reverse is true as well and we should probably have testing that validates the round trip

mjeffryes (approved):
Could you update the description on the PR to actually describe the change before committing so we get a more useful commit message in the git history?

dynamic/parameterize/value.go
mjeffryes: ah thanks for clarifying. (Sounds like there's still a possible round trip test, but only values -> args -> values, not args -> values -> args)

PR #2214, opened 2024-07-20
Translate installation examples

### LINKED ISSUE #2212: Convert installation examples from upstream index.md
Configuration examples should render as Pulumi examples, including:

- Code choosers
- Pulumi.yaml config

### LINKED ISSUE #2213: Configuration page: Render field names in camelCase
Pulumi config fields are camelCase. When we translate the upstream `index.md` file, we should translate all field names to camelCase.

guineveresaenger:
This pull request iterates over the upstream configuration page, finds all HCL code blocks, and translates the examples therein. Because this is the configuration page, there is added functionality to translate the `Pulumi.yaml` config file as well, showing the configuration file for a default provider.

Also translates all field names to `camelCase`. _Technically_, the translation mechanism uses `reformatText` with `language: "nodejs"` since to the best of my knowledge that is equivalent to expected config field syntax.

Fixes #2212 
Fixes #2213


pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+		if nextNewLine == -1 {
+			//Write the inline block
+			returnContent = returnContent + contentStr[block.start:block.end] + codeFence + "\n"
+			continue
+		}
+		fenceLanguage := contentStr[block.start : block.start+nextNewLine+1]
+		code := contentStr[block.start+nextNewLine+1 : block.end]
+		// Only convert code blocks that we have reasonable suspicion of actually being Terraform.
+		if fenceLanguage == "```terraform\n" || fenceLanguage == "```hcl\n" ||
+			(fenceLanguage == "```\n" && guessIsHCL(code)) {
```
iwahbe, line 219: This tripple must be duplicated in other places. Can we pull it out into a function?

```go
func isHCL(code string) bool {
    return strings.HasPrefix(code, "```terraform\n") || 
        strings.HasPrefix(code, "```hcl\n") || guessIsHCL(code)
}

File: pkg/tfgen/installation_docs.go [hunk 2]
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+
+			//Generate the main.tf file for converting the config file, since --convert-examples doesn't support
```
iwahbe, line 221:
```suggestion
			// Generate the main.tf file for converting the config file, since --convert-examples doesn't support
```

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+			// the Pulumi.yaml config file
+			err := os.WriteFile(filepath.Join(outDir, "main.tf"), []byte(code), 0600)
```
iwahbe, line 223: Should this be a `tmp` file? If not, should there be a `defer` block to remove it? I don't think we want to leave a `.tf` file in our output tree.

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+			}
+
+			// Prepare examples
+			// Make an example to record in the cliConverter.
+			fileName := fmt.Sprintf("configuration-installation-%d", i)
+			examples[fileName] = code
+
+			// Convert to PCL with convertViaPulumiCLI.
+			// This gives us a map of translatedExamples.
+			translatedExamples, err := g.cliConverter().convertViaPulumiCLI(examples, []tfbridge.ProviderInfo{
```
iwahbe, line 235: Doesn't `examples` grow with each iteration of the loop? Do we want/need it to?

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+			g.cliConverter().pcls[code] = translatedExample{
+				PCL: translatedExamples[fileName].PCL,
+			}
+
+			exPath := examplePath{fullPath: fileName}
+			conversionResult := g.coverageTracker.getOrCreateExample(
+				exPath.String(), code)
+
+			langs := genLanguageToSlice(g.language)
+			chooserStart := `{{< chooser language "typescript,python,go,csharp,java,yaml" >}}` + "\n"
```
iwahbe, line 252:
```suggestion
			const chooserStart = `{{< chooser language "typescript,python,go,csharp,java,yaml" >}}` + "\n"
```

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+			returnContent = returnContent + chooserStart
+			chooserEnd := "{{< /chooser >}}\n"
```
iwahbe, line 254:
```suggestion
			const chooserEnd = "{{< /chooser >}}\n"
```

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+
+			// Generate each language in turn, so that we can mark up the output with the correct Hugo shortcodes.
+			// TODO: we want to use pulumi code choosers in the future, but resourcedocsgen does not support this yet.
```
iwahbe, line 257: Please make sure this `TODO` links to an issue in the comment.

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+				choosableStart := fmt.Sprintf("{{%% choosable language %s %%}}\n", lang)
+				choosableEnd := "\n{{% /choosable %}}\n"
+
+				// Generate the Pulumi.yaml config file for each language
+				configFile, err := writeConfigYamlViaCLI(outDir, lang, mappingsArgs)
+				if err != nil {
+					return "", err
+				}
+				// Generate example itself
+				convertedLang, err := g.convertHCL(conversionResult, code, exPath.String(), langSlice)
```
iwahbe, line 269: This is easier to read if we just inline `langSlice`.
```suggestion
				convertedLang, err := g.convertHCL(conversionResult, code, exPath.String(), []string{lang})
```

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+				if err != nil {
+					return "", err
+				}
+				returnContent = returnContent + choosableStart + configFile + convertedLang + choosableEnd
+			}
+			returnContent += chooserEnd
+			startIndex = block.end + len(codeFence)
+		} else {
```
iwahbe, line 277: This block is so long that I would normally like to see a comment explaining what the `else` is referring to. Instead, I would like this function to be broken up into understandable chunks.

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+	startIndex := 0
+	outDir, err := os.MkdirTemp("", "configuration-example")
+	if err != nil {
+		return "", err
+	}
+	// In order to translate the examples with the currently-being-generated version of our provider, we will obtain
+	// mappings from the cli converter.
+	mappingsDir := filepath.Join(outDir, "mappings")
+	mappings := []tfbridge.ProviderInfo{
+		g.cliConverter().info,
```
t0yv0, line 175: g.cliConverter() is stateful and is coupled to a specific sequence of Start/Finish converting examples in the schema, which is not what this is doing. I'm worried about reusing the same instance.

pkg/tfgen/installation_docs_test.go
```diff
@@ -184,3 +189,83 @@ func TestApplyEditRules(t *testing.T) {
...
...    type: simple:resource\n    name: a_resource\n    properties:\n      renamedInput1: hello\n      inputTwo: true\noutputs:\n  someOutput: ${aResource.result}\n```\n{{% /choosable %}}\n{{% choosable language java %}}\n```yaml\n# Pulumi.yaml provider configuration file\nname: configuration-example\nruntime: java\n\n```\n```java\npackage generated_program;\n\nimport com.pulumi.Context;\nimport com.pulumi.Pulumi;\nimport com.pulumi.core.Output;\nimport com.pulumi.simple.resource;\nimport com.pulumi.simple.ResourceArgs;\nimport java.util.List;\nimport java.util.ArrayList;\nimport java.util.Map;\nimport java.io.File;\nimport java.nio.file.Files;\nimport java.nio.file.Paths;\n\npublic class App {\n    public static void main(String[] args) {\n        Pulumi.run(App::stack);\n    }\n\n    public static void stack(Context ctx) {\n        var aResource = new Resource(\"aResource\", ResourceArgs.builder()\n            .renamedInput1(\"hello\")\n            .inputTwo(true)\n            .build());\n\n        ctx.export(\"someOutput\", aResource.result());\n    }\n}\n```\n{{% /choosable %}}\n{{< /chooser >}}\n\n\n## Configuration Reference\n\nThe following configuration inputs are supported:",
```
t0yv0, line 247: Could we use hexops/autogold to make this multiline and readable?

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+	}
+	// Remove the temp directory's random number from project name
+	regex := regexp.MustCompile(`configuration-example.*`)
+	pulumiYAMLFileBytes = regex.ReplaceAll(pulumiYAMLFileBytes, []byte("configuration-example"))
+
+	// Add descriptive code comment
+	configFile := "```yaml\n" +
+		"# Pulumi.yaml provider configuration file\n" +
+		string(pulumiYAMLFileBytes) + "\n```\n"
+	return configFile, nil
```
t0yv0, line 334: Are these configuration files ever different from non-primitive files? Can we just embed Pulumi.yaml per supported language? They look really primitive and there's not a lot going on in the generator in https://github.com/pulumi/pulumi-converter-terraform/blob/7879de9e9ac884ea9575d8c47866bcb2d6c26161/pkg/convert/tf.go#L2898

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
 	}
 	return contentBytes, nil
 }
+
+func translateCodeBlocks(contentStr string, g *Generator) (string, error) {
```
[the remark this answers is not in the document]
t0yv0: QQ. How is this function different from convertExamples?

```
func (g *Generator) convertExamples(docs string, path examplePath) string {
```

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +144,192 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("couldn't find pulumi path")
+	}
+
+	cmdArgs := []string{
+		"convert",
+		"--from",
+		"terraform",
+		"--language",
+		lang,
```
t0yv0, line 306: Oof. You're converting from HCL to the target language directly via `pulumi convert`. We don't support this elsewhere in the bridge yet. Instead, we convert `TF->PCL` via `pulumi convert` and then we convert in-memory from PCL to JavaScript etc.

t0yv0:
The other thing to worry about is pinning - we need to make sure this under `pulumi convert enabled` flag, and that providers builds are pinned for deterministic builds.

t0yv0:
Reminder to self: write/find tickets on finishing transitioning the PCL->{language} translation bits to `pulumi convert`.

t0yv0:
e.g. https://github.com/pulumi/pulumi-terraform-bridge/blob/t0yv0%2Fdocument-sets/pkg/tfgen/convert_cli.go#L517

t0yv0:
Filed https://github.com/pulumi/pulumi-terraform-bridge/issues/2241 :) FYI.

pkg/tfgen/installation_docs.go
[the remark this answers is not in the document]
guineveresaenger: > How is this function different from convertExamples?
convertExamples is meant to be called on the schema, no? These files are not written to the schema at all. There is no `examplePath`.

> Do we need to take the entire *Generator, or can we restrict ourselves to only taking a couple of fields or functions?

I'm under the impression that we need the mappings file from the cliConverter in order to translate any examples with the current provider state; also we'll need the generator languages.

t0yv0:
LMK when ready for another review 🙏

pkg/tfgen/installation_docs.go
guineveresaenger: I think this is okay, because `registry-docs` is a separate flag of `tfgen`. We wouldn't generate registry docs and resource docs in the same call.

pkg/tfgen/installation_docs.go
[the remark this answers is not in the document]
guineveresaenger: I broke up this function and I left a long comment as to why this PR choses to use the cliConverter approach. It _seems_ possible to use regular `pulumi convert` as well here - but especially if we want to refactor the way API docs are rendered as well, then having a way to bulk convert/lookup will be necessary anyhow.

pkg/tfgen/installation_docs_test.go
guineveresaenger: Unfortunately, autogold flails here too, because the string in question contains back ticks. I've created files the represent the input and expected transformation instead for improved readability.

pkg/tfgen/installation_docs.go
```diff
@@ -124,3 +139,132 @@ func applyEditRules(contentBytes []byte, docFile *DocFile) ([]byte, error) {
...
+	// Add descriptive code comment
+	pulumiYAMLFile = "```yaml\n" +
+		"# Pulumi.yaml provider configuration file\n" +
+		pulumiYAMLFile + "\n```\n"
+	return pulumiYAMLFile
+}
+
+func convertExample(g *Generator, code string, exampleNumber int) (string, error) {
+	// Make an example to record in the cliConverter.
+	converter := g.cliConverter()
```
t0yv0, line 218: g.cliConvereter() is caching :

```
	if g.cliConverterState != nil {
		return g.cliConverterState
	}
```

You want a non-caching version here (to add one) to rule out any stateful interference that'd be nice.

t0yv0:
Suggested edits https://github.com/pulumi/pulumi-terraform-bridge/pull/2287 to make interfacing cliConverter easier. Sorry it's a bit half-baked as I've not the time to pass the tests. WDYT? 

I think the other bit here is making sure that when cliConverter is not enabled/available, we fall back on something reasonable like the previous behavior. 

Besides 2287 it's looking good to me, there's some code duplication around parsing codeblocks but it's really tedious to factor out these bits from existing code so I'm ok with a little bit of duplication temporarily to unblock the feature.

guineveresaenger:
@t0yv0 and I discussed and clarified that the cliConverter is invoked by a different entrypoint to the tfgen cli, using
`./bin/pulumi-tfgen-<provider> registry-docs --out docs/`, therefore will not be creating unstable builds.

Nonetheless I incorporated some of the cleanups from #2287 to make for a smoother, more straightforward read. This does drop the conversion stats for the installation docs file, but I think that is acceptable given we should have quick insight into conversion failures.

t0yv0 (approved):
Thank you! Indeed, given a new entry-point I was missing concerns around accidental state sharing or assuming Pulumi CLI are moot. 🚢

t0yv0:
Besides a smoother read you got 2x reduction in the number of times it shells out to Pulumi CLI since last version, I think, but that's immaterial in the grand scheme of things since there are not a lot of examples here. Let's ship this.

PR #2229, opened 2024-07-22
Handle legacy suffix in rawname when obtaining possible docs sources

### LINKED ISSUE #2228: RenameResourceWithAlias and RenameDataSource should support modern upstream docs structure
The update for https://github.com/pulumi/pulumi-mongodbatlas/issues/536 fails `tfgen` with the message 

```
error: failed to gather package metadata: problem gathering data sources: 1 error occurred:
	* could not find docs for data source 'mongodbatlas_x509_authentication_database_user' (aliased or renamed). Override the Docs property in the data source mapping. See type tfbridge.DocInfo for details.
```

But the [upstream doc exists](https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/docs/data-sources/x509_authentication_database_user.md), and the `renames` section in `bridge-metadata.json` [was deprecated](https://github.com/pulumi/pulumi-terraform-bridge/pull/1698). Edited to clarify: this was meant to say the bridge metadata's `renames` section does not affect docs discovery.

But then I looked at the release notes and noticed [upstream migrated to new docs structure](https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2425).

It turns out that we [do not support the new docs formate](https://github.com/pulumi/pulumi-terraform-bridge/blob/18db7c57211af0eb7dfa26a70b145edc4d33a88f/pkg/tfbridge/info/rename.go#L90) when finding assigning the docs source for legacy docs.

So what happened with pulumi-mongodbatlas was the following:
1. Upstream renamed docs to end in `.md`
2. Data source `mongodbatlas_x509_authentication_database_user` had been renamed on our end (see entry in bridge-metadata.json)
3. Renames only look for docs sources ending in `.html.markdown` if docs source is nil
4. => Bridge cannot find docs for renamed resources that have docs ending in `.md`.

The workaround for the upgrade is to explicitly name the docs source. However, this means a manual entry in `resources.go` and it also means manual intervention in the upgrade itself. The bridge should handle both docs formats.

guineveresaenger:
Fixes #2228.

Removes the potentially incorrect setting of `DocInfo.Source` in `renames.go` and instead trims the `RenamedEntitySuffix` before generating possible docs names in `getMarkdownNames`.

edit: interesting test failure; this should not affect token aliasing itself. 

edit2: because `renames.go` no longer sets a possible docs source, the test in `tokens_test` had to be adjusted. It now shows that the bridge honors provided docs sources as well as default ones for renamed resources.

### DIFF
--- a/pkg/tfbridge/tokens_test.go
+++ b/pkg/tfbridge/tokens_test.go
@@ -17,6 +17,7 @@ package tfbridge_test
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/info"
 	"testing"
 
 	"github.com/hexops/autogold/v2"
@@ -507,6 +508,13 @@ func testTokenAliasing(t *testing.T) {
 					"pkg_mod1_d1": nil,
 				},
 			}).Shim(),
+			Resources: map[string]*info.Resource{
+				"pkg_mod1_r1": {
+					Docs: &tfbridge.DocInfo{
+						Source: "manually_provided_docs_source.md",
+					},
+				},
+			},
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/tfbridge/tokens_test.go
```diff
@@ -557,7 +569,7 @@ func testTokenAliasing(t *testing.T) {
 		"pkg_mod1_r2_legacy": {
 			Tok:                "pkg:index/mod1R2:Mod1R2",
 			DeprecationMessage: "pkg.index/mod1r2.Mod1R2 has been deprecated in favor of pkg.mod1/r2.R2",
-			Docs:               &tfbridge.DocInfo{Source: "kg_mod1_r2.html.markdown"},
+			//Docs:               &tfbridge.DocInfo{Source: "kg_mod1_r2.html.markdown"},
```
iwahbe, line 572: Should `Docs:` be commented out?

iwahbe (approved):
I think this is good, but we need to remove the commented out code before we merge.

PR #2243, opened 2024-07-24
fix(pf): remove __defaults entry from inputs

### LINKED ISSUE #2189: Importing PluginFramework based resources generates extraneous __defaults

corymhall:
PF resources should not be injecting __defaults during
import as they do not support it, but it is happening through an
unintended side-effect of code reuse from sdkv2.

I could think of a couple other ways to handle this:

1. Duplicate the code in the pf and remove the code specific to
injecting __defaults
2. Removing where we inject an empty __defaults. See #2236

Decided to go with the simplest approach of just doing some post
processing since touching the `sdkv2` code seems to be a much bigger
effort.

fixes #2189, fixes pulumi/pulumi-aws#4204

### DIFF
--- a/pf/tfbridge/provider_read.go
+++ b/pf/tfbridge/provider_read.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/propertyvalue"
 )
 
 // Read the current live state associated with a resource. Enough state must be include in the inputs to uniquely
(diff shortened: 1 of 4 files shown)


pf/tfbridge/provider_read.go
```diff
@@ -90,12 +90,35 @@ func (p *provider) ReadWithContext(
...
+		deleteDefaultsKey(&result.Inputs)
 	}
 
 	return result, ignoredStatus, err
 }
 
+// deleteDefaultsKey removes the `__defaults: []` entry from all objects recursively
+// The `__defaults` key is something used in sdkv2 and is not handled here in pf. Because
+// of some code reuse between sdkv2 & pf the `__defaults` key is getting inserted
+func deleteDefaultsKey(inputs *resource.PropertyMap) {
```
t0yv0, line 102: Have a look at propertyvalue.TransformPropertyValue that also goes under secrets etc and handles the recursion hopefully properly without any further bugs. Might be easier to review, we're just transforming everything and we're dropping __defaults whenever we find them.

corymhall: Done!

pf/tfbridge/provider_read.go
```diff
@@ -90,12 +91,36 @@ func (p *provider) ReadWithContext(
...
+	i := *inputs
+	delete(i, "__defaults")
+	for _, key := range i.StableKeys() {
```
t0yv0, line 106: This is good. NIt: the for loop and top level delete may have been avoided by `propertyvalue.TransformPropertyValue` on `resource.NewObjectProperty(inputs)`. But this works too.

t0yv0 (approved):

iwahbe (approved):
1 nit, otherwise LGTM

pf/tfbridge/provider_read.go
corymhall: It looks like `TransformPropertyValue` won't  run the `transform` function on the top level value, only on each value in the map downward.

PR #2250, opened 2024-07-25
Improve the error messages for `checks.go`

iwahbe:
I discovered this when working on
https://github.com/pulumi/pulumi-terraform-bridge/pull/2245. This improves the quality of the error message for some cases, making it more actionable.

This PR also allows users to override the `Secret` value to remove the sensitive error. This is logically consistent with `.Type` overrides.

### DIFF
--- a/pf/internal/check/check_test.go
+++ b/pf/internal/check/check_test.go
@@ -36,7 +36,7 @@ import (
 
 func TestMissingIDProperty(t *testing.T) {
 	stderr, err := test(t, tfbridge.ProviderInfo{
-		P: pfbridge.ShimProvider(testProvider{}),
+		P: pfbridge.ShimProvider(testProvider{missingID: true}),
 	})
 
 	assert.Equal(t, "error: Resource test_res has a problem: no \"id\" attribute. "+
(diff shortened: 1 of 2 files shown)


pf/internal/check/check_test.go
```diff
@@ -90,14 +127,26 @@ func test(t *testing.T, info tfbridge.ProviderInfo) (string, error) {
 }
 
 type (
-	testProvider struct{ provider.Provider }
-	testResource struct{ resource.Resource }
+	testProvider struct {
+		provider.Provider
+
+		missingID, sensativeID bool
```
guineveresaenger, line 133:
```suggestion
		missingID, sensitiveID bool
```

pf/internal/check/checks.go
```diff
@@ -78,26 +75,79 @@ func checkIDProperties(sink diag.Sink, info tfbridge.ProviderInfo, isPFResource
...
+	}
+	return msg
+}
+
+type errSensativeID struct {
+	token string
+}
+
+func (err errSensativeID) Error() string {
+	msg := `"id" attribute is sensitive, but cannot be kept secret.`
```
guineveresaenger, line 96: Would it be worth it to say why it cannot be kept secret?

guineveresaenger (approved):
I don't generally block on spelling, but I'd really love it if all instances of "sensitive" were spelled the same. 🙏

pf/internal/check/check_test.go
iwahbe: > I don't generally block on spelling

Please do let me know if you spot a spelling mistake. I'll fix it before merging!

pf/internal/check/checks.go
iwahbe: The best explanation for why is https://github.com/pulumi/pulumi/pull/8948, but it's not a very good explanation. I'll open an issue.

iwahbe: https://github.com/pulumi/pulumi/issues/16818 is a slightly better explanation.

Issue #2227, opened 2024-07-22
Configuration page: Skip irrelevant sections


guineveresaenger:
Some upstream provider configuration pages include [terraform-specific logging instructions](https://registry.terraform.io/providers/PagerDuty/pagerduty/latest/docs#debugging-provider-output-using-logs) or even [contributing guidelines](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs#testing-and-development) and we do not want to include those.

We should do two things here:

1. Add default functionality to the `--registry-docs` flag that skips common sections (such as logging)
2. Allow for docEditRules to pass to the `plainDocsParser` that can take a section title and elide the section

PR #2292, opened 2024-08-09
Skip sections by header with goldmark

guineveresaenger:
This pull request introduces a Markdown transformer that can in conjunction with a parser/renderer remove unwanted sections from a markdown document by screening for words in the section header. This functionality is aimed at and implemented for the installation-configuration document, but it can be used as a docsEditRule by any provider on any file name. It is implemented as a docsEditRule in this pull request as well.

Note: A test was accidentally using writeFile instead of assert; this is fixed.
Additionally, the tests verify whether transformed results are equal in HTML, to account for parse idiosyncracies.

Resolves https://github.com/pulumi/pulumi-terraform-bridge/issues/2227.


pkg/tfgen/installation_docs.go
```diff
@@ -245,3 +250,106 @@ func convertExample(g *Generator, code string, exampleNumber int) (string, error
...
+	}
+}
+
+type shouldSkipHeaderFunc = func(headerText string) bool
+
+// SkipSectionByHeaderContent takes a document's content byte array and a callback function that contains logic on
+// whether a given header should be skipped.
+// It uses "github.com/yuin/goldmark" parser in conjunction with "github.com/teekennedy/goldmark-markdown" renderer
+// to walk the markdown document tree and apply the SectionSkipper's transformer
+func SkipSectionByHeaderContent(content []byte, shouldSkipHeader shouldSkipHeaderFunc) ([]byte, error) {
```
iwahbe, line 303: I don't love code comments that re-iterate how a function is implemented or what is in the function signature.

If you want to include general description of how a function works, I would move that to below the signature, since it is for people who are editing the function, not for function callers.

```suggestion
// SkipSectionByHeaderContent removes headers where shouldSkipHeader(header) returns true, along with any text under the header.
//
// content is assumed to be Github flavored markdown when parsing.
//
// shouldSkipHeader is called on the raw header text, like this:
//
//    shouldSkipHeader("My Header")
//
// *not* like this:
//
//    // This is wrong
//    shouldSkipHeader("## My Header\n")
func SkipSectionByHeaderContent(content []byte, shouldSkipHeader shouldSkipHeaderFunc) ([]byte, error) {
	// <I don't think this is necessary, since the code already says this>
	// It uses "github.com/yuin/goldmark" parser in conjunction with "github.com/teekennedy/goldmark-markdown" renderer
	// to walk the markdown document tree and apply the SectionSkipper's transformer
```

pkg/tfgen/installation_docs.go
```diff
@@ -245,3 +250,106 @@ func convertExample(g *Generator, code string, exampleNumber int) (string, error
...
+		goldmark.WithExtensions(extension.GFM),
+		goldmark.WithParserOptions(
+			parser.WithASTTransformers(
+				util.Prioritized(recognizeHeaderAfterHTML{}, 2000),
+				util.Prioritized(sectionSkipper, 500),
+			),
+		),
+		goldmark.WithRenderer(markdown.NewRenderer()),
+	)
+	buf := bytes.Buffer{}
```
iwahbe, line 318: Nit: When we are using a zero value, I prefer `var buf bytes.Buffer`. You are free to ignore me here.

pkg/tfgen/installation_docs.go
```diff
@@ -245,3 +250,106 @@ func convertExample(g *Generator, code string, exampleNumber int) (string, error
...
+	// Convert parses the source, applies transformers, and renders output to buf
+	err := gm.Convert(content, &buf)
+	if err != nil {
+		log.Fatalf("Encountered Markdown conversion error: %v", err)
+	}
+	return buf.Bytes(), nil
```
iwahbe, line 324: Why would we panic here when we can return the error?
```suggestion
	err := gm.Convert(content, &buf)
	return buf.Bytes(), err
```

pkg/tfgen/installation_docs.go
```diff
@@ -245,3 +250,106 @@ func convertExample(g *Generator, code string, exampleNumber int) (string, error
...
+}
+
+// Edit Rule for skipping headers.
+func skipSectionHeadersEdit(docFile *DocFile) tfbridge.DocsEdit {
+	defaultHeaderSkipRegexps := getDefaultHeadersToSkip()
+	return tfbridge.DocsEdit{
+		Path: docFile.FileName,
```
iwahbe, line 331: It feels weird to pass a `DocFile` (with `.Contents`) to produce a `skipSectionHeadersEdit`. If we already have `.Contents`, why wouldn't we just call transform the content here?
```suggestion
// Edit Rule for skipping headers.
func skipSectionHeadersEdit(filePath string) tfbridge.DocsEdit {
	defaultHeaderSkipRegexps := getDefaultHeadersToSkip()
	return tfbridge.DocsEdit{
		Path: filePath,
```

pkg/tfgen/installation_docs.go
```diff
@@ -245,3 +250,106 @@ func convertExample(g *Generator, code string, exampleNumber int) (string, error
 	exampleContent += chooserEnd
 	return exampleContent, nil
 }
+
+type SectionSkipper struct {
```
iwahbe, line 254: This should not be public.
```suggestion
type sectionSkipper struct {
```

pkg/tfgen/installation_docs.go
iwahbe, line 303: I don't see why an external user would want to call this function. Let's keep it private.

iwahbe (commented):
`github.com/teekennedy/goldmark-markdown` seems like a great find. If it works well, it looks like a viable path forward for docs generation.

pkg/tfgen/installation_docs.go
guineveresaenger: This is public on purpose - when a provider would manually want to skip a more specific header, you'd be able to use this function to more easily create a docsEdit, similar to skipSectionHeadersEdit.

guineveresaenger: I generally find callback functions hard to read, and since I do expect this function to be used externally as part of a DocsEdit rule, it would be helpful to include information about the callback specifics. Naming only goes so far.

iwahbe: We can make it more obvious that `shouldSkipHeaderFunc` is a callback by inlining the private type:

```patch
-func SkipSectionByHeaderContent(content []byte, shouldSkipHeader shouldSkipHeaderFunc) ([]byte, error) {
+func SkipSectionByHeaderContent(
+    content []byte, 
+    shouldSkipHeader func(header string) (shouldSkip bool),
+) ([]byte, error) {
```

We should probably do this anyway to avoid exposing a public function that takes a private type (and to minamise the pollution of the `tfgen` namespace.

---

I'm happy to include info on how the callback is used. My suggestion included:

```patch
+// content is assumed to be Github flavored markdown when parsing.
+//
+// shouldSkipHeader is called on the raw header text, like this:
+//
+//    shouldSkipHeader("My Header")
+//
+// *not* like this:
+//
+//    // This is wrong
+//    shouldSkipHeader("## My Header\n")
```

guineveresaenger: Does this look better to you?

pkg/tfgen/installation_docs.go
```diff
@@ -245,3 +250,111 @@ func convertExample(g *Generator, code string, exampleNumber int) (string, error
...
+// content is assumed to be Github flavored markdown when parsing.
+//
+// shouldSkipHeader is called on the raw header text, like this:
+//
+//	shouldSkipHeader("My Header")
+//
+// *not* like this:
+//
+//	// This is wrong
+//	shouldSkipHeader("## My Header\n")
```
danielrbradley, line 307: Perhaps a full example of calling this function would help demonstrate how to write the callback as well as indicate how the header is matched ...

```suggestion
// SkipSectionByHeaderContent removes headers where shouldSkipHeader(header) returns true,
// along with any text under the header.
// content is assumed to be Github flavored markdown when parsing.
//
// shouldSkipHeader is called on the parsed header text.
// 
// Example of removing the first header and it's context
//
//	result := SkipSectionByHeaderContent(byte[](`
//	## First Header
//	
//	First content
//	
//	## Second Header
//	
//	Second content
//	`, func(headerText string) bool) ([]byte, error) {
//		return headerText == "First Header"
//	})
// 
// The result will only now contain the second result:
//	## Second Header
//	
//	Second content
```

PR #2306, opened 2024-08-12
Add a goldmark parser extension for first class sections

iwahbe:
Adds section as a first class node kind in goldmark.

The test shows how sections are useful for manipulating the AST, since they encapsulate a header and it's children.


guineveresaenger (approved):
This allows us to clean up the logic for skipSectionByHeader.

PR #2310, opened 2024-08-13
Fix typechecking booleans

t0yv0:
The type-checking pass had a bug where it failed to flag passing a string to a slot of type boolean. This is now fixed so that appropriate warnings are generated.

Fixes  https://github.com/pulumi/pulumi-aws/issues/4342 P1

### DIFF
--- a/pkg/tfbridge/validate_input_types.go
+++ b/pkg/tfbridge/validate_input_types.go
@@ -157,9 +157,8 @@ func (v *PulumiInputValidator) validatePropertyValue(
 	}
 
 	switch typeSpec.Type {
-	case "bool":
-		// The bridge permits coalescing strings to booleans, hence skip strings.
-		if !propertyValue.IsBool() && !propertyValue.IsString() {
+	case "boolean":
+		if !propertyValue.IsBool() {
 			return []TypeFailure{{
 				ResourcePath: propertyPath.String(),
 				Reason: fmt.Sprintf(
--- a/pkg/tfbridge/tests/provider_test.go
+++ b/pkg/tfbridge/tests/provider_test.go
@@ -825,8 +825,11 @@ func TestTypeCheckingMistypedBooleansWithUnknowns(t *testing.T) {
 		},
 	}, newTestProviderOptions{})
 
+	reason := "expected boolean type, got string type. " +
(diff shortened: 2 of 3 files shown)


pkg/tfbridge/validate_input_types.go
```diff
@@ -157,9 +157,8 @@ func (v *PulumiInputValidator) validatePropertyValue(
 	}
 
 	switch typeSpec.Type {
-	case "bool":
-		// The bridge permits coalescing strings to booleans, hence skip strings.
-		if !propertyValue.IsBool() && !propertyValue.IsString() {
+	case "boolean":
```
t0yv0, line 160: This was "boolean" not "bool" apparently.

pkg/tfbridge/validate_input_types.go
```diff
@@ -157,9 +157,8 @@ func (v *PulumiInputValidator) validatePropertyValue(
 	}
 
 	switch typeSpec.Type {
-	case "bool":
-		// The bridge permits coalescing strings to booleans, hence skip strings.
-		if !propertyValue.IsBool() && !propertyValue.IsString() {
```
t0yv0, line 162: I've searched the code and apparently over here https://github.com/pulumi/pulumi-terraform-bridge/blob/t0yv0%2Ffix-typechecking-booleans/pkg/tfbridge/schema.go#L431 the bridge is not parsing strings when the slot is of type boolean, at least not yet. Removing `!propertyValue.IsBool()` condition matches this behavior.

corymhall (approved):

pkg/tfbridge/validate_input_types.go
iwahbe: Can we add an error to `validate.go` to prevent `bool`? The bridge should prevent this kind of user error.

iwahbe (approved):

pkg/tfbridge/validate_input_types.go
t0yv0: Er, not sure what you're suggesting. Pulumi Package Schema mandates "boolean" but we weren't using enums in this code so made a typo. The "bool" branch will never be hit in practice. Or do you mean something else?

t0yv0: In the interest of time, I'll merge but address this as a follow up if needed.

PR #2312, opened 2024-08-14
Refactor example manipulation to use an AST

iwahbe:


iwahbe:
To merge, I think we need to add tests for splicing together external examples (I don't think that situation is under test at all right now). Otherwise I think this is ready to merge.

pkg/tfgen/parse/meta.go
```diff
@@ -0,0 +1,78 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package parse
+
+import (
+	"github.com/yuin/goldmark"
+	meta "github.com/yuin/goldmark-meta"
```
guineveresaenger, line 19: nice find! Feel free to ignore me on this, but is there a reason we need to alias this here? I have to admit I went back and forth a bit on "meta" being a bit overloaded as a word in general, and leaving it as `goldmark-meta` at least tells the future maintainer that it's a parser thing.

pkg/tfgen/docs.go
```diff
@@ -372,34 +373,17 @@ var (
...
-
-	offset := len(text) - len(bytes)
-	contract.Assertf(offset >= 0, "The offset generated by chopping of the front-matter cannot be negative")
+	bytes := []byte(content)
 
 	gm := goldmark.New(goldmark.WithExtensions(parse.TFRegistryExtension))
 
 	headers := []int{}
-	parse.WalkNode(gm.Parser().Parse(gmtext.NewReader(bytes)), func(heading *gmast.Heading) {
+	parse.WalkNode(gm.Parser().Parse(text.NewReader(bytes)), func(heading *ast.Heading) {
```
guineveresaenger, line 386: 💟

pkg/tfgen/docs.go
```diff
@@ -515,19 +497,110 @@ func (p *tfMarkdownParser) readFile(name string) ([]byte, error) {
...
+			header.RemoveChildren(header)
+			header.AppendChild(header, ast.NewString(matches[2]))
+		}
+
+		nestedExampleSections = append(nestedExampleSections, node)
+		return ast.WalkContinue, nil
+	})
+	contract.AssertNoErrorf(err, "impossible - the walker doesn't return an error")
+
+	// TODO: transform H4 sections that contain code snippets into H3 sections.
```
guineveresaenger, line 565: This looks like it would be blocking a merge, no?

pkg/tfgen/parse/meta.go
iwahbe: We don't need the alias... it's actually implied. `-` is not valid in Go's identifier syntax, so it just drops the `goldmark-` part:

```go
import "github.com/yuin/goldmark-meta"

...

func myFunc(meta.Extension) {}
```

pkg/tfgen/docs_test.go
```diff
@@ -1360,11 +1307,24 @@ Basic usage:`
...
+		).Convert([]byte(input), &out)
+		require.NoError(t, err)
+
+		exp := make([]string, len(expected))
+		for i, e := range expected {
+			exp[i] = strings.Join(e, "\n")
+		}
 
-		assert.Equal(t, expected, actual)
+		assertEqualHTML(t, strings.Join(exp, "\n"), out.String())
```
guineveresaenger, line 1327: 😀

pkg/tfgen/docs.go
```diff
@@ -515,19 +497,135 @@ func (p *tfMarkdownParser) readFile(name string) ([]byte, error) {
...
+	} else {
+		e.replaceExamples(node, reader, pc)
+	}
+}
+
+var exampleHeaderRegexp = regexp.MustCompile(`(?i)^(Example Usage\s*)(?:(?:(?:for|of|[\pP]+)\s*)?(.*?)\s*)?$`)
+
+func (e exampleTransformer) reformatExamples(node *ast.Document, reader text.Reader, pc parser.Context) {
+	// First, find the canonical "Example Usage" section and all other docs sections.
+	var cannonicalExampleSection *section.Section
```
guineveresaenger, line 558:
```suggestion
	var canonicalExampleSection *section.Section
```
and similar below

pkg/tfgen/docs.go
```diff
@@ -372,34 +373,17 @@ var (
 	listMarkerRegex         = regexp.MustCompile("[-*+]")
 )
 
-func trimFrontMatter(text []byte) []byte {
```
guineveresaenger, line 375: I think this removal may be a merge conflict at this point :(

guineveresaenger (approved):
I'd love to get the typo fixed but this looks great!

pkg/tfgen/parse/extension.go
```diff
@@ -40,12 +44,29 @@ var TFRegistryExtension goldmark.Extender = tfRegistryExtension{}
...
 func (s tfRegistryExtension) Extend(md goldmark.Markdown) {
+	md.SetRenderer(markdown.NewRenderer())
 	extension.GFM.Extend(md)     // GitHub Flavored Markdown
 	section.Extension.Extend(md) // AST defined sections
 	meta.Extension.Extend(md)    // Support for YAML metadata blocks
 	md.Parser().AddOptions(parser.WithASTTransformers(
 		util.Prioritized(recognizeHeaderAfterHTML{}, 902),
 	))
+	md.Renderer().AddOptions(renderer.WithNodeRenderers(
+		util.Prioritized(tableRenderer{md.Renderer()}, 499),
```
guineveresaenger, line 55: Could we add a short version of the commit description as a code comment here? Just like "we do this so the SDKs have docs that make sense"

pkg/tfgen/parse/extension.go
```diff
@@ -40,12 +44,29 @@ var TFRegistryExtension goldmark.Extender = tfRegistryExtension{}
...
+		util.Prioritized(renderType{
+			kind: ast.KindString,
```
guineveresaenger, line 57: Can we keep the explanation here as well?

guineveresaenger (approved):
nice find! I also like that the string renderer is refactored into the extension. 👍

iwahbe:
I ran a downstream test run and reviewed the results. There are still wholes in our test coverage. I saw a bunch of positive changes, which makes me think this PR is worth merging. These include (but are not limited to):

Better parsing removes `- - -` from GCP (for every resource), inlining footnotes (necessary since we chop up the upstream document), and preventing the `## # Resource: ...` pattern we see in many upstream providers.

---

Unfortunately, the PR also introduces some blocking regressions:

Previously, every title was Title Cased. We have accidentally dropped that behavior.
Links are rendered as absolute, so `https://...` is now rendered as `[https://...](https://...)`. This is irrelevant to HTML but makes the markdown less readable.

Most concerning, some providers leaked their frontmatter into the upstream docs, for example newrelic and nomad:

- https://github.com/pulumi/pulumi-newrelic/pull/859/files: `sdk/dotnet/DataPartitionRule.cs`
-  https://github.com/pulumi/pulumi-nomad/pull/407/files#diff-3954e5395e691069e93a8adf679180973d1475eb3092c2a25e738c72c1c332c7: `sdk/dotnet/GetAclRoles.cs`

iwahbe:
Marking as a draft since this contains known regressions.

PR #2316, opened 2024-08-15
Infer Map[string] when the Map's element type is unspecified

iwahbe:
SDKv2 providers allow specifying a map type without a backing element type:

```go
schema.Schema{
	Type:     schema.TypeMap,
	Required: true,
}
```

Previously, the bridge interpreted this as `map[any]`. This does not agree with TF's interpretation: `map[string]`. This PR changes the bridge to correctly interpret `map[???]` as `map[string]`.

Fixes https://github.com/pulumi/pulumi-nomad/issues/389

### DIFF
--- a/pkg/tests/internal/tfcheck/tf_test.go
+++ b/pkg/tests/internal/tfcheck/tf_test.go
@@ -54,6 +54,67 @@ resource "test_resource" "test" {
 	t.Logf(driver.GetState(t))
 }
 
+// TestTfMapMissingElem shows that maps missing Elem types are equivalent to specifying:
+//
+//	Elem: &schema.Schema{Type: schema.TypeString}
+//
+// Previously, the bridge treated a missing map element type as `schema.TypeAny` instead
+// of `schema.TypeString`, which caused provider panics. For example:
+//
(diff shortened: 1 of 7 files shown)


t0yv0:
One more case missed in https://github.com/pulumi/pulumi-terraform-bridge/issues/1828 it seems.

iwahbe:
> Great find! Did you test this in nomad?
> 
> Can we also add an integration test for the panic?

I tested this in nomad, the error case becomes unrepresentable. I haven't added an integration test for the panic in nomad, and we shouldn't need to:

The bridge translation ensures that (TF) `Schema{Type: Map}` is translated to (Pulumi) `{ Type: Map, Elem: {Type: String}}`. As long as that holds (and it does, with tests) then SDKs (and YAML) will prevent sending malformed inputs. If a user tries to send malformed inputs, then [Type-check incoming values against the Pulumi Package Schema #1328](https://github.com/pulumi/pulumi-terraform-bridge/issues/1328) should let them know that what they are doing is wrong (and catch the panic).

VenelinMartinov (approved):
That makes sense, you are saying the tfgen test is sufficient here. Sounds good

PR #2317, opened 2024-08-15
Add a test for examples replacement

iwahbe:
I'm adding test coverage in preparation for https://github.com/pulumi/pulumi-terraform-bridge/pull/2312.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -491,6 +491,10 @@ type tfMarkdownParser struct {
 	editRules editRules
 
 	ret entityDocs
+
+	// readFile allows tests to mock out external files. It should not be set outside
+	// of test cases.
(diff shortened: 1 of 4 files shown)


VenelinMartinov (approved):
lgtm

pkg/tfgen/docs.go
```diff
@@ -502,13 +506,20 @@ const (
 	sectionImports             = 5
 )
 
-func (p *tfMarkdownParser) parseSupplementaryExamples() (string, error) {
+func (p *tfMarkdownParser) readFile(name string) ([]byte, error) {
+	if p.readFileFunc != nil {
```
guineveresaenger, line 510: Can we add a comment that makes it clear this if branch is hit by test cases only?

guineveresaenger (approved):

PR #2330, opened 2024-08-21
Change Go's import path to a dynamic specific path with version

iwahbe:
This change provides 2 benefits:

1. It prevents a static SDK from conflicting with a dynamic SDK.
2. It correctly includes the major version of the generated SDK in the path.

### DIFF
--- a/dynamic/info.go
+++ b/dynamic/info.go
@@ -16,7 +16,6 @@ package main
 
 import (
 	"context"
-	"fmt"
 	"path"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


VenelinMartinov (approved):

dynamic/info.go
```diff
@@ -60,8 +59,8 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
 		Java: &tfbridge.JavaInfo{ /* Java does not have a RespectSchemaVersion flag */ },
 		Golang: &tfbridge.GolangInfo{
 			ImportBasePath: path.Join(
-				fmt.Sprintf("github.com/pulumi/pulumi-%[1]s/sdk/", p.Name()),
-				tfbridge.GetModuleMajorVersion("0.0.0"),
+				"github.com/pulumi/pulumi-terraform-provider/sdks",
+				tfbridge.GetModuleMajorVersion(p.Version()),
```
julienp, line 63: I like having a dedicated scope for these providers, but I wonder if the name should also come before the version number?

We'll have

`github.com/pulumi/pulumi-terraform-provider/sdks/v3/something`
`github.com/pulumi/pulumi-terraform-provider/sdks/v5/someotherthing`

which looks a little weird to me, since there might not be a v3/someotherthing for example.

Maybe we could have `github.com/pulumi/pulumi-terraform-provider/sdks/something/v3` instead?

iwahbe: I like that idea.

julienp (approved):

PR #2334, opened 2024-08-21
Correctly map `pkg_index` tokens

iwahbe:
This PR improves the error message when it fails to map tokens (first commit), then removes the `"_"` appended to the end of the prefix so tokens that are of the form `<pkg>_index` can be correctly mapped.

Fixes https://github.com/pulumi/pulumi-terraform-provider/issues/13

### DIFF
--- a/dynamic/info.go
+++ b/dynamic/info.go
@@ -84,7 +84,11 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
 		return prov, err
 	}
 
-	prov.MustComputeTokens(tokens.SingleModule(p.Name()+"_", "index", tokens.MakeStandard(p.Name())))
+	err := prov.ComputeTokens(tokens.SingleModule(p.Name()+"_", "index", tokens.MakeStandard(p.Name())))
+	if err != nil {
+		return prov, err
+	}
+
 	prov.SetAutonaming(255, "-")
 
 	return prov, nil
--- a/dynamic/info.go
+++ b/dynamic/info.go
@@ -84,7 +84,8 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
 		return prov, err
 	}
 
-	err := prov.ComputeTokens(tokens.SingleModule(p.Name()+"_", "index", tokens.MakeStandard(p.Name())))
+	err := prov.ComputeTokens(tokens.SingleModule(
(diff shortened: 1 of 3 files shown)


dynamic/info.go
```diff
@@ -84,7 +84,8 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
 		return prov, err
 	}
 
-	err := prov.ComputeTokens(tokens.SingleModule(p.Name()+"_", "index", tokens.MakeStandard(p.Name())))
+	err := prov.ComputeTokens(tokens.SingleModule(
+		prov.GetResourcePrefix(), "index", tokens.MakeStandard(p.Name())))
```
mjeffryes, line 88: I can see how this fixes this specific error, and I can see that specific case tested in the new test, but I'm not sure I follow why we don't need the "_" for other resources?

pkg/tfbridge/tokens_test.go
```diff
@@ -87,6 +87,24 @@ func TestTokensSingleModule(t *testing.T) {
 	}, info.Resources)
 }
 
+func TestTokenWithModuleName(t *testing.T) {
+	info := tfbridge.ProviderInfo{
+		Name: "foo",
+		P: (&schema.Provider{
+			ResourcesMap: schema.ResourceMap{
+				"foo_index": nil,
```
mjeffryes, line 95: should we be testing some other names against this strategy eg. maybe the list fromt he previous test:
```
"foo_fizz_buzz":       {Tok: "foo:index:FizzBuzz"},
"foo_bar_hello_world": {Tok: "foo:index:BarHelloPulumi"},
"foo_bar":             {Tok: "foo:index:Bar"},
```

mjeffryes (approved):
I trust your familiarity with the naming code to get this correct, but would feel better if there were more testcases checked for the exact strategy used by the dynamic provider.

pkg/tfbridge/tokens_test.go
iwahbe: This strategy is pretty well tested in:

https://github.com/pulumi/pulumi-terraform-bridge/blob/610224f512cdc81b79ef771f36ba6ac9e9d3387c/pkg/tfbridge/tokens_test.go#L36

dynamic/info.go
iwahbe: This is basically an instance of "we were calling `tokens.SingleModule` wrong. It already knows to split on `"_"`.

PR #2335, opened 2024-08-21
Ensure all sections are marked for skipping by ast.Walk

guineveresaenger:
It turns out that removing nodes while walking the document tree results in lost pointers to subsequent nodes. 

This pull request gathers nodes for removal, and removes them subsequently to ast.Walk(). It adjusts the test to check that multiple, same-level sections get skipped accordingly.


iwahbe (approved):
Thanks for fixing this. LGTM

Issue #2314, opened 2024-08-14
Introduce an import heuristic to drop entries that fail validators


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Import for bridged providers may return data from `Read()` that does not pass validations. This causes Pulumi to report warnings.

Since the issue does not manifest as badly in TF as in Pulumi, the underlying Read() implementations are unlikely to be fixed upstream at scale. The suggestion therefore is to introduce heuristic-like code in the bridge that tries to autocorrect Read() results and drop entries that fail to pass validators. A similar tactic was applied in https://github.com/pulumi/pulumi-terraform-bridge/pull/1948

Initial implementation in https://github.com/pulumi/pulumi-terraform-bridge/pull/2337

Todo:
- [ ] Account for nested properties.

### Affected area/feature

Affects import. See also:

- [import developer docs](https://pulumi-developer-docs.readthedocs.io/en/latest/architecture/import.html)
- https://github.com/pulumi/pulumi-terraform-bridge/issues/2028

t0yv0:
@corymhall I think https://github.com/pulumi/pulumi-terraform-bridge/pull/2337 has this mostly done but didn't tackle nested properties, that sounds right? Leaving it in the backlog for now until we have more evidence like linode 373 that this needs fixing.

corymhall:
Yep that is correct.

guineveresaenger:
It looks like https://github.com/pulumi/pulumi-terraform-bridge/issues/3456 is another instance of nested properties hitting this behavior.

The failing fields sit at two nested depths the #2337 heuristic doesn't reach:
  - nodeConfig.windowsNodeConfig.osversion — nested block, 2 levels deep
  - nodePools[].nodeConfig.windowsNodeConfig.osversion — list-of-blocks, 3 levels deep

Both are Computed+Optional StringInSlice enums whose Read value ("") fails the validator — same shape as the top-level case #2337 fixed, just nested. Equivalent Terraform config (google_container_cluster with lifecycle { ignore_changes = [node_pool] }) imports and plans cleanly because TF core only validates user-written HCL.

PR #2337, opened 2024-08-21
feat: run validate during import and drop invalid inputs

corymhall:
> This is a reroll of #2277 which was reverted in #2299
> I've added additional handing for secret values

During an import operation property values are read from the cloud and
turned into inputs which then becomes part of the generated code. For
some properties (Computed & Optional) the cloud can return values that
fail validation. Since these are computed & optional the provider will
allow the user to not specify a value. In these cases the import would
succeed, but if the user took the generated code as is and attempted to
run pulumi up it would fail due to the validation errors.

Rather than generate invalid inputs we will instead run Validate during import
and then processes any resulting validation errors. If any properties
fail validation (and they match our other cases) then we remove those
inputs so that the generated program code will not contain the invalid
inputs

NOTE: we want to (for now at least) scope this as narrowly as possible because we do not want to
create a scenario where the resulting inputs create a different error
(e.g. removing an invalid required property). In these cases we will just allow the invalid configuration
through and the user will have to decide what to do.

fixes pulumi/pulumi-aws#4054

### DIFF
--- a/pkg/tests/schema_pulumi_test.go
+++ b/pkg/tests/schema_pulumi_test.go
@@ -3,11 +3,13 @@ package tests
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
 
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hexops/autogold/v2"
@@ -16,7 +18,9 @@ import (
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	"github.com/pulumi/pulumi/sdk/v3/go/auto/optpreview"
 	"github.com/pulumi/pulumi/sdk/v3/go/auto/optrefresh"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
 )
 
 func TestBasic(t *testing.T) {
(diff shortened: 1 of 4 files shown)


pkg/tfbridge/schema.go
```diff
@@ -351,6 +351,12 @@ func (ctx *conversionContext) makeTerraformInput(
 	tfs shim.Schema,
 	ps *SchemaInfo,
 ) (interface{}, error) {
+
+	if v.IsSecret() {
+		// We shouldn't see secrets, but if we do then we should ignore them
```
t0yv0, line 356: This is somewhat reasonable, because we can't transform first-class secrets into TF we should not be handling them. There is though the need to handle secrets through black-box TF transformations, and it's been done elsewhere .. I'm wondering how you're triggering this code path where secrets are actually pushed in here, and should't the outer secret handling take care of that, if it's not taking care of the secrets perhaps it's also not secreting the results properly on the output side.

pkg/tfbridge/provider.go
```diff
@@ -1433,6 +1444,73 @@ func (p *Provider) Read(ctx context.Context, req *pulumirpc.ReadRequest) (*pulum
...
+		// - dropping non-computed properties will cause a diff on the next preview
+		if !schemaAtPath.Optional() || !schemaAtPath.Computed() {
+			continue
+		}
+		pp, err := resource.ParsePropertyPath(path.valuePath)
+		if err != nil {
+			logger.Debug(fmt.Sprintf("could not parse property path %q for validation error: %s", path.valuePath, err.Error()))
+			continue
+		}
+		logger.Warn(fmt.Sprintf("property at path %q failed validation and was dropped from generated input", pp.String()))
```
t0yv0, line 1509: The warning will not be user-actionable. I think this might better be a Debug?

pkg/tfbridge/provider.go
```diff
@@ -1433,6 +1444,73 @@ func (p *Provider) Read(ctx context.Context, req *pulumirpc.ReadRequest) (*pulum
...
+		return
+	}
+
+	rescfg := MakeTerraformConfigFromInputs(ctx, p.tf, tfInputs)
+	_, errs := p.tf.ValidateResource(ctx, tfName, rescfg)
+	for _, e := range errs {
+
+		path, _, _ := parseCheckError(schema, schemaInfos, e)
+
+		// do not process errors on nested types
```
t0yv0, line 1483: There's a few scenarios unblocked by nested types, let's backlog and link the TODO here.

pkg/tfbridge/provider.go
```diff
@@ -1433,6 +1444,73 @@ func (p *Provider) Read(ctx context.Context, req *pulumirpc.ReadRequest) (*pulum
...
+// processImportValidationErrors runs `Validate` and then processes any resulting validation errors.
+// This only needs to run during an `Import` operation because the values read from the cloud
+// during import will be turned into the inputs in the generated code. Sometimes the cloud returns invalid
+// input values which will cause the next `up` to fail.
+//
+// To try and avoid this failure case, we run the validate functions for the generated inputs and
+// if any fail (and they match our other cases) then we remove those inputs so that the generated
+// program code will not contain the invalid inputs
+//
+// NOTE: we want to (for now at least) scope this as narrowly as possible because we do not want to
```
t0yv0, line 1456: This is looking reasonable.

pkg/tests/schema_pulumi_test.go
```diff
@@ -2592,3 +2596,360 @@ Resources:
...
+						err := rd.Set("collection_prop", tc.cloudVal)
+						assert.NoError(t, err)
+						err = rd.Set("other_prop", "test")
+						assert.NoError(t, err)
+						return nil
+					},
+				},
+			}
+			tfp := &schema.Provider{ResourcesMap: resMap}
+			bridgedProvider := pulcheck.BridgedProvider(t, "prov", tfp, pulcheck.DisablePlanResourceChange())
```
t0yv0, line 2913: Brilliant, this is managing to test end to end somehow. Big step up for the validity of these tests!

pkg/tests/schema_pulumi_test.go
```diff
@@ -2592,3 +2596,360 @@ Resources:
...
+				assert.NotContains(t, imp.Stdout, "One or more imported inputs failed to validate")
+
+				f, err := os.OpenFile(filepath.Join(pt.CurrentStack().Workspace().WorkDir(), "Pulumi.yaml"), os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
+				assert.NoError(t, err)
+				defer f.Close()
+				_, err = f.WriteString(string(contents))
+				assert.NoError(t, err)
+
+				// run preview using the generated file
+				pt.Preview(optpreview.Diff(), optpreview.ExpectNoChanges())
```
t0yv0, line 2951: 💯

pkg/tfbridge/provider.go
```diff
@@ -1433,6 +1444,73 @@ func (p *Provider) Read(ctx context.Context, req *pulumirpc.ReadRequest) (*pulum
...
+// create a scenario where the resulting inputs create a different error
+// (e.g. removing an invalid required property). In these cases we will just allow the invalid configuration
+// through and the user will have to decide what to do.
+func (p *Provider) processImportValidationErrors(
+	ctx context.Context,
+	urn resource.URN,
+	tfName string,
+	inputs resource.PropertyMap,
```
t0yv0, line 1464: Maybe just https://github.com/pulumi/pulumi-terraform-bridge/blob/t0yv0%2Fautofill/unstable/propertyvalue/propertyvalue.go#L94 RemoveSecrets over the value here, if that's where secrets are coming from? This code should be obviously safe to ignore them.

pkg/tfbridge/schema.go
corymhall: This is where I am testing it https://github.com/pulumi/pulumi-terraform-bridge/blob/b2e46fef4662978fc9b924fafe7f930bc837b0c4/pkg/tfbridge/provider_test.go#L5738-L5753

Also see https://pulumi.slack.com/archives/C02FXTZEZ6W/p1723217439575279?thread_ts=1723201028.986739&cid=C02FXTZEZ6W

t0yv0: I've added a suggestion to strip secrets specifically in the newly added code, it's more obviously safe there.

t0yv0 (approved):

pkg/tfbridge/provider.go
corymhall: It looks like this works too, and I like it much better!

pkg/tfbridge/provider.go
corymhall: I'll just add a comment on https://github.com/pulumi/pulumi-terraform-bridge/issues/2314 that this PR has addressed part of it, but that we still need to address nested types.

pkg/tfbridge/provider.go
corymhall: Done! Let me know what you think.

t0yv0: That's perfect, ty.

pkg/tfbridge/provider.go
```diff
@@ -1433,6 +1444,76 @@ func (p *Provider) Read(ctx context.Context, req *pulumirpc.ReadRequest) (*pulum
...
+) {
+	inputValueWithoutSecrets := propertyvalue.RemoveSecrets(resource.NewObjectProperty(inputs))
+	inputsWithoutSecrets := inputValueWithoutSecrets.ObjectValue()
+	logger := GetLogger(ctx)
+	tfInputs, _, err := makeTerraformInputsWithOptions(ctx,
+		&PulumiResource{URN: urn, Properties: inputs},
+		p.configValues, inputsWithoutSecrets, inputsWithoutSecrets, schema, schemaInfos,
+		makeTerraformInputsOptions{DisableTFDefaults: true, UnknownCollectionsSupported: p.tf.SupportsUnknownCollections()})
+	if err != nil {
+		return
```
iwahbe, line 1476: This seems like the perfect place to log a warning.
```suggestion
		logger.Warn("Failed to makeTerraformInputsWithOptions. <insert useful information about inputs here>")
		return
```

If we want this to fail silently, then we should still log to debug.

iwahbe (approved):
LGTM

pkg/tfbridge/provider.go
corymhall: Added a debug log

PR #2341, opened 2024-08-22
Add tf as a code identifier to isHCL func

guineveresaenger:
Some provider docs like Artifactory use ```tf as a code block identifier. 

This PR adds that so the bridge can recognize it for example translation.


flostadler (approved):
LGTM

This is a good list of code identifiers btw: https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md

PR #2346, opened 2024-08-23
Clean up type error display

iwahbe:
Previously, an error message looked like this:

```
    warning: Unexpected type at field "skipMetadataApiCheck":
```

The same error now looks like this:

```

                 Unexpected type at field "skipMetadataApiCheck":

             Type checking is still experimental. If you believe that a warning is incorrect,
```

There is only one `warning:` block, and every item is nested within the warning block. The heart of the error: the type failures take up the "center" of the terminal, separated by space from the rest of the display.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -588,41 +588,74 @@ func (p *Provider) typeCheckConfig(
 	// for now we are just going to log warnings if there are failures.
 	// over time we may want to turn these into actual errors
 	validateShouldError := cmdutil.IsTruthy(os.Getenv("PULUMI_ERROR_CONFIG_TYPE_CHECKER"))
+
+	// If we don't have a schema, then we don't attempt to type check the config at
+	// all.
+	if p.pulumiSchemaSpec == nil {
+		logger.Debug("p.pulumiSchemaSpec == nil, skipping type checking config")
+		return nil
+	}
+
+	iv := NewInputValidator(urn, *p.pulumiSchemaSpec, true)
+	typeFailures := iv.ValidateConfig(news)
+	if validateShouldError {
+		return p.convertTypeFailures(urn, typeFailures)
+	}
+
+	// If we don't have any type errors, we can just return.
+	if len(typeFailures) == 0 {
+		return nil
+	}
+
+	// warningSpaces is same length as the string that prefixes a [logger.Warn] invocation.
+	//
+	//                    "warning: "
+	const warningSpaces = "         "
+	prefix := func(indent int) string {
+		if indent == 0 {
+			return ""
+		}
+		const fourSpaces = "    "
+		return warningSpaces + strings.Repeat(fourSpaces, indent-1)
+	}
+
+	var msg strings.Builder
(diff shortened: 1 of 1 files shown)


iwahbe:
This is almost entirely a cosmetic change. The only non-display change is that we will no longer warn `Type checking failures:` when we are emitting a structured error.

pkg/tfbridge/provider.go
```diff
@@ -588,41 +588,74 @@ func (p *Provider) typeCheckConfig(
...
+
+	// If we don't have any type errors, we can just return.
+	if len(typeFailures) == 0 {
+		return nil
+	}
+
+	// warningSpaces is same length as the string that prefixes a [logger.Warn] invocation.
+	//
+	//                    "warning: "
+	const warningSpaces = "         "
```
guineveresaenger, line 613: Is there a reason tabs won't work here and for `fourSpaces`? I know we can't get tabs to be the same exact spacing as `warning: ` but they're more visible when reading this code.

pkg/tfbridge/provider.go
```diff
@@ -588,41 +588,74 @@ func (p *Provider) typeCheckConfig(
...
+	msg.WriteString(prefix(1) + "Type checking is still experimental. If you believe that a warning is incorrect,\n" +
+		prefix(1) + "please let us know by creating an " +
+		"issue at https://github.com/pulumi/pulumi-terraform-bridge/issues.\n" +
+		prefix(1) + "This will become a hard error in the future.",
+	)
+	logger.Warn(msg.String())
+	return nil
+}
+
+func (p *Provider) convertTypeFailures(urn resource.URN, tf []TypeFailure) *pulumirpc.CheckResponse {
```
guineveresaenger, line 641: Could we please not use `tf` as a shorthand here - it's really easy to think "Terraform" when reading it.

pkg/tfbridge/provider.go
```diff
@@ -588,41 +588,74 @@ func (p *Provider) typeCheckConfig(
...
+	// all.
+	if p.pulumiSchemaSpec == nil {
+		logger.Debug("p.pulumiSchemaSpec == nil, skipping type checking config")
+		return nil
+	}
+
+	iv := NewInputValidator(urn, *p.pulumiSchemaSpec, true)
+	typeFailures := iv.ValidateConfig(news)
+	if validateShouldError {
+		return p.convertTypeFailures(urn, typeFailures)
```
guineveresaenger, line 602: thank you for breaking this away 💟

guineveresaenger (approved):
A few non-blocking nits.

With regards to your comment on behavioral change: If I understand correctly, the `warning` that was emitted on failure before was basically redundant?

pkg/tfbridge/provider.go
iwahbe: Sure.

pkg/tfbridge/provider.go
iwahbe: Tabs didn't display the same way for me in my in-editor (emacs/vterm) terminal and my separate terminal (Alacritty).

Issue #2307, opened 2024-08-13
Type check failure: Improve error message


guineveresaenger:
Our type check failure messages should have a bit more detail.

```
error: `resourcename` resource `foo` has a problem: Expected Object Type. Expected object, got string. Examine values at `foo`.
```

Significantly, the bridge does transform some field values into strings, so the result is additionally confusing for users.

pulumi/pulumi-aws#4360 shows that it can be difficult to make sense of the standard error message  for type check failures. The message in the linked issue isn't incorrect, but it should make more helpful suggestions. At the very least we could mention that a resource's Constructor Example shows the complete shape of a resource and its fields.

Issue #2339, opened 2024-08-22
Type checking failure -- "skipMetadataApiCheck"


AshTharappel:
### What happened?

getting this warning. the argument passed in is a boolean type, but the check thinks its a string

this is on the `AWS.Provider` resource

```
    warning: Type checking failed:
    warning: Unexpected type at field "skipMetadataApiCheck":
               expected boolean type, got string type
    warning: Type checking is still experimental. If you believe that a warning is incorrect,
    please let us know by creating an issue at https://github.com/pulumi/pulumi-terraform-bridge/issues.
    This will become a hard error in the future. 
```

### Example

here is the code block we are using to create the provider

```ts
  const provider = new Provider(accountName, {
    region: 'us-east-1',
    assumeRole: {
      roleArn: `arn:aws:iam::${accountId}:role/${roleName}`,
      sessionName: `${accountName}-session`,
    },
    skipCredentialsValidation: true,
    skipMetadataApiCheck: true,
    skipRegionValidation: true,
    skipRequestingAccountId: true,
  });
```

### Output of `pulumi about`

```
CLI
Version      3.129.0
Go Version   go1.22.6
Go Compiler  gc

Plugins
KIND      NAME    VERSION
resource  aws     6.50.0
language  nodejs  unknown
resource  random  4.16.3

Host
OS       darwin
Version  14.5
Arch     arm64

This project is written in nodejs: executable='/opt/homebrew/bin/node' version='v22.6.0'

Backend
Name           mm-tharappela3
URL            s3://rs-pulumi-state-852480429243
User           ashwin.tharappel
Organizations
Token type     personal

Dependencies:
NAME                VERSION
@pulumi/aws         6.50.0
@pulumi/pulumi      3.129.0
@rs-crossguard/aws  0.1.7
@rs-pulumi/aws      1.1.3
@ws-pulumi/aws      0.0.21
rs-accounts         1.1.1
@types/jest         29.5.12
@types/node         22.5.0
gts                 5.3.1
jest                29.7.0
ts-jest             29.2.4
typescript          5.5.4

Pulumi locates its logs in /var/folders/0n/0llmscvn1z7b6h4cq2qr31d40000gq/T/ by default
warning: Failed to get information about the current stack: No current stack
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

iwahbe:
[...]

PR #2347, opened 2024-08-23
Display invalid values when error checking

iwahbe:
This makes it much easier to understand type errors. This is motivated by https://github.com/pulumi/pulumi-terraform-bridge/issues/2339. 

The `.Reason` section of `TypeFailures` used to look like this:

```
```

They now look like this:

```
expected boolean type, got "true" of type string
```

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2307.

### DIFF
--- a/pkg/tfbridge/provider_test.go
+++ b/pkg/tfbridge/provider_test.go
@@ -1202,7 +1202,7 @@ func TestCheckWarnings(t *testing.T) {
 		},
 	}
 
-	// we need the pschema for type checking
+	// we need the schema for type checking
 	pulumiSchemaSpec := &pschema.PackageSpec{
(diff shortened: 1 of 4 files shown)


iwahbe:
> Does this respect secretness?

That is an good question. 

My first response:

> It does not, and I'm not sure it should. Ideally, we would respect secretness if and only if `--show-secrets` is passed for the `pulumi {up,preview}`. If that option is passed to providers, then I'm not aware of it.

My second response:

> It doesn't matter, since the bridge doesn't get secrets anyway. Handling secrets seems like a good idea to me, since we do want to eventually introduce support for them (and the type describes secrets), but we don't accept secrets now.

PR #2348, opened 2024-08-23
Emit warning on example translation failure

guineveresaenger:
Installation docs build failed for pulumi-alicloud and others when an example could not be translated. This PR continues on a failed example translation (just like for API docs) and emits a warning instead.

Additionally, we write a default message to the code field in the missing example.

### DIFF
--- a/pkg/tfgen/convert_cli.go
+++ b/pkg/tfgen/convert_cli.go
@@ -643,17 +643,17 @@ func (cc *cliConverter) singleExampleFromHCLToPCL(path, hclCode string) (transla
 
 // Function for one-off example conversions PCL --> supported language (nodejs, yaml, etc)
 func (cc *cliConverter) singleExampleFromPCLToLanguage(example translatedExample, lang string) (string, error) {
+	var err error
+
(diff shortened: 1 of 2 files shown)


pkg/tfgen/convert_cli.go
```diff
@@ -643,17 +643,17 @@ func (cc *cliConverter) singleExampleFromHCLToPCL(path, hclCode string) (transla
...
 	}
-	source, diags, err := cc.convertPCL(example.PCL, lang)
-	if err != nil {
-		return "", err
-	}
+	source, diags, _ := cc.convertPCL(example.PCL, lang)
 	diags = cc.postProcessDiagnostics(diags.Extend(example.Diagnostics))
 	if diags.HasErrors() {
-		return "", fmt.Errorf("Failed to convert an example: %s", diags.Error())
+		source = "Example unavailable in this language\n"
```
iwahbe, line 654:
```suggestion
		source = "Example currently unavailable in this language\n"
```

iwahbe (approved):

PR #2349, opened 2024-08-23
Add default patterns to strip TF version references and tests

guineveresaenger:
Adds a more pinpointed removal of TF version mentions, because a simple replace of "Terraform" --> "Pulumi" leaves these nonsensical.

There will likely be more patterns added here.

Contains relevant tests.

### DIFF
--- a/pkg/tfgen/installation_docs.go
+++ b/pkg/tfgen/installation_docs.go
@@ -376,7 +376,7 @@ func getDefaultHeadersToSkip() []*regexp.Regexp {
 func getTfVersionsToRemove() []*regexp.Regexp {
 	tfVersionsToRemove := []*regexp.Regexp{
 		regexp.MustCompile(`It requires [tT]erraform [v0-9]+\.?[0-9]?\.?[0-9]? or later.`),
-		regexp.MustCompile(`(?s)For [tT]erraform [v0-9]+\.?[0-9]?\.?[0-9]? and later:`),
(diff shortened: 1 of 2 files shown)


mjeffryes (approved):

pkg/tfgen/installation_docs.go
```diff
@@ -372,3 +372,24 @@ func getDefaultHeadersToSkip() []*regexp.Regexp {
...
+		regexp.MustCompile(`(?s)(For )?[tT]erraform [v0-9]+\.?[0-9]?\.?[0-9]? and (later|earlier):`),
+	}
+	return tfVersionsToRemove
+}
+
+func removeTfVersionMentions(docFile string) tfbridge.DocsEdit {
+	return tfbridge.DocsEdit{
+		Path: docFile,
+		Edit: func(_ string, content []byte) ([]byte, error) {
+			for _, tfVersion := range getTfVersionsToRemove() {
```
iwahbe, line 388: This compiles the regular expressions in the `.Edit` call, which is not efficient.
```suggestion
	tfVersionsToRemove := getTfVersionsToRemove()
	return tfbridge.DocsEdit{
		Path: docFile,
		Edit: func(_ string, content []byte) ([]byte, error) {
			for _, tfVersion := range tfVersionsToRemove {
```

iwahbe (approved):

PR #2360, opened 2024-08-26
Simplify using MuxWith with PF

t0yv0:
Currently `info.Provider` allows overriding resource or function implementations by specifying `MuxWith` section with a
list of `MuxProvider` instances. This PR adds a function to easily wrap a Plugin Framework based provider into a
`MuxProvider` instance:

    pf.NewMuxProvider(context.Context, info.Provider, *ProviderMetadata) (info.MuxProvider, error)

This also enables bridge developers to more easily create end-to-end tests against providers implemented with the Plugin
Framework.

### DIFF
--- a/pf/tests/muxwith_test.go
+++ b/pf/tests/muxwith_test.go
@@ -1,0 +1,67 @@
+// Copyright 2016-2023, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


guineveresaenger (approved):
side note: I see in CI we're skipping non-linux based tests, including MacOS. Where/why is this happening? I don't see the origin of that message in this code.

t0yv0:
Not seeing this, curious if you have a link?

Look I'm still not linter-clean, let me update.

PR #2361, opened 2024-08-26
fix: strip upstream frontmatter

guineveresaenger:
When generating Pulumi frontmatter, we do not need to obtain the `title` content by parsing the upstream file. We can use the provider name instead.

This fixes ~a bug~ behavior where the previous code was interpreting comments inside of the upstream frontmatter as H1. Example: https://github.com/kreuzwerker/terraform-provider-docker/blob/master/docs/index.md?plain=1#L2 - this never made it to actually get filed as a bug; it was discovered during content generation.

Additionally, H1 title removal is now done via Goldmark transformer.

- **Strip upstream frontmatter by default and use provider name to generate Pulumi frontmatter title. Add tests.**
- **Add capitalization**


pkg/tfgen/installation_docs.go
```diff
@@ -70,15 +70,20 @@ func writeFrontMatter(providerName string) string {
 		title)
 }
 
-func getBodyAndTitle(content string) (string, string) {
-	// The first header in `index.md` is the package name, of the format `# Foo Provider`.
+func stripUpstreamFrontMatter(content string) string {
```
iwahbe, line 73: There is already a `trimFrontMatter` function in `pkg/tfgen/docs.go`. Can we just use that?

Also, it looks like this function doesn't just strip frontmatter, it also removes the first H1 encountered. Can we do this with goldmark instead of manual parsing?

For example, this is a valid markdown H1:

```md
H1
===
```

We don't need to write code to understand `===`.

iwahbe (commented):
The change you describe looks good. I wouldn't call it a "refactor", since it sounds like its a bug fix. Can you link the bug?

I'd like to require that a function called `stipUpstreamFrontMatter` strips the frontmatter from a document (and nothing else).

pkg/tfgen/installation_docs.go
guineveresaenger: Thanks for pointing out that function! I....had to adjust the trim index at the end because it would deliver a little `-` otherwise - but I added a test that should keep it behaving.

iwahbe:
Let me know when you want me to take another look.

pkg/tfgen/installation_docs.go
```diff
@@ -251,6 +246,50 @@ func convertExample(g *Generator, code string, exampleNumber int) (string, error
...
 }
 
+type titleRemover struct {
+}
+
+var _ parser.ASTTransformer = titleRemover{}
+
+func (tr titleRemover) Transform(node *ast.Document, reader text.Reader, pc parser.Context) {
+	// Walk to find sections that should be skipped.
+	// Walk() loses information on subsequent nodes when nodes are removed during the walk, so we only gather them here.
```
iwahbe, line 256: Leftover from copy & paste?
```suggestion
```

pkg/tfgen/installation_docs.go
```diff
@@ -251,6 +246,50 @@ func convertExample(g *Generator, code string, exampleNumber int) (string, error
...
+				parent := n.Parent()
+				if parent == nil {
+					panic("PARENT IS NIL")
+				}
+				// Removal here is safe, as we want to remove only the first header anyway.
+				n.Parent().RemoveChild(parent, header)
+				return ast.WalkStop, nil
+			}
+		}
+		return ast.WalkContinue, nil
```
iwahbe, line 270: Nit: Could we keep this left aligned? I think it's easier to read.
```suggestion
		header, ok := n.(*ast.Heading)
		if !ok || header.Level != 1 || !entering {
			return ast.WalkContinue, nil
		}
		
		parent := n.Parent()
		contract.Assertf(parent != nil, "parent cannot be nil")
		// Removal here is safe, as we want to remove only the first header anyway.
		n.Parent().RemoveChild(parent, header)
		return ast.WalkStop, nil
```

iwahbe (approved):
I've never been sure if LGTM means "looks good to me" or "lets get this merged", but both apply. LGTM

PR #2368, opened 2024-08-27
Fix schema generation for dynamic bridged providers

t0yv0:
This fixes the dynamic bridged provider inability to pass input parameters to resources that use blocks in their schema.

As part of the fix, doc comments are added around shim.Schema encoding clarifying the uses of Optional and the
conventions for blocks specifically.

Fixes https://github.com/pulumi/pulumi-terraform-provider/issues/8

### DIFF
--- a/pf/proto/block.go
+++ b/pf/proto/block.go
@@ -114,12 +114,12 @@ func (m blockSchema) Elem() interface{} {
 	}
 }
 
-func (s blockSchema) Optional() bool {
-	return !s.Required()
+func (m blockSchema) Optional() bool {
+	return !m.Required()
 }
 
-func (s blockSchema) Required() bool {
-	return s.MinItems() > 0
+func (m blockSchema) Required() bool {
+	return m.MinItems() > 0
 }
 
 func (m blockSchema) Default() interface{}                { return nil }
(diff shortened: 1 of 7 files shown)


pf/tests/schemashim_test.go
```diff
@@ -74,12 +85,48 @@ func TestSchemaShimRepresentations(t *testing.T) {
       }
     }
   }
+}`),
+			autogold.Expect(`{
```
t0yv0, line 89: This expansion of the test is optional but it helped me sanity-check the schemas generated for each case by PF.

pf/proto/block.go
```diff
@@ -114,8 +114,14 @@ func (m blockSchema) Elem() interface{} {
 	}
 }
 
-func (m blockSchema) Optional() bool                      { return false }
-func (m blockSchema) Required() bool                      { return false }
+func (s blockSchema) Optional() bool {
+	return !s.Required()
```
t0yv0, line 118: This is the key fix mirroring the values from `blockSchema` in PF.

iwahbe (approved):
Thanks for the fix and the documentation. LGTM!

iwahbe:
To unblock CI, I think you need to run `make -C dynamic test_accept` to regenerate test files.

PR #2371, opened 2024-08-28
Display dynamic provider stacktrace on upstream provider panic

iwahbe:
We ensure that panic messages are added to the returned error so that they are displayed to the user.

Fixes https://github.com/pulumi/pulumi-terraform-provider/issues/22.


VenelinMartinov (approved):
lgtm

PR #2374, opened 2024-08-29
Refactor the type checker to it's own package

iwahbe:
This PR is best reviewed commit by commit. This is a pure refactor, there is no change in functionality.

- 0c6245cf46532780eb26f0543a5137f709258004 moves the `validate_input_types*` to a `typechecker` package located under `pkg/tfbridge`. The type checker is fully independent from other bridge code, and this makes the interface much easer to distinguish (because public vs private is enforced).
- 3d7918f16243bc1b41cb5b2d2b5a271c8adde208 does some type renames because type checker types will be prefixed by `typechecker`.
- 19038ebeb9c03113a17ef898a7cbd6e3ace8336b removes the `urn` argument to `typechecker.New` (previously `tfbridge.NewTypeValidator`). The argument is only stored and never used.
- b60f64f569a2049a1305574c4bc644d72e42676f removes the intermediate variable when a new type checker is created.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -46,6 +46,7 @@ import (
 	"google.golang.org/protobuf/types/known/emptypb"
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/info"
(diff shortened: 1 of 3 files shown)


corymhall (approved):
Love it!

PR #2381, opened 2024-08-30
Convert examples into serialized PCL form

andrzejressel:
Part 2 of https://github.com/pulumi/pulumi/pull/17124


t0yv0:
Thank you! This is an interesting contribution! We historically had some hesitation adding new languages like PCL to the Pulumi Package Schema as that's distributed many places and imposes some stability requirements. We'll discuss with the team though as the trade-offs are different as of today.

Also I've not seen this "protobuf" language before, need to dig in, is that simply serialized PCL?

Couple of notes I wanted to point out if you need to get unblocked before this lands. You can currently use this trick to extract JSON-formatted example conversions (PCL + translations to target languages).

```COVERAGE_OUTPUT_DIR=$PWD/cov make tfgen```

You can always go from PCL to TypeScript etc using `pulumi convert`.

If we do go forward with this we should also consider including source HCL instead of PCL in the Pulumi Package Schema, as that's more stable and PCL can be derived from it.

andrzejressel:
These are serialized pcl structures. I don't like that most of testing seems to be here and not in the main pulumi repo. I think I'll maybe add another language that will be json form (it already exists in pulumi/pulumi MR) so that it can be reviewable in pulumi/pulumi

t0yv0:
Hi andrzejressel I had a quick discussion internally and it's currently difficult for our team to take this change, as we need to be careful not to break the registry integration and other tools that consume Pulumi Package Schema. Are you unblocked for your use case with the workaround I provided above?

We could also consider publishing the PCLs side-by-side with provider releases for you to consume, instead of extending Pulumi Package Schema with them.

PR #2382, opened 2024-09-01
Handle code blocks in import section

andrzejressel:
Fix for https://github.com/pulumi/pulumi-docker/issues/1201

I've tested it with docker provider and it seems to be working fine apart from this warning:

```
warning: unable to convert HCL example for Pulumi entity '#/resources/docker:index/serviceConfig:ServiceConfig'. The example will be dropped from any generated docs or SDKs: 1 error occurred:
        * [csharp, go, java, python, typescript, yaml] <nil>: pcl.BindProgram failed: example.pp:2,17-41: unknown package 'std'; unknown package 'std'; 
```

which removes example from ServiceConfig, but I think this it not related to this change.


pkg/tfgen/test_data/parse-imports/docker_container-expected.md
```diff
@@ -0,0 +1,39 @@
...
+```sh
+#!/bin/bash
+docker run --name foo -p8080:80 -d nginx 
+# prints the container ID 
+9a550c0f0163d39d77222d3efd58701b625d47676c25c686c95b5b92d1cba6fd
+```
+
+you provide the definition for the resource as follows
+
+```terraform
```
guineveresaenger, line 21: We do not, as a rule, display Terraform code on our documentation. This code block would have to undergo example translation using `pulumi convert` as provided by this file's functions: https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfgen/convert_cli.go

guineveresaenger:
Hi @andrzejressel - thank you for your pull request. I'm trying to run tests for you but I'm seeing a setup issue on our end for the tests. 
While we're waiting for that, I've run this change against one of our other providers (pulumi-aws), and I see an index out of bounds panic on line 1269. You may find the `findCodeBlock` function helpful here - also know that we're trying to move away from a line-by-line string-based parsing approach and use an AST-based Markdown parser instead, but haven't gotten to the import section yet.

Additionally, all Terraform code will need to be translated to Pulumi code. We do not, as a rule, display Terraform code in these docs.

This is an incredibly gnarly part of our codebase, so thank you so much for willing to get involved here! 💟 I'm happy to help and answer questions, you can also find me on the [community Slack for Pulumi](https://slack.pulumi.com/).

pkg/tfgen/test_data/parse-imports/docker_container-expected.md
andrzejressel: This is not displayed - it is converted into provider code in next step somewhere - in any case docker doc ended being correct - this terraform block was converted into multiple code blocks.

guineveresaenger:
I've rebased and we should be able to get some better test output here.

andrzejressel:
Should be fine now

guineveresaenger:
I will take a look next week - thank you for your patience!

pkg/tfgen/docs.go
```diff
@@ -1206,65 +1242,68 @@ func (p *tfMarkdownParser) parseImports(subsection []string) {
...
-				importDocString = importDocString + section
+			importDocString = importDocString + "\n"
+			continue
+		}
+
+		// Handle script blocks - replace terraform import with pulumi import
+		if strings.HasPrefix(trimmedSection, "```") {
+			initial := true
+			for {
+				section = subsection[i]
```
guineveresaenger, line 1282: This line panics when running this code against pulumi-gcp.

guineveresaenger (changes requested):
I've tried this against one of our other providers - `pulumi-gcp`, and I unfortunately still see a panic. There's something about the counter in the loop that's not quite right.

PR #2390, opened 2024-09-04
Fix custom timeout handling under PlanResourceChange

t0yv0:
Under PlanResourceChange the way custom timeout option is handled is not properly communicating the value to the underlying TF provider. This is now fixed and the existing test made to pass.

See also: https://www.pulumi.com/docs/concepts/options/customtimeouts/

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2386

New warnings are introduced for the case when a user tries to set a customTimeouts option on a resource that does not support it. The customTiemeouts in this case are a no-op. I believe this should be a warning to guide the user but we might also consider downgrading severity to an INFO or debug level message.

### DIFF
--- a/pkg/tfbridge/log/log.go
+++ b/pkg/tfbridge/log/log.go
@@ -1,0 +1,74 @@
+// Copyright 2016-2024, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package log
+
+import (
+	"context"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/logging"
+)
+
+// Send logs or status logs to the user.
+//
+// Logged messages are pre-associated with the resource they are called from.
+type Logger interface {
+	Log
+
+	// Convert to sending ephemeral status logs to the user.
+	Status() Log
+}
+
+// The set of logs available to show to the user
+type Log interface {
+	Debug(msg string)
+	Info(msg string)
+	Warn(msg string)
+	Error(msg string)
+}
+
+// Get access to the [Logger] associated with this context.
+func GetLogger(ctx context.Context) Logger {
+	logger := ctx.Value(logging.CtxKey)
+	contract.Assertf(logger != nil, "Cannot call GetLogger on a context that is not equipped with a Logger")
+	return newLoggerAdapter(logger)
+}
+
+func newLoggerAdapter(logger any) Logger {
+	uLogger, ok := logger.(untypedLogger)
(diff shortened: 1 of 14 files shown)


t0yv0:
https://github.com/pulumi/terraform-plugin-sdk/blob/upstream-v2.33.0/helper/schema/core_schema.go#L320 is the interesting bit that defines the schema for timeouts piecemeal, unless the user already defined "timeouts" attribute or block. 

I've not verified how this works in TF but would be interesting:

- if user specifies a "timeouts" block or attribute, does it just get fully custom handling (i.e. we shouldn't be setting custom timeout overrides at all as that's always a no-op)
 
- if user does not specify a "timeouts" block or attribute, but the auto-generated one only support e.g. Create timeouts, is it true that Update timeouts can never work then? Sounds plausible but I've not confirmed in TF.

pkg/tfbridge/log/log.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/logging"
+)
+
+// Send logs or status logs to the user.
+//
+// Logged messages are pre-associated with the resource they are called from.
+type Logger interface {
```
t0yv0, line 28: This needs its own package so that it's usable from tfshim/* modules without imposing a circular pacakge dependency build error.

VenelinMartinov (approved):
lgtm

pkg/tfshim/sdk-v2/provider2.go
```diff
@@ -147,6 +151,73 @@ type planResourceChangeImpl struct {
...
+// some resources do not seem to allow customizing timeouts for certain operations, and their impliedType will not have
+// the corresponding slot. Warn the user if the custom timeout is a no-op.
+func (p *planResourceChangeImpl) configWithTimeouts(
+	ctx context.Context,
+	configWithoutTimeouts map[string]any,
+	topts shim.TimeoutOptions,
+	impliedType cty.Type,
+) map[string]any {
+	config := map[string]any{}
+	for k, v := range configWithoutTimeouts {
```
iwahbe, line 166: What is `configWithoutTimeouts`? Is it the TF config translation of the Pulumi config? A comment explaining that would be helpful.

It would be amazing if `configWithoutTimeouts` was it's own type, but that's a painful pattern in Go.

pkg/tfshim/sdk-v2/provider2.go
```diff
@@ -168,7 +239,9 @@ func (p *planResourceChangeImpl) Diff(
 	if err != nil {
 		return nil, err
 	}
-	cfg, err := recoverAndCoerceCtyValueWithSchema(res.CoreConfigSchema(), config.Config)
+
+	cfg, err := recoverAndCoerceCtyValueWithSchema(res.CoreConfigSchema(),
+		p.configWithTimeouts(ctx, config.Config, opts.TimeoutOptions, ty))
```
iwahbe, line 244: It looks like this is only called during `Diff`. Doesn't this need to be called during update as well?

pkg/tfshim/sdk-v2/provider2.go
t0yv0: Yeah. I can use ResourceConfig which is what's needed, but I always forget how to construct that. All the helper types aren't quite there yet to be truly usable. Added some commits. 

Generally I agree with your sentiment, just qualifying it with this idea that I sweat the signatures a bit less for functions that are private file-local single-use functions, e.g. the smaller the scope the less I worry about this. But I'm quite open to efforts to raise the bar on code quality in this codebase so if you have plans here happy to play along or follow a style guide.

PR #2409, opened 2024-09-10
OmitType and TypeName to support sharing types

t0yv0:
Introduce OmitType and TypeName flags to enforce type sharing.

Some of the upstream providers generate very large concrete schemata. TF is not being materially affected, just high RAM demands for in-memory processing. The example is inspired by QuickSight types in AWS. Pulumi is affected significantly. In Pulumi the default projection is going to generate named types for every instance of the shared schema. This leads to SDK bloat and issues with "filename too long." 

With this change it is possible for the provider maintainer opt into explicit sharing of types, and ensure that the type names for the shared types have shorter meaningful prefixes.

At definition type the user can specify the type name to generate, which can be very short, and replace the automatically implied ReallyLongPrefixedTypeName like this:

```go
"visuals": {
	Elem: &info.Schema{
		TypeName: tfbridge.Ref("Visual"),
	},
},
```

At reference time in another resource, the user can reuse an already generated type by token. This already worked before this change but had the downside of still generating unused helper types and causing SDK bloat.

```go
"visuals": {
	Elem: &info.Schema{
		Type:     "testprov:index/Visual:Visual",
	},
},
```

With this change it is possible to instruct the bridge to stop generating the unused helper types:

```go
"visuals": {
	Elem: &info.Schema{
		Type:     "testprov:index/Visual:Visual",
		OmitType: true
	},
},
```

### DIFF
--- a/pkg/tfbridge/info/info.go
+++ b/pkg/tfbridge/info/info.go
@@ -448,6 +448,8 @@ type Schema struct {
 
 	// Used together with [Type] to omit generating any Pulumi types whatsoever for the current property, and
 	// instead use the object type identified by the token setup in [Type].
+	//
+	// It is an error to set [OmitType] to true without specifying [Type].
 	OmitType bool
 
 	// alternative types that can be used instead of the override.
--- a/pkg/tfbridge/info/info.go
+++ b/pkg/tfbridge/info/info.go
@@ -450,6 +450,8 @@ type Schema struct {
 	// instead use the object type identified by the token setup in [Type].
 	//
 	// It is an error to set [OmitType] to true without specifying [Type].
+	//
+	// Experimental.
 	OmitType bool
 
 	// alternative types that can be used instead of the override.
@@ -517,6 +519,8 @@ type Schema struct {
 	// prefix and the name of the property. Use [TypeName] to override this decision when the default names for
 	// nested properties are too long or otherwise undesirable. The choice will further affect the automatically
 	// generated names for any properties nested under the current one.
+	//
+	// Experimental.
 	TypeName *string
 }
 
--- a/pkg/tfbridge/info/info.go
+++ b/pkg/tfbridge/info/info.go
@@ -520,6 +520,14 @@ type Schema struct {
 	// nested properties are too long or otherwise undesirable. The choice will further affect the automatically
 	// generated names for any properties nested under the current one.
 	//
+	// Example use:
+	//
+	//     TypeName: tfbridge.Ref("Visual")
+	//
+	// Note that the type name, and not the full token like "aws:quicksight/Visual:Visual" is specified. The token
+	// will be picked based on the current module ("quicksight" in the above example) where the parent resource or
+	// data source is found.
+	//
 	// Experimental.
 	TypeName *string
 }
--- a/pkg/tfbridge/info/info.go
+++ b/pkg/tfbridge/info/info.go
@@ -509,11 +509,13 @@ type Schema struct {
 	// whether or not to treat this property as secret
 	Secret *bool
 
-	// If specified, resets the type name prefix for any types that Pulumi needs to generate to represent the schema
-	// of the current property. Normally the names for helper object types are built from concatenating fragments
-	// representing the path to the type in the schema. The default strategy can lead to unacceptably long type
-	// names, reducing the SDK usability. Using TypePrefixOverride allows the maintainer to get shorter type names.
-	TypePrefixOverride *string
+	// Specifies the exact name to use for the generated type.
+	//
+	// When generating types for properties, by default Pulumi picks reasonable names based on the property path
+	// prefix and the name of the property. Use [TypeName] to override this decision when the default names for
+	// nested properties are too long or otherwise undesirable. The choice will further affect the automatically
+	// generated names for any properties nested under the current one.
+	TypeName *string
 }
 
 // Config represents a synthetic configuration variable that is Pulumi-only, and not passed to Terraform.
--- a/pkg/tfbridge/info/info.go
+++ b/pkg/tfbridge/info/info.go
@@ -441,8 +441,14 @@ type Schema struct {
 	// a name to override the default when targeting C#; "" uses the default.
 	CSharpName string
 
-	// a type to override the default; "" uses the default.
+	// An optional Pulumi type token to use for the Pulumi type projection of the current property. When unset, the
+	// default behavior is to generate fresh named Pulumi types as needed to represent the schema. To force the use
+	// of a known type and avoid generating unnecessary types, use both [Type] and [OmitType].
 	Type tokens.Type
+
+	// Used together with [Type] to omit generating any Pulumi types whatsoever for the current property, and
+	// instead use the object type identified by the token setup in [Type].
+	OmitType bool
 
 	// alternative types that can be used instead of the override.
 	AltTypes []tokens.Type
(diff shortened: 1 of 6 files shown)


t0yv0:
This does not check that the underlying type is structurally equal to the new replacement type. We could add this check.

pkg/tfgen/generate.go
```diff
@@ -549,6 +564,15 @@ func (t *propertyType) equals(other *propertyType) bool {
...
 	}
+	switch {
+	case t.typePrefixOverride != nil && other.typePrefixOverride == nil:
+		return false
+	case t.typePrefixOverride == nil && other.typePrefixOverride != nil:
+		return false
+	case t.typePrefixOverride != nil && other.typePrefixOverride != nil &&
+		*t.typePrefixOverride != *other.typePrefixOverride:
+		return false
+	}
```
iwahbe, line 575: Nit:

I think this is clearer as an `if` statement. It looks odd to me to have 3 branches that all lead to the same body.

```go
	if (t.typePrefixOverride != nil && other.typePrefixOverride == nil) ||
		(t.typePrefixOverride == nil && other.typePrefixOverride != nil) ||
		(t.typePrefixOverride != nil && other.typePrefixOverride != nil &&
			*t.typePrefixOverride != *other.typePrefixOverride) {
		return false
	}
```

That said, this is just an inequality check made painful by nil values. Nitpicking, I would express this as the inverse of an equality check:

```suggestion
if eq := t.typePrefixOverride == other.typePrefixOverride || 
	(t.typePrefixOverride != nil && other.typePrefixOverride != nil &&
		*t.typePrefixOverride == *other.typePrefixOverride); !eq {
	return false
}
```

pkg/tfbridge/info/info.go
```diff
@@ -441,9 +441,15 @@ type Schema struct {
...
 
-	// a type to override the default; "" uses the default.
+	// An optional Pulumi type token to use for the Pulumi type projection of the current property. When unset, the
+	// default behavior is to generate fresh named Pulumi types as needed to represent the schema. To force the use
+	// of a known type and avoid generating unnecessary types, use both [Type] and [OmitType].
 	Type tokens.Type
 
+	// Used together with [Type] to omit generating any Pulumi types whatsoever for the current property, and
+	// instead use the object type identified by the token setup in [Type].
+	OmitType bool
```
iwahbe, line 451: Let's constrain our value space as much as possible here:
```suggestion
	// Used together with [Type] to omit generating any Pulumi types whatsoever for the current property, and
	// instead use the object type identified by the token setup in [Type].
	//
	// It is an error to set [OmitType] to true without specifying [Type].
	OmitType bool
```

pkg/tfgen/generate.go
```diff
@@ -456,10 +461,20 @@ func getDocsFromSchemaMap(key string, schemaMap shim.SchemaMap) string {
 func (g *Generator) makeObjectPropertyType(typePath paths.TypePath,
 	res shim.Resource, info *tfbridge.SchemaInfo,
 	out bool, entityDocs entityDocs) (*propertyType, error) {
+
+	// If the user supplied an explicit Type token override, omit generating types and short-circuit.
+	if info != nil && info.OmitType && info.Type != "" {
+		return &propertyType{typ: info.Type}, nil
+	}
```
iwahbe, line 468:
```suggestion
	if info != nil && info.OmitType {
		if info.Type == "" {
			return nil, errors.New("Cannot set info.OmitType without setting info.Type")
		}
		return &propertyType{typ: info.Type}, nil
	}
```

pkg/tfgen/generate_schema_test.go
```diff
@@ -116,6 +121,167 @@ func TestCSharpMiniRandom(t *testing.T) {
 	bridgetesting.AssertEqualsJSONFile(t, "test_data/minirandom-schema-csharp.json", schema)
 }
 
+// Test the ability to force type sharing. Some of the upstream providers generate very large concrete schemata by in
+// Go, with TF not being materially affected. The example is inspired by QuickSight types in AWS. In Pulumi the default
+// projection is going to generate named types for every instance of the shared schema. This may lead to SDK bloat. Test
+// the ability of the provider author to curb the bloat and force an explit sharing.
```
iwahbe, line 127:
```suggestion
// Test the ability to force type sharing. Some of the upstream providers generate very large concrete schemata in
// Go, with TF not being materially affected. The example is inspired by QuickSight types in AWS. In Pulumi the default
// projection is going to generate named types for every instance of the shared schema. This may lead to SDK bloat. Test
// the ability of the provider author to curb the bloat and force an explit sharing.
```

iwahbe (commented):
This is a very surgical change, which is great. I would mark the new fields on `info.Schema` as experimental in their respective doc comments, just to hedge against future iteration on the design.

VenelinMartinov (approved):
Very nice! You mentioned tools to help provider maintainers spot these repeated types in the schemas. Is the plan to add these to upstream upgrades or tfgen or something else?

t0yv0:
I've spent a lot of time last night getting Quicksight sharing rolled out to AWS against this functionality and found it can be made to work but it is quite awkward. 

#4449

This PR reduces Quicksight from 30,000 to 7,000 types - there is still some work to specify sharing. 

Couple of learnings there:

1. the user might want to recent the name of the auto-generated type and not just the prefix, since it is otherwise hard to predict the sharing token.

2. perhaps the better interface would be:

```
{ ShareTypeAs: "<token>" }
```

In AWS I built a stateful facade that emulates this by emitting `{Prefix:""}` on the first invocation and `{Type: "<token>", OmitType: true}`.

3. I've been relying heavily on a shared type detector from a different PR that operates on TypeSpec equality. It sounds like we could really use shim.Schema equality instead. 

4. Once the provider builder confirms that certain types are shared, it can be extremely daunting to specify chasing down all the copies in complicated cases like Quicksight. 

Perhaps we can automate the following flow. 

1. `go generate` to compute a type sharing specification and check in as a yaml file. This specification specifies where the shared types are found and what are their suggested tokens. 

2. Provider maintainer audits this to rename tokens as needed and "accept" that the discovered sharing is desirable, e.g. confirm that the identical types were generated from a shared func() invocation upstrema. 

3. At resources.go time the sharing specification is loaded from disk and executed against ProviderInfo to find all copies of shared root types and specified in the corresponding info.Schema ; then tfgen proceeds to generate a schema with sharing

A question is what to do with dynamically bridged providers. It might be desirable there, since they are subject to type explosion as well. I think once we are certain of this functionality, we should default to detecting sharing in dynamically bridged providers, picking arbitrary names for the shared types.

pkg/tfgen/generate.go
t0yv0: This is an F# habit  that may be out of place but I find individual cases more tractable than a large bool expr.

iwahbe (commented):
LGTM

pkg/tfbridge/info/info.go
```diff
@@ -502,6 +512,16 @@ type Schema struct {
...
 	Secret *bool
+
+	// Specifies the exact name to use for the generated type.
+	//
+	// When generating types for properties, by default Pulumi picks reasonable names based on the property path
+	// prefix and the name of the property. Use [TypeName] to override this decision when the default names for
+	// nested properties are too long or otherwise undesirable. The choice will further affect the automatically
+	// generated names for any properties nested under the current one.
+	//
+	// Experimental.
```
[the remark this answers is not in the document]
t0yv0: Good idea, added.

iwahbe (approved):

PR #2419, opened 2024-09-17
Add an Overview subheader to logically separate Installation and general provider info

guineveresaenger:
This pull request is the result of discovering #2420.

This fix:

- Avoids the rendering issues for unheadered text as outlined in #2420 by adding a `##` header
- Visually and logically separates the provider summary text from the installation section.
- Strips the H1 header earlier in the doc transformation so we can verify remaining content
- Strips the "schema generated by tfplugindocs" HTML comment

TL;DR: while this is on its face a quick fix, we _do_ want to separate the installation from the general text, so adding the Overview header is something we'd want to do anyway.

Screenshots:

Before:
Screenshot 2024-09-17 at 11 16 15 AM

After:
Screenshot 2024-09-17 at 11 16 34 AM


iwahbe (changes requested):
We need to make the overview section conditional on the remaining body not starting with a H2. We don't want to add:

```md
## Overview

## Examples
```

iwahbe (commented):
Can we add a test that shows we drop `## Overview` it has no content.

guineveresaenger:
`## Overview` doesn't get written if there's no content. https://github.com/pulumi/pulumi-terraform-bridge/pull/2419/files#diff-dbba61023eb9aed059bd34c79844b0f76a82c2a8b296702b9a56cae17ad35a41 should show that but I see how the input doc isn't too descriptive of what the desired behavior is. I've added another test that verifies the output with a bit more context.

pkg/tfgen/installation_docs_test.go
```diff
@@ -157,6 +157,57 @@ func TestWriteInstallationInstructions(t *testing.T) {
 	})
 }
 
+func TestWriteOverviewHeader(t *testing.T) {
```
iwahbe, line 160: I'd love if this was a bit higher level, probably by the scope of one function. We shouldn't need to add the header to the body in the test content.

pkg/tfgen/installation_docs_test.go
```diff
@@ -157,6 +157,57 @@ func TestWriteInstallationInstructions(t *testing.T) {
...
+		name:     "Does Not Write For Empty Overview With Context",
+		input:    readTestFile(t, "write-overview-header/empty-overview-with-context-input.md"),
+		expected: readTestFile(t, "write-overview-header/empty-overview-with-context-expected.md"),
+	}
+	t.Run(contextTest.name, func(t *testing.T) {
+		t.Parallel()
+		text, err := removeTitle([]byte(contextTest.input))
+		require.NoError(t, err)
+		header := writeOverviewHeader(text)
+		actual := header + string(text)
```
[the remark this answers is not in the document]
guineveresaenger: This follows how the surrounding code operates here. It mirrors the context surrounding this function, so we can test on an actual input file, rather than just assert that `write/getOverview` does the right thing.

pkg/tfgen/installation_docs_test.go
guineveresaenger: The next level up is testing the whole file conversion. I will add this in another pull request today.
Adding the header to the body here provides more surrounding context, which I was hoping would provide more clarity.

PR #2421, opened 2024-09-17
Add more regex filters to installation docsgen

guineveresaenger:
Rather than handling these in individual providers, they're best served here.

Adds a few test cases to guard against any regression.

- **Match TF mentions found in gitlab provider and add test cases**
- **Add a header to skip as seen in hcloud**

### DIFF
--- a/pkg/tfgen/installation_docs.go
+++ b/pkg/tfgen/installation_docs.go
@@ -401,6 +401,7 @@ func getDefaultHeadersToSkip() []*regexp.Regexp {
 		regexp.MustCompile("[Dd]ebugging"),
 		regexp.MustCompile("[Tt]erraform CLI"),
 		regexp.MustCompile("[Tt]erraform Cloud"),
+		regexp.MustCompile("Delete Protection"),
 	}
 	return defaultHeaderSkipRegexps
 }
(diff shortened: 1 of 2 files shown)


pkg/tfgen/installation_docs.go
```diff
@@ -401,14 +401,16 @@ func getDefaultHeadersToSkip() []*regexp.Regexp {
...
 		regexp.MustCompile("[Tt]erraform Cloud"),
+		regexp.MustCompile("Delete Protection"),
 	}
 	return defaultHeaderSkipRegexps
 }
 
 func getTfVersionsToRemove() []*regexp.Regexp {
 	tfVersionsToRemove := []*regexp.Regexp{
-		regexp.MustCompile(`It requires [tT]erraform [v0-9]+\.?[0-9]?\.?[0-9]? or later.`),
+		regexp.MustCompile(`(It|This provider) requires(.*|at least) [tT]erraform [v0-9]+\.?[0-9]?\.?[0-9]?(.*|or later).`),
```
iwahbe, line 411:
```suggestion
		regexp.MustCompile(`(It|This provider) requires( at least)? [tT]erraform [v0-9]+\.?[0-9]?\.?[0-9]?( or later)?.`),
```

I think that `(.*|at least)` is match anything or "at least", which I think is equivalent to `.*`. I suspect the more specific clause never triggers, since the spacing is wrong.

iwahbe (approved):
I'd rather see tests with a bit more context (not just the desired string, into nothing).

PR #2423, opened 2024-09-17
fix: remove frontmatter whitespace when trimming

guineveresaenger:
[Some providers](https://github.com/CiscoDevNet/terraform-provider-ise/blob/main/docs/index.md?plain=1#L1C1-L14C1) have poorly formatted files. We can fix this by removing the whitespace before trimming off the remaining frontmatter.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -373,7 +373,9 @@ var (
 )
 
 func trimFrontMatter(text []byte) []byte {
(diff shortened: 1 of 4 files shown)


iwahbe (commented):
Tests show that this effects the output far more then I would expect. I think we need to drill into that before merging.

guineveresaenger:
Hm, this is truly unexpected. I can strip the whitespace separately to avoid Jenga, but oof...?

PR #2424, opened 2024-09-18
Only check mapped resources

iwahbe:
We should only validate resources we will use. Discovered as part of https://github.com/pulumi/pulumi-databricks/issues/595.


VenelinMartinov (approved):
lgtm

PR #2425, opened 2024-09-18
Update pu/pu to v3.133.0

iwahbe:


t0yv0 (approved):
Nice, thanks! We used to have automation for this might worth checking up on but I guess you already have.

iwahbe:
> Nice, thanks! We used to have automation for this might worth checking up on but I guess you already have.

It runs on Thursday, but I was hoping this could roll out today. I don't believe that our tooling does anything to fix breaking changes.

t0yv0:
AH no, there were breaking changes. Missed that. Indeed.

guineveresaenger (approved):

PR #2435, opened 2024-09-20
Add option to filter edit rules by location in docs conversion pipeline and apply for installation docs

guineveresaenger:
This code is required to successfully apply the edit rules needed for https://github.com/pulumi/pulumi-libvirt/pull/426.

It covers the case for when the upstream file's code sections need to be adjusted for code conversion to work.

It adds a complete index file test which tests the entire `plainDocsParser` function.
In support of https://github.com/pulumi/home/issues/3598.

- **Run provider-supplied edit rules before example conversion. Add complete file tests, with and without provider-supplied edits**
- **clean up tests**


pkg/tfgen/installation_docs.go
```diff
@@ -128,17 +138,25 @@ func stripSchemaGeneratedByTFPluginDocs(content []byte) []byte {
...
+	contentBytes, err := edits.apply(docFile, contentBytes)
+	if err != nil {
+		return nil, err
+	}
+	return contentBytes, nil
+}
+func applyInstallationEditRules(contentBytes []byte, docFile string) ([]byte, error) {
+	var edits editRules
 	// Additional edit rules for installation files
 	edits = append(edits,
```
iwahbe, line 153: If we have a fixed set of edit rules, we don't need to `append`.

```suggestion
	edits := editRules{
```

iwahbe (commented):
When edit rules are applied is important. Can we add a note to `info.DocsEdit` that tells the user "your edit will be applied before any other edits"?

I think the change is OK as is, but it also feels like this makes edit rules application even more complicated and special for installation docs.

We can merge as is if we really need to, but I'd love if we took the opportunity to re-unify behavior for edit rules. For that, see 👇.

---

Instead of making bridge supplied edit rules special, I propose that we:

1. Add a field to `info.DocsEdit` to control where the edit rule is applied. Initially, I would suggest an enum with: `EditPreTranslation` (the zero value, since this is what we do for normal pages) and `EditPostTranslation`.
2. Edit `apply` to also take a phase (pre or post) and then filter edit rules applied.
3. Move the existing "special" edit rules for installation docs into `defaultEditRules` (in edit_rules.go), specifying `EditPostTranslation` as needed.

guineveresaenger:
Hmm, I like this idea, _and_ I think that edit rules should be applied mostly _after_ code conversion. The main case for applying them pre-conversion is to fix any mistakes in the provided code, as was discovered here. Having both options would not only allow us to fix incorrect TF code and make it convertable (convertible? 🚗 ), but also give us far more freedom in translating "Terraform" strings for "Pulumi" ones.

pkg/tfbridge/info/info.go
```diff
@@ -251,7 +258,8 @@ type DocsEdit struct {
 	// The function that performs the edit on the file bytes.
 	//
 	// Must not be nil.
-	Edit func(path string, content []byte) ([]byte, error)
+	Edit  func(path string, content []byte) ([]byte, error)
+	Phase EditPhase
```
iwahbe, line 262:
```suggestion
	// Phase determines when the edit rule will run.
	//
	// The default phase is [PreCodeTranslation].
	Phase EditPhase
```

pkg/tfbridge/info/info.go
```diff
@@ -233,6 +233,13 @@ type DocsPath struct {
 	TfToken string
 }
 
+type EditPhase int
+
+const (
+	PreCodeTranslation EditPhase = iota
+	PostCodeTranslation
```
iwahbe, line 240: Any new object we export should have a comment explaining what it is *and* how to use it.
```suggestion
	// PreCodeTranslation directs a [DocsEdit] to occur before resource example code is translated.
	PreCodeTranslation EditPhase = iota
	// <useful comment here - links to DocsEdit>
	// TODO[<link appropriate issue>]: Right now, PostCodeTranslation is only called on installation docs.
	PostCodeTranslation
```

pkg/tfgen/edit_rules.go
```diff
@@ -29,24 +30,52 @@ func defaultEditRules() editRules {
...
 		fixupImports(),
 		// Replace content such as "jdoe@hashicorp.com" with "jdoe@example.com"
-		reReplace("@hashicorp.com", "@example.com"),
+		reReplace("@hashicorp.com", "@example.com", info.PreCodeTranslation),
+
+		// The following edit rules may be applied after translating the code sections in a document.
+		// Its primary use case is for the docs translation approach spearheaded in installation_docs.go.
+		// These edit rules allow us to safely transform certain strings that we would otherwise need in the
+		// code translation or nested type discovery process.
+		// These rules are currently only called when generating installation docs.
```
iwahbe, line 44: Can we create a TODO here and link it to an issue.

```suggestion
		//
		// These rules are currently only called when generating installation docs. 
		// TODO[<issue here>]: Call info.PostCodeTranslation rules on all docs.
```

iwahbe (approved):
I have some comment nits, but otherwise LGTM

PR #2438, opened 2024-09-24
Revert "Remove vendored plugin server"

iwahbe:
Reverts pulumi/pulumi-terraform-bridge#2410


t0yv0 (approved):
Happy to revert, hoping it's fairly self-contained.

PR #2456, opened 2024-09-30
Add more shim representation tests

t0yv0:
This PR simply adds test cases around tfgen and shim representations of various schemata found in SDKv2 and PF. For PF I went through the docs to cover each major advertised case. What's of special interest is pinning down when Pulumi generates object types. 

It seems to be happening in two cases:

1. when TF specifies object types directly in PF: 
       - Single Nested Attribute
       - Object Attribute
       - Single Nested Block

2. when TF specifies a block with MaxItems=1 that is a list or set in TF but is flattened to an object in Pulumi:
       - In SDKv2 this is typically a list-nested block with MaxItems=1, very common
       - In PF this may be a block with List Validators indicating that at most 1 item is supported; seems rare

It seems helpful to pin down what shim.Schema captures in each of these cases.

### DIFF
--- a/pf/tests/schemashim_test.go
+++ b/pf/tests/schemashim_test.go
@@ -14,15 +14,25 @@
 
 package tfbridgetests
 
+// Test how various PF-based schemata translate to the shim.Schema layer. Excerpts of the resulting Pulumi Package
+// Schema are included for reasoning convenience.
+//
+// References:
+//
+// https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes
+// https://developer.hashicorp.com/terraform/plugin/framework/handling-data/blocks
+
 import (
 	"context"
 	"encoding/json"
 	"io"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


t0yv0:
Guess could be interesting to add for PF set-of-object and list-of-object cases for attributes.

t0yv0:
Hmm linter is happy locally but not in CI. Perhaps not aligned on linter versions. 

```
 ::high file=pkg/tfshim/sdk-v2/cty.go,line=164,col=32::G115: integer overflow conversion uint -> int64 (gosec)
  
  level=warning msg="[config_reader] The output format `github-actions` is deprecated, please use `colored-line-number`"
```

VenelinMartinov (approved):
LGTM.

Linking https://github.com/pulumi/pulumi-terraform-bridge/pull/2231 as it covered some of the same grounds and had some interesting observations on similarities/differences between various TF Types represented in the Shim layer.

iwahbe (approved):

Issue #1922, opened 2024-05-03
Decoupling from pulumi/pkg dependency


t0yv0:
Logging a conversation on what would it take to not build-depend on pulumi/pkg as alternative to https://github.com/pulumi/pulumi/issues/12139

Lighter build-dependencies mean fewer issues with conflicts and faster builds, but likely do not affect the quality of end-user artifacts since Go does tree-shaking well at build time.

Overview of existing references:

```
Exported grep results:

pkg/tfbridge/x/tokens.go:28:	"github.com/pulumi/pulumi/pkg/v3/codegen/cgstrings"
pkg/tfbridge/tokens/tokens.go:24:	"github.com/pulumi/pulumi/pkg/v3/codegen/cgstrings"
pkg/tfbridge/serve.go:20:	"github.com/pulumi/pulumi/pkg/v3/resource/provider"
pkg/tfbridge/info/provider.go:20:	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
pkg/tfbridge/info/provider.go:21:	"github.com/pulumi/pulumi/pkg/v3/resource/provider"
pkg/tfbridge/info/info.go:26:	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
pkg/tfbridge/info/info.go:27:	"github.com/pulumi/pulumi/pkg/v3/resource/provider"
pkg/tfbridge/validate_input_types.go:7:	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
pkg/tfbridge/log.go:25:	"github.com/pulumi/pulumi/pkg/v3/resource/provider"
pkg/tfbridge/provider.go:27:	pschema "github.com/pulumi/pulumi/pkg/v3/codegen/schema"
pkg/tfbridge/provider.go:43:	"github.com/pulumi/pulumi/pkg/v3/resource/provider"
pkg/tf2pulumi/convert/eject.go:28:	"github.com/pulumi/pulumi/pkg/v3/codegen/convert"
pkg/tf2pulumi/convert/eject.go:29:	"github.com/pulumi/pulumi/pkg/v3/codegen/hcl2/syntax"
pkg/tf2pulumi/convert/eject.go:30:	"github.com/pulumi/pulumi/pkg/v3/codegen/pcl"
pkg/tf2pulumi/convert/eject.go:31:	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
[...]

PR #2483, opened 2024-10-14
Add HasPulumiProviderName check for use in terraform converter.

brandonpollack23:
This will be used in GetProviderInfo in converter-terraform.

In that case, the main.tf package info will be used to get the repo and package in the open tofu repositories (as opposed to just the raw name) to be passed to the GetMapper command.

### DIFF
--- a/pkg/tf2pulumi/il/plugin_info.go
+++ b/pkg/tf2pulumi/il/plugin_info.go
@@ -158,6 +158,12 @@ var pulumiNames = map[string]string{
 	"template": "terraform-template",
 }
 
+// HasPulumiProviderName returns true if the given Terraform provider has a corresponding Pulumi provider name.
+func HasPulumiProviderName(terraformProviderName string) bool {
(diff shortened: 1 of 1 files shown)


t0yv0:
Hi @brandonpollack23 we're trying to deprecate and remove pkg/tf2pulumi so that https://github.com/pulumi/pulumi-converter-terraform does not depend on the bridge. Would that be possible? Or perhaps this PR is needed as part of that larger change?

t0yv0:
Looks like some balls got dropped on this but https://github.com/pulumi/pulumi-terraform-bridge/pull/2004 was one attempt. This decoupling can also be instrumental in https://github.com/pulumi/pulumi-terraform-bridge/issues/1922

https://github.com/pulumi/pulumi-terraform-bridge/issues/2037 completed internally for Pulumi-managed providers but we still need to do a small rollout to external stakeholders to make this the default behavior.

PR #2490, opened 2024-10-16
More info on compute ID mapping error

guineveresaenger:
This is a papercut tweak to remind provider maintainers on what to do for ComputeID mapping errors.

I'm happy if we don't take this; but it confused me in a recent upgrade so I thought I'd address it.

### DIFF
--- a/docs/resource-ids.md
+++ b/docs/resource-ids.md
@@ -60,7 +60,21 @@ If the type of the `"id"` attribute is not coercible to a string, you must set `
 error: Resource test_res has a problem: no "id" attribute. To map this resource consider specifying ResourceInfo.ComputeID
 ```
 
-If the resource simply doesn't have an `"id"` attribute, you will need to set `ResourceInfo.ComputeID`. If you want to delegate the ID field in Pulumi to another attribute, you should use `tfbridge.DelegateIDField` to produce a `ResourceInfo.ComputeID` compatible function. Otherwise you can pass in any function that complies with:
+If the resource simply doesn't have an `"id"` attribute, you will need to set `ResourceInfo.ComputeID`. 
+
+```go
+"test_res": {ComputeID: computeIDField("id")}
+```
+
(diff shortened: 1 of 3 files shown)


docs/resource-ids.md
```diff
@@ -60,7 +60,21 @@ If the type of the `"id"` attribute is not coercible to a string, you must set `
 error: Resource test_res has a problem: no "id" attribute. To map this resource consider specifying ResourceInfo.ComputeID
 ```
 
-If the resource simply doesn't have an `"id"` attribute, you will need to set `ResourceInfo.ComputeID`. If you want to delegate the ID field in Pulumi to another attribute, you should use `tfbridge.DelegateIDField` to produce a `ResourceInfo.ComputeID` compatible function. Otherwise you can pass in any function that complies with:
+If the resource simply doesn't have an `"id"` attribute, you will need to set `ResourceInfo.ComputeID`. 
+
+```go
+"test_res": {ComputeID: computeIDField("id")}
```
t0yv0, line 66: I find this confusing, what does `computeIDField` identifier resolve to?

pkg/pf/internal/check/check_test.go
```diff
@@ -100,7 +100,7 @@ func TestMissingIDProperty(t *testing.T) {
 	})
 
 	assert.Equal(t, "error: Resource test_res has a problem: no \"id\" attribute. "+
-		"To map this resource consider specifying ResourceInfo.ComputeID\n", stderr)
+		"To map this resource consider specifying ResourceInfo.ComputeID to a valid field on the upstream resource\n", stderr)
```
t0yv0, line 103: Can be multiple fields. I've found many AWS resources being added have composite IDs https://github.com/pulumi/pulumi-aws/blob/master/provider/resources.go#L5860

t0yv0 (approved):

iwahbe:
Thanks @guineveresaenger.

PR #2492, opened 2024-10-16
Automatically propagate bridge updates on providers when a new bridge version is tagged

guineveresaenger:
This pull request changes the automatic upgrade of providers to run on tagging of the main bridge version, since we are deprecating the `pf` module.


blampe (approved):
Great catch!

PR #2502, opened 2024-10-18
Test int->string type overrides with cross-tests

### LINKED ISSUE #2048: Testing: Add an integration test for type overrides

iwahbe:
Stacked PRs:
 * __->__#2502
 * #2503
 * #2501
 * #2500

--- --- ---

### Test int->string type overrides with cross-tests

Fixes #2048

### DIFF
--- a/pkg/tfbridge/tests/provider_test.go
+++ b/pkg/tfbridge/tests/provider_test.go
@@ -5,20 +5,46 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
+	"strconv"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/pulumi/providertest/replay"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
+	"github.com/zclconf/go-cty/cty"
 
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/internal/tests/cross-tests"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/info"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen"
 	shimv2 "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/sdk-v2"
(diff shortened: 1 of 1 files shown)


VenelinMartinov (commented):
Don't want to block this on the refactoring work.

If you want to, you can implement this as a "manual" cross-test like so: https://github.com/pulumi/pulumi-terraform-bridge/blob/bf282ab5f0663bf7813ed248a4e9c81c36ee4587/pkg/tests/schema_pulumi_test.go#L1593

pkg/tfbridge/tests/provider_test.go
```diff
@@ -5,21 +5,47 @@ import (
...
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen"
 	shimv2 "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/sdk-v2"
 )
 
+func TestIntToStringOverride(t *testing.T) {
+	t.Parallel()
+
+	const largeInt = math.MaxInt64 / 2
+
+	crosstests.Create(t,
```
t0yv0, line 33: GPT conversation.

I am designing a Go library to test that Pulumi and Terraform programs behave identically. I want to design a Golang
method that checks that for a given set of inputs provided by the user, Pulumi and Terraform will create resources with
identical data flowing into the provider code handling resource creation. How can I name this method so the name is
short but meaningful?



Here are a few suggestions for naming the method:

1. =VerifyIdenticalOutputs=
2. =CompareResourceData=
3. =CheckConvergedStates=
4. =ValidateOutputsMatch=
5. =MatchProviderInputs=

Each name emphasizes different aspects, such as verification, comparison, checking, validation, or matching, while aiming to convey the core functionality concisely.

t0yv0: That didn't help. GPT is bad at naming things just as we are. `crosstests.VerifyCreate` possibly.

t0yv0 (approved):

pkg/tfbridge/tests/provider_test.go
iwahbe: Maybe rename `crosstests` to `crosstest`, then it's `crosstest.Create`, which reads well. We can do naming in a follow-up PR. I want to keep the pattern consistent between SDKv2 and PF based cross-tests.

t0yv0: Curious you mention this, why are SDKv2 and PF based cross tests any different? Seems like you can take a somehow derived TF provider (SDKv2 or PF), and some inputs, and cross-check TF vs Pulumi behavior, without caring which one it is.

t0yv0: We probably have some legacy gunk to make it happen though.. but in principle..

t0yv0: NVM, I'm a little tired, perhaps that's exactly what you are doing already

PR #2503, opened 2024-10-18
Implement `runCreateInputCheck` in terms of `Create`

iwahbe:
Stacked PRs:
 * #2502
 * __->__#2503
 * #2501
 * #2500

--- --- ---

### DIFF
--- a/pkg/internal/tests/cross-tests/create.go
+++ b/pkg/internal/tests/cross-tests/create.go
@@ -55,7 +55,10 @@ func Create(
 
 	makeResource := func(writeTo *result) *schema.Resource {
 		return &schema.Resource{
-			Schema: resource,
+			Schema:         resource,
+			SchemaVersion:  opts.stateUpgrader.schemaVersion,
+			StateUpgraders: opts.stateUpgrader.stateUpgraders,
+			Timeouts:       opts.timeouts,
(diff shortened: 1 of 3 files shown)


pkg/internal/tests/cross-tests/create.go
```diff
@@ -103,17 +106,49 @@ func Create(
 	assertCtyValEqual(t, "RawConfig", tfResult.data.GetRawConfig(), puResult.data.GetRawConfig())
 	assertCtyValEqual(t, "RawPlan", tfResult.data.GetRawPlan(), puResult.data.GetRawPlan())
 	assertCtyValEqual(t, "RawState", tfResult.data.GetRawState(), puResult.data.GetRawState())
+
+	for k := range resource {
+		// TODO: make this recursive
```
t0yv0, line 111: What does this mean here? How can we get to 100%?

t0yv0: If I remember right, we are comparing some eeky structures here coming from the TFSDK, these structures resist directly comparing for equality but this is something we need for the purposes of the tests. It is a little tedious to get right but sounds like it should be doable within an hour or two. Am I understanding right? Need a hand here perhaps?

t0yv0 (approved):
Looks like my comments are out of scope for the refactor.

PR #2507, opened 2024-10-22
provider index doc: Skip headers that mention Contributing

guineveresaenger:
This pull request globalizes skipping sections whose headers mention Contributing, in the `docs/_index.md` file.

It does so by adding Contributing/contributing to the denylist of words found in a header in an index.md file.

Example: https://github.com/pulumi/pulumi-vsphere/pull/607


iwahbe (approved):
This is fine, but I think we need a test before we merge.

guineveresaenger:
Can you explain what kind of test you'd like to see? 
The function itself is under test. This just adds a new pattern.

iwahbe:
> What kind of test would you like to see? The function itself is under test. This just adds a new pattern.

I'd like to see a docs page level test that would fail if this PR were reverted. It doesn't need to be compliacated.

PR #2510, opened 2024-10-24
Migrate cross-tests from input_cross_test.go to tests/provider_test.go

iwahbe:
This PR migrates our tests away from input_cross_tests.go to the public `crosstests.Create`, removing `runCreateInputCheck` in the process. The only tests removed were tests of `runCreateInputCheck` itself.


pkg/tfbridge/tests/provider_test.go
```diff
@@ -129,6 +130,363 @@ func TestInputsConfigModeEqual(t *testing.T) {
...
+				ForceNew: true,
+				StateFunc: func(v interface{}) string {
+					return v.(string) + " world"
+				},
+			},
+		},
+		cty.ObjectVal(map[string]cty.Value{
+			"test": cty.StringVal("hello"),
+		}),
+		crosstests.InferPulumiValue(),
```
t0yv0, line 154: Nit: would be cool if the interface was such that when omitted, it infers. So as not to spell this out. But perhaps spelling out is better as it calls attention to this.

pkg/tfbridge/tests/provider_test.go
```diff
@@ -129,6 +130,363 @@ func TestInputsConfigModeEqual(t *testing.T) {
 	}
 }
 
+// TestStateFunc ensures that resources with a StateFunc set on their schema are correctly
```
t0yv0, line 133: Neat, this is neat arrangement of tests.

pkg/tfbridge/tests/provider_test.go
```diff
@@ -129,6 +130,363 @@ func TestInputsConfigModeEqual(t *testing.T) {
...
+				}),
+				crosstests.InferPulumiValue(),
+			)
+		})
+	}
+}
+
+// TestAccCloudWatch failed with PlanResourceChange to do a simple Create preview because the state upgrade was
+// unexpectedly called with nil state. Emulate this here to test it does not fail.
+func TestCreateDoesNotPanicWithStateUpgraders(t *testing.T) {
```
t0yv0, line 228: It is 100% surprising that state upgrade is at all called in Create. A comment should probably address this. Do we believe that TF calls state upgraders on Create an that is why Pulumi does it?

pkg/tfbridge/tests/provider_test.go
```diff
@@ -129,6 +130,363 @@ func TestInputsConfigModeEqual(t *testing.T) {
...
+			{
+				Type:    resourceRuleV0().CoreConfigSchema().ImpliedType(),
+				Upgrade: resourceRuleUpgradeV0,
+				Version: 0,
+			},
+		}),
+	)
+}
+
+func TestTimeouts(t *testing.T) {
```
t0yv0, line 275: This is interesting but this only calls create timeouts. Are we sure that asserts are strong enough in `crosstests` to check that timeouts are propagated similarly? I recall there was quite a complicated story around timeouts. Perhaps out of scope here, but we can x-ref some issues and comment on this.

pkg/tfbridge/tests/provider_test.go
```diff
@@ -129,6 +130,363 @@ func TestInputsConfigModeEqual(t *testing.T) {
...
+			},
+		},
+		cty.ObjectVal(map[string]cty.Value{
+			"f0": cty.StringVal(""),
+		}),
+		crosstests.InferPulumiValue(),
+	)
+}
+
+func TestInputsNesqtedBlocksEmpty(t *testing.T) {
```
t0yv0, line 355: Typo.

t0yv0 (approved):

pkg/tfbridge/tests/provider_test.go
iwahbe: I'm not 100% sure. I'll add an explicit timeout validation block in `assertResourceDataEqual` to verify.

t0yv0: Could also be just something to backlog into the quality epic.

pkg/tfbridge/tests/provider_test.go
iwahbe: New test looks like this:

```go
// TestCreateDoesNotInvokeStateUpgraders ensures that state upgrade machinery is not
// invoked during Create operations.
func TestCreateDoesNotInvokeStateUpgraders(t *testing.T) {
	t.Parallel()
	resource := func() *schema.Resource {
		return &schema.Resource{Schema: map[string]*schema.Schema{
			"f0": {
				Type:     schema.TypeString,
				Optional: true,
			},
		}}
	}

	upgradeFunc := func(ctx context.Context, rawState map[string]any, meta any) (map[string]any, error) {
		panic("State upgraders should not be called during create")
	}

	crosstests.Create(t,
		resource().Schema,
			"f0": cty.StringVal("default"),
		}),
		crosstests.InferPulumiValue(),
		crosstests.CreateStateUpgrader(1, []schema.StateUpgrader{
			{
				Type:    resource().CoreConfigSchema().ImpliedType(),
				Upgrade: upgradeFunc,
				Version: 0,
			},
		}),
	)
}
```

pkg/tfbridge/tests/provider_test.go
iwahbe: I prefer spelling it out, but you are now the second person to make that comment, so I'll reconsider. That will be in a separate PR.

PR #2511, opened 2024-10-24
Fix shims

blampe:
## Problem
When running `go list -m` in a provider upgraded to use bridge v3.93.1 you will get this error: 

```
go: github.com/hashicorp/terraform-provider-tls@v0.0.0: invalid version: unknown revision v0.0.0
go: github.com/hashicorp/terraform-provider-tls/shim@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
go: github.com/terraform-providers/terraform-provider-random/randomshim@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

Ian called this out in https://github.com/pulumi/pulumi-terraform-bridge/pull/2473#discussion_r1799354257. My response there is true in the sense that providers _do still build_ just fine -- `go build`, `go test` etc. don't touch these unused modules. The caveat I overlooked is that some IDE integrations (or more specifically any tooling built around `go list`) _will_ try to resolve all of these direct and indirect dependencies.

When we consolidated modules, we moved `replace` directives for these shims out of the test module and into the root module.

## Option 0: Bring back all the modules

I beg you not to do this but I do want to call it out as an option if the team would prefer it. I outline two more "idiomatic" options below.

## Option 1: Handle replacements in the consumer

As a reminder, the `replace` directive only applies to the module where it's defined -- primarily to keep Go's minimal version selection algorithm simple and fast, but also to give downstream modules complete control of their builds. Hence `replace` is best suited for temporary overrides, for example to point at a local branch.

Permanent `replace` directives do happen and we use them quite a lot, though. If a module truly needs a permanent `replace` to work properly (which, to be clear, should be avoided!), its consumers need to mirror the same `replace`. 

In other words, Option 1 is to not make any changes. Downstream projects can resolve this issue by copying these lines into `go.mod`:

```
replace github.com/hashicorp/terraform-provider-tls/shim => github.com/pulumi/pulumi-terraform-bridge/v3/pkg/pf/tests/internal/tlsshim v0.0.0-20241025174446-f865647b85ed

replace github.com/hashicorp/terraform-provider-tls => github.com/hashicorp/terraform-provider-tls v1.2.1-0.20210201152640-49d7590ccf3f

replace github.com/terraform-providers/terraform-provider-random/randomshim => github.com/pulumi/pulumi-terraform-bridge/v3/pkg/pf/tests/internal/randomshim v0.0.0-20241025174446-f865647b85ed
```

Clearly this is burdensome. For a real-world data point, this is the same strategy adopted by the k8s ecosystem and it has made dependency management with k8s notoriously painful.

We should not do this.

## Option 2: Fork instead of abusing modules

`tlsshim` and `randomshim` are their own modules because they intentionally trick the compiler into exposing internal terraform packages. This `internal` import is only legal if your module is identified as `github.com/terraform-providers/terraform-provider-random`:

https://github.com/pulumi/pulumi-terraform-bridge/blob/7dfbcd601c7392fbf04bdec4c83a56b1ef115bc6/pkg/pf/tests/internal/randomshim/shim.go#L17-L24

This is hacky but it matches the guidance we gives to users, and it is _workable_ as long as the module consuming these shims doesn't have downstream consumers of its own. This is the case with most providers, but tfbridge is different because it _is_ meant to be consumed as a library.

A more typical remedy for consuming private types is to fork the dependency and expose what you need. We could fork these two repositories under the `pulumi` org, but new repos come with their own overhead.

Instead, I opted to simply check-in the upstream random and tls providers. I preserved the commits we were previously pinned to ([c6e90de46687](https://github.com/hashicorp/terraform-provider-random/commit/c6e90de466877b3b0d6dd39c7573c1cfedf0a58d) and [afdd54107aba](https://github.com/hashicorp/terraform-provider-tls/commit/afdd54107aba371f7e030b2d935cceb1a6050fe9), respectively) so licensing is unaffected.

This way our `go.mod` will reflect a dependency on a public module that is still consumable downstream (e.g. `terraform-provider-random`), and we will use forks that expose the internals when testing.

This PR implements Option 2.


t0yv0:
This is fine in my book. Also it might be possible to drop the tests that depend on these without meaningful loss of coverage. This was developed before we had a system going that qualifies bridge releases  through checking if TLS and Random etc test will pass on these, and before we had any significant integration tests written against test providers. IN the long term specifically crafted test providers are the way. This is fine for now though AFAIK.

go.mod
```diff
@@ -51,7 +51,7 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/stretchr/testify v1.9.0
 	github.com/teekennedy/goldmark-markdown v0.3.0
-	github.com/terraform-providers/terraform-provider-random/randomshim v0.0.0-00010101000000-000000000000
```
t0yv0, line 54: Will actual random provider build with this fine? It has its own dependency on this, and its own linker trick.

t0yv0 (approved):

iwahbe (approved):
Let's take this change now, since we want a release that unbreaks users ASAP.

I agree with @t0yv0: We can stop relying on these tests soon without meaningful loss of coverage.

guineveresaenger (approved):
I would deeply love this change and I would also like a followup issue that tracks the consideration of removing the tests that depend on this code in favor of more modern testing options.

blampe:
Created https://github.com/pulumi/pulumi-terraform-bridge/issues/2518 to track the followup.

PR #2519, opened 2024-10-28
Enable typechecker to allow stringified boolean property values on Type: "boolean"

### LINKED ISSUE #2339: Type checking failure -- "skipMetadataApiCheck"

guineveresaenger:
This pull request allows for strings called "true" or "false" that are set on the "boolean" type to pass the type checker.

This change is necessary because the JSON-style config encoding, which powers the TypeScript, Python, .NET, and Java SDKS,
returns a string for config values with a type override to "boolean". The engine can handle this regarding functionality,
but the type checker sends an unnecessary warning to affected SDKs.

Fixes #2339

For additional context, see changes made in https://github.com/pulumi/pulumi-terraform-bridge/pull/2310.

### DIFF
--- a/pkg/tfbridge/typechecker/typechecker.go
+++ b/pkg/tfbridge/typechecker/typechecker.go
@@ -162,7 +162,8 @@ func (v *TypeChecker) validatePropertyValue(
 	switch typeSpec.Type {
 	case "boolean":
 		// Check for strings that are values "true" or "false".
-		// These are handled as booleans in the bridge, so they should be skipped by the type checker.
+		//TODO: Remove the boolString condition when https://github.com/pulumi/pulumi-terraform-bridge/issues/2520
+		// is resolved. This is a workaround for the config encoding not honoring type overrides.
 		var boolString bool
 		if propertyValue.IsString() {
 			if propertyValue.StringValue() == "true" || propertyValue.StringValue() == "false" {
--- a/pkg/tfbridge/typechecker/typechecker.go
+++ b/pkg/tfbridge/typechecker/typechecker.go
@@ -161,8 +161,15 @@ func (v *TypeChecker) validatePropertyValue(
 
 	switch typeSpec.Type {
 	case "boolean":
-		// The bridge permits the strings "true" and "false" to read as boolean, so allow strings.
-		if !propertyValue.IsBool() && !propertyValue.IsString() {
+		// Check for strings that are values "true" or "false".
+		// These are handled as booleans in the bridge, so they should be skipped by the type checker.
+		var boolString bool
+		if propertyValue.IsString() {
+			if propertyValue.StringValue() == "true" || propertyValue.StringValue() == "false" {
+				boolString = true
+			}
+		}
+		if !propertyValue.IsBool() && !boolString {
 			return []Failure{newTypeFailure(propertyPath, typeSpec.Type, propertyValue)}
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/typechecker/typechecker.go
```diff
@@ -161,7 +161,15 @@ func (v *TypeChecker) validatePropertyValue(
 
 	switch typeSpec.Type {
 	case "boolean":
-		if !propertyValue.IsBool() {
+		// Check for strings that are values "true" or "false".
```
t0yv0, line 164: Let's try to avoid leaving confusing comments for future maintainers in the source. 

"These are handled as booleans in the bridge, so they should be skipped by the type checker."

This is not right. https://github.com/pulumi/pulumi-terraform-bridge/issues/2520 is the actual root cause that trips up the type-checker, and it would be great to have a TODO comment linking the issue in here and explaining the workaround.

pkg/tfbridge/typechecker/typechecker_test.go
```diff
@@ -1631,6 +1631,11 @@ func TestValidateConfigType(t *testing.T) {
 				})),
 			}),
 		},
+		{
```
t0yv0, line 1634: The test isn't great. It exercises the changed code path workaround which is good but it does assert something that will need to be retracted once the root cause is fixed. I think if we add a todo to the root cause ticket in here that'll help us remember later on. Thank you!

t0yv0 (approved):

PR #2553, opened 2024-10-30
Enable TestConfigureSimpleValues

iwahbe:
I'm unable to get the flake on my computer, even with `--count 1000`. The test wasn't previously running in parallel, so it shouldn't be an interaction with other tests.

I've added better asserts so we can see which result is unexpectedly nil. I think our best chance at fixing the test is to turn the test back on and then re-engage if/when it flakes again.

Part of investigating https://github.com/pulumi/pulumi-terraform-bridge/issues/2530


VenelinMartinov (approved):
Let's not close the issue - I don't think we've fixed the problem yet.

PR #2562, opened 2024-10-31
Fix provider name discrepancy in index doc

guineveresaenger:
This pull request swaps out use of `g.info.Name` for `g.pkg.Name().String()` when determining the name of the provider for title and download link purposes.
`g.info.Name` refers to the upstream Terraform provider name. We need to use the Pulumi package name instead. 
Fortunately the bridge already knows how to do this, so this pull request swaps out the package name getter and adjusts the parser test provider to have a pkg name.

Fixes https://github.com/pulumi/pulumi-f5bigip/issues/598.
PR in that provider with these changes: https://github.com/pulumi/pulumi-f5bigip/pull/601

### DIFF
--- a/pkg/tfgen/installation_docs.go
+++ b/pkg/tfgen/installation_docs.go
@@ -35,7 +35,7 @@ func plainDocsParser(docFile *DocFile, g *Generator) ([]byte, error) {
 
 	// Add pulumi-specific front matter
 	// Generate pulumi-specific front matter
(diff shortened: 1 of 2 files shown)


iwahbe (approved):
We need a test that shows, when both `Name` and `pkg` are specified, which one should end up in the generated file.

Once we have a test here, LGTM.

iwahbe (approved):
We discussed off of GH. The change is under test and safe to merge.

PR #2565, opened 2024-11-01
tfcheck.execCmd: only log an error for a non-nil error

iwahbe:
Currently, we always log the full output of any executed program, which results in hyper-verbose logging. Since CI is run with `-v`, GitHub struggles to load our CI logs. We can get by only logging the full result of `execCmd` when it actually errors.

### DIFF
--- a/pkg/tests/tfcheck/exec.go
+++ b/pkg/tests/tfcheck/exec.go
@@ -23,6 +23,14 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func (d *TfDriver) execTf(t pulcheck.T, args ...string) ([]byte, error) {
+	cmd, err := execCmd(t, d.cwd, []string{d.formatReattachEnvVar()}, getTFCommand(), args...)
+	if stderr := cmd.Stderr.(*bytes.Buffer).String(); len(stderr) > 0 {
+		t.Logf("%q stderr:\n%s\n", cmd.String(), stderr)
+	}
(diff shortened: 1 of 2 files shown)


pkg/tests/tfcheck/exec.go
```diff
@@ -18,22 +18,24 @@ import (
...
-	cmd.Env = append(cmd.Env, environ...)
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
+	cmd.Dir = wdir
+	cmd.Env = append(os.Environ(), environ...)
+	t.Logf("%s", cmd.String())
 	err := cmd.Run()
-	t.Logf("error from `%s %s`\n\nStdout:\n%s\n\nStderr:\n%s\n\n",
-		program, strings.Join(args, " "), stdout.String(), stderr.String())
+	if err != nil {
```
flostadler, line 36: Are there cases where we could be interested in log output even if there was no error? In that case it would be great to have a flag/env variable to control the behavior here (`Verbose`?)

iwahbe: I can't think of any case where we would be interested in stdout if we did not error, since that is used later in the test pipeline. Stderr is never used, so I added a log of that if it's non-empty: https://github.com/pulumi/pulumi-terraform-bridge/pull/2565/commits/3f56c3f0fe108ef9b1af7437a188de2b11495d70.

flostadler (approved):

PR #2569, opened 2024-11-01
Remove specific PF testing step in CI

iwahbe:
Right now, our PF tests run twice in CI: once in the `make test` step and once in the `make -C pkg/pf test` step. This adds a needless 2 minutes to our CI time.

The test duplication was caused when we moved `pf` into `pkg`.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -47,8 +47,6 @@ jobs:
         run: cd pkg/pf && make build
       - name: Test
         run: make test
-      - name: Test PF
-        run: cd pkg/pf && make test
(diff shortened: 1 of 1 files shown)


iwahbe:
https://github.com/pulumi/pulumi-terraform-bridge/pull/2570 should show us that PF tests continue to run.

iwahbe:
Observe that https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/11629577268/job/32386813103?pr=2570 fails in the `Test`, before the `Test PF` step runs.

VenelinMartinov (approved):
nice

t0yv0 (approved):

PR #2579, opened 2024-11-01
Post-code-translation edit rules: Configuration Schema and datasource

guineveresaenger:
Papercuts discovered when generating Snowflake's registry `_index.md` file.

The string "D/datasource" should be translated to "function", and "Configuration Schema" would translate to "Configuration Configuration Reference" without tis change.

They're worth having as global rules in my opinion.

- **Add datasource -> function replace edit rule**
- **Add replace rule for '# Configuration Schema' because some providers are extra special**


mjeffryes (approved):
Seems reasonable to me especially wrt consistency with other edits we already do.  I have some worries about the "data sources" -> function generally; seems like some one might talk about data sources in one of these docs and mean something different than the TF concept.

PR #2591, opened 2024-11-05
Add lookup for pulumi provider renames.

brandonpollack23:


iwahbe (approved):
I'm not sure how accurate this will be, since we don't maintain `pulumiNames`. That said, this is a small function, so I'm happy to merge it if it helps.

PR #2599, opened 2024-11-06
Use Provider Display Name for index doc

### LINKED ISSUE #2586: Use the info.DisplayName for the title of the generated installation docs
Some providers like https://www.pulumi.com/registry/packages/oci/ or https://www.pulumi.com/registry/packages/junipermist/ don't have a terribly obvious `info.Name` so we'd rather use `info.DisplayName` for things like the title of the installation/configuration pages in the registry.

guineveresaenger:
This pull request leverages the provider's info.DisplayName field to populate the index file's frontmatter (title) as well as the installation instructions.

It will fall back to using the package name as supplied by the generator if DisplayName is not set.

For generating installation instructions, the package name continues to be used for the packages, but the display name, if set, will be used for "The Foo provider is available as a package...." line.

Fixes #2586.

### DIFF
--- a/pkg/tfgen/installation_docs_test.go
+++ b/pkg/tfgen/installation_docs_test.go
@@ -108,6 +108,50 @@ func TestPlainDocsParser(t *testing.T) {
 	}
 }
 
+func TestDisplayNameFallback(t *testing.T) {
+	t.Parallel()
+
+	type testCase struct {
+		// The name of the test case.
+		name        string
+		displayName string
+		pkgName     string
+		expected    string
+	}
+
+	tests := []testCase{
+		{
+			name:        "Uses Display Name",
+			displayName: "Unicorn",
+			pkgName:     "Horse",
+			expected:    "Unicorn",
+		},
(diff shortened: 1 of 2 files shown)


pkg/tfgen/installation_docs_test.go
```diff
@@ -186,28 +230,50 @@ func TestWriteInstallationInstructions(t *testing.T) {
...
-		t.Parallel()
-		actual := writeInstallationInstructions(tc.goImportBasePath, tc.packageName)
-		require.Equal(t, tc.expected, actual)
-	})
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			if runtime.GOOS == "windows" {
+				t.Skipf("Skipping on Windows due to a test setup issue")
+			}
```
mjeffryes, line 271: TIL we test tfgen on windowws

mjeffryes (approved):
Thanks for this!

PR #2600, opened 2024-11-06
Improve autonaming docstrings

t0yv0:
A few stylistic fixes, typos, cross-linking and adding missing doc strings around the autonaming feature.


t0yv0:
I was on a call with @rshade trying to figure out how to use `From:` which apparently is now deprecated, so wanted to do some small fixes around this.

rshade (approved):
LGTM

iwahbe (approved):
Thanks for taking the time!

PR #2603, opened 2024-11-08
Add table renderer to render tables as Markdown. Add tests.

guineveresaenger:
The markdown renderer we use does [not support rendering out tables](https://github.com/teekennedy/goldmark-markdown/issues/19). This pull request adds a table renderer, since the raw HTML rendering is lossy and does not handle inline code blocks well.

With this change, a table will properly render out inline code blocks inside table cells, and correctly handle tables as separate elements.

Example from NewRelic's `docs/_index.md` page with this change applied:

Screenshot 2024-11-07 at 5 07 42 PM


pkg/tfgen/parse/extension.go
```diff
@@ -85,3 +121,98 @@ func WalkNode[T ast.Node](node ast.Node, f func(T)) {
...
+				content := strings.TrimSpace(cell.String())
+				if inHeader {
+					header = append(header, content)
+				} else {
+					rows[len(rows)-1] = append(rows[len(rows)-1], content)
+				}
+			}
+			return ast.WalkSkipChildren, nil
+		default:
+			return ast.WalkStop, fmt.Errorf("unexpected node in a table: %s", n.Kind().String())
```
iwahbe, line 203: I'm pretty sure that `%s` implies `.String()` when available.
```suggestion
			return ast.WalkStop, fmt.Errorf("unexpected node in a table: %s", n.Kind())
```

iwahbe (approved):
You're missing a `make tidy`, but otherwise this looks solid.

PR #2625, opened 2024-11-14
Move capitalization logic into getPackageDisplayName

guineveresaenger:
This pull request clarifies the purpose and use of a package name for display reasons.

Prior to this change, when generating front matter, a provider's display name would be put through a capitalization logic that would capitalize each word in the name, even if the word was in all caps already. E.g.:

`Palo Alto Networks Cloud NGFW For AWS Provider` -> `Palo Alto Networks Cloud Ngfw For Aws Provider`

This change ensures that whatever is passed into `info.DisplayName` in the provider remains unaltered. It also unblocks the setting up of automatic index docs for https://github.com/pulumi/pulumi-cloudngfwaws.

I've also updated the function name to clarify it is specifically about finding or generating a display name.

### DIFF
--- a/pkg/tfgen/installation_docs.go
+++ b/pkg/tfgen/installation_docs.go
@@ -33,11 +33,11 @@ func plainDocsParser(docFile *DocFile, g *Generator) ([]byte, error) {
 	// Get file content without front matter
 	content := trimFrontMatter(contentBytes)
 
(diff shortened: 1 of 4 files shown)


iwahbe (approved):
Makes perfect sense. LGTM

Issue #2466, opened 2024-10-08
[docs] Tables rendered as such strip the next header


guineveresaenger:
### What happened?

Trying to run `registry-docs` on any provider with a table element in the `index.md` results in missing headers.

### Example

Examples: pulumi-ise, pulumi-sdwan, pulumi-newrelic.

1. Add `registryDocs: true` to `.ci-mgmt.yaml`
2. `make ci-mgmt`
3. `make tfgen && make build_registry_docs`
4. Observe:
```
<td>A path to a PEM-encoded certificate authority used to verify the remote agent's certificate. The `NEW_RELIC_API_CACERT` environment variable can also be used.</td>
</tr>
</tbody>
</table>
##
``` 

when it should be 
```
 </tr>
 </tbody>
 </table>

## Authentication Requirements
```

### Output of `pulumi about`

n/a; bridge at tip of default branch.

### Additional context

Possibly related: https://github.com/teekennedy/goldmark-markdown/issues/19

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

guineveresaenger:
Update:

It is my suspicion that a Renderer is not being called correctly, causing Walk() to lose place in the document. 

- Running a test like so:
```golang
m := goldmark.New(goldmark.WithExtensions(TFRegistryExtension))
	reader := text.NewReader([]byte(tt.input))
	doc := m.Parser().Parse(reader)
	doc.Dump([]byte(tt.input), 0)
	err := m.Renderer().Render(&out, []byte(tt.input), doc)
	fmt.Println(out.String())
```
shows the parser correctly detects Table nodes, but the `out.String()` reader shows the truncated output as shown in previous comment.

- `tableRenderer` wraps a `goldmark-markdown` renderer, which we then register the local tableRenderer.render function to.

- We call the "regular" renderer inside our render func to help with table cell rendering. This is fine and does not transform the source tree.

- However, the "tablewriter" from https://github.com/olekukonko/tablewriter does not seem to implement Render in a way that allows the writer (?) or the render context(set on Nodes; needs investigation) to keep track of where in the source byte array we are, and the source content, or the reference to it, gets dropped.
``` golang
table := tablewriter.NewWriter(writer) # this is our writer that we pass in alongside the Table node
table.SetHeader(header)
[...]

PR #2632, opened 2024-11-18
Fix table renderer

guineveresaenger:
This pull request does a couple of things:

- Implement a render function for each subset of Table
- Expand the `tableRenderer` struct to hold state
- initialize a separate Renderer for the table renderer, since the logic for cell rendering calls back into the outer renderer and trying to use the same renderer lost track of content.
- add tests to reflect a table in the middle of a document to ensure text after the table continues to get rendered
- ad an overall doc test that contains a table.

PR using this change: https://github.com/pulumi/pulumi-sdwan/pull/114
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2466

### DIFF
--- a/pkg/tfgen/installation_docs.go
+++ b/pkg/tfgen/installation_docs.go
@@ -44,6 +44,7 @@ func plainDocsParser(docFile *DocFile, g *Generator) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
+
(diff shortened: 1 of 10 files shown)


iwahbe (approved):
I actually like this solution better then mine. Also, it works :)

LGTM

PR #2645, opened 2024-11-21
Add `info.Schema` option: `XAlwaysIncludeInImport`

iwahbe:
This allows a provider author to specify that a field should *always* be included in an import's inputs when present - even when the input is it's zero value.

This is related to fixing https://github.com/pulumi/pulumi-cloudflare/issues/957

### DIFF
--- a/pkg/tfbridge/info/info.go
+++ b/pkg/tfbridge/info/info.go
@@ -560,6 +560,40 @@ type Schema struct {
 	//
 	// Experimental.
 	TypeName *string
+
+	// XAlwaysIncludeInImport prevents the field from pruning zero values when generating import inputs.
+	//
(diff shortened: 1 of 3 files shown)


iwahbe:
I have manually tested that this enables fixing https://github.com/pulumi/pulumi-cloudflare/issues/957.

VenelinMartinov (commented):
Awesome, thanks for the detailed doc. Are you adding tests as part of the PR?

iwahbe:
> Awesome, thanks for the detailed doc. Are you adding tests as part of the PR?

Yep. I had a note that they would be added. I just pushed the tests.

VenelinMartinov (approved):
💯

PR #2653, opened 2024-11-21
Remove top-level pf folder

guineveresaenger:
We have migrated the PF functionality to live inside of `pkg/pf` instead of its own module.

This pull request removes the deprecated files.

Reference:
https://github.com/pulumi/pulumi-terraform-bridge/pull/2480
https://github.com/pulumi/ci-mgmt/pull/1103


blampe:
There are still a couple things to do before it's "safe" to remove this:
1. Cut one final `pf/...` (and `x/...` and `testing/..`) tag so users will pick up the "Deprecated" tag on the module.
2. Wait a sufficiently long amount of time for users to notice pf is deprecated and point at the new module.

Step (2) means technically we might want to keep this around forever to avoid breaking third-parties years from now.

guineveresaenger:
Summary of internal discussion:

- Cut a final `deprecated` release for pf
- Then remove this code(?)

iwahbe (changes requested):
We need to wait at least one more release, then verify that this won't break people before merging.

iwahbe (approved):

iwahbe:
https://github.com/pulumi/pulumi-terraform-bridge/pull/2690 did this.

PR #2661, opened 2024-11-22
Create a resource that depend on the provider in `crosstests.Configure`

iwahbe:
This will hopefully help address
https://github.com/pulumi/pulumi-terraform-bridge/issues/2530.


VenelinMartinov (approved):

iwahbe:
I'm seeing repeated test failures on windows, which is really weird given that this is a test that only runs on linux.

PR #2664, opened 2024-11-22
Download upstream docs for dynamically bridged provider

### LINKED ISSUE #2607: Generate Any TF Provider schemas with full examples
## Hello!

## Issue details

Currently, schemas generated from `pulumi-terraform-provider` are not aware of their upstream git repo, and use only in-memory docs. When generating docs for the registry, we should run the full documentation pipeline used when generating docs for our other providers. We can do this by setting a flag on our `get-schema` call:

```console
$ pulumi package get-schema terraform-provider $PROVIDER_NAME --docs=full --upstream-repo=$UPSTREAM_REPO
```

The provider would need to download `$UPSTREAM_REPO`'s `docs` or `website` folder and then point docs generation at it.

guineveresaenger:
This pull request adds logic that enables us to generate resource docs for a dynamically bridged provider using the upstream dependency.

Usage for remote providers:
`pulumi package get-schema terraform-provider <registry address> <version> fullDocs=<true|false>`. 

Adds a parameterized arg field  to `pulumi package get-schema terraform-provider <registry address> <version>` called `fullDocs`, which when set to `true` will instruct the bridge to `git clone --depth 1 -b <version> <terraform provider github repo> <local dir for dynamic docs>`. This allows us to keep the exact same docs logic we have established. The shallow clone targets the exact doc version we need. (*)

For this, we infer the github repo from the OpenTofu org name. It is based on the assumption that `registry.opentofu.org/org/foo` is based on a provider that lives at `github.com/org/terraform-provider-foo`. OpenTofu says their protocol follows that of the HashiCorp Terraform Registry which [requires an org/user and a provider name of the format `terraform-provider-foo`](https://developer.hashicorp.com/terraform/registry/providers/publishing) and we have historical evidence that GitHub is the most commonly used source host for terraform providers. See also https://github.com/opentofu/registry/issues/1337.

Usage for local providers:
`pulumi package get-schema terraform-provider <local-path> upstreamRepoPath=<localPath>`. Here, the local docs path gets read directly into `upstreamRepoPath`. 

The one thing that is perhaps missing here is to expand the remote args to allow to take a source repo as an additional argument. However, our internal use case is primarily one of automation, so this is not an immediate need.

(*)Cloning docs for the AWS terraform provider at v5.70.0 has the following time performance:
```
git clone --depth 1 -b v5.70.0  awsDir  
2.06s user 
1.62s system 
34% cpu 
10.713 total
```

Fixes #2607


dynamic/main.go
```diff
@@ -62,14 +62,23 @@ func initialSetup() (info.Provider, pfbridge.ProviderMetadata, func() error) {
...
+			}
+
 			packageSchema, err := tfgen.GenerateSchemaWithOptions(tfgen.GenerateSchemaOptions{
 				ProviderInfo: info,
 				DiagnosticsSink: diag.DefaultSink(os.Stdout, os.Stderr, diag.FormatOptions{
 					Color: colors.Always,
 				}),
-				XInMemoryDocs: true,
+				XInMemoryDocs:            schemaDocsOnly,
+				XLoadUpstreamRepoForDocs: fullDocs,
```
iwahbe, line 81: Instead of keeping track of both `schemaDocsOnly` and `fullDocs` (and keeping them in sync), we can simplify to one variable.
```suggestion
				XInMemoryDocs:            !fullDocs,
				XLoadUpstreamRepoForDocs: fullDocs,
```

dynamic/parameterize/args.go
```diff
@@ -61,6 +72,6 @@ func ParseArgs(args []string) (Args, error) {
 		remote.Name = args[0]
 		return Args{Remote: &remote}, nil
 	default:
-		return Args{}, fmt.Errorf("expected to be parameterized by 1-2 arguments: <name> [version]")
+		return Args{}, fmt.Errorf("expected to be parameterized by 1-3 arguments: <name> [version] fullDocs")
```
iwahbe, line 75: By convention, the `[]` mean optional.
```suggestion
		return Args{}, fmt.Errorf("expected to be parameterized by 1-3 arguments: <name> [version] [fullDocs]")
```

dynamic/info.go
```diff
@@ -33,13 +33,17 @@ import (
 
 func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value) (tfbridge.ProviderInfo, error) {
 	provider := proto.New(ctx, p)
+	// TODO: handle docsgen for a local dynamic provider
+	remoteURL := value.Remote.URL
```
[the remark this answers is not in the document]
guineveresaenger: hence the TODO 😝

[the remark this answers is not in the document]
iwahbe: I'm never sure what TODOs are meant to stay in with an issue after merge and which ones are meant as a "must fix before merging".

pkg/tfgen/generate.go
```diff
@@ -944,6 +944,7 @@ func (g *Generator) generateSchemaResult(ctx context.Context) (*GenerateSchemaRe
 	if err != nil {
 		return nil, err
 	}
+
```
iwahbe, line 947: Let's not have extra changes here.
```suggestion
```

pkg/tfbridge/info/info.go
```diff
@@ -129,6 +129,9 @@ type Provider struct {
 	// - Name
 	// - TFProviderModuleVersion
 	//
+	// For dynamically bridged providers, when the fullDocs option is set on the parameterized inputs,
+	// this field will be automatically set to a temporary directory containing a shallow clone of the upstream repo.
```
iwahbe, line 133: Who is this comment for?

dynamic/info.go
iwahbe, line 1: We skip all examples on L46. We don't want to do that when we are generating full examples.

dynamic/info.go
```diff
@@ -83,6 +83,22 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
...
+		// we infer the Terraform provider's source code repository via the following assumptions:
+		// - The provider's source code is hosted at github.com
+		// - The provider's github org, for providers, is the namespace field of the registry name
+		// Example:
+		//
+		// opentofu.org/provider/hashicorp/random -> "hashicorp" is deduced to be the github org.
+		// Note that this will only work for the provider (not the module) protocol.
+		urlFields := strings.Split(value.Remote.URL, "/")
+		ghOrg := urlFields[len(urlFields)-2]
+		prov.GitHubOrg = ghOrg
```
iwahbe, line 100: We should set `prov.Repository`.
```suggestion
		urlFields := strings.Split(value.Remote.URL, "/")
		prov.GitHubOrg = urlFields[len(urlFields)-2]
		prov.Repository = fmt.Sprintf("https://github.com/%s/terraform-provider-%s",
			prov.GitHubOrg, urlFields[len(urlFields)-1])
```

pkg/tfbridge/info/info.go
guineveresaenger: Maintainers. It's set implicitly here, so it makes sense to document this behavior.

dynamic/info.go
guineveresaenger: Thank you for pointing out this wrinkle.

It turns out that's...not what happens, or perhaps I misunderstand.

The examples aren't being _translated_. They're not being skipped; they're in fact still in the schema. (In fact, I'm not entirely sure I understand the purpose/implementation detail of this setting, as setting it to `false` has no effect on the generated schema.)

guineveresaenger: Update: It turns out that `convertExamples` is never called. The `SkipExamples` field isn't actually ever used in the call stack.
Was this field set to _disable_ behavior? It appears that something isn't hooked up correctly.

iwahbe: It's a flag I turned on in the hope it would reduce time spent on example generation. If it doesn't do *anything at all*, feel free to remove it.

It was *supposed* to skip converting examples.

pkg/tfbridge/info/info.go
iwahbe: I wouldn't say its *set implicitly*. It is *set* in `dynamic/info.go`. If you are working in `/dynamic`, then you will see where it is set. If you are implementing a bridged provider, then this is not helpful for you, since it's relevant only to dynamic providers. `info/info` feels like the wrong place for this comment.

This feels spiritually equivalent to annotating a function with where it is called, which we don't and shouldn't do.

dynamic/main.go
```diff
@@ -147,6 +149,35 @@ func initialSetup() (info.Provider, pfbridge.ProviderMetadata, func() error) {
...
 
+			if args.Remote != nil {
+				fullDocs = args.Remote.Docs
+				if fullDocs {
+					// Write the upstream files at this version to a temporary directory
+					tmpDir, err := os.MkdirTemp("", "upstreamRepoDir")
+					if err != nil {
+						return plugin.ParameterizeResponse{}, err
+					}
+					versionWithPrefix := "v" + info.Version
```
iwahbe, line 160: Instead of calling this `versionWithPrefix`, let's call it what we assume it to be: the tag at which we are downloading the provider.
```suggestion
					tag := "v" + info.Version
```

dynamic/parameterize/args.go
```diff
@@ -31,27 +31,46 @@ type RemoteArgs struct {
 	Name string
 	// Version is the (possibly empty) version constraint on the provider.
 	Version string
+	// Docs is the (possibly empty) argument to download and write docs into the schema
```
iwahbe, line 34: What does "(possibly empty)" mean in this context? I'm also not sure why this is an "argument"

```suggestion
	// Docs indicates if full schema documentation should be generated.
```

dynamic/parameterize/args.go
```diff
@@ -31,27 +31,46 @@ type RemoteArgs struct {
...
+	Docs bool
 }
 
 // LocalArgs represents a local TF provider referenced by path.
 type LocalArgs struct {
 	// Path is the path to the provider binary. It can be relative or absolute.
 	Path string
+	// DocsLocation is the path to the provider documentation.
+	DocsLocation string
```
iwahbe, line 43: Maybe we should call this `UpstreamRepoPath`, since that's what it is. 

In any case, its not the path to docs, but to the repo that contains the docs.

```suggestion
	// UpstreamRepoPath (if provided) is the local path to the dynamically bridged Terraform provider's repo.
	//
	// If set, full documentation will be generated for the provider. If not set, only documentation from the TF provider's schema will be used.
	UpstreamRepoPath string
```

dynamic/parameterize/args.go
```diff
@@ -31,27 +31,46 @@ type RemoteArgs struct {
...
 }
 
 func ParseArgs(args []string) (Args, error) {
 	// Check for a leading '.' or '/' to indicate a path
+
```
iwahbe, line 48:
```suggestion
```

dynamic/parameterize/args.go
```diff
@@ -31,27 +31,46 @@ type RemoteArgs struct {
...
 	if len(args) >= 1 &&
 		(strings.HasPrefix(args[0], "./") || strings.HasPrefix(args[0], "/")) {
 		if len(args) > 1 {
-			return Args{}, fmt.Errorf("path based providers are only parameterized by 1 argument: <path>")
+			docsArg := args[1]
+			docsLocation, found := strings.CutPrefix(docsArg, "docsLocation=")
+			if !found {
+				return Args{}, fmt.Errorf("path based providers are only parameterized by 2 arguments: <path> [docsLocation]")
```
iwahbe, line 55: Let's validate that `docsLocation` exists here.

iwahbe, line 55: If we are going to require a prefix of `docsLocation=`, it needs to be documented.
```suggestion
				return Args{}, fmt.Errorf("path based providers are only parameterized by 2 arguments: <path> [docsLocation=<localPath>]")
```

dynamic/testdata/TestSchemaGenerationFullDocs/hashicorp/random-3.6.3.golden
```diff
@@ -0,0 +1,938 @@
...
...ore_destroy` lifecycle flag set, to avoid conflicts with unique names during the brief period where both the old and new resources exist concurrently.\n\n## Example Usage\n\n```terraform\n# The following example shows how to generate a random priority\n# between 1 and 50000 for a aws_alb_listener_rule resource:\n\nresource \"random_integer\" \"priority\" {\n  min = 1\n  max = 50000\n  keepers = {\n    # Generate a new integer each time we switch to a new listener ARN\n    listener_arn = var.listener_arn\n  }\n}\n\nresource \"aws_alb_listener_rule\" \"main\" {\n  listener_arn = random_integer.priority.keepers.listener_arn\n  priority     = random_integer.priority.result\n\n  action {\n    type             = \"forward\"\n    target_group_arn = var.target_group_arn\n  }\n  # ... (other aws_alb_listener_rule arguments) ...\n}\n```\n\n## Import\n\nRandom integers can be imported using the result, min, and max, with an\n\noptional seed. This can be used to replace a config value with a value\n\ninterpolated from the random provider without experiencing diffs.\n\nExample (values are separated by a ,):\n\n```sh\n$ pulumi import random:index/integer:Integer priority 15390,1,50000\n```\n\n",
```
iwahbe, line 226: It looks like we are not converting examples, which we definitely need to do.

dynamic/info.go
guineveresaenger: Update: It turns out this flag _does_ skip example conversion! But we weren't calling into that code at all, because we weren't calling tfgen.Generate().

So we can use it here, but only if we call the full Generate() process.

dynamic/info.go
guineveresaenger: I wonder if there's a better way to differentiate between "this provider has pulumi source code" and "this provider only has terraform source code".
While the [docs are pretty clear that the field refers to the source code repo](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/info/info.go#L70), for Pulumi-maintained providers this field is expected to be a Pulumi provider repository as opposed to a Terraform one.
And elsewhere does [express this difference](https://github.com/pulumi/pulumi-terraform-bridge/blob/2a8b877dbe500f33980524c68855f0e732c3d6aa/pkg/tfgen/generate_schema.go#L588).

guineveresaenger:
This may not be the way we want to do this, but I changed the function signature of `Generate()` and `UnstableGenerateFromSchema` to return a package schema so that we can access it from an outside library.

Alternatives:
- Implementing a separate `MakeSchemaWithExamples(generator)` function in `generate.go` that takes a pre-created Generator as an argument and returns a package schema for consumption by `dynamic`
- Generating the "lightweight" schema first, as before, and then making a separate call to `UnstableGenerateFromSchema`.

Unfortunately, adding extra options to the GeneratorOpts only gets us halfway - we must call Generate() or its parts somewhere in `dynamic`.

I've gone with changing the function signature to avoid writing yet another generator function with a narrowly scoped purpose, but am not married to the idea.

Makefile
```diff
@@ -6,7 +6,7 @@ export PULUMI_DISABLE_AUTOMATIC_PLUGIN_ACQUISITION := true
 PROJECT_DIR := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
 
 install_plugins::
-	pulumi plugin install converter terraform 1.0.19
+	pulumi plugin install converter terraform 1.0.20
```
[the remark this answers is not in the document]
guineveresaenger: We need this bump, otherwise the test won't pass.

dynamic/provider_test.go
```diff
@@ -457,6 +457,48 @@ func TestSchemaGeneration(t *testing.T) { //nolint:paralleltest
...
+		name     string
+		version  string
+		fullDocs bool
+	}
+
+	tc := testCase{
+		name:     "hashicorp/random",
+		version:  "3.6.3",
+		fullDocs: true,
+	}
```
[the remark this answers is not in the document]
guineveresaenger: > Why are we defining a local struct only to instantiate one copy?
- Test information at a glance
- More easily add subtests in the future if needed
- Allows us to run individual tests by name when using an integrated IDE test framework
- Makes the name of subtest far more obvious for running it in the terminal

Is there a downside here?

dynamic/info.go
```diff
@@ -33,17 +33,17 @@ import (
...
+		P:              provider,
+		Name:           p.Name(),
+		Version:        p.Version(),
+		Description:    "A Pulumi provider dynamically bridged from " + p.Name() + ".",
+		Publisher:      "Pulumi",
 		ResourcePrefix: inferResourcePrefix(provider),
 
 		// To avoid bogging down schema generation speed, we skip all examples.
-		SkipExamples: func(tfbridge.SkipExamplesArgs) bool { return true },
+		SkipExamples: func(tfbridge.SkipExamplesArgs) bool { return false },
```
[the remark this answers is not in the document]
guineveresaenger: We never go to example generation in the normal case. But point taken on the default.

dynamic/parameterize/args.go
```diff
@@ -61,6 +102,6 @@ func ParseArgs(args []string) (Args, error) {
 		remote.Name = args[0]
 		return Args{Remote: &remote}, nil
 	default:
-		return Args{}, fmt.Errorf("expected to be parameterized by 1-2 arguments: <name> [version]")
+		return Args{}, fmt.Errorf("expected to be parameterized by 1-3 arguments: <name> [version] [fullDocs]")
```
iwahbe, line 105: Let's be consistent with `upstreamRepoPath` for local providers.

```suggestion
		return Args{}, fmt.Errorf("expected to be parameterized by 1-3 arguments: <name> [version] [fullDocs=<true|false>]")
```

iwahbe (approved):

dynamic/provider_test.go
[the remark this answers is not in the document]
iwahbe: Verbosity and the chance of rot. It's an abstraction we aren't using.

dynamic/testdata/TestSchemaGeneration/Backblaze/b2-0.8.9.golden
```diff
@@ -1,8 +1,8 @@
 {
     "name": "b2",
-    "version": "0.8.9",
```
iwahbe, line 3: This is a regression.

PR #2705, opened 2024-12-10
Fix ImportBasePath for dynamic bridged providers

Frassle:
The current code generates an import base path like `github.com/pulumi/pulumi-terraform-provider/sdks/something/v3`. This is a _module_ path (ends with a version number) not a package import path, we always nest our packages one level below the module.

The correct import base path should be `github.com/pulumi/pulumi-terraform-provider/sdks/something/v3/something`.

### DIFF
--- a/dynamic/info.go
+++ b/dynamic/info.go
@@ -67,6 +67,7 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
 				"github.com/pulumi/pulumi-terraform-provider/sdks/go",
 				p.Name(),
 				tfbridge.GetModuleMajorVersion(p.Version()),
+				p.Name(),
 			),
 			RootPackageName:              p.Name(),
 			LiftSingleValueMethodReturns: true,
--- a/dynamic/testdata/TestSchemaGeneration/Backblaze/b2-0.8.9.golden
+++ b/dynamic/testdata/TestSchemaGeneration/Backblaze/b2-0.8.9.golden
@@ -14,7 +14,7 @@
             "respectSchemaVersion": true
         },
         "go": {
-            "importBasePath": "github.com/pulumi/pulumi-terraform-provider/sdks/go/b2",
+            "importBasePath": "github.com/pulumi/pulumi-terraform-provider/sdks/go/b2/b2",
             "rootPackageName": "b2",
             "liftSingleValueMethodReturns": true,
             "generateExtraInputTypes": true,
--- a/dynamic/testdata/TestSchemaGeneration/databricks/databricks-1.50.0.golden
+++ b/dynamic/testdata/TestSchemaGeneration/databricks/databricks-1.50.0.golden
@@ -14,7 +14,7 @@
             "respectSchemaVersion": true
         },
         "go": {
-            "importBasePath": "github.com/pulumi/pulumi-terraform-provider/sdks/go/databricks",
+            "importBasePath": "github.com/pulumi/pulumi-terraform-provider/sdks/go/databricks/databricks",
             "rootPackageName": "databricks",
             "liftSingleValueMethodReturns": true,
             "generateExtraInputTypes": true,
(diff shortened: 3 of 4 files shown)


brandonpollack23:
This looks like itll fix the issue that I was running into

iwahbe:
`github.com/pulumi/pulumi-terraform-provider/sdks/something/v3` is a nicer import path then `github.com/pulumi/pulumi-terraform-provider/sdks/something/v3/something`. Why do we need the double nesting?

Not having the double nesting worked in the past. Has something changed?

brandonpollack23:
> `github.com/pulumi/pulumi-terraform-provider/sdks/something/v3` is a nicer import path then `github.com/pulumi/pulumi-terraform-provider/sdks/something/v3/something`. Why do we need the double nesting?
> 
> > we always nest our packages one level below the module.
> 
> We don't do this for Any TF Provider code.
> 
> Not having the double nesting worked in the past. Has something changed?

Fraser's answer will almost certainly be more pointed but this has to do with the way we generate sdks locally in pulumi and the go code is expected to be nested due to the possiblity of multiple parameterizations and is how conformance tests (and the generated sdks are expected to) behave.

It also avoids name overlap

Frassle:
> We don't do this for Any TF Provider code.

Normal TF providers use a different sdk-gen to dynamic TF providers. When we generate a normal TF provider codegen doesn't output any go.mod and leaves it up to provider authors to add that to their repo wherever they want.

For parameterised packages we do output a go.mod and we output it at a specific place just above the provider package, so we end up with:
```
sdks
└── pkg
    ├── go.mod
    └── pkg
        └── init.go
```

If you set the ImportBasePath to `github.com/pulumi/pulumi-terraform-provider/sdks/pkg/v3` we'd need to get rid of the double nesting in sdk-gen to instead have:
```
sdks
└── pkg
    ├── go.mod
    └── init.go
```

We're not doing that. It doesn't even match existing sdk layouts.

> Not having the double nesting worked in the past.

We had a bug where we were generating two go.mods before for parameterised packages, one at the root as expected and another inside the package directory. The one in the package directory _could_ be used with the import path as given, although it didn't match up if actually pushed to GitHub.

iwahbe:
> If you set the ImportBasePath to github.com/pulumi/pulumi-terraform-provider/sdks/pkg/v3 we'd need to get rid of the double nesting

We should do that. It's marginally better for users and the double-nesting seems to have no benefit beyond path dependency. I doubt we are going to change it now.

> It doesn't even match existing sdk layouts.

It shouldn't matter. Codegen should just trust the import path.

iwahbe (approved):
This change will need to come with docs updates:

https://github.com/pulumi/pulumi-terraform-provider/blob/6016223e96aa54a6ab7a0e3cfb0aac02d5c53235/docs/_index.md#L75-L79

https://github.com/pulumi/pulumi-terraform-provider/blob/6016223e96aa54a6ab7a0e3cfb0aac02d5c53235/docs/_index.md#L171-L176

If [`ImportBasePath`](https://github.com/pulumi/pulumi/blob/4e68c21c2ba8584048218bbcabf4285351ce66fa/pkg/codegen/go/importer.go#L25-L28) has correctness requirements, they need to be documented. Ideally, `pulumi package use` should warn (or fail) for an invalid `ImportBasePath`.

Frassle:
> It shouldn't matter. Codegen should just trust the import path.

It does mostly, but when we're trying to infer the module path _just_ from the import path we don't have enough info unless we can assume things are laid out exactly as we expect (which means double nesting) or if the schema gives us the module path (which tfbridge doesn't), but then...

> We should do that. 

I think we could if we had the module path from the schema as well, but this all requires codegen fixes to change the sdk layout based on those two fields. 

> If [ImportBasePath](https://github.com/pulumi/pulumi/blob/4e68c21c2ba8584048218bbcabf4285351ce66fa/pkg/codegen/go/importer.go#L25-L28) has correctness requirements...

That's probably the better fix here.

> This change will need to come with docs updates:

Sure, I can add that.

PR #2717, opened 2024-12-11
Write _index.md to local docs dir on Any TF Provider

guineveresaenger:
Update: Rebased against https://github.com/pulumi/pulumi-terraform-bridge/pull/2742 which adds Cobra. Blocked until that merges.

This pull request adds a `registryDocs` Generator to the initialSetup call for a dynamic provider.

Adjustments to accommodate this change:

- Because the generation instructions will necessarily include the string `terraform`, the section needs to be added after the edit rules are applied.
- The logic to determine which installation instruction to generate is hinged on `info.Repository`, which is the source repo of the provider. If it doesn't contain "pulumi" then we're dealing with an Any TF provider.
- `_index.md` is written to a user-specified (via flag) `--indexDocOutDir`. This allows us to write to any location in a file system. Leaving this blank prompts the file to only get written to the virtual filesystem in the bridge.
- An integration test is added that verifies generation and translation of the `_index.md` file` from hashicorp/random.


pkg/tfgen/installation_docs.go
```diff
@@ -106,24 +110,40 @@ func writeFrontMatter(providerDisplayName string) string {
...
+			cSharpName,
+			goImportBasePath,
+		)
+	} else {
+		instructions = fmt.Sprintf("## Generate Provider\n\n"+
+			"The %[1]s provider must be installed as a Local Package by following the "+
+			"[instructions for Any Terraform Provider]"+
+			"(https://www.pulumi.com/registry/packages/terraform-provider/):\n\n"+
+			"```bash\n"+
+			"pulumi package gen-sdk terraform-provider %[2]s/%[3]s\n"+
```
iwahbe, line 139: `gen-sdk` is a command for provider authors. `use` is what we want package consumers to use.
```suggestion
			"pulumi package use terraform-provider %[2]s/%[3]s\n"+
```

guineveresaenger: yep, wasn't sure which! thank you!

dynamic/main.go
```diff
@@ -94,6 +95,34 @@ func initialSetup() (info.Provider, pfbridge.ProviderMetadata, func() error) {
 				info.SchemaPostProcessor(&packageSchema.PackageSpec)
 			}
 
+			if fullDocs {
```
iwahbe, line 98: Why are we associating "generate a full schema" with "generate a docs page"? Can we allow them to be set independently?
```suggestion
			if indexDocsOutDir != "" {
```

dynamic/main.go
```diff
@@ -94,6 +95,34 @@ func initialSetup() (info.Provider, pfbridge.ProviderMetadata, func() error) {
...
+				// Create a custom generator for registry docs (_index.md).
+				root := afero.NewMemMapFs()
```
iwahbe, line 100: AFAIK, we are just doing this in memory if `indexDocsOutDir` is `""`? We will never save it to disk. Why would we bother doing this at all?

dynamic/parameterize/args.go
```diff
@@ -102,6 +132,6 @@ func ParseArgs(args []string) (Args, error) {
 		remote.Name = args[0]
 		return Args{Remote: &remote}, nil
 	default:
-		return Args{}, fmt.Errorf("expected to be parameterized by 1-3 arguments: <name> [version] [fullDocs=<true|false>]")
+		return Args{}, fmt.Errorf("expected to be parameterized by 1-4 arguments: <name> [version] [fullDocs=<true|false>] [indexDocOutDir=<path/to/dir>]")
```
iwahbe, line 135: We are getting to the point where we should just slap cobra on this. We can do that in a follow-up PR.

pkg/tfgen/installation_docs.go
```diff
@@ -106,24 +110,50 @@ func writeFrontMatter(providerDisplayName string) string {
...
+//
+// A dynamically bridged provider receives the following instead:
+// ****
+// ## Generate Provider
+//
+// The Foo provider must be installed as a Local Package by following the instructions for Any Terraform Provider: (link)
+// ```bash
+// pulumi package add terraform-provider org/foo
+// ```
+// ****
```
iwahbe, line 122: Go's preferred way to add examples is with indentation.
```suggestion
//	## Generate Provider
//	
//	The Foo provider must be installed as a Local Package by following the instructions for Any Terraform Provider: (link)
//	 ```bash
//	pulumi package add terraform-provider org/foo
//	```
```

pkg/tfgen/installation_docs.go
```diff
@@ -106,24 +110,50 @@ func writeFrontMatter(providerDisplayName string) string {
...
-			"* .NET: [`Pulumi.%[3]s`](https://www.nuget.org/packages/Pulumi.%[3]s)\n"+
-			"* Java: [`com.pulumi/%[2]s`](https://central.sonatype.com/artifact/com.pulumi/%[2]s)\n\n",
-		displayName,
-		pkgName,
-		cSharpName,
-		goImportBasePath,
-	)
+	var instructions string
+	if strings.Contains(sourceRepo, "pulumi") {
+		instructions = fmt.Sprintf(
```
iwahbe, line 130: You can't return the `if` expression (since it isn't an expression), but we should return in it.
```suggestion
	if strings.Contains(sourceRepo, "pulumi") {
		return fmt.Sprintf(
```

dynamic/main.go
guineveresaenger: This is meant as an internal feature to generate docs for the registry, where my expectation is we'd always want to do it together.
I almost added it as a default part of the `fullDocs` flag originally, but wanted to be able to specify an outDir (and not write to a physical file if outDir is unspecified).

dynamic/main.go
guineveresaenger: AIUI, `NewMemMapFs` _is_ the in-memory option.

pkg/bridgetest/README.md
iwahbe, line 1: Should be removed. In favor of `unstable/testutils`.

pkg/tfgen/installation_docs.go
```diff
@@ -95,23 +99,38 @@ func writeFrontMatter(providerDisplayName string) string {
...
+//	Installation
+//	The Foo provider is available as a package in all Pulumi languages:
+//
+//	JavaScript/TypeScript: @pulumi/foo
+//	Python: pulumi-foo
+//	Go: github.com/pulumi/pulumi-foo/sdk/v3/go/foo
+//	.NET: Pulumi.foo
+//	Java: com.pulumi/foo
+//
+// ****
```
iwahbe, line 114:
```suggestion
//
//	Installation
//	The Foo provider is available as a package in all Pulumi languages:
//
//	JavaScript/TypeScript: @pulumi/foo
//	Python: pulumi-foo
//	Go: github.com/pulumi/pulumi-foo/sdk/v3/go/foo
//	.NET: Pulumi.foo
//	Java: com.pulumi/foo
//
```

iwahbe (approved):
2 nits, but otherwise LGTM.

PR #2727, opened 2024-12-11
Do not wipe the version when generating schemas for dynamic providers

guineveresaenger:
#2664 introduced using the full Generator for schema gen. 
This added a call to UnstableGenerateFromSchema, which removes the version field from the package spec, a feature intended to make local builds stable.
This pull request re-adds the version in `SchemaPostProcessor`.

### DIFF
--- a/dynamic/main.go
+++ b/dynamic/main.go
@@ -62,7 +62,6 @@ func initialSetup() (info.Provider, pfbridge.ProviderMetadata, func() error) {
 			spec.Attribution = ""
 			spec.Provider = schema.ResourceSpec{}
 			spec.Language = nil
-			spec.Version = version.Version()
 		},
 	}
 
@@ -90,6 +89,7 @@ func initialSetup() (info.Provider, pfbridge.ProviderMetadata, func() error) {
 			if err != nil {
 				return nil, err
 			}
+			packageSchema.PackageSpec.Version = info.Version
 
 			if info.SchemaPostProcessor != nil {
 				info.SchemaPostProcessor(&packageSchema.PackageSpec)
@@ -132,7 +132,6 @@ func initialSetup() (info.Provider, pfbridge.ProviderMetadata, func() error) {
 			if err != nil {
 				return plugin.ParameterizeResponse{}, err
 			}
-
 			v, err := semver.Parse(p.Version())
 			if err != nil {
 				return plugin.ParameterizeResponse{}, err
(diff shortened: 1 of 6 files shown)


pkg/tfgen/generate.go
```diff
@@ -1009,9 +1009,11 @@ func (g *Generator) UnstableGenerateFromSchema(genSchemaResult *GenerateSchemaRe
...
 		files["_index.md"] = content
 	case Schema:
-		// Omit the version so that the spec is stable if the version is e.g. derived from the current Git commit hash.
-		pulumiPackageSpec.Version = ""
-
+		// For pulumi-owned (not dynamically bridged) providers, omit the version so that the spec is stable
+		// if the version is e.g. derived from the current Git commit hash.
+		if strings.Contains(g.info.Repository, "pulumi") {
+			pulumiPackageSpec.Version = ""
+		}
```
iwahbe, line 1016: The check doesn't do what the comment says. Specifically, it doesn't test pulumi-owned. I'm concerned that this will break external users who set `.Repository` to the wrong value (or didn't set it). Can we add restoring the version to the post-process script instead?

```patch
 dynamic/info.go | 1 +
 dynamic/main.go | 1 +
 2 files changed, 2 insertions(+)

diff --git a/dynamic/info.go b/dynamic/info.go
index d5b05fbc..76fea320 100644
--- a/dynamic/info.go
+++ b/dynamic/info.go
@@ -79,6 +79,7 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
 				},
 				Parameter: value.Marshal(),
 			}
+			spec.Version = p.Version()
 		},
 	}
 	// Add presumed best-effort GitHub org to the provider info.
diff --git a/dynamic/main.go b/dynamic/main.go
index 0a97b970..ce8e387b 100644
--- a/dynamic/main.go
+++ b/dynamic/main.go
@@ -62,6 +62,7 @@ func initialSetup() (info.Provider, pfbridge.ProviderMetadata, func() error) {
 			spec.Attribution = ""
 			spec.Provider = schema.ResourceSpec{}
 			spec.Language = nil
+			spec.Version = version.Version()
 		},
 	}
 
```

guineveresaenger:
Hmmm, this doesn't actually seem like it fixes the schemas.

dynamic/main.go
```diff
@@ -131,7 +132,6 @@ func initialSetup() (info.Provider, pfbridge.ProviderMetadata, func() error) {
 			if err != nil {
 				return plugin.ParameterizeResponse{}, err
 			}
-
```
iwahbe, line 134: Please restore the original whitespace here.

iwahbe (approved):

PR #2733, opened 2024-12-12
Protect `declaredRuntimeMetadata` with a `sync.Mutex`

### LINKED ISSUE #2728: Concurrent map write in declareRuntimeMetadata
### What happened?

Suspect this code:

https://github.com/pulumi/pulumi-terraform-bridge/blob/f8f86a61c2f8ca7cdbe9d4ce5f49a755be4acc37/pkg/tfbridge/metadata.go#L37

To trigger "concurrent map writes" panic in the test suite in https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/12283355952/job/34282617115?pr=2718#step:9:2935

Would it be possible to remove this shared state ?

### Example

See above.

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

iwahbe:
This should fix the immediate issue. We should probably refactor this to avoid the global, but that can be done later.

Fixes #2728

### DIFF
--- a/pkg/tfbridge/metadata.go
+++ b/pkg/tfbridge/metadata.go
@@ -15,6 +15,8 @@
 package tfbridge
 
 import (
+	"sync"
+
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/info"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata"
 )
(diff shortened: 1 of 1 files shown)


t0yv0:
Excellent ty.

t0yv0 (approved):
There's also sync.Map or somesuch FWIW.

iwahbe:
> There's also sync.Map or somesuch FWIW.

`sync.Map` is a `map[any]any` under the hood, I'd rather have the type safety then the concurrency safety.

PR #2738, opened 2024-12-12
Call configure in sequence (not in parallel)

iwahbe:
It looks like GCP has [introduced a dependency](https://github.com/hashicorp/terraform-provider-google-beta/pull/8747) on configuring SDK and then PF into their provider. This commit is being used to check if ordering will fix the panics that the GCP upgrade is seeing in tests.


VenelinMartinov (commented):
How does this impact AWS? I remember there were some performance issues with Configure there. Very fuzzy on the details but I believe one of the Configure calls might hang sometimes

iwahbe:
@t0yv0 Might have a better memory for AWS's performance issues.

t0yv0:
Yeah we might pay few more milliseconds on startup for this in AWS. I guess the behavior here could be opt-in so only GCP opts in? Or do you think it's generally the right thing to do for most providers? 

I suspect in the case of AWS what we might actually want is removing the PF Configure call entirely or making it a no-op in AWS. I think it might still work if both PF and SDKv2 providers in the AWS internals are wrapping the same state bag for the configuration information.

t0yv0:
I think we need to own gating CI on pulumi-aws wrt the performance issues, so we just come back to this convo as needed when breaching thresholds. I'm taking a quick look.

PR #2755, opened 2024-12-16
chore: remove dynamic module

t0yv0:
This downgrades code under `./dynamic` from a Go module to a regular package.


dynamic/go.mod
```diff
@@ -1,240 +0,0 @@
-module github.com/pulumi/pulumi-terraform-bridge/dynamic
-
-go 1.22.3
-
-replace (
-	// Use OpenTofu's fork of hcl (copied from OpenTofu).
-	github.com/hashicorp/hcl/v2 => github.com/opentofu/hcl/v2 v2.0.0-20240416130056-03228b26f391
```
t0yv0, line 7: This rewrite is lost in the change. We can reintroduce it specifically scoped to the bin/pulumi-resource-terraform-provider build by using a Go workspace to build it, if necessary. Wanted to confirm if it is necessary.

Another option is to reintroduce it at top-level go.mod but it seems that would be an actual change to the bridge as currently `github.com/hashicorp/hcl/v2` is used elsewhere without the replace.

go.mod
```diff
@@ -23,7 +23,7 @@ require (
 	github.com/hashicorp/hcl v1.0.0
 	github.com/hashicorp/hcl/v2 v2.22.0
 	github.com/hashicorp/hil v0.0.0-20190212132231-97b3a9cdfa93
-	github.com/hashicorp/terraform-plugin-framework v1.7.0
+	github.com/hashicorp/terraform-plugin-framework v1.12.0
```
t0yv0, line 26: This is possibly a big jump that we might need to ship separately.

blampe (approved):
mo modules mo problems

iwahbe:
> @iwahbe second opinion on the risks outlined. I'm tempted to try this as-is but LMK if you think we need to be more conservative on this.

If tests pass, then I think it's safe.

iwahbe:
I'm pretty sure this change will result in the dynamic tests being run twice: once with all other tests and once with the "Test dynamic" test action. We can probably remove the explicit test dynamic in CI.

t0yv0:
I'll send a follow-up PR.

PR #2775, opened 2024-12-18
Fix the concurrent writes in TestDetailedDiffMap

iwahbe:
Calling `pb.NewResource` mutates the associated `.ResourceSchema`. This is safe in sequence (since the mutation is idempotent), but not in parallel. By turning the schemas into schema generators, we avoid concurrent writes.

An example of the failure is https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/12396464885/job/34605333073?pr=2774.


VenelinMartinov (approved):
Can we move the NewResource call to the outer loop instead? Seems like an more readable solution here.

PR #2781, opened 2024-12-19
Do not shell out to `git` to download Terraform repos

iwahbe:
This avoid depending on in-path binaries when generating documentation.


guineveresaenger (approved):
Looks like this works and it's a perfect example of why a library is better than shelling out. 

LGTM

PR #2809, opened 2025-01-03
Document versioning practices for dynamic package

guineveresaenger:
Followup for https://github.com/pulumi/pulumi-terraform-provider/issues/54.

Clarifies where and how we're linking in the version to the dynamic binary.

### DIFF
--- a/dynamic/version/version.go
+++ b/dynamic/version/version.go
@@ -15,6 +15,11 @@
 package version
 
 // version is initialized by the Go linker to contain the semver of this build.
+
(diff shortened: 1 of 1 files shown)


dynamic/version/version.go
```diff
@@ -15,6 +15,11 @@
 package version
 
 // version is initialized by the Go linker to contain the semver of this build.
+
```
iwahbe, line 18: The original comment on `version` should still be attached to the variable.
```suggestion
//
```

iwahbe (approved):
1 nit, then LGTM.

PR #2817, opened 2025-01-08
Improve the error messages in dynamic/internal/shim/grpcutil/replay

iwahbe:
This was motivated while trying to add a test for https://github.com/pulumi/pulumi-terraform-bridge/pull/2815. Beyond adding comments, the major change is how the failure for differences in nested fields looks. It used to look like this:

```console
go test -test.run \^TestConfigInDestroy\$
panic: fatal: An assertion has failed: field config does not match logged field \
	&{{} [] [] 0x14000410008} != &{{} [] [] 0x14000410008}
```

Observe that both sides of the `!=` are identical.

It now looks like this:

```console
$ go test -test.run \^TestConfigInDestroy\$
panic: fatal: An assertion has failed: field "terraformv6_pulumi config msgpack" does not match logged field \
	[134 83 71 ...] != [134 82 69 ...]
```

The path to the nested field is specified (`"terraformv6_pulumi config msgpack"`) and we see a difference in the value displayed.

### DIFF
--- a/dynamic/internal/shim/grpcutil/replay.go
+++ b/dynamic/internal/shim/grpcutil/replay.go
@@ -29,6 +29,11 @@ import (
 
 // LogReplayProvider is a provider that replays logs from a previous run.
 type LogReplayProvider struct {
+	// LogReplayProvider is safe to copy because methodLogs is a by-reference type.
+	// This allows mutating methods on LogReplayProvider to take a by-value copy of
+	// LogReplayProvider while sharing the same reference to the underlying data.
+
+	// methodLogs is the set of un-replayed method logs for the provider.
 	methodLogs map[string][]grpcLog
 }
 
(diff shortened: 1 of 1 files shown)


t0yv0:
Could we make `[134 83 71 ...] != [134 82 69 ...]` human friendly?

t0yv0 (approved):

iwahbe:
> Could we make `[134 83 71 ...] != [134 82 69 ...]` human friendly?

That's how `%q` formats a byte array[^1], and unchanged by this PR[^2]. We can write a specialized formatter for `[]byte` to write hex, but I don't think there is much of a benefit there.

I think improving the display of types like `[]byte` is out of the scope for this PR.

[^1]: `[]byte` is the correct type here
[^2]: For types at the top level. Nested types didn't display before this PR.

PR #2818, opened 2025-01-08
Disable the `make lint test` part of weekly-pulumi-update

iwahbe:
We don't need to run `make lint test` as part of creating the update PR, since these steps are already run as part of CI on the PR itself.


flostadler (approved):
LGTM
Might be worth replacing this with renovate

Issue #2819, opened 2025-01-08
Ensure that an empty block doesn't have a chooser, ideally rolling up empty heading blocks


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

If we are converting an empty block from TF like this:

```tf
```

We currently convert it into this:

```
{{< chooser language "typescript,python,go,csharp,java,yaml" >}}
{{% choosable language typescript %}}

{{% /choosable %}}
{{% choosable language python %}}

{{% /choosable %}}
{{% choosable language csharp %}}

{{% /choosable %}}
{{% choosable language go %}}

{{% /choosable %}}
{{% choosable language yaml %}}

{{% /choosable %}}
{{% choosable language java %}}

{{% /choosable %}}
{{< /chooser >}}
```

We should instead just drop the block entirely if there is no code content.

PR #2822, opened 2025-01-10
Do not render empty code choosers on generated index page

guineveresaenger:
This pull request does the following:

1. Prevents the generation of code choosers with empty content
1. Writes a YAML config file with no code choosers if the example is config-only
1. Checks the `pclExample` object for empty pulumi YAML or PCL explicitly and before attempting to convert, removing the empty string check from the converter functions
1. Adds tests to illustrate all four situations: 
    - valid provider config + example
    - invalid code
    - valid provider config but no example
    - valid example but no provider config
 1. Adds `exampleUnavailable` placeholder constant for use when conversion fails on an individual language
 1. Adds a `successfulConversion` flag that flips to True as soon as any one language registers as having content other than the `exampleUnavailable` message. The idea is that if even only one pulumi language converted, we want to display the code chooser.
 
 Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2819

### DIFF
--- a/pkg/tfgen/convert_cli.go
+++ b/pkg/tfgen/convert_cli.go
@@ -45,6 +45,10 @@ import (
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen/internal/autofill"
 )
 
+const (
+	exampleUnavailable = "Example currently unavailable in this language\n"
+)
+
 func cliConverterEnabled() bool {
 	return cmdutil.IsTruthy(os.Getenv("PULUMI_CONVERT"))
 }
@@ -645,13 +649,14 @@ func (cc *cliConverter) singleExampleFromHCLToPCL(path, hclCode string) (transla
 func (cc *cliConverter) singleExampleFromPCLToLanguage(example translatedExample, lang string) (string, error) {
 	var err error
 
-	if example.PCL == "" {
-		return "", nil
-	}
 	source, diags, _ := cc.convertPCL(example.PCL, lang)
 	diags = cc.postProcessDiagnostics(diags.Extend(example.Diagnostics))
 	if diags.HasErrors() {
 		err = fmt.Errorf("conversion errors: %s", diags.Error())
+	}
+
+	if source == "" {
+		source = exampleUnavailable
 	}
 	source = "```" + lang + "\n" + source + "```"
 	return source, err
--- a/pkg/tfgen/convert_cli.go
+++ b/pkg/tfgen/convert_cli.go
@@ -651,8 +651,7 @@ func (cc *cliConverter) singleExampleFromPCLToLanguage(example translatedExample
 	source, diags, _ := cc.convertPCL(example.PCL, lang)
 	diags = cc.postProcessDiagnostics(diags.Extend(example.Diagnostics))
 	if diags.HasErrors() {
-		source = "Example currently unavailable in this language\n"
-		err = fmt.Errorf("failed to convert an example: %s", diags.Error())
+		err = fmt.Errorf("conversion errors: %s", diags.Error())
 	}
 	source = "```" + lang + "\n" + source + "```"
 	return source, err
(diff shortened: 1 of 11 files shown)


pkg/tfgen/installation_docs.go
```diff
@@ -477,3 +503,39 @@ func getProviderDisplayName(g *Generator) string {
...
+	if checkExamplesEmpty(contentStr) {
+		mybytes, err := SkipSectionByHeaderContent([]byte(contentStr), func(headerText string) bool {
+			return headerText == "Example Usage"
+		})
+		contentStr = string(mybytes)
+		return contentStr, err
+	}
+	return contentStr, nil
+
+}
```
iwahbe, line 517: 3 nits:
1. Let's operate in `[]byte` space (instead of `string`) space as much as possible, since that's what `goldmark` uses.
2. Postfixing a variable name by its type should be avoided if possible.
3. Let's favor returning early if possible.
```suggestion
func removeEmptyExamplesSection(content []byte) ([]byte, error) {
	const title = "Example Usage"
	if !isMdSectionEmpty(title, content) {
		return content
	}
	isExamplesSection := func(s string) bool { return title == s }
	return SkipSectionByHeaderContent(content, isExamplesSection)
}
```

pkg/tfgen/test_data/installation-docs/example-only-expected.md
```diff
@@ -0,0 +1,132 @@
+This example is invalid and should not be translated for this test to pass
```
iwahbe, line 1: It looks like the example was translated. Do we need a more invalid example?

pkg/tfgen/test_data/installation-docs/invalid-example-expected.md
```diff
@@ -0,0 +1,9 @@
+This example is invalid and should not be translated for this test to pass
+
+## Example Usage
+
+
+
```
iwahbe, line 6: I thought a goal of this PR was to remove empty `## Example Usage` sections like this.

guineveresaenger: Yes! This test is only for the examples conversion. We would still have an empty Examples section after this process.

pkg/tfgen/installation_docs.go
```diff
@@ -477,3 +503,35 @@ func getProviderDisplayName(g *Generator) string {
...
+func isMarkdownSectionEmpty(title string, contentBytes []byte) bool {
+	gm := goldmark.New(goldmark.WithExtensions(parse.TFRegistryExtension))
+	astNode := gm.Parser().Parse(text.NewReader(contentBytes))
+
+	isEmpty := false
+
+	err := ast.Walk(astNode, func(n ast.Node, entering bool) (ast.WalkStatus, error) {
+		if section, ok := n.(*section.Section); ok && entering {
+			sectionText := section.Text(contentBytes)
+			// A little confusingly, we check if the section text _only_ contains the title, "Example Usage".
```
iwahbe, line 525:
```suggestion
			// A little confusingly, we check if the section text _only_ contains the title.
```

pkg/tfgen/convert_cli.go
```diff
@@ -645,14 +649,14 @@ func (cc *cliConverter) singleExampleFromHCLToPCL(path, hclCode string) (transla
...
 	diags = cc.postProcessDiagnostics(diags.Extend(example.Diagnostics))
 	if diags.HasErrors() {
-		source = "Example currently unavailable in this language\n"
-		err = fmt.Errorf("failed to convert an example: %s", diags.Error())
+		err = fmt.Errorf("conversion errors: %s", diags.Error())
+	}
+
+	if source == "" {
+		source = exampleUnavailable
 	}
```
iwahbe, line 660: I don't understand why this change makes sense:

If there are no errors (`!diag.HasErrors()`), but `source == ""`, should we really set `source = exampleUnavailable`.

guineveresaenger: Two reasons:

1. It is possible for `diag` to have no errors (for example, when the input is empty), which results in `source` being empty. We can return an empty string, or declare that we currently have no example.

1. It is also possible for `diag` to _have_ errors but the conversion being successful regardless. 

Therefore it makes more sense to check for "what are we looking to display to the user" and if it comes up empty, say that.

iwahbe (approved):

PR #2832, opened 2025-01-14
Test sharding - take 2

iwahbe:
Follow-up work on #2810.

This PR aimed to validate the more complex test optimization strategy of #2810 (pre-calculated perfect allocation; implemented by github.com/pulumi/sharder) by demonstrating the effects of a simpler strategy (random allocation; implemented by github.com/pulumi/shard).

Initial testing showed that both strategies were bounded by the 15 minute job [`Test and Lint / test (1.22.x, windows-latest, DEFAULT, 0)`](https://github.com/pulumi/pulumi-terraform-bridge/actions/runs/12770018688/job/35593903461?pr=2832#logs), which is not effected by sharding. 

Since both approaches appear identical, we will merge the simpler strategy.


VenelinMartinov (approved):
LGTM, thanks for checking this again. I'd wrongly assumed the results would be similar to before the recent refactoring of tests.
Simpler is definately better

PR #2883, opened 2025-02-03
Add deprecation note to provider landing page for recently deprecated providers"

guineveresaenger:
This pull request adds a list of deprecated providers to the landing page docsgen so that the page can display a deprecation warning with relevant links.

Currently, the list only contains pulumi-civo; we will add to it as we deprecate.

Part of https://github.com/pulumi/pulumi-civo/issues/696.

Screenshot of expected landing page as soon as we add Civo to the list of dynamically bridged providers:

Screenshot 2025-02-03 at 1 58 31 PM

- **Add deprecation note to installation instructions and add Civo to a hardcoded list of recently-deprecated**
- **Add test to show that a provider returned by getDeprecateProviderNames contains a deprecation note**

### DIFF
--- a/pkg/tfgen/installation_docs_test.go
+++ b/pkg/tfgen/installation_docs_test.go
@@ -306,14 +306,16 @@ func TestWriteInstallationInstructions(t *testing.T) {
 		},
 		{
 			name: "Generates Deprecation Note If Provider on Deprecated List",
-			expected: autogold.Expect("~> **NOTE:** This provider has recently been deprecated by Pulumi.\n" +
-				"Going forward, it is available as a [Local Package](https://www.pulumi.com/blog/any-terraform-provider/) " +
-				"instead.\n" +
-				"Please see the [provider's repository](https://github.com/pulumi/pulumi-civo) for details.\n\n" +
+			expected: autogold.Expect(
 				"## Generate Provider\n\n" +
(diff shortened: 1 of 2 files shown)


pkg/tfgen/installation_docs_test.go
```diff
@@ -304,6 +304,22 @@ func TestWriteInstallationInstructions(t *testing.T) {
 			repository:       "terraform-provider-testcase",
 			ghOrg:            "unicorncorp",
 		},
+		{
+			name: "Generates Deprecation Note If Provider on Deprecated List",
+			expected: autogold.Expect("~> **NOTE:** This provider has recently been deprecated by Pulumi.\n" +
+				"Going forward, it is available as a [Local Package](https://www.pulumi.com/blog/any-terraform-provider/) " +
+				"instead.\n" +
+				"Please see the [provider's repository](https://github.com/pulumi/pulumi-civo) for details.\n\n" +
```
mjeffryes, line 312: This note seems like it could be confusing if you're new to using civo with Pulumi; it's sounds like we're saying "don't use this it's deprecated!" 

Instead, I might suggest moving the note after the installation instructions and saying something along the lines of: "This provider was previously published as @pulumi/civo"; however that package is no longer being updated. Going forward, it is available as a Local Package instead. Please see the <repo name> for migration details."

mjeffryes (approved):

PR #2905, opened 2025-02-21
Bump `pulumi/pulumi` to 3.152.0

lunaris:
This change bumps the version of `pulumi/pulumi` used to the latest release, 3.151.0. This is mostly a mechanical change, but the main motivation is that the `convert.Mapper` interface we rely on for generating mappings from Terraform to Pulumi now supports parameterized hints, which are essential in the face of dynamically bridged providers. See pulumi/pulumi#18671 for more context.

### DIFF
--- a/tools/go.mod
+++ b/tools/go.mod
@@ -5,7 +5,7 @@ go 1.22.4
 require (
 	github.com/hashicorp/hcl/v2 v2.17.0
 	github.com/hexops/autogold/v2 v2.2.1
-	github.com/pulumi/terraform v1.4.0
+	github.com/pulumi/terraform v0.12.1-0.20230322133416-a268cd0892c9
 	github.com/spf13/afero v1.11.0
 	github.com/stretchr/testify v1.10.0
 )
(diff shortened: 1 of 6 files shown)


tools/go.mod
```diff
@@ -5,7 +5,7 @@ go 1.22.4
 require (
 	github.com/hashicorp/hcl/v2 v2.17.0
 	github.com/hexops/autogold/v2 v2.2.1
-	github.com/pulumi/terraform v1.4.0
+	github.com/pulumi/terraform v0.12.1-0.20230322133416-a268cd0892c9
```
lunaris, line 8: It remains to be seen whether I leave this in or remove it. There is no (as far as I can tell) 1.4.0 tag on `pulumi/terraform`, so the fact that this is still building for some of us/some of our CI pipelines is (I assume) because of caching? So I've pulled this back to the last actually tagged version I can find. But perhaps I'm being an idiot and missing something obvious.

t0yv0: Does the tool still work? I'm surprised we still have a separate go.mod here but I do have the history of how this came to be in my head. I think the only reason this tool is around is that we tried to avoid proliferating the dependency on pulumi/terraform out into all the bridged providers.

t0yv0: It might be lacking an acc tests, I can sling one if you need it.

t0yv0 (approved):

tools/go.mod
t0yv0: If it doesn't work it can cause a bit of example dropout.

PR #2907, opened 2025-02-21
Update terraform-plugin-sdk to our version corresponding to upstream v2.36.1

### LINKED ISSUE #2906: Upgrade Terraform SDK dependency
Once https://github.com/pulumi/terraform-plugin-sdk/issues/41 is done we need to upgrade the dependency here so the bridge can support ephemeral resources.

guineveresaenger:
Fixes #2906.

Notably, this is the change that brings us Ephemeral Resources. It is a prerequisite for supporting these in our providers.


t0yv0:
Continuing in https://github.com/pulumi/pulumi-terraform-bridge/pull/2909

PR #2909, opened 2025-02-24
Upgrade TF plugin sdk to v2.36.1

t0yv0:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2906

This is required to compile code referencing ephemeral resources even as these are not yet supported in the bridge.


guineveresaenger (approved):
thank you for finishing this up! LGTM!

PR #2918, opened 2025-02-27
chore: fix downstream_checks.py

t0yv0:
Relates to https://github.com/pulumi/pulumi-terraform-bridge/issues/2917

The script was confusing to me. It printed SKIPPED for jobs that failed. It is too easy to miss failure. 

It was writing MISSING for cases where some jobs may still be in progress.

It also tried to handle providers such as awsx that are not bridged - best exclude these.

### DIFF
--- a/scripts/downstream_checks.py
+++ b/scripts/downstream_checks.py
@@ -1,9 +1,14 @@
 from typing import Any, Iterator
 import argparse
 import subprocess as sp
-
 import requests
 
+NOT_BRIDGED_PROVIDERS = [
+    "aws-apigateway",
+    "awsx",
+    "eks",
+    "terraform-module",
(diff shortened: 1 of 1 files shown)


scripts/downstream_checks.py
```diff
@@ -1,9 +1,14 @@
...
 import subprocess as sp
-
 import requests
 
+NOT_BRIDGED_PROVIDERS = [
+    "aws-apigateway",
+    "awsx",
+    "eks",
+    "terraform-module",
+]
```
guineveresaenger, line 11: yes! I ran into this and completely forgot to follow up. Thank you.

guineveresaenger (approved):
LGTM

PR #2943, opened 2025-03-12
[WIP] Attempt to repro behaviour of getting an update diff due to usage of unicode

Zaid-Ajaj:
Trying to repro the behaviour of #2898 where we are getting an update diff without any change to the program. 

I expected the added test to fail based on my assumption that resolved state used explicit unicode characters which differed from the program inputs but the test actually passes. investigating this further

### DIFF
--- a/pkg/tfbridge/diff_test.go
+++ b/pkg/tfbridge/diff_test.go
@@ -479,6 +479,31 @@ func TestSimpleUpdate(t *testing.T) {
 		pulumirpc.DiffResponse_DIFF_SOME)
 }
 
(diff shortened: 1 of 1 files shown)


t0yv0:
@Zaid-Ajaj this is not the right part of the codebase unfortunately to repro this as I think all the resources are powered by plugin framework: https://github.com/port-labs/terraform-provider-port-labs/blob/main/provider/provider.go#L134 (unless I am mixing up my versions and we are debugging an older version).

Zaid-Ajaj:
Closing since #2898 seems to be an upstream issue

PR #2944, opened 2025-03-12
Repro gitlab diff issue

corymhall:
see #2948

### DIFF
--- a/pkg/pf/tests/diff_test/diff_test.go
+++ b/pkg/pf/tests/diff_test/diff_test.go
@@ -2,10 +2,8 @@ package tfbridgetests
 
 import (
 	"context"
-	"regexp"
 	"testing"
 
-	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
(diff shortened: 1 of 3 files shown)


pkg/pf/tests/diff_test/diff_test.go
```diff
@@ -59,6 +68,275 @@ Resources:
 `).Equal(t, diff.PulumiOut)
 }
 
+func TestPFGitlabDiffRepro(t *testing.T) {
+	t.Parallel()
+
+	getSchema := func(withNew bool) rschema.Schema {
```
corymhall, line 74: pulled the exact schema from the gitlab provider.

pkg/pf/tests/diff_test/diff_test.go
```diff
@@ -59,6 +68,275 @@ Resources:
...
+
+	autogold.Expect(`
+Terraform used the selected providers to generate the following execution
+plan. Resource actions are indicated with the following symbols:
++/- create replacement and then destroy
+
+Terraform will perform the following actions:
+
+  # testprovider_test.res must be replaced
++/- resource "testprovider_test" "res" {
```
corymhall, line 288: The cross tests show the Terraform diff with a replace plan, same as pulumi. Why would this be different than what we get using a real Terraform program?

PR #2945, opened 2025-03-12
Preserve TF raw state as deltas in Pulumi state

### LINKED ISSUE #1667: RawState distorted by schema-aware transformations

t0yv0:
This change fixes a long-standing problem of resource state data being distorted by the bridge, which is particularly sensitive around provider upgrade scenarios where schemas are changing and the TF provider expects to receive the older state record verbatim to auto-migrate it to the new schema. Before the change states were stored as PropertyMap and converted to TF representation on-read against the currently known provider schema, which may be different from the schema at the write time. The fix is two-fold:

- at write time, compute the delta between TF raw state and Pulumi PropertyMap and store it in the PropertyMap
- at read time, if the delta is found, reconstruct the original byte-for-byte Terraform RawState and pass it to Terraform to migrate. 

The change is feature-flagged and we aim for a gradual rollout but eventually make the new behavior the default.

- Set EnableRawStateDelta: true on info.Provider to enable for a provider
- Set PULUMI_RAW_STATE_DELTA_ENABLED=true to dynamically enable (or disable) this behavior 

One particular technical complication is around Set structures, which Pulumi does not model. For the deltas to work correctly the set element ordering in Pulumi array representation needs to be consistent between the TF and Pulumi states. This is accomplished for the SDKv2 bridge by preserving the cty.Value-canonical ordering but for the PF bridge by preserving the "pulumi-natural" ordering of set elements.

Fixes #1667 

There notable known limitations on #1667 that this approach does not address.

1. pre-existing Pulumi states do not have delta markers and running updated providers does not add them; so if there is an existing bug in a provider where it is panicking because an incorrect data is passed to it, this bug will not be magically fixed by running a yet newer version of the provider with the delta support. Options to resolve here are manually editing the state, or possibly releasing a v-prev provider with delta support and forcing an update to rewrite the state. Once https://github.com/pulumi/pulumi/issues/19307 is available it could also greatly enhance the story of fixing existing states, as it would unlock the ability of the provider to update the state independently of resource updates/CRUD lifecycle.
 
2. storing deltas alongside property maps greatly complicates debugging and manually editing provider states; if you edit the state but the delta does not match the provider will error out. Currently the error message recommends deleting the delta as that might just work. If that is insufficient, instead of editing the deltas the users can probably edit the PropertyMap state proper until it is of an acceptable shape for the provider to stop failing.


pkg/tfbridge/schema.go
```diff
@@ -1336,7 +1371,35 @@ func makeTerraformStateWithOpts(
 		return nil, err
 	}
 
-	return res.TF.InstanceState(id, inputs, meta)
+	instanceState, err := res.TF.InstanceState(id, inputs, meta)
```
t0yv0, line 1374: Need to exclude "raw" value from passing as meta here.

pkg/tfbridge/schema.go
```diff
@@ -1040,7 +1040,42 @@ func MakeTerraformResult(
...
+					"  p-map: %s\n"+
+					"  infl: %#v\n"+
+					"  error: %v",
+					stc.Value().GoString(),
+					pv.String(),
+					infl,
+					err))
+				contract.AssertNoErrorf(err, "Failed marshaling raw state")
+			}
+			metaMap["raw"] = inflEnc
```
t0yv0, line 1075: Need to check for clobbering "raw" here if it is already set.

t0yv0:
Nit: "TransformOutputs" is currently after the raw state deltas are computed, not before, which can invalidate rawState deltas.

t0yv0:
Wondering out loud: might be working too hard to preserve cty.Type; upon preliminary reading, all that is needed is making sure the "raw state JSON" as recorded by the writing provider, passes to the UpgradeState method to the currently running provider. Typically this is done with `ctyjson.Marshal` that takes a cty.Type here:

https://github.com/hashicorp/go-cty/blob/v1.4.1-0.20200414143053-d3edf31b6320/cty/json/value.go#L31

But does it actually use the cty.Type very much.. It might be that it does not. It appears at least when serializing nulls to JSON it drops the cty.Type for those. So we might not need to work quite so hard to preserve the exact cty.Value, what we need instead is that we get the same raw JSON on extraction as on writing.

t0yv0:
Inching a little closer, now need to ignore secrets.

t0yv0:
- test and fix tuples
- test assets/archives
- pass existing test suite
- run downstream checks on providers
- e2e test state upgraders
- e2e test refresh
- do we need hash indices for sets to account for reordering?
- tests against MakeTerraformResult obtained data to make sure all distortions can be recovered from
- can we drop most of the typing information from inflections to make them more space-efficient?
- plugin framework support + tests

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,639 @@
...
+	"github.com/hashicorp/go-cty/cty"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+
+	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/walk"
+)
+
+type rawStateInflections struct {
+	TypedNull *typedNull        `json:"null,omitempty"`
```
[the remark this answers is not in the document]
t0yv0: It's a null with a type. cty.Value has these.

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,639 @@
...
+	Pluralize *pluralize        `json:"plu,omitempty"`
+	Map       *mapInflections   `json:"map,omitempty"`
```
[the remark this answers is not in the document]
t0yv0: Can you provide an example that'd need more? I was fleshing out the test set and examples are helpful.

pkg/tfbridge/rawstate_test.go
```diff
@@ -0,0 +1,395 @@
...
+			}),
+			pv: resource.NewObjectProperty(resource.PropertyMap{
+				"prop": resource.NewNullProperty(),
+			}),
+			cv: cty.ObjectVal(map[string]cty.Value{
+				"prop": cty.SetValEmpty(cty.Object(map[string]cty.Type{
+					"str": cty.String,
+				})),
+			}),
+		},
```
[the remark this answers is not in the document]
t0yv0: Yeah absolutely, can you help me think through some edge cases? Renames for sure. Changes in type?

t0yv0:
Trying to pass tests with turnaround asserts at writetime is proving to be useful in sussing out edge cases. Working on it.

t0yv0:
Good progress - the bridge test suite passed with turnaround panics enabled on write. This is promising. Still some work needed.

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,768 @@
...
+
+		for k := range pvElements {
+			if _, ok := keySetWithCtyValueMatches[k]; !ok {
+				infl.ignore(k)
+			}
+		}
+
+		return rawStateInflections{Obj: &infl}, nil
+	case v.Type().IsTupleType():
+		panic("TODO TypeType")
```
t0yv0, line 733: Tuples should be possible in principle in resource states? Perhaps wrapped under DynamicValue ?

 https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/tuple says it is limited.  Curious there are no bridge tests exercising these.

pkg/tfbridge/rawstate.go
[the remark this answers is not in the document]
VenelinMartinov: We discussed this sync - an example would be an int in TF which is decoded as a string on our side. Or a float decoded as an int, etc.

pkg/tfbridge/rawstate_test.go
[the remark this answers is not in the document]
VenelinMartinov: These are the two main ones. MaxItemsOne overrides might be interesting too

pkg/pf/tfbridge/resource_state.go
```diff
@@ -21,15 +21,23 @@ import (
...
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/convert"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 )
 
-const metaKey = "__meta"
+const (
+	metaKey          = "__meta"
```
t0yv0, line 38: Can share these constants via a shared package.

t0yv0:
Added PF implementation, adding some tests for it next.

t0yv0:
Concerns: 

interactions with hooks - PreStateUpgradeHooks
any other hooks affected ?

t0yv0:
Running downstream tests on this version.

t0yv0:
Checking up the check results:

FAILED: https://github.com/pulumi/pulumi-datadog/pull/823
UNKNOWN https://github.com/pulumi/pulumi-gcp/pull/3135
FAILED: https://github.com/pulumi/pulumi-auth0/pull/828
UNKNOWN https://github.com/pulumi/pulumi-aws/pull/5374
FAILED: https://github.com/pulumi/pulumi-docker/pull/1402
UNKNOWN https://github.com/pulumi/pulumi-azure/pull/3148
FAILED: https://github.com/pulumi/pulumi-random/pull/1772
FAILED: https://github.com/pulumi/pulumi-meraki/pull/312
UNKNOWN https://github.com/pulumi/pulumi-oci/pull/763
WAITING onelogin

t0yv0:
Meraki and Datadog failures are meaningful for this PR. Studying them to lift to local tests in this repo.

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,867 @@
...
+type RawStateDelta struct {
+	TypedNull *typedNullDelta `json:"null,omitempty"`
+	Pluralize *pluralizeDelta `json:"plu,omitempty"`
+	Map       *mapDelta       `json:"map,omitempty"`
+	Obj       *objDelta       `json:"obj,omitempty"`
+	Array     *arrayDelta     `json:"arr,omitempty"`
+	Set       *setDelta       `json:"set,omitempty"`
+	Asset     *assetDelta     `json:"asset,omitempty"`
+	Num       *numDelta       `json:"num,omitempty"`
+	Replace   *replaceDelta   `json:"replace,omitempty"`
```
iwahbe, line 70: I think this needs naming information so that if we introduce a name override we can correctly reconstruct the TF name from this representation.

t0yv0: That's handled in Obj.

t0yv0:
OK Meraki test confirms that cty.SetVal reorders elements, so we cannot store set element inflections by positional index. Instead we need to be able to correlate PropertyValue inside a set to the hash of the set value, so we record hash-indexed deltas and check for hash collisions, or some such. Interesting.

t0yv0:
Interesting. The problem on set ordering is not fully general as it appeared. I can pass the meraki test under a slight alteration. Since we are using cty.Value as an intermediate vehicle, we have this problem, cty.Value reorders set elements by hash first and then original order. Unlike cty.Value, tftypes.Value does not reorder elements.  t

```go
func (u *upgradedResourceState) ToPropertyMap(ctx context.Context, rh *resourceHandle) (resource.PropertyMap, error) {
	stateType := rh.schema.Type(ctx)

	// Adding a reordering pass here fixes the problem!!
	stateValue, err := reorderStateSetsByCtyHashes(stateType, u.state.Value)
	if err != nil {
		return nil, err
	}

	pmStrict, err := convert.DecodePropertyMapWithOptions(ctx, rh.decoder, stateValue, convert.DecodeOptions{
		PreserveNull: true,
	})
	if err != nil {
		return nil, err
	}

	pm, err := convert.DecodePropertyMapWithOptions(ctx, rh.decoder, stateValue, convert.DecodeOptions{
		PreserveNull: false,
	})
	if err != nil {
		return nil, err
	}

	s, err := convertStateToCtyValue(rh.schema.Type(ctx), stateValue)
	if err != nil {
		return nil, err
	}

	// This needs to make sure the set element ordering is consistent between pmStrict and s, as well as between
	// pmStrict and pm.
	delta := tfbridge.RawStateComputeDelta(ctx, rh.schema.Shim(), rh.pulumiResourceInfo.GetFields(), pmStrict, s)
	pm[rawStateDeltaKey] = delta.ToPropertyValue()

	return updateMeta(pm, metaState{
		SchemaVersion: u.state.TFSchemaVersion,
		PrivateState:  u.state.Private,
	})
}
```

t0yv0:
Reordering set elements by their cty hashes will in fact be an observable change in PF as we will start recording values in that new order. This is already happening for SDKv2 though. In here: https://github.com/pulumi/pulumi-terraform-bridge/blob/eb62c34a2fe42e79bbae29fdf13fbe1002a3a6c5/pkg/tfshim/sdk-v2/provider2.go#L116

When we record set results for SDKv2 we record them in the cty-canonical order. Interesting.

t0yv0:
I think this and other downstream bugs point to the most promising solution here is to remove cty.Value as the vehicle for the deltas. Instead, target RawState TF JSON format, with the caveat that the representation of sets in this format must match PropertyMap set element ordering. 

How this would work:

- in the SDKv2 case we can already assume PropertyValue sets are cty.Value-ordered
- in the PF case PropertyValue sets are tftypes.Value-ordered so we need to make sure that is kept in JSON

What it solves:

- it should solve the meraki bug
- it should solve the additional bugs pertaining to "inconsisent List/Set element" - these are constraints coming from cty.Value intermediate form
- it should make deltas further minimal as there is no need to record cty.Type for nulls and such

pkg/tfshim/shim.go
```diff
@@ -343,3 +351,22 @@ type DiffOptions struct {
...
+//
+//	https://github.com/hashicorp/terraform-plugin-go/blob/v0.26.0/tfprotov5/internal/tfplugin5/tfplugin5.pb.go#L519
+//	https://github.com/hashicorp/terraform-plugin-go/blob/v0.26.0/tfprotov6/state.go#L35
+type RawState any
+
+type ProviderWithRawStateSupport interface {
+	Provider
+
+	// Construct a new empty state when creating a resource.
+	NewEmptyState(ctx context.Context, t string) InstanceState
```
t0yv0, line 368: Reminder to self: I'm not sure this got woven in properly, double-check.

pkg/tfshim/shim.go
```diff
@@ -343,3 +351,22 @@ type DiffOptions struct {
...
 type IgnoreChanges = func() map[string]struct{}
+
+// RawState is the raw un-encoded Terraform state, without type information. It is passed as-is for providers to
+// upgrade and run migrations on.
+//
+// The representation is the result of parsing the JSON format accepted on the gRPC Terraform protocol:
+//
+//	https://github.com/hashicorp/terraform-plugin-go/blob/v0.26.0/tfprotov5/internal/tfplugin5/tfplugin5.pb.go#L519
+//	https://github.com/hashicorp/terraform-plugin-go/blob/v0.26.0/tfprotov6/state.go#L35
+type RawState any
```
t0yv0, line 362: Go is making me do typing mistakes here as it automatically accepts anything as RawState... I might refactor this to hide the implementation behind a struct or such so that conversion is always explicit.

t0yv0:
Local tests passing again. Extracted https://github.com/pulumi/pulumi-terraform-bridge/pull/2985 as a starter change to rebase on.

t0yv0:
junyptermis failed 

FAILED: https://github.com/pulumi/pulumi-junipermist/pull/349

I suspect might not be related to this change though:
https://github.com/pulumi/pulumi-junipermist/issues/350

t0yv0:
These are related to over-fitting tests that do not like a new __internal field being added:

FAILED: https://github.com/pulumi/pulumi-auth0/pull/836
FAILED: https://github.com/pulumi/pulumi-random/pull/1778


These are failing for unrelated reasons but possibly masking real failures:

FAILED: https://github.com/pulumi/pulumi-gcp/pull/3155
WAITING docker
WAITING aws

t0yv0:
python scripts/downstream_checks.py --hash b733e4cbd1fbdc802cb2659949b46ab2014b6ec1 | grep -v SUCCESS

Discovered a real failure in gcp:

FAILED: https://github.com/pulumi/pulumi-gcp/pull/3160

Spurious failure: pulumi-random
WAITING aws, docker, azure

t0yv0:
Interesting finding from downstream checks in AWS - the interaction with TransformOutputs can be problematic, we do TransformOutputs to recompute tags_all in AWS for example. That is now fixed. We first TransformOuputs and then compute the delta on the result, not otherwise, for the `delta(p, t)` is not the same as `delta(p, t')` and does not apply cleanly to the transformed property map.

This also surfaces the need to make delta algorithm more precise as to not assume that two different scalars are identical, but inject a replace delta in this case, since transforms can arbitrarily edit things in the property map.

t0yv0:
Running another set of downstream checks after fixing this.

t0yv0:
Allright, in SDKv2 TransformOutputs still works inconsistently with this PR, which fails AWS tests. The previous commits only fixed it for PF.

t0yv0:
Reworking the SDKv2 integration fixes some panics in AWS, that is progress.

Still being stymied by tag-related churn in AWS PF-based resources. Looking into it, something in tag normalization code may need changing on the AWS side, but it is confusing.

t0yv0:
Downstream checks failed today because the bridge was out-of-date. Rebased and retrying.

t0yv0:
Another round:

Benign:
FAILED: https://github.com/pulumi/pulumi-random/pull/1787

To investigate:
FAILED: https://github.com/pulumi/pulumi-cloudflare/pull/1128

Compile failure: GCP

```
# github.com/pulumi/pulumi-gcp/provider/v8
Error: ./labels.go:34:23: invalid operation: cannot index pConfig.AsMap() (value of type property.Map)
Error: ./labels.go:43:22: cannot range over labels.AsMap() (value of type property.Map)
Error: ./labels.go:44:18: invalid operation: cannot index programLabels (variable of type property.Map)
Error: ./labels.go:70:30: invalid operation: cannot index programLabels (variable of type property.Map)
```

Waiting:

UNKNOWN https://github.com/pulumi/pulumi-aws/pull/5434
UNKNOWN https://github.com/pulumi/pulumi-azure/pull/3172

Needs bridge upgrade:
WAITING docker

t0yv0:
The good news is that all failing tests in the AWS provider pertain to tagging. That is no other failures were detected.

pkg/pf/tfbridge/provider_upgrade_resource_state.go
```diff
@@ -28,21 +31,27 @@ import (
 func (p *provider) UpgradeResourceState(
 	ctx context.Context,
 	rh *resourceHandle,
-	st *resourceState,
+	st *rawResourceState,
 ) (*upgradedResourceState, error) {
-	if st.TFSchemaVersion >= rh.schema.ResourceSchemaVersion() {
```
t0yv0, line 33: One change here is that this always now run the upgrade method even when versions indicate that the resource state being upgraded is on a current enough schema. I wonder whether this is wrong though in terms of parity with Terraform. Let me check if Terraform runs the upgrade method always or not.

t0yv0: https://github.com/opentofu/opentofu/blob/4083fe2cc4a483fc19d9769e0eb19ebee2386b33/internal/tofu/upgrade_resource_state.go#L28  seems like this is in fact conditional.. so the check needs to be brought back, and worked around/tested.

pkg/tfbridge/schema.go
```diff
@@ -1315,6 +1323,40 @@ func makeTerraformStateWithOpts(
...
+			logger.Debug(fmt.Sprintf("Failed recover raw state:\n"+
+				"  %q: %#v\n"+
+				"  error: %v",
+				reservedkeys.RawStateDelta,
+				delta.Marshal().String(),
+				err))
+			contract.AssertNoErrorf(err, "Failed to recover raw state")
+		}
+
+		return providerWithRawStateSupport.UpgradeState(ctx, res.TFName, recoveredRawState, meta)
```
t0yv0, line 1357: Reading this again, this does not seem right, in light of the comment on PF and the implementation note there linking opentofu behavior. The Upgrading of the state (in terms of calling the state upgrade method on the TF provider) needs to be discretionary on the version mismatch between current schema and the state being analyzed. If the schema is matching, what may happen instead is simply recovering cty.Value against the current schema. 


Also reading related prior changes in https://github.com/pulumi/pulumi-terraform-bridge/commit/3d9c23a7651d974e66819f07b530f328aa28255f

t0yv0: The prior logic has 
https://github.com/pulumi/pulumi-terraform-bridge/blob/9ff35dbdae818c512cd9b9205a87ac4f664ec0e1/pkg/tfshim/sdk-v2/provider2.go#L42

Which calls
recoverAndCoerceCtyValueWithSchema

and normalizeBlockCollections; 

so e.g. state recovery proceeds in this manner:

makeTerraformInputsWithOptions translates PropertyMap to a slightly incorrect map[string]any
recoverAndCoerceCtyValueWithSchema tries to recover cty.Value from it but enforce the schema
normalizeBlockCollections further does fixup

Hm. This doesn't feel entirely optimal, we probably need to retain this behavior for a bit until we have a more direct and reliable means of recovering the state. One design is:

if schema versions indicate upgrade --> use deltas to recover RawState JSON and call the TF state Upgrade method
otherwise call old Instance State as before. 

This has a downside that scenarios where resource version is not changing but MaxItems=1 is flipping around are not fixed up.

If we are to fixup those scenarios, and yet not call TF UpgradeResourceState method when TF would not call it, we would need to recover cty.Value or similar from the rawstate representation and current schema. And proceed with that. I think that might be possible. Thinking if the RawState representation is in any way lossy for this purpose. Might need to strengthen the checks that happen at write-time to make sure of this.

dynamic/testdata/TestPrimitiveTypes/update.golden
```diff
@@ -1,5 +1,26 @@
 {
   "properties": {
+    "__pulumi_raw_state_delta": {
+      "obj": {
+        "ps": {
+          "attrBoolComputed": {
+            "replace": {
```
VenelinMartinov, line 7: Is this inteded to be a replace? I was under the impression replace is a fallback in case the algorithm doesn't work.

pkg/tfbridge/schema_test.go
```diff
@@ -2972,6 +2978,10 @@ func TestExtractSchemaInputsNestedMaxItemsOne(t *testing.T) {
 
 			outs, err := plugin.UnmarshalProperties(resp.GetProperties(), plugin.MarshalOptions{})
 			assert.NoError(t, err)
+
+			// Ignore special keys as the values can get complicated.
+			delete(outs, reservedkeys.Meta)
```
VenelinMartinov, line 2983: What has changed in the `__meta` key due to this PR? Why are we deleting it here?

pkg/tfshim/shim.go
```diff
@@ -343,3 +357,28 @@ type DiffOptions struct {
...
+func (x *RawState) UnmarshalJSON(raw []byte) error {
+	*x = raw
+	return nil
+}
+
+type ProviderWithRawStateSupport interface {
+	Provider
+
+	// Ensure raw state is upgraded to the current resource schema version.
+	UpgradeState(ctx context.Context, t string, state RawState, meta map[string]any) (InstanceState, error)
```
VenelinMartinov, line 383: If we don't hide away the RawStateDeltas in `makeTerraformState` isn't this redundant? Why does the current path for upgrading state not work with the delta state?

pkg/pf/tfbridge/provider_upgrade_resource_state.go
t0yv0: Actually reading the code carefully today, TF does always call the upgrade method even if current schema numbers match. This is great. This makes implementation easier for us I just need to annotate.

pkg/tfshim/shim.go
t0yv0: Discussed in person - at this point would love to keep responsibility of recovering RawState outside of tfshim layer so it gets passed explicitly down to the layer; does sdk-v1 support it ? not yet but it might/should.

VenelinMartinov: Answered sync, Anton was arguing that this method should be visible on the shim layer as the shim behaviour is different if the provider supports it.

dynamic/testdata/TestPrimitiveTypes/update.golden
```diff
@@ -1,5 +1,26 @@
...
+              "raw": null
```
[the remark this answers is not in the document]
t0yv0: Investigating, one issue here is that the test in question is very indirect.

It uses this provider: 

https://github.com/pulumi/pulumi-terraform-bridge/blob/031d18691ab62c1a3ba0e97d8e64aed83304e084/pkg/pf/proto/empty.go#L29

Coming from here:

https://github.com/pulumi/pulumi-terraform-bridge/blob/031d18691ab62c1a3ba0e97d8e64aed83304e084/dynamic/main.go#L66

https://github.com/pulumi/pulumi-terraform-bridge/blob/031d18691ab62c1a3ba0e97d8e64aed83304e084/dynamic/provider_test.go#L377

https://github.com/pulumi/pulumi-terraform-bridge/blob/031d18691ab62c1a3ba0e97d8e64aed83304e084/dynamic/provider_test.go#L70

Since the underlying provider does not return anything on the TF gRPC layer, the Pulumi-returned values do not match and the delta tries to remove them. There is a slight problem here that the removal is rendered as inserting nulls as opposed to simply removing. On the off-chance that null entries are significant somewhere, this isn't quite right. I will add some tests to the delta to see how it can render removals of a property without actually rendering it as null replacement.

pkg/tfbridge/info/info.go
```diff
@@ -177,6 +177,10 @@ type Provider struct {
 	// EnableAccuratePFBridgePreview makes the Plugin Framework bridge use an experimental feature
 	// to generate more accurate diffs and previews for resources
 	EnableAccuratePFBridgePreview bool
+
+	// Newer versions of the bridge preserve Terraform raw state by saving the delta between Pulumi state and
+	// Terraform raw state into the state file. Setting this to true disables the feature.
+	DisableRawStateDelta bool
```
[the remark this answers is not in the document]
t0yv0: https://github.com/pulumi/pulumi-terraform-bridge/issues/3003 yes let us do this. I flipped the switch.

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+// schema aware and may rename fields or even change their type, and the V-prev schema is not available at read time.
+//
+// It is possible to store RawState alongside PropertyMap in Pulumi state, but this approach would double the required
+// storage. More pertinently, it would complicate editing state files in scenarios requiring state repair.
+//
+// Instead of storing RawState directly, the chosen approach uses RawStateDelta representing the difference between a
+// PropertyMap and RawState, so that RawState need not be stored itself. In typical scenarios RawStateDelta is fairly
+// small. It falls back to storing the entirety of RawState if an efficient encoding cannot be computed.
+//
+// Invariant: this struct represents a serializable union type. At most one of the fields must be set. The zero value
```
VenelinMartinov, line 56: Do we need any asserts around this or is it the job of the implementer to maintain?

pkg/tfbridge/provider.go
```diff
@@ -1124,7 +1124,7 @@ func (p *Provider) Diff(ctx context.Context, req *pulumirpc.DiffRequest) (*pulum
 		return nil, err
 	}
 
-	state, err := makeTerraformStateWithOpts(ctx, res, req.GetId(), olds,
+	state, err := makeTerraformStateWithOpts(ctx, p.tf, res, req.GetId(), olds,
```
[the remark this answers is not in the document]
t0yv0: Fixed this.

pkg/tfbridge/schema_test.go
```diff
@@ -762,12 +767,14 @@ func TestInjectingCustomTimeouts(t *testing.T) {
...
 
 			state, err = makeTerraformStateWithOpts(
-				ctx, Resource{TF: res, Schema: &ResourceInfo{}}, state.ID(), props,
+				ctx, prov, resInst, state.ID(), props,
 				makeTerraformStateOptions{defaultZeroSchemaVersion: true})
 			assert.NoError(t, err)
 			assert.NotNil(t, state)
 
-			assert.Equal(t, "0", state.Meta()["schema_version"])
+			// A bit of discrepancy here: sdkv2 will upgrade the resource state but sdk2 will not;
```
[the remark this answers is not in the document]
t0yv0: Yes.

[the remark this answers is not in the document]
t0yv0: Typo

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+			return rawstate.Array(), nil
+		default:
+			v, err := d.Pluralize.Inner.recoverRepr(pv)
+			if err != nil {
+				return rawstate.Null(), err
+			}
+			return rawstate.Array(v), nil
+		}
+	case d.Map != nil:
+		if !pv.IsObject() {
```
VenelinMartinov, line 104: Should this be handling a null propertyValue? Might come up as a result of manual edits to the state perhaps

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+		for k, v := range pm {
+			element, err := d.Map.ElementDeltas[k].recoverRepr(v)
+			if err != nil {
+				return rawstate.Null(), err
+			}
+			recovered[string(k)] = element
+		}
+		return rawstate.Object(recovered), nil
+	case d.Obj != nil:
+		if !pv.IsObject() {
```
VenelinMartinov, line 118: Should this handle the null case too?

pkg/tfbridge/schema_test.go
t0yv0: Leftover from old version where deltas were nested under __meta; removing.

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+			for _, ign := range d.Obj.Ignored {
+				ignoredSet[ign] = struct{}{}
+			}
+		}
+		for k, v := range pm {
+			if reservedkeys.IsBridgeReservedKey(string(k)) {
+				continue
+			}
+
+			if _, ign := ignoredSet[k]; ign {
```
VenelinMartinov, line 136: When do these come up?

pkg/pf/tfbridge/provider_upgrade_resource_state.go
t0yv0: We might need to have it under flag though.

pkg/tfshim/sdk-v2/provider2.go
```diff
@@ -78,11 +79,24 @@ func (r *v2Resource2) DecodeTimeouts(config shim.ResourceConfig) (*shim.Resource
 type v2InstanceState2 struct {
 	resourceType string
 	stateValue   cty.Value
+
 	// Also known as private state.
 	meta map[string]interface{}
+
+	// The state has passed through the state upgrade method and does not need to do it again.
+	isUpgraded bool
```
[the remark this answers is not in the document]
t0yv0: This is compensating for a bit of a legacy code conundrum. 

We have Diff, Refresh, Apply call ensureStateUpgraded (previously they called upgradeState). That is because if the caller uses Resource.InstanceState() to get an InstanceState handle, then it is *not* upgraded, unless we are talking about nil state indicating a Create that counts as upgraded; and therefore Diff, Refresh, Apply need to pass it through the upgrader.

On the other hand if the caller uses UpgradeState then that method actually passes through the upgrader before it reaches Diff, Refresh, Apply, and this flag avoids having it run twice.

If we used different types for upgraded/not-upgraded state this could read cleaner but it's a bigger refactor I've postponed here.

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+		}
+		v := json.Number(pv.StringValue())
+		return rawstate.Number(v), nil
+	default:
+		contract.Failf("RawStateDelta.Recover does not recognize this rawStateDelta case")
+		return rawstate.Null(), errors.New("impossible")
+	}
+}
+
+func rawStateRecoverNatural(pv resource.PropertyValue) (rawstate.Builder, error) {
```
VenelinMartinov, line 227: Sounds like we are imposing some restrictions on the type of state edits users can apply - for example, they can no longer add an object property to the state without a matching delta. Should we document this path externally somewhere? Is there some pulumi CLI facility which might help users during such edits?

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+	ctx context.Context,
+	schemaMap shim.SchemaMap, // top-level schema for a resource
+	schemaInfos map[string]*SchemaInfo, // top-level schema overrides for a resource
+	outMap resource.PropertyMap,
+	instanceState shim.InstanceState,
+) error {
+	// If called in a pulumi preview e.g. Create(preview=true) or in a continue-on-error scenario, bail because the
+	// code for deltas cannot process unknowns.
+	if resource.NewObjectProperty(outMap).ContainsUnknowns() {
+		return nil
```
VenelinMartinov, line 455: Shouldn't this return an error? Seems like this is called in the wrong place

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+	}
+
+	rawStateRecoveredBytes, err := json.Marshal(rawStateRecovered)
+	if err != nil {
+		return err
+	}
+
+	if !bytes.Equal(rawStateRecoveredBytes, rawStateWithoutTimeoutsBytes) {
+		return fmt.Errorf("recovered raw state does not byte-for-byte match the original raw state")
+		// return fmt.Errorf("recovered raw state does not byte-for-byte match the original raw state\n"+
```
VenelinMartinov, line 513: Commented out code

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+			return RawStateDelta{}, errors.New(
+				"Expected array length parity for PropertyValue and matching cty.Value",
+			)
+		}
+
+		if len(pvElements) == 0 {
+			return RawStateDelta{ArrayOrSet: &arrayOrSetDelta{}}, nil
+		}
+
+		arrayInfl := arrayOrSetDelta{}
```
VenelinMartinov, line 683: The naming here is probably left over from when this was `inflections` - arrayDelta is better

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+			return RawStateDelta{}, errors.New(
+				"Expected map length parity for PropertyValue and matching cty.Value",
+			)
+		}
+
+		if len(pvElements) == 0 {
+			return RawStateDelta{Map: &mapDelta{}}, nil
+		}
+
+		mapInfl := mapDelta{}
```
VenelinMartinov, line 711: Leftover inflection name, mapDelta is better

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+			return RawStateDelta{}, errors.New(
+				"Expected array length parity for PropertyValue and matching Set cty.Value",
+			)
+		}
+
+		if len(pvElements) == 0 {
+			return RawStateDelta{ArrayOrSet: &arrayOrSetDelta{}}, nil
+		}
+
+		setInfl := arrayOrSetDelta{}
```
VenelinMartinov, line 763: setDelta name

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+		return RawStateDelta{ArrayOrSet: &setInfl}, nil
+
+	case v.Type().IsObjectType() && !v.IsNull():
+		if !pv.IsObject() {
+			return RawStateDelta{}, errors.New("Expected an Object PropertyValue for an Object cty.Value")
+		}
+
+		elements := v.AsValueMap()
+		pvElements := pv.ObjectValue()
+		infl := objDelta{}
```
VenelinMartinov, line 780: delta name

pkg/tfbridge/rawstate.go
```diff
@@ -0,0 +1,826 @@
...
+		handledKeys := map[resource.PropertyKey]struct{}{}
+
+		for k, v := range elements {
+			subPath := path.GetAttr(k)
+			keyRaw, err := TerraformToPulumiNameAtPath(subPath, ih.schemaMap, ih.schemaInfos)
```
VenelinMartinov, line 785: If I understand this correctly you are using both the TF schema and the Pulumi SchemaInfo at encoding time but these might be different at decoding time. I think you need to compare this to the result of TerraformToPulumiName(nilSchema, nilSchemaInfo) and add a rename if the result is different.

pkg/tfbridge/rawstate_test.go
```diff
@@ -0,0 +1,1226 @@
...
+			if tcase.schema != nil {
+				ih.schemaMap = sdkv2.NewSchemaMap(map[string]*schema.Schema{
+					"prop": tcase.schema,
+				})
+
+				cv = cty.ObjectVal(map[string]cty.Value{"prop": cv})
+				pv = resource.NewObjectProperty(resource.PropertyMap{"prop": pv})
+
+				ih.schemaInfos = map[string]*info.Schema{
+					"prop": {Name: "prop"},
```
VenelinMartinov, line 363: Is this to avoid pluralizations?

pkg/tfbridge/rawstate_test.go
```diff
@@ -0,0 +1,1226 @@
...
+	// Check that it does not mutate the inputs.
+	fl := big.NewFloat(1.252235135353451351345134)
+	rawStateReducePrecision(cty.NumberVal(fl))
+	assert.Equal(t, 0, big.NewFloat(1.252235135353451351345134).Cmp(fl))
+}
+
+// For each situation when MakeTerraformResult introduces a distortion between the natural encoding of a TF value as a
+// Pulumi value, rawstate needs to be able to compute delta to reverse the process and reconstruct the TF value.
+//
+// It is useful to look at coverage reports produced solely from this test matrix to check that it covers interesting
```
VenelinMartinov, line 542: very cool

pkg/pf/tfbridge/provider_read.go
```diff
@@ -64,9 +66,9 @@ func (p *provider) ReadWithContext(
 			return plugin.ReadResult{}, 0, err
 		}
 
-		result, err = p.readResource(ctx, &rh, currentStateMap)
+		result, resultingState, err = p.refreshResource(ctx, &rh, currentStateMap)
```
VenelinMartinov, line 69: Which method is called on `.get`? Last time I checked we can't distinguish between a refresh and a `.get` with partial state on the protocol - this might make it worse.

pkg/tfshim/sdk-v2/provider2.go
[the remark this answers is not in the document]
t0yv0: I've taken a pass to introduce constructors to make sure this is always intentionally set. Found cases where we missed setting `t`.

pkg/tfbridge/rawstate_test.go
t0yv0: Yeah . Added comment.

pkg/pf/tfbridge/provider_read.go
t0yv0: Sigh we need to finally get this into devdocs. I keep forgetting and I need a reference.

pkg/tfbridge/rawstate.go
t0yv0: https://github.com/pulumi/pulumi-terraform-bridge/blob/8c0d79f2d35e9182f611ba5cab43d74cda27499c/pkg/tfbridge/rawstate.go#L697 when computing deltas we never emit d.Map if the PV value is a null. You are correct that if folks edit the state the delta may get out of sync and stop being applicable. This is not supported and will hit the error case. I hesitate on supporting that since it is hard in general. If there is demand we can reconsider.

pkg/tfbridge/rawstate.go
t0yv0: Nope https://github.com/pulumi/pulumi-terraform-bridge/blob/8c0d79f2d35e9182f611ba5cab43d74cda27499c/pkg/tfbridge/rawstate.go#L777

pkg/tfbridge/rawstate.go
t0yv0: I would suggest users running into this simply delete the _raw_state_delta and try again on the off-chance that succeeds.

pkg/tfbridge/rawstate.go
t0yv0: I've added a test. This may come up if Pulumi object has keys that are not in TF schema, so the TF object does not have anything at that key at all. I have not been able to repro it through MakeTeraraformResult but remains a theoretical possibility we need to handle.

dynamic/testdata/TestPrimitiveTypes/update.golden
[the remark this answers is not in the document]
t0yv0: Okay this is all by design. The attrNumberComputed was in TF schema and set by TF to null, but Pulumi failed to set it to null, so the delta is a null insertion. In contrast, attrIntComputed is set by Pulumi so there is no delta for it as it matches the TF - this was packing `"attr_int_computed":tftypes.Number<"256">` against `"attrIntComputed": 256` and decided it was a natural delta.

pkg/tfbridge/rawstate.go
t0yv0: No asserts, just convention. Asserts a little awkward in Go for constructing this kind of thing, but maybe we could come up with something, such as (&objDelta{Obj: .. }).make(). For now just a convention.

pkg/tfshim/rawstate/rawstate.go
```diff
@@ -0,0 +1,39 @@
+package rawstate
```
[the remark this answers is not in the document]
t0yv0: Good point, just was not thinking clearly. I will move up.

pkg/pf/tfbridge/resource_state.go
```diff
@@ -217,3 +179,159 @@ func updateMeta(m resource.PropertyMap, newMeta metaState) (resource.PropertyMap
...
+			SchemaVersion: stateVersion,
+			PrivateState:  parsedMeta.PrivateState,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("PreStateUpgradeHook failed to update schema version: %w", err)
+		}
+	}
+
+	if stateVersion > rh.schema.ResourceSchemaVersion() {
+		return nil, fmt.Errorf("The current state of %s was created by a newer provider version "+
```
t0yv0, line 228: Match TOFU behavior here, seems reasonable?

pkg/pf/tfbridge/provider_read.go
t0yv0: I found this:

ReadStep.Apply()
https://github.com/pulumi/pulumi/blob/8ad78da96621cfeece6fed63fd3be820091f45df/pkg/resource/deploy/step.go#L877

RefreshStep.Apply()
https://github.com/pulumi/pulumi/blob/8ad78da96621cfeece6fed63fd3be820091f45df/pkg/resource/deploy/step.go#L1034

ImportStep.Apply()
https://github.com/pulumi/pulumi/blob/8ad78da96621cfeece6fed63fd3be820091f45df/pkg/resource/deploy/step.go#L1291

https://github.com/pulumi/pulumi/issues/17656 we still don't have official docs sadly.

I think it checks out that R.get() scenario seems very similar to refresh scenario, the provider does not know the context in which it's operating. There might be a case to make that R.get should run TF importer which this doesn't run currently. 

What do you think in the current changes makes this worse in particular?

t0yv0: What the changes are trying to do is grab resultingState value so they can compute a delta against that and have it written into state.

pkg/tfbridge/rawstate.go
VenelinMartinov: I see, yeah, that makes sense.

VenelinMartinov (commented):
Can we also make sure we have a sufficient number of integration tests which cover some of the corner cases? We might need to work with this code soon, e.g. when the engine implements Provider Upgrades and it'd be a shame if we had to re-do all of this work then.

VenelinMartinov:
I am only aware of these integration tests for upgrades in the SDKv2 bridge: https://github.com/pulumi/pulumi-terraform-bridge/blob/366cbff413c89f0b9d6fb731d8ab57ec6b5dfc27/pkg/tests/upgrade_scenario_test.go#L36

Given that we disabled the upgrade cross-tests, which were broken, AFAIK these are the only tests for upgrades. The tests in `upgrade_scenario_test` mostly cover MaxItemsOne changes. That's pretty essential but is not the only type of schema change we are handling here. We discussed a few more and I am sure you are aware of many others, e.g. property renames, type changes, something which triggers the DeltaReplace scenario etc.

You ran into a few different issues with downstream tests - these are all integration tests and it'd be great to pull some of these into the bridge so we don't have to do the provider outer loop next time we are working in the area.

LMK if these already exist here.

t0yv0:
If I understand correctly from our sync today, the remaining work is:

## strengthen state upgrade tests

https://github.com/pulumi/pulumi-terraform-bridge/pull/3006 is WIP

I have found these tests:

```
pkg/tfshim/sdk-v2/provider2_test.go:21:func TestProvider2UpgradeResourceState(t *testing.T) {


pkg/tests/create_test.go:180:func TestCreateDoesNotInvokeStateUpgraders(t *testing.T) {
pkg/tests/upgrade_scenario_test.go:36:func TestUpgrade_StateUpgraders(t *testing.T) {
pkg/tests/upgrade_scenario_test.go:140:func TestUpgrade_Pulumi_Removes_MaxItems1(t *testing.T) {
pkg/tests/upgrade_scenario_test.go:235:func TestUpgrade_Pulumi_Adds_MaxItems1(t *testing.T) {
pkg/tests/upgrade_scenario_test.go:330:func TestUpgrade_Upstream_Adds_MaxItems1(t *testing.T) {
pkg/tests/upgrade_scenario_test.go:340:func TestUpgrade_Refresh(t *testing.T) {
pkg/tests/upgrade_scenario_test.go:432:func TestUpgrade_Upstream_Removes_MaxItems1(t *testing.T) {
pkg/tests/schema_pulumi_test.go:305:func TestSDKv2AliasesSchemaUpgrade(t *testing.T) {

pkg/pf/tests/schema_and_program_test.go:456:func TestStateUpgrade(t *testing.T) {
pkg/pf/tests/schema_and_program_test.go:560:func TestPFAliasesSchemaUpgrade(t *testing.T) {
pkg/pf/tests/prestateupgradehook_test.go:28:func TestPreStateUpgradeHook(t *testing.T) {
pkg/pf/tests/provider_diff_test.go:167:func TestDiffVersionUpgrade(t *testing.T) {
pkg/pf/tests/diff_test/diff_upgrade_test.go:20:func TestPFUpgradeRenamedProp(t *testing.T) {
pkg/pf/tests/diff_test/diff_upgrade_test.go:93:func TestPFUpgradeMovedProp(t *testing.T) {

pkg/internal/tests/cross-tests/upgrade_state_check.go:111:func runUpgradeStateInputCheck(t T, tc upgradeStateTestCase) {
pkg/internal/tests/cross-tests/diff_cross_test.go:1739:func TestDiffProviderUpgradeBasic(t *testing.T) {
pkg/internal/tests/cross-tests/diff_cross_test.go:1791:func TestStateUpgradeSet(t *testing.T) {
pkg/internal/tests/cross-tests/diff_cross_test.go:1837:func TestDiffProviderUpgradeMaxItemsOneChanged(t *testing.T) {
```

I would like to move upgrade_scenario_test.go into cross-tests and test these properties:


```
SDKv2 and PF:

Applied on up, applied on refresh

   State upgrader applied on up:
     version - same
         Upgrader always called
         pulumi rename
         pulumi typechange
     version - upgrade
         pulumi rename
         pulumi typechange
         terraform type change + custom migrate function
     version - downgrade
         opentofu does hard failure
```

Once this is completed for PF as well,  we can close these:

https://github.com/pulumi/pulumi-terraform-bridge/issues/2984
https://github.com/pulumi/pulumi-terraform-bridge/issues/1736

This will provide a better basis for this PR.

## Adjust assertions/errors for corrupt deltas

The feedback was that when delta/PropertyMap integrity is broken on read, we need to direct the user as the user may be in a state repair scenario. We can direct them to delete the deltas and edit PropertyMap as needed.

## Incorporate downstream check learnings

As discussed this may be difficult but can check on TransformOutptus sequence ordering and also perhaps find an example where cty.Value-induced set re-ordering is interesting (this was a learning from Meraki) failure. This perhaps is going too deep though.

## Tags learnings from pulumi-aws

I think we need to solve that there, not in the bridge.

## Numeric precision learnings

We might be able to backlog something around preserving precision for large numbers, that was one of the learnings from downstrem checks.

pkg/pf/tests/upgrade_state_cross_test.go
```diff
@@ -1109,3 +1092,9 @@ func TestPFUpgrade_Downgrading(t *testing.T) {
 		autogold.Expect([]ct.UpgradeStateTrace{}).Equal(t, result.PulumiUpgrades)
 	})
 }
+
+func skipUnlessDeltasEnabled(t *testing.T) {
```
VenelinMartinov, line 1096: Perhaps this should only apply to CI? `if os.LookupEnv("CI")`

It can be surprising to run a test and it gets auto-skipped locally.

pkg/internal/tests/cross-tests/upgrade_state_check_test.go
```diff
@@ -401,3 +414,9 @@ func skipUnlessLinux(t T) {
 		t.Skip("Skipping on non-Linux platforms as our CI does not yet install Terraform CLI required for these tests")
 	}
 }
+
+func skipUnlessDeltasEnabled(t T) {
+	if d, ok := os.LookupEnv("PULUMI_RAW_STATE_DELTA_ENABLED"); !ok || !cmdutil.IsTruthy(d) {
```
VenelinMartinov, line 419: Should this only apply to CI?

pkg/pf/internal/providerbuilder/build_resource.go
```diff
@@ -26,10 +26,13 @@ import (
...
 
 type NewResourceArgs struct {
-	// Name is the name of the resource. Defaults to "test".
-	Name           string
+	Name string
+
 	ResourceSchema schema.Schema
 
+	// Do not inject a Computed String "id" attribute into the schema.
+	CustomID bool
```
VenelinMartinov, line 34: Is this used anywhere? AFAIK, the pulumi engine balks if it sees resources without an `id` prop.

pkg/tfshim/shim.go
```diff
@@ -25,6 +27,11 @@ type InstanceState interface {
 	Meta() map[string]interface{}
 }
 
+// Newer versions of the bridge want to interact with a cty.Value representation of the state.
+type InstanceStateWithCtyValue interface {
+	Value() cty.Value
```
VenelinMartinov, line 32: Should this return a `valueshim.Value` instead? We don't actually need a `cty.Value` at the call site and `PF` could implement this too if it was a `valueshim.Value`

VenelinMartinov (approved):
Awesome, a few minor comments but glad we got there!

pkg/pf/tests/upgrade_state_cross_test.go
t0yv0: It should tell you why though? If you don't skip it presumably it would fail. I think I prefer the current way.

pkg/tfshim/shim.go
t0yv0: That's a nice observation. Yes we can do that.

pkg/tfbridge/rawstate_test.go
iwahbe, line 1: It seems like it would be worth writing a property based test here. If we could generate a `cty.Value` (We can already [generate](https://github.com/pulumi/pulumi-terraform-bridge/blob/aa252f8b3e7536b8794b34bd33d645840fdae7f0/pkg/internal/tests/cross-tests/rapid_tv_gen.go#L26-L32) a `tftypes.Value` - maybe we can translate from that) and then assert on round-tripping that, I think that would provide a useful fallback on round tripping values.

iwahbe (commented):
I have some suggestions, but I don't see anything major. Strategically, my only recommendation is that we enable testing roundtripping by default (warning and requesting an error to be reported) before we enable the feature by default.

pkg/tfbridge/provider.go
```diff
@@ -1407,11 +1419,31 @@ func (p *Provider) Read(ctx context.Context, req *pulumirpc.ReadRequest) (*pulum
...
+		state, err = makeTerraformStateViaUpgrade(ctx, pNew, res, props)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unmarshaling %s's instance state via upgrade", urn)
+		}
+	} else {
+		state, err = makeTerraformStateWithOpts(ctx, res, req.GetId(), props, makeTerraformStateOptions(opts))
+		if err != nil {
+			return nil, errors.Wrapf(err, "unmarshaling %s's instance state", urn)
+		}
 	}
```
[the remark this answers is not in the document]
t0yv0: It started off that way but got the exact reverse feedback from @VenelinMartinov ; I think I'll leave it. Style nits.

Issue #2949, opened 2025-03-13
Release v3.105.0


t0yv0:
Preparing the v3.105.0 release, currently getting these results:

```
python scripts/downstream_checks.py --hash de5510662e1c87d0851b19405314503376fdb81b
```

FAILED: https://github.com/pulumi/pulumi-aws/pull/5294
FAILED: https://github.com/pulumi/pulumi-azuredevops/pull/626

WAITING random
WAITING auth0
WAITING onelogin
WAITING gitlab

t0yv0:
in AWS:

Error: ../../../../go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.104.1-0.20250313133841-de5510662e1c/pkg/tfshim/sdk-v2/schema.go:129:14: s.tf.WriteOnly undefined (type *"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema".Schema has no field or method WriteOnly)


Looks like something to do with the SDK override in go.mod. Need to double check.

t0yv0:
Same in pulumi-azure-devops.

t0yv0:
Random panicked with:

https://github.com/pulumi/pulumi-random/actions/runs/13839680005/job/38723675572#step:5:1925

```
panic: runtime error: slice bounds out of range [917:914]

goroutine 1 [running]:
github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen.(*Generator).convertExamplesInner(0xc000983b08, {0xc00079fc00, 0x392}, {{0xc0005f4360?, 0xc0001181b0?}, {0xc0005fb780?, 0xe?}}, 0xc0008c95d0, 0x0)
	/home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.104.1-0.20250313133841-de5510662e1c/pkg/tfgen/docs.go:1654 +0x8f1
github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfgen.(*cliConverter).StartConvertingExamples(0xc000854848, {0xc00079fc00, 0x392}, {{0xc0005f4360?, 0x1acdf20?}, {0xc0005fb780?, 0x1acdf20?}})
	/home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v3@v3.104.1-0.20250313133841-de5510662e1c/pkg/tfgen/convert_cli.go:136 +0xe3
[...]

PR #2951, opened 2025-03-13
Revert "Use goldmark to find code blocks in `docs.go`"

t0yv0:
Reverts pulumi/pulumi-terraform-bridge#2939

This regressed auth0 builds and caused panics. Reverting it removes the panics.
https://github.com/pulumi/pulumi-terraform-bridge/issues/2949

I tried fixing forward something along these lines, note that section.Section always has ast.Heading as first child and it may not have any lines in it somehow:

```go


func findCodeBlocks(docs []byte) []codeBlock {
	rootNode := goldmark.New(goldmark.WithExtensions(parse.TFRegistryExtension)).
		Parser().Parse(gmtext.NewReader(docs))

	var codeBlocks []codeBlock
	parse.WalkNode(rootNode, func(cb *ast.FencedCodeBlock) {
		lines := cb.Lines()

		headerStart := -1
		for p := cb.Parent(); p != nil; p = p.Parent() {
			if s, ok := p.(*section.Section); ok {

				start := -1
				for c := s.FirstChild(); c != nil; c = c.NextSibling() {
					if c.Lines().Len() > 0 {
						start = c.Lines().At(0).Start
						break
					}
				}

				if start == -1 {
					panic("!")
				}

				headerStart = bytes.LastIndexByte(docs[:start], '\n') + 1
				break
			}
		}

		codeBlocks = append(codeBlocks, codeBlock{
			start:       bytes.LastIndexByte(docs[:lines.At(0).Start-1], '\n') + 1,
			end:         lines.At(lines.Len() - 1).Stop,
			headerStart: headerStart,
		})
	})
	return codeBlocks
}

```

This fix gets past the panic but the change still drops all import sections from Auth0. I am not sure how sensitive we are to import sections but I think we are. So I do not think we should fix forward just yet...

I realize reverting this opens another panic in dynamic provider.


t0yv0:
With respect to how this would work ideally.. Perhaps what could be useful is defensive code in this section (lots of `error`) and find a way to gracefully proceed instead of panicking. This combined with data quality checks. I have not looked here for ages - we have some infrastructure I think for deciding if an example renderer failed and some metrics and alerts on that, but I am not sure if the failure in question here can be counted against a code example metric in particular.

guineveresaenger:
We should definitely recover the result from auth0 failure and add test case.

guineveresaenger (approved):
sigh. LGTM, thank you for trying to fix forward

PR #2971, opened 2025-03-26
Update pulumi/pulumi to v3.158.0

guineveresaenger:
This skips one version of pulumi.

Also updates pulumi-yaml and pulumi-java to their respective `latest`.

Small adjustments to GetPluginPath and ResolvePlugin.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2958.


Makefile
```diff
@@ -6,7 +6,7 @@ export PULUMI_DISABLE_AUTOMATIC_PLUGIN_ACQUISITION := true
 PROJECT_DIR := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
 
 install_plugins::
-	pulumi plugin install converter terraform 1.0.20
+	pulumi plugin install converter terraform 1.1.0
```
t0yv0, line 9: Why is this necessary, curious? Adds to the scope of the change. Perhaps separate PR?

pkg/pf/internal/plugin/provider_server.go
```diff
@@ -561,10 +561,11 @@ func (p *providerServer) Construct(ctx context.Context,
 		}
 		propertyDependencies[resource.PropertyKey(name)] = urns
 	}
+	protect := req.GetProtect()
 	options := pl.ConstructOptions{
 		Aliases:              aliases,
 		Dependencies:         dependencies,
-		Protect:              req.GetProtect(),
+		Protect:              &protect,
```
t0yv0, line 568: GetProtect() papers over the nil I think.  A bit dangerous here.

pkg/tfgen/generate_schema.go
```diff
@@ -832,9 +832,7 @@ func (g *schemaGenerator) genResourceType(mod tokens.Module, res *resourceType)
 
 	for _, a := range res.info.Aliases {
 		spec.Aliases = append(spec.Aliases, pschema.AliasSpec{
-			Name:    a.Name,
-			Project: a.Project,
-			Type:    a.Type,
+			Type: *a.Type,
```
t0yv0, line 835: This can needlessly panic. A bit dangerous.

pkg/tfgen/installation_docs_test.go
```diff
@@ -424,13 +424,13 @@ func TestTranslateCodeBlocks(t *testing.T) {
 		name       string
 		contentStr string
 		g          *Generator
-		expected   string
+		expected   autogold.Value
 	}
 	p := tfbridge.ProviderInfo{
-		Name: "simple",
```
t0yv0, line 430: Why change from simple? Actually the delta on the "simple" provider is interesting to investigate.

Makefile
guineveresaenger: Local test runs do not enforce this constraint unless you run `make`, which is difficult to do for an individual test. A lot of the pain here comes from different environments doing different things. :/

pkg/pf/internal/plugin/provider_server.go
guineveresaenger: GetProtect() does have a nil check as of [this change](https://github.com/pulumi/pulumi/pull/18924/files#diff-e38a8413bb3b6d73725cb4349bdb60867cc13b78ac22b4efa722812972f7e7b6R2669)

Regardless, it looks like we can call the field directly now, so we can do that instead.

pkg/tfgen/installation_docs_test.go
guineveresaenger: It is because of a change in the converter that doesn't allow for "fake" providers. Since this code path calls out to the command line, the converter attempts to download a provider called "simple", and it results in breaking Java codegen in this test.

The delta is not really all that interesting, because it is caused by the fact that this provider doesn't exist in real life.

guineveresaenger: @lunaris and I did investigate for quite a while on what happened here on Friday.

lunaris: The challenge here now is that the converter will call out to Terraform's registry in order to try and resolve versions for providers that it thinks are outside the Pulumiverse. It's always done this, but this check is now earlier in the call chain in order to facilitate converting "any TF provider". We do this by dynamically bridging providers not in the Pulumiverse to e.g. facilitate generating mappings. This test now fails because "simple" is not in the Pulumiverse, but it is also not a "real" TF provider, so version resolution (and consequently producing a hint for mapping, which powers part of this test case) fails. There are several options here, none of them (on the surface) ideal, but enumerating in case people have thoughts/better ideas than I:

* Don't fail mapping hint generation for TF providers we can't resolve. I think this is fine, but it does make it trickier for `convert` to produce errors when e.g. the user has genuinely mentioned a provider that (as far as we're concerned) doesn't exist. That said, I think fixing that gets us further in this test (though it may fail later on due to missing/incompletely-mocked schema, as discussed elsewhere e.g. in the case of Java imports)
* More completely mock out the processes that power this test -- we already mock out mappings (passing a JSON file on the CLI to the converter, so that it can just use those mappings as-is without remote calls), but we don't mock out the TF registry. We could look at doing that somehow here (likely not pretty, but possible)?
* Use a "real" provider so that the TF registry lookups succeed. This is the approach currently in this PR I believe. Again, not ideal, but I think (?) the simplest change?

Makefile
t0yv0: Good point on enforcement, let us fix this.  This kind of change needs to be a separate PR otherwise it is hard to see which change comes from what IMO.

pkg/tfgen/installation_docs_test.go
t0yv0: Reasonable but maybe a different PR? Too many moving parts at once. Also it needs to pin a version of openstack such as in here or why not use one that's already there perhaps.

https://github.com/pulumi/pulumi-terraform-bridge/blob/4c35e78f1e3979b8617260614f9636523ec3664d/Makefile#L8

Makefile
guineveresaenger: Yeah, completely agreed.

t0yv0:
Landed https://github.com/pulumi/pulumi-terraform-bridge/pull/2973 but there might be some interesting residual things in this PR that are worth merging up. There is in particular an upgrade to the TF converter version we use in tests. Note that it is not currently coupled with the converter version deployed in the bridged providers (pinned in .ci-mgmt.yaml) but both could use updating I guess.

PR #2981, opened 2025-04-07
Improve test isolation for TestTranslateCodeBlocks

### LINKED ISSUE #2974: testing: TestTranslateCodeBlocks insufficiently isolated
I am having trouble passing TestTranslateCodeBlocks on main as of this morning. We have some isolation issues: running this test modifies the global pulumi plugin cache to auto-install these providers:

```
pulumi plugin ls                                                                                                                                                                    ~/code/pulumi-terraform-bridge
NAME                KIND       VERSION  SIZE   INSTALLED     LAST USED
terraform           converter  1.1.0    40 MB  1 minute ago  1 minute ago
terraform-provider  resource   0.8.1    69 MB  1 minute ago  1 minute ago
```

Additionally it uses a floating version of the `simple` terraform provider here:

https://github.com/pulumi/pulumi-terraform-bridge/blob/0ab89aeca2a9e838887bf0351fb7109bb16ef541/pkg/tfgen/test_data/installation-docs/configuration.md?plain=1#L12

So in effect the test runs `pulumi` CLI under the hood, which engages 'any-terraform-provider' to auto-bridge the provider and run the conversion.

I am adding this to track a Skip here, we have some options to move forward. The scope of the test is extremely broad, which is also why the test is slow, perhaps we could replace with a more narrow test. If we keep it as a broad acceptance test, we should pin and isolate all dependencies.

guineveresaenger:
This pull request fixes up `TestTranslateCodeBlocks`:

- Provides a fully assigned Generator with a mock test host to the test case. This: 
    - prevents the underlying code from attempting to re-create a Generator for each Convert call, resulting in very quick rate limiting
    - additionally prevents the automatic downloading of the terraform converter plugin 
- Ensures the test provider names in the test files match the test provider name in the test (`simple`).
- Removes the `required_providers` section as unnecessary

As a result of https://github.com/pulumi/pulumi-converter-terraform/pull/319, which is now in use by the bridge, we also no longer download the terraform-provider for a nonexisting provider.

Fixes #2974

### DIFF
--- a/pkg/tfgen/installation_docs_test.go
+++ b/pkg/tfgen/installation_docs_test.go
@@ -425,7 +425,7 @@ func TestTranslateCodeBlocks(t *testing.T) {
 		contentStr string
 		g          *Generator
 	}
-	p := tfbridge.ProviderInfo{
+	providerInfo := tfbridge.ProviderInfo{
 		Name: "simple",
 		P: sdkv2.NewProvider(&schema.Provider{
 			ResourcesMap: map[string]*schema.Resource{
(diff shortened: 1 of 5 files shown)


t0yv0:
Feel free to check this in but this does not fix 2974.

guineveresaenger:
You're right, this PR is mainly a response to changes in the converter, which add _additional_ network calls when encountering a `required_provider` section, which this test doesn't need.

Going to add my thoughts on #2974 over there - had a few thoughts.

t0yv0:
Your real problem in this test case is

```
Configure the OpenStack Provider
provider "simple-provider" {
  user_name   = "admin"
  tenant_name = "admin"
  password    = "pwd"
  auth_url    = "http://myauthurl:5000/v3"
  region      = "RegionOne"
}
```

Should be probably this:

```
provider "simple"
```

As it thinks provider-simple is very different from the one needed for simple_ resources.

t0yv0:
It would be great to understand what this fixes, and possibly take this as an improvement but https://github.com/pulumi/pulumi-converter-terraform/pull/319 remains a problem as of this morning:

```
anton@anton-mbp-m3> pulumi plugin ls                                                                                                     ~/code/pulumi-terraform-bridge
NAME       KIND       VERSION  SIZE   INSTALLED      LAST USED
terraform  converter  1.1.0    40 MB  8 seconds ago  8 seconds ago
```

The test is not isolated as it auto-installs the converter at latest version.

Am I perhaps running this wrong?

guineveresaenger:
Hm 🤔 - this test should now be exactly as isolated as `TestConvertViaPulumiCLI`. I thought this meant the only way it would install the converter was by going through the converter install in `Makefile` but it appears I was wrong.

I'm not sure that it is a significant problem anymore though - the test will not attempt to install the latest converter, and instead abide by the converter version specified in Make. So if tests are run via Make, which is what CI uses, this should be OK.

t0yv0 (approved):
If it's isolated in CI that's probably good enough for now.. Sounds like the installation of a concrete version is pinned in Makefile. I'd like to fix someday so that the tests could be invoked directly through Go w/o Makefile.

PR #2985, opened 2025-04-08
Add DecodeOptions to pkg/convert

t0yv0:
Adds an ability to preserve object property nulls to pkg/convert. As part of #1667 solution the bridge is adding deltas that encode the difference between TF RawState and Pulumi PropertyMap state so that the TF RawState can be reconstructed byte-for-byte at read time and passed to TF state upgraders as is. The RawState will be passed to UpgradeResourceState request method as in: https://github.com/hashicorp/terraform-plugin-go/blob/v0.26.0/tfprotov6/resource.go#L114

The question here is around object types with missing attributes. In the RawState representation, is there any semantic difference between `"foo": null` and omitting "foo" from the RawState? The change errs on the side that any such nulls may be significant.

The PR does not change behavior but introduces options to be used in the #1667 solution.

### DIFF
--- a/pkg/convert/adapter.go
+++ b/pkg/convert/adapter.go
@@ -45,12 +45,12 @@ type adaptedDecoder[T Decoder] struct {
 	decoder T
 }
 
-func (d adaptedDecoder[T]) toPropertyValue(v tftypes.Value) (resource.PropertyValue, error) {
+func (d adaptedDecoder[T]) toPropertyValue(v tftypes.Value, dopts DecodeOptions) (resource.PropertyValue, error) {
 	adapted, err := d.adapter(v)
 	if err != nil {
 		return resource.PropertyValue{}, fmt.Errorf("failed to adapt for %T: %w", d.decoder, err)
 	}
-	return decode(d.decoder, adapted)
+	return decode(d.decoder, adapted, dopts)
 }
 
 func newIntOverrideStringEncoder() Encoder {
--- a/pkg/convert/bool.go
+++ b/pkg/convert/bool.go
@@ -48,7 +48,7 @@ func (*boolEncoder) fromPropertyValue(p resource.PropertyValue) (tftypes.Value, 
 	return tftypes.NewValue(tftypes.Bool, p.BoolValue()), nil
 }
 
-func (*boolDecoder) toPropertyValue(v tftypes.Value) (resource.PropertyValue, error) {
+func (*boolDecoder) toPropertyValue(v tftypes.Value, _ DecodeOptions) (resource.PropertyValue, error) {
 	if !v.IsKnown() {
 		return unknownProperty(), nil
 	}
--- a/pkg/convert/convert.go
+++ b/pkg/convert/convert.go
@@ -100,7 +100,7 @@ type Decoder interface {
 	//
 	// This is because toPropertyValue is not required to handle [tftypes.Value]
 	// modifiers, like secrets and unknown values.
-	toPropertyValue(tftypes.Value) (resource.PropertyValue, error)
+	toPropertyValue(value tftypes.Value, dopts DecodeOptions) (resource.PropertyValue, error)
 }
 
 func NewObjectDecoder(os ObjectSchema) (Decoder, error) {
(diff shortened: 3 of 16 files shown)


VenelinMartinov (commented):
The changes in this PR seem fine but can you please add more detail as to why we distinguish explicit null from missing for Objects?

What meaning do we ascribe to the explicit nulls? Given that this is intended for consumption only on our side would it not be better to have an explicit marker for flattened properties which distinguishes null from empty on the TF side? Something like `__isEmpty: true` or the other way around to mark the non-default case.

t0yv0:
A little experiment:

```

func TestNullPreservation(t *testing.T) {
	t.Parallel()
	prov := func(v int) *pb.Provider {
		return pb.NewProvider(pb.NewProviderArgs{
			AllResources: []pb.Resource{
				pb.NewResource(pb.NewResourceArgs{
					Name: "res",
					ResourceSchema: rschema.Schema{
						Version: int64(v),
							"n": rschema.StringAttribute{
								Computed: true,
							},
							"s": rschema.StringAttribute{
								Optional: true,
								Computed: true,
							},
							"x": rschema.StringAttribute{
								Optional: true,
							},
						},
					},
					CreateFunc: func(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
						if 1+2 == 4 {
							resp.Diagnostics = resp.State.SetAttribute(ctx, path.Root("s"), nil)
						}
						resp.Diagnostics = resp.State.SetAttribute(ctx, path.Root("n"), "OK")
					},
					UpdateFunc: func(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
						panic(req.State.Raw.String())
						resp.Diagnostics = resp.State.SetAttribute(ctx, path.Root("x"), "OK")
					},
					UpgradeStateFunc: func(ctx context.Context) map[int64]resource.StateUpgrader {
						return map[int64]resource.StateUpgrader{
							0: resource.StateUpgrader{
								StateUpgrader: func(ctx context.Context, usr1 resource.UpgradeStateRequest, usr2 *resource.UpgradeStateResponse) {
									panic(string(usr1.RawState.JSON))
								},
							},
						}
					},
				}),
			},
		})
	}

	driver := tfcheck.NewTfDriver(t, t.TempDir(), prov(0).TypeName, prov(0))

	driver.Write(t, `
resource "testprovider_res" "test" {
}
output "n_val" {
	value = testprovider_res.test.n
}
output "s_val" {
	value = testprovider_res.test.s
}
`)

	plan, err := driver.Plan(t)
	require.NoError(t, err)
	err = driver.Apply(t, plan)
	require.NoError(t, err)

	state := driver.GetState(t)

	driver = tfcheck.NewTfDriver(t, t.TempDir(), prov(1).TypeName, prov(1))
	driver.SetState(t, state)

	driver.Write(t, `
resource "testprovider_res" "test" {
  x = "OK"
}
output "n_val" {
	value = testprovider_res.test.n
}
output "s_val" {
	value = testprovider_res.test.s
}
`)
	plan, err = driver.Plan(t)
	require.NoError(t, err)

	err = driver.Apply(t, plan)
	require.NoError(t, err)

	//require.Equal(t, "OK", driver.GetOutput(t, "n_val"))
	//require.Equal(t, "", driver.GetOutput(t, "s_val"))
}

```

Results in:

```
panic: {
	                        "id": null,
	                        "n": "OK",
	                        "s": null,
	                        "x": null
	                    }
```

t0yv0:
It is inconclusive to me if nulls in RawState in TF are actually semantically meaningful but the conservative approximation is to include them; ctyjson.Marshal includes them and per the test above they can be observed it seems.

t0yv0:
As to why preserving them in PropertyMap. This is done in current PF code when computing the deltas, to pass on to the delta compressor. There is a little bit of code-smell there, let me revisit if this can be gotten rid of.

VenelinMartinov (approved):

t0yv0:
Actually I found a way to resolve this without the code smell. I've improved the delta computation to take into account when a value is dropped from Pulumi but present in TF raw state, like a null, and it can handle it now without having to use this trick.

PR #2988, opened 2025-04-09
Update pu/pu to 3.161 and remove StreamInvoke

Frassle:
This updates pulumi/pulumi/pkg and sdk to v3.161 and removes all the StreamInvoke code that's also been removed from pulumi core.


VenelinMartinov (approved):
Looks fine, seems like this was never used?

Frassle:
> seems like this was never used?

Was only ever used in the kubernetes provider.

PR #2995, opened 2025-04-10
Do not run Upgrade during Read-for-import

t0yv0:
This is a subtle change in behavior for Plugin Framework implementation of resource import. Before the change the implementation of import executed UpgradeResourceState on the underlying provider, but it no longer does so. Upgrading the imported state is unnecessary. It should also save a redundant `ToPropertyMap()` call.


t0yv0:
Trying to write a test here I realized this is actually a mis-interpretation of the current code. This does not make sense outside of the context of https://github.com/pulumi/pulumi-terraform-bridge/pull/2945 - redoing it there.

PR #3004, opened 2025-04-17
Update pulumi/pulumi/{sdk/pkg} to 3.163.0

julienp:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/3002

3.163.0 changed the Node.js and Python GeneratePackage  functions, they now take a reference loader.


VenelinMartinov (approved):
lgtm

PR #3021, opened 2025-04-22
chore: fixup newPulumiTest

t0yv0:
Per https://github.com/pulumi/pulumi-terraform-bridge/pull/3016#discussion_r2054658146 there is now a capability to start a PF provider directly for testing without calling main functions. Capability is coming from https://github.com/pulumi/pulumi-terraform-bridge/pull/2075 and is only working for PF-based providers at the moment. Still using this capability makes the given test helper code much simpler as muxing need not be involved.

### DIFF
--- a/pkg/pf/tests/pulumi_test.go
+++ b/pkg/pf/tests/pulumi_test.go
@@ -15,58 +15,23 @@
 package tfbridgetests
 
 import (
-	"context"
 	"testing"
 
-	"github.com/hashicorp/terraform-plugin-framework/provider"
-	sdkv2schema "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/pulumi/providertest/pulumitest"
 	"github.com/stretchr/testify/require"
 
-	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/internal/tests/pulcheck"
-	pf "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/pf/tfbridge"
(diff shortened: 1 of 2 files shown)


pkg/pf/tests/pulumi_test.go
```diff
@@ -15,58 +15,23 @@
...
 	"github.com/pulumi/providertest/pulumitest"
 	"github.com/stretchr/testify/require"
 
-	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/internal/tests/pulcheck"
-	pf "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/pf/tfbridge"
-	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/info"
-	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/tokens"
-	sdkv2shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim/sdk-v2"
+	pb "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/pf/internal/providerbuilder"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/pf/tests/pulcheck"
```
t0yv0, line 24: I had to swap out a different pulcheck. The fact that there is two is a smell. The original one does not work, because tfbridge methods for starting provider servers still do not accept the PF providers in the info.P position. This should be something we can maybe fix later, but a bigger change.

VenelinMartinov (approved):

PR #3026, opened 2025-04-23
chore: do not fail-fast when testing

t0yv0:
When making some of the more extensive changes to the bridge, it is helpful to have a list of all failing tests. Running all the tests locally takes a fair bit of time so currently in practice we often rely on CI to identify test failures. It is helpful to identify all of them instead of just the first one.


VenelinMartinov (approved):

guineveresaenger (commented):
Do we want this in all cases though? 

I think this is the lesser evil; can't hurt to try it out,

t0yv0:
It can increase CI costs.

PR #3047, opened 2025-05-02
Update the `go-version` so we test supported go versions

iwahbe:
We should test the Go versions we actually support: 1.23 & 1.24.


iwahbe:
Screenshot 2025-05-02 at 12 57 21 PM

Interestingly enough, even though we theoretically installed a `go 1.22.x` version to test against, it looks like we actually wound up with go1.23.

PR #3054, opened 2025-05-07
Upgrade pulumi to v3.167.0

t0yv0:
Also upgrade github.com/pulumi/pulumi-java/pkg v1.11.0 as that is required to be build-compatible with latest Pulumi. 

Changes around `pschema.BindSpec` options are now satisifed by opting into AllowDanglingReferences: true which I believe was the prior behavior. We need https://github.com/pulumi/pulumi-aws/issues/5250 and similar to be solved in providers before we can tighten this to the strict mode.


guineveresaenger (approved):
LGTM

One thing we could add here is re-enable https://github.com/pulumi/pulumi-terraform-bridge/blob/main/dynamic/provider_test.go#L420 since https://github.com/pulumi/pulumi/issues/19274 was fixed and is in 3.167.

But happy to do as a followup as well.

go.mod
```diff
@@ -220,8 +220,8 @@ require (
 	github.com/pkg/term v1.1.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/posener/complete v1.2.3 // indirect
-	github.com/pulumi/pulumi/pkg/v3 v3.166.0
-	github.com/pulumi/pulumi/sdk/v3 v3.166.0
+	github.com/pulumi/pulumi/pkg/v3 v3.167.1-0.20250507103745-1e3ee619cf4d
```
t0yv0, line 223: Hmm that's not right I wonder what happened here. It broke our automation though. Let me revert quick.

Issue #3078, opened 2025-05-16
DynamicPseudoType is not yet supported


t0yv0:
### What happened?

A user reports hitting this in cloudflare: https://github.com/pulumi/pulumi-cloudflare/issues/1197#issuecomment-2886931529

The problem is here which is a miss:

https://github.com/pulumi/pulumi-terraform-bridge/blob/d7f6622fc1793217bb6246bb078d071f5c367c1f/pkg/pf/internal/pfutils/value_to_json.go#L94

In particular this https://github.com/pulumi/pulumi-terraform-bridge/pull/2362 feature was incomplete and missed an update test.

Researching a possible fix a bit, we might want to extend valueshim to handle dynamic types and remove the package that currently contains the error to reduce duplication. 

The question becomes *how* should the dynamic types be represented in RawState.

What we are trying to do is build a tfprotov6.RawState to send to a providers' UpgradeResourceState method. terraform-plugin-go has the inverse logic (parsing this JSON form) but not serializing it:

https://github.com/hashicorp/terraform-plugin-go/blob/v0.26.0/tfprotov6/state.go#L45
https://github.com/hashicorp/terraform-plugin-go/blob/v0.26.0/tftypes/value_json.go#L21
https://github.com/hashicorp/terraform-plugin-go/blob/v0.26.0/tftypes/value_json.go#L194

This is interesting how this data is processed:

https://github.com/opentofu/opentofu/blob/4083fe2cc4a483fc19d9769e0eb19ebee2386b33/internal/tofu/upgrade_resource_state.go#L28
https://github.com/opentofu/opentofu/blob/4083fe2cc4a483fc19d9769e0eb19ebee2386b33/internal/states/instance_object_src.go#L42
https://github.com/opentofu/opentofu/blob/4083fe2cc4a483fc19d9769e0eb19ebee2386b33/internal/states/instance_object_src.go#L89
https://github.com/opentofu/opentofu/blob/4083fe2cc4a483fc19d9769e0eb19ebee2386b33/internal/states/instance_object.go#L113

I think ctyjson.Umarshal() and ctyjson.Marshal cover the correct form.

https://github.com/hashicorp/go-cty/blob/v1.4.1-0.20200414143053-d3edf31b6320/cty/json/marshal.go#L182

Both seem to point out that the serialized form for DynamicPseudoType is `{"type": ..., "value": "..."}` in Raw JSON, that is the concrete type of the value is written out into this form.

[...]

PR #3079, opened 2025-05-16
WIP: fix parsing dynamic pseudo type

t0yv0:
Toward fixing https://github.com/pulumi/pulumi-terraform-bridge/issues/3078

Surprisingly this still cannot pass the test because it is now doing: 

```priorState.Value.Diff(plannedStateValue)```

and that somehow blows up in the presence of values with DynamicPseudoType type.

I get this:

```
        stderr: error: AttributeName("dyn"): Error converting tftypes.Number<"42"> (value2) at "AttributeName(\"dyn\")": can't unmarshal tftypes.Number into *string, expected string
```


t0yv0:
Reminder to self: tuples don't yet fully integrate through RawStateDelta so tuple paths are not there.

PR #3118, opened 2025-06-04
Add NoDanglingReferences option

### LINKED ISSUE #3116: Allow configuring `AllowDangingReferences` in schema generation
## Hello!

## Issue details

When we validate the schema [here](https://github.com/pulumi/pulumi-terraform-bridge/blob/92748b5518d84600fabbbf16ec14792470c6c8fb/pkg/tfgen/generate_schema.go#L445-L445) we currently hardcode `AllowDanglingReferences` to `true`. In AWS v7 we are removing all dangling references so it would be nice to be able to set this to `false` so we don't have to add a separate test for this.

t0yv0:
Historically providers such as pulumi-aws generated schemata that had type references that cannot resolve, and relied
on the quirks of the Node SDK generator to make use of these references to link in overlay types.

This is going away in upcoming major release of pulumi-aws, and newer providers should be strict about generating the
schema.

When the new flag NoDanglingReferences is set to `true`, bridge will perform strict schema validity checks before
emitting a schema during `make tfgen`.

Fixes #3116

### DIFF
--- a/pkg/tfbridge/info/info.go
+++ b/pkg/tfbridge/info/info.go
@@ -186,6 +186,9 @@ type Provider struct {
 
 	// DisableRequiredWithDefaultTurningOptional disables making required properties optional if they have a default value.
 	DisableRequiredWithDefaultTurningOptional bool
+
+	// Check generated schema for dangling references. Newer providers should opt into this.
+	NoDanglingReferences bool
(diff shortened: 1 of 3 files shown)


corymhall (approved):
Nice!

PR #3147, opened 2025-08-06
Add --resources flag to filter dynamic provider resources

iwahbe:
This PR introduces a `--include` flag that allows users to specify which resources and datasources should be included when generating a dynamic Pulumi provider from a Terraform provider. This enables creating focused providers that contain only the subset of resources/datasources needed for a particular use case.

## Changes

The implementation adds a `Includes` field to both `Args` and `Value` structs to track the list of resource/datasources names to include. When specified, the filtering is applied in the `providerInfo` function by leveraging the existing `IgnoreMappings` mechanism - all resources/datasources not in the include list are added to `IgnoreMappings`, effectively excluding them from the generated provider.

Additionally, the `fixup.Default` function was modified to respect `IgnoreMappings`, ensuring ignored resources are completely skipped during fixup processing and eliminating the need for manual cleanup code.

Users can now run commands like:

    pulumi package get-schema terraform-provider -- hashicorp/aws --include aws_instance,aws_vpc             

This results in a focused AWS provider containing only the specified resources, reducing the generated provider size and improving development ergonomics when working with large Terraform providers.

Fixes https://github.com/pulumi/pulumi-terraform-provider/issues/91

### DIFF
--- a/dynamic/info.go
+++ b/dynamic/info.go
@@ -123,20 +123,6 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
 		}
 
 		prov.IgnoreMappings = ignoreMappings
-
-		// Remove ignored resources from the maps (fixup may have pre-populated them)
-		//
-		// TODO: fixup.Default should ignore ignored mappings, so we shouldn't need this.
-		if prov.Resources != nil {
-			for _, ignored := range ignoreMappings {
-				delete(prov.Resources, ignored)
-			}
-		}
-		if prov.DataSources != nil {
-			for _, ignored := range ignoreMappings {
-				delete(prov.DataSources, ignored)
-			}
-		}
 	}
 
 	if err := fixup.Default(&prov); err != nil {
--- a/dynamic/info.go
+++ b/dynamic/info.go
@@ -101,10 +101,10 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
 	}
 
 	// Apply resource filtering if specified
-	if len(value.Resources) > 0 {
+	if len(value.Includes) > 0 {
 		// Create a set of resources to include
 		includeSet := make(map[string]bool)
-		for _, resource := range value.Resources {
+		for _, resource := range value.Includes {
 			includeSet[resource] = true
 		}
 
--- a/dynamic/info.go
+++ b/dynamic/info.go
@@ -100,6 +100,45 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
 		prov.Repository = "https://github.com/" + ghOrg + "/terraform-provider-" + name
 	}
 
+	// Apply resource filtering if specified
+	if len(value.Resources) > 0 {
+		// Create a set of resources to include
+		includeSet := make(map[string]bool)
+		for _, resource := range value.Resources {
+			includeSet[resource] = true
+		}
+
+		// Add all resources NOT in the include set to IgnoreMappings
+		var ignoreMappings []string
+		for key := range provider.ResourcesMap().Range {
+			if !includeSet[key] {
+				ignoreMappings = append(ignoreMappings, key)
+			}
+		}
+		// Also check data sources
+		for key := range provider.DataSourcesMap().Range {
+			if !includeSet[key] {
+				ignoreMappings = append(ignoreMappings, key)
+			}
+		}
+
+		prov.IgnoreMappings = ignoreMappings
+
+		// Remove ignored resources from the maps (fixup may have pre-populated them)
+		//
+		// TODO: fixup.Default should ignore ignored mappings, so we shouldn't need this.
+		if prov.Resources != nil {
+			for _, ignored := range ignoreMappings {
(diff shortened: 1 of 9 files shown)


dynamic/info.go
```diff
@@ -100,6 +100,31 @@ func providerInfo(ctx context.Context, p run.Provider, value parameterize.Value)
...
+		var ignoreMappings []string
+		for key := range provider.ResourcesMap().Range {
+			if !includeSet[key] {
+				ignoreMappings = append(ignoreMappings, key)
+			}
+		}
+		// Also check data sources
+		for key := range provider.DataSourcesMap().Range {
+			if !includeSet[key] {
+				ignoreMappings = append(ignoreMappings, key)
```
guineveresaenger, line 121: It looks like we're only filtering for resources on L#104. Why would we need to add datasources keys to ignored mappings? Should we not allow to also filter by data source?

dynamic/info_test.go
```diff
@@ -144,6 +144,131 @@ func TestFixHyphenToken(t *testing.T) {
...
+		for tfName := range info.Resources {
+			actualResources = append(actualResources, tfName)
+		}
+		for tfName := range info.DataSources {
+			actualResources = append(actualResources, tfName)
+		}
+		assert.ElementsMatch(t, expectedResources, actualResources)
+	})
+
+	t.Run("single resource filter", func(t *testing.T) {
```
guineveresaenger:
```suggestion
	t.Run("single resource filter has single resource", func(t *testing.T) {
```

dynamic/info_test.go
```diff
@@ -144,6 +144,131 @@ func TestFixHyphenToken(t *testing.T) {
...
+		for tfName := range info.Resources {
+			actualResources = append(actualResources, tfName)
+		}
+		for tfName := range info.DataSources {
+			actualResources = append(actualResources, tfName)
+		}
+		assert.ElementsMatch(t, expectedResources, actualResources)
+	})
+
+	t.Run("non-existent resource", func(t *testing.T) {
```
guineveresaenger:
```suggestion
	t.Run("non-existent resource does not affect filter", func(t *testing.T) {
```

dynamic/info_test.go
```diff
@@ -144,6 +144,131 @@ func TestFixHyphenToken(t *testing.T) {
 	}, p.Resources)
 }
 
+func TestResourceFiltering(t *testing.T) {
```
guineveresaenger: Since we're adding data sources to ignoreMappings, can we add an explicit test?

dynamic/internal/fixup/properties.go
```diff
@@ -41,13 +41,26 @@ import (
...
 
+func newFixCtx(p *info.Provider) fixCtx {
+	ignoredMappings := make(map[string]struct{}, len(p.IgnoreMappings))
+	for _, v := range p.IgnoreMappings {
+		ignoredMappings[v] = struct{}{}
+	}
+	return fixCtx{ignoredMappings}
+}
+
+type fixCtx struct {
```
guineveresaenger, line 60:
```suggestion
type generatedResources struct {
```

Something that conveys "hey, we don't necessarily generate all resources"...

dynamic/parameterize/args.go
```diff
@@ -80,6 +84,8 @@ func ParseArgs(ctx context.Context, a []string) (Args, error) {
 		"Specify a local file path to the root of the Git repository of the provider being dynamically bridged")
 	cmd.Flags().StringVar(&indexDocOutDir, "indexDocOutDir", "",
 		"Specify a local output directory for the provider's _index.md file")
+	cmd.Flags().StringSliceVar(&resources, "resources", nil,
+		"Comma-separated list of resource names to include in the provider (e.g. aws_instance,aws_vpc)")
```
guineveresaenger: Might explain more about the exclusivity here.

dynamic/info.go
iwahbe: `IgnoredMappings` applies to both resources and datasources. I've changed `Resources` to `Selectors`. It doesn't really distinguish between resources and datasources.

iwahbe:
> This feature sounds less like a "filter" and more of a "selection". Specifically, if a user specifies, say, four resources, we only generate those resources, and nothing else, if I understand correctly.

I've changed the command to `--include`, since that makes a bit more sense. It makes it clearer that what it does and that it applies to both datasources and resources.

dynamic/internal/fixup/properties.go
iwahbe: I'd happy with `fixCtx`, where the inner field is `ignoredMapping`. We have a context, and a property on the context.

guineveresaenger (approved):
Thank you; this is a lot clearer now!

I like the rename to `Includes` and while I still I worry that it may send the wrong message, as in "wait, do I have to specify includeAll or some such", I think it follows allowlist patterns well enough to not pick at it any further.

Ideally it would be `IncludeResourcesAndDatasources` but even I think that is wayyyyy too long, especially for a CLI flag. 😝

PR #3150, opened 2025-08-07
Fix panic when --help is passed to dynamic provider parameterization

iwahbe:
Add proper error handling to detect when help text is displayed instead of actual command execution. This prevents panics and provides appropriate error responses when users request help documentation.

Changes:
- Add gRPC status error handling for argument validation
- Track whether command actually executed vs help display
- Return informative error when help is shown

Fixes https://github.com/pulumi/pulumi-terraform-provider/issues/80

### DIFF
--- a/dynamic/parameterize/args.go
+++ b/dynamic/parameterize/args.go
@@ -67,7 +67,7 @@ type LocalArgs struct {
 	IndexDocOutDir string
 }
 
-func ParseArgs(ctx context.Context, a []string) (Args, error) {
+func ParseArgs(ctx context.Context, cliArgs []string) (Args, error) {
 	var args Args
 	var fullDocs bool
 	var upstreamRepoPath string
@@ -90,10 +90,10 @@ func ParseArgs(ctx context.Context, a []string) (Args, error) {
 		},
 		Args: func(cmd *cobra.Command, args []string) error {
 			err := cobra.RangeArgs(1, 2)(cmd, args)
-			if err == nil {
-				return nil
+			if err != nil {
+				return status.Error(codes.InvalidArgument, err.Error())
 			}
-			return status.Error(codes.InvalidArgument, err.Error())
+			return nil
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


dynamic/parameterize/args.go
```diff
@@ -72,14 +74,27 @@ func ParseArgs(ctx context.Context, a []string) (Args, error) {
...
 			args, err = parseArgs(cmd.Context(), a, fullDocs, upstreamRepoPath, indexDocOutDir, includes, providerName)
 			return err
 		},
-		Args: cobra.RangeArgs(1, 2),
+		Args: func(cmd *cobra.Command, args []string) error {
+			err := cobra.RangeArgs(1, 2)(cmd, args)
+			if err == nil {
+				return nil
+			}
+			return status.Error(codes.InvalidArgument, err.Error())
```
guineveresaenger: Does this strictly need to be a grpc error?
And could we rewrite this to the more common "if err != nil" or is there some subtlety I'm missing?

dynamic/parameterize/args.go
```diff
@@ -72,14 +74,27 @@ func ParseArgs(ctx context.Context, a []string) (Args, error) {
 	var indexDocOutDir string
 	var includes []string
 	var providerName string
+
+	// If --help is included in `a`, then `RunE` will never be executed.
```
guineveresaenger, line 78: nit because I stumbled over this a bit: I see `a` on line 85. Can we rename it to "args"? 🙏

guineveresaenger (approved):
Comments are non-blocking. Thank you for fixing the panic.

dynamic/parameterize/args.go
iwahbe: It doesn't, but it is more "correct". I live in hope that the CLI will start recognizing these error codes and make the display look better.

guineveresaenger: bless 🙏

PR #3153, opened 2025-08-07
Remove info.SkipExamples

### LINKED ISSUE #3152: Remove experimental `skipExamples` feature
The `skipExamples` feature should be removed.

- It was marked as experimental upon implementation
- When invoked, it drops the entire resource description from the schema (not just the examples section) so it never worked quite as expected
- We have the option of using DocsEdit to skip examples generation by omitting the entire examples section pre-schemagen, so we're not really losing any functionality
- Only [two providers currently use the feature](https://github.com/search?q=org%3Apulumi+SkipExamples&type=code) and either use case looks unnecessary at this point
   - https://github.com/pulumi/pulumi-alicloud/issues/1044
   - https://github.com/pulumi/pulumi-aws/pull/5727 

Found in the bridge in these locations et al:
- https://github.com/pulumi/pulumi-terraform-bridge/blob/50f06f70b5fc612e6a071659490a4c7252510ded/pkg/tfbridge/info/info.go#L137-L142
- https://github.com/pulumi/pulumi-terraform-bridge/blob/50f06f70b5fc612e6a071659490a4c7252510ded/pkg/tfbridge/info/info.go#L1379-L1389

guineveresaenger:
This pull request removes the experimental feature of info.SkipExamples.

It was intended as a more granular example-skipping feature for use on individual resources in providers.

Only two providers are using this currently, and using the feature results in loss of documentation. For more context, see https://github.com/pulumi/pulumi-aws/issues/5691.

Blocked by https://github.com/pulumi/pulumi-aws/issues/5691, https://github.com/pulumi/pulumi-alicloud/issues/1044.

Fixes #3152.

### DIFF
--- a/pkg/tfbridge/info.go
+++ b/pkg/tfbridge/info.go
@@ -394,9 +394,6 @@ func ConfigBoolValue(vars resource.PropertyMap, prop resource.PropertyKey, envs 
 	return false
 }
 
-// EXPERIMENTAL: the signature may change in minor releases.
-type SkipExamplesArgs = info.SkipExamplesArgs
-
 // If specified, the hook will run just prior to executing Terraform state upgrades to transform the resource state as
 // stored in Pulumi. It can be used to perform idempotent corrections on corrupt state and to compensate for
 // Terraform-level state upgrade not working as expected. Returns the corrected resource state and version. To be used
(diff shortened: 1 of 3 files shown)


corymhall (approved):
🚀

PR #3159, opened 2025-08-18
Fix secret value logged when type checking fails

### LINKED ISSUE #3158: secret value leaked to console in error message
### What happened?

when mistakenly passing the output from get_secret_object to the secret_string field on an aws.secretsmanager.SecretVersion resource, you see the beginning of the plaintext value of the json object printed in the error message (where you see `<MANUALLY-REDACTED>`):

```
stderr: Command failed with exit code 255: pulumi preview --diff --exec-agent pulumi/actions@v6 --color auto --suppress-progress --exec-kind auto.local --event-log /tmp/automation-logs-preview-AUBT9z/eventlog.txt --stack ***/health-systems-data-management/dev --non-interactive
  warning: Unexpected type at field "secretString": 
             expected string type, got <MANUALLY-REDACTED>... of type object
  error: aws:secretsmanager/secretVersion:SecretVersion resource 'XP_PMA_DB_PARAMS' has a problem: Attribute must be a single value, not a map. Examine values at 'XP_PMA_DB_PARAMS.secretString'.
```
pulumi cli `v3.188.0`

### Example

```
import pulumi
import pulumi_aws as aws

config = pulumi.Config()

secret = aws.secretsmanager.Secret(
    "secret",
    name="secret"
)

aws.secretsmanager.SecretVersion(
    "version",
    secret_id=secret.arn,
    secret_string=config.get_secret_object("secret_key", "default_value")
)
```
### Output of `pulumi about`

pulumi/actions@v6

### Additional context

_No response_

### Contributing

corymhall:
In the typechecker we have to unwrap the secret value to do the nested
type checking, but if it fails we log the value for the user. While it
is useful to see the actual data to know what shape the data is in, we
have to prevent secrets from being logged. They will still see the
`type`, they just won't be able to see the actual value.

fixes #3158

### DIFF
--- a/pkg/tfbridge/typechecker/typechecker.go
+++ b/pkg/tfbridge/typechecker/typechecker.go
@@ -56,6 +56,7 @@ func (v *TypeChecker) validatePropertyMap(
 	propertyMap resource.PropertyMap,
 	propertyTypes map[string]pschema.PropertySpec,
 	propertyPath resource.PropertyPath,
+	isSecret bool,
 ) []Failure {
 	stableKeys := propertyMap.StableKeys()
 	failures := []Failure{}
@@ -77,7 +78,7 @@ func (v *TypeChecker) validatePropertyMap(
 			continue
 		}
 		subPropertyPath := append(propertyPath, string(objectKey))
-		failure := v.validatePropertyValue(objectValue, objType.TypeSpec, subPropertyPath)
+		failure := v.validatePropertyValue(objectValue, objType.TypeSpec, subPropertyPath, isSecret)
 		if failure != nil {
 			failures = append(failures, failure...)
 		}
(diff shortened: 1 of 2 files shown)


pkg/tfbridge/typechecker/typechecker.go
```diff
@@ -240,11 +243,17 @@ func (v *TypeChecker) validatePropertyValue(
 func newTypeFailure(
 	path resource.PropertyPath,
 	expectedType string, actualValue resource.PropertyValue,
```
blampe, line 245: If this is a `resource.PropertyValue` shouldn't it already have `IsSecret` available?

corymhall: We have to unwrap the secret to validate nested values so we lose that tracking. 

https://github.com/pulumi/pulumi-terraform-bridge/blob/6584dc5059c722a6cc662b1d324b24cc44a12f5b/pkg/tfbridge/typechecker/typechecker.go#L120-L123

blampe: Yuck! That would explain it.

blampe (approved):

Issue #1918, opened 2024-05-01
Revisit schema inflectors 


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Bridged providers currently keep rebuilding the schema for every target language. That is pulumi-tfgen-azure python --out sdk/python/ will regenerate the schema from scratch, including converting examples, when pulumi-tfgen-azure schema --out provider/cmd/pulumi-resource-azure has already done the work of generating schema.json.

The reason for this is that schema rendering and doc comment rendering are DIFFERENT and not byte-for-byte the same. The differences boil down to inflecting code references:

```
-        Please refer to VpcSubnet to manage the subnets under a Linode VPC.
+        Please refer to linode.VpcSubnet to manage the subnets under a Linode VPC.
```

```
-	// Controls whether traffic is accepted or dropped by this rule (`ACCEPT`, `DROP`). Overrides the Firewall’s inboundPolicy if this is an inbound rule, or the outboundPolicy if this is an outbound rule.
+	// Controls whether traffic is accepted or dropped by this rule (`ACCEPT`, `DROP`). Overrides the Firewall’s inbound_policy if this is an inbound rule, or the outbound_policy if this is an outbound rule.
```

Originally researched in https://github.com/pulumi/pulumi-terraform-bridge/pull/1701

@danielrbradley brings up that moving bridged providers to `pulumi package gen-sdk` is beneficial, one recently important reason is enabling a change in how versioning is done. If we were to go ahead with this naively, the schema inflector functionality would be lost.

What options should we pursue?

- drop this functionality
- generate schema.json, schema-python.json etc and feed them into pulumi package gen-sdk
- factor out the inflectors and reintroduce them in the language generators (part of `pulumi package gen-sdk`)

danielrbradley:
One possible design that could let each language implement their own inflectors would be to annotate properties which are likely to be language specific so they can be identified within the description field and updated where required.

Inserting a `<span>` tag around replacable elements would provide a backward compatible approach as spans with no classes should have no visible impact if left in the code.

Attributes would then be used within the span to indicate additional information about the type of replacement to perform e.g. `data-pulumi-naming="resource:[TOKEN]"` or `data-pulumi-naming="property:[PROP_PATH]"`

## Examples

```diff
-        Please refer to VpcSubnet to manage the subnets under a Linode VPC.
[...]

PR #3164, opened 2025-08-21
Remove line wrapper for rawDoc and adjust tests

guineveresaenger:
This pull request removes the `genRawDocComment` line wrapper function, which was used to add newlines to inline doc comments, specifically `prop.rawdoc` strings, which originate directly from the TF schema (not the upstream markdown docs). These docs, like their markdown siblings, are read into the schema and then used to document the SDK.

These added newlines seem to serve no other purpose than to render the generated SDK files with newlines in the docs. 

This change should be made because:
- Most of these doc strings aren't that long as to render the comment useless
- it is otherwise unnecessary: the Pulumi registry handles its own line wrapping, and so doo do IDEs that use Intellisense to pull down docs descriptions for fields and functions
- The schema is simplified, losing a bunch of extra newline characters
- Any docs formatting should be done during SDK gen, not schema gen

This change is a prerequisite/nice-to-have for https://github.com/pulumi/pulumi-terraform-bridge/issues/1918, because it allows us to mark up doc strings with necessary HTML for registry rendering and variable inflection.


iwahbe (approved):
🔥

PR #3184, opened 2025-09-10
Add Java gen as a bridge functionality

### LINKED ISSUE #3183: Generate Java SDK via the bridge
Currently, we use the java-gen binary in bridged providers to generate the Java SDK.
We have the ability to run java-gen via `pulumi package gen-sdk` now, which is how the other SDK languages are handled in the bridge.

The bridge should accept `java` as a valid language and run the appropriate CLI command to generate Java.

Rollout steps:
1. Implement bridge functionality
1. Release bridge and rollout to providers
1. Change Makefile via ci-mgmt to call java-gen via bridge binary and rollout to providers

Related: https://github.com/pulumi/ci-mgmt/issues/1636

Prerequisite for: https://github.com/pulumi/pulumi-terraform-bridge/issues/1918

guineveresaenger:
This pull request adds the ability to generate the Java SDK via `pulumi-tfgen-foo --java`.

It additionally introduces a breaking change to the `info.Java` struct by adding an `Overlay` field. I believe this is a breaking change we can afford to take, as the struct already advertises that its use is for overlay functionality, but this was never implemented. We can also choose to not support overlays via the bridge for Java for now, since I'm unaware of any of our bridged providers having them.

Sample changes for pulumi-random:
https://github.com/pulumi/pulumi-random/pull/1863

Sample changes in pulumi-aws (warning; very noisy, as the autogenerated-file-warning has changed):
https://github.com/pulumi/pulumi-aws/pull/5817

TL;DR:
- We see a new, empty `Pulumi.yaml` file, same as for the other SDKs
- Autogenerated warning has changed
- A nice bonus: Java inline docs no longer emit spurious PulumiCodeChooser tags, as the bridge filters that for us now.
- Functional code remains the same

---
- **Add Java as a valid bridge language for SDK generation**
- **Add overlay info for Java**

Fixes #3183

### DIFF
--- a/pkg/tfbridge/info/info.go
+++ b/pkg/tfbridge/info/info.go
@@ -914,11 +914,10 @@ type CSharp struct {
 	RespectSchemaVersion bool
 }
 
-// See https://github.com/pulumi/pulumi-java/blob/main/pkg/codegen/java/package_info.go#L35C1-L108C1 documenting
-// supported options.
+// Java contains optional overlay information for Java code-generation.
 type Java struct {
-	BasePackage string // the Base package for the Java SDK
-
+	BasePackage string   // the Base package for the Java SDK
+	Overlay     *Overlay // optional overlay information for augmented code-generation.
 	// If set to "gradle" enables a generation of a basic set of
 	// Gradle build files.
 	BuildFiles string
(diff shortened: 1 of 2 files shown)


Graham-Pedersen (approved):
Ship it!!!

blampe (commented):
In your example diff on p-random it outputs an empty sdk/java/Pulumi.yaml file. That seems odd... Is it expected?

blampe (approved):
> It additionally introduces a breaking change to the info.Java strict by adding an Overlay field.

Isn't it back-compat since we're just adding something?

PR #3188, opened 2025-09-13
Mark up docs schema, and add language specific filter

### LINKED ISSUE #1918: Revisit schema inflectors

guineveresaenger:
This pull request does several things:

1. Changes the `fixUpPropertyReference` function to mark up any such reference with a <span> tag that contains language-specific keys and values so a language-specific filter function can identify which value should be used for the language selected.
The span looks as follows in the schema:
`\u003cspan pulumi-lang-nodejs=\"`minUpper`\" pulumi-lang-dotnet=\"`MinUpper`\" pulumi-lang-go=\"`minUpper`\" pulumi-lang-python=\"`min_upper`\" pulumi-lang-yaml=\"`minUpper`\" pulumi-lang-java=\"`minUpper`\"\u003e`min_upper`\u003c/span\u003e`
2. Starts to fix a bug in `fixUpPropertyReference` where Dotnet was generating some property names in `snake_case` (TODO: proper capitalization for these property names is still missing) - see [example](https://github.com/pulumi/pulumi-random/pull/1869/files#diff-81676d671ec0f669c06325b0f7c93214146c124ccbf80cf9a624b183148b6059R51)
3. Creates a filter function that reads the provider's main schema from file and filters it for the correct language selectors on these spans and on the Pulumi Code Chooser (to avoid repeating examples translation), instead of re-generating the schema from scratch.
4. Adds new tests for `fixUpPropertyReference`; adds tests for the filter function.

In this [pulumi-random PR](https://github.com/pulumi/pulumi-random/pull/1869), you can see the changes to the SDK are minimal (the azure -> azurerm is unrelated; the Java changes are due to https://github.com/pulumi/pulumi-terraform-bridge/pull/3184 and haven't been rolled out to providers.

Additional benefits:
- We could use these spans in registry docsgen to fix up the broken default variable inflections there
- Build time for the AWS Node SDK went from 130s to 26s; other SDKs are similar; smaller providers build the SDK almost instantaneously
- Prompted us to build the Java SDK via pulumi package gen-sdk, reducing much future maintainer toil - it jsut comes for free with the Pulumi binary!
- Paves the way for decoupling SDKgen from the terraform bridge in the future.

This pull request does not address the following:

- some conversion stat cleanup should happen for the SDK gen pass - we're still seeing conversion stat output, which is nonsensical. This can be in a separate pull request IMO.

[I ran this against the AWS provider as well](https://github.com/pulumi/pulumi-aws/pull/5825). The SDK inline documentation has a few extra empty section headers with this change, because we're not running the `description` field through the docs generator twice. **I think this is the correct bargain to make** - after this change, we can fix the docs rendering in _one_ place, and have the schema be the source of truth for everything.

Fixes #1918.


pkg/tfgen/docs.go
```diff
@@ -2183,42 +2183,76 @@ func (c infoContext) fixupPropertyReference(text string) string {
...
-			case Golang, Python:
-				// Use `ec2.Instance` format
-				return open + modname + resname.String() + close
-			default:
-				// Use `aws.ec2.Instance` format
-				return open + c.pkg.String() + "." + modname + resname.String() + close
-			}
+
+			// Build our span
+			// <span pulumi-lang-nodejs="firstProperty" pulumi-lang-go="FirstProperty" ...>firstProperty</span>
```
corymhall: Looking at the linked issues it looks like there was some agreement on a different format

https://github.com/pulumi/pulumi/issues/16099#issuecomment-2637017949

```html
<pulumi ref="#/resources/aws:s3:Bucket/properties/bucketName" />
```

pkg/tfgen/docs.go
```diff
@@ -2183,42 +2183,90 @@ func (c infoContext) fixupPropertyReference(text string) string {
...
+			// Use `ec2.Instance` format
+			goAndPyFormat := open + modname + resname.String() + close
+			// Use `aws.ec2.Instance` format
+			allOtherLangs := open + c.pkg.String() + "." + modname + resname.String() + close
+
+			// We use the NodeJS default for registry docs.
+			if c.language == RegistryDocs {
+				return allOtherLangs
+			} else {
```
Graham-Pedersen: I agree with the linter! Ditch the else block here since it doesn't do anything

pkg/tfgen/docs.go
```diff
@@ -2183,42 +2183,90 @@ func (c infoContext) fixupPropertyReference(text string) string {
...
+				return buildSpan(
+					allOtherLangs,
+					allOtherLangs,
+					goFormat,
+					pyFormat,
+					allOtherLangs,
+					allOtherLangs,
+					allOtherLangs,
```
Graham-Pedersen: Ditto, also I feel like I just saw this code above. Can this be a method to avoid code repetition?

pkg/tfgen/docs.go
```diff
@@ -2183,42 +2183,90 @@ func (c infoContext) fixupPropertyReference(text string) string {
...
+				camelCaseFormat,
+				camelCaseFormat,
+				match,
+			)
 		}
 	})
 }
 
+func buildSpan(node, dotnet, golang, python, yaml, java, unmodified string) string {
+	return fmt.Sprintf(`<span pulumi-lang-nodejs="%s" pulumi-lang-dotnet="%s" pulumi-lang-go="%s" pulumi-lang-python="%s" pulumi-lang-yaml="%s" pulumi-lang-java="%s">%s</span>`,
```
Graham-Pedersen: Yeah just pull the format into a static string and then format that!

pkg/tfgen/docs.go
```diff
@@ -2183,42 +2183,90 @@ func (c infoContext) fixupPropertyReference(text string) string {
...
+		node,
+		dotnet,
+		golang,
+		python,
+		yaml,
+		java,
+		unmodified,
```
Graham-Pedersen: I don't love these magic lists. Can we name them and have them be static or functions? Might be tricky depending on the way you are trying to compose them but there could be like a "BuildLanguageSpanInPulumiStandardOrder". Just a thought

pkg/tfgen/docs.go
```diff
@@ -2183,42 +2183,90 @@ func (c infoContext) fixupPropertyReference(text string) string {
...
+	)
+
```
Graham-Pedersen: super nit!

pkg/tfgen/generate.go
```diff
@@ -1042,15 +1043,102 @@ type GenerateOptions struct {
...
-	return g.UnstableGenerateFromSchema(genSchemaResult)
+	if g.language == "schema" || g.language == "registry-docs" || g.language == "pulumi" {
+
+		genSchemaResult, err := g.generateSchemaResult(context.Background())
+		if err != nil {
+			return nil, err
+		}
+		// Now push the schema through the rest of the generator.
+		return g.UnstableGenerateFromSchema(genSchemaResult)
+	} else {
```
Graham-Pedersen: I think you can skip this else too since you are guaranteed to return out of this block.

pkg/tfgen/generate.go
```diff
@@ -1042,15 +1043,102 @@ type GenerateOptions struct {
...
+		schemaBytes, err := os.ReadFile(fmt.Sprintf("provider/cmd/pulumi-resource-%s/schema.json", g.info.Name))
+		if err != nil {
+			return nil, err
+		}
+
+		// Generate the language-specific bytes
+		languageSchemaBytes := g.FilterSchemaByLanguage(schemaBytes)
+
+		// Parse the filtered schema bytes back into PackageSpec
+		var languagePackageSpec pschema.PackageSpec
```
Graham-Pedersen: Tiny question about my go knowledge, what does this statement do? It looks like you are declaring two variables or something weird that I don't quite follow

pkg/tfgen/generate.go
```diff
@@ -1042,15 +1043,102 @@ type GenerateOptions struct {
...
+		err = json.Unmarshal(languageSchemaBytes, &languagePackageSpec)
+		if err != nil {
+			return nil, err
+		}
+		// In order to ensure stability, the docs schema, which we're using as a source for this filter function,
+		// removes the Version field in UnstableGenerateFromSchema.
+		// For our local schemas, we want to add the version back in.
+		languagePackageSpec.Version = g.version
```
Graham-Pedersen: I read this comment, and I am still confused what this version does and why this matters. Can you elaborate a little bit?

pkg/tfgen/testdata/TestFilterSchemaByLanguage/Handles_property_names_that_are_surrounded_by_back_ticks_AND_double_quotes.golden
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "random",
+  "description": "A Pulumi package to safely use randomness in Pulumi programs.",
+  "keywords": [
+    "pulumi",
+    "random"
+  ],
+  "homepage": "https://pulumi.io",
+  "license": "Apache-2.0",
```
Graham-Pedersen, line 9: I love all the testing you added! Just wanted to call out how great that is to have some more unit test coverage!

pkg/tfgen/generate.go
guineveresaenger: I'm declaring `languagePackageSpec` which is a variable of type `pschema.PackageSpec`. Just one variable!
analogous to `var hello string`.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/pulumi/pulumi-terraform-bridge/v3
 
-go 1.23.11
+go 1.24.0
```
blampe: Why the bump? I thought we follow the same policy as pu/pu and try to stay 2 releases behind latest.

pkg/tests/schema_generation_test.go
```diff
@@ -153,6 +153,43 @@ func Test_Generate(t *testing.T) {
 		root = afero.NewBasePathFs(afero.NewOsFs(), absOutDir)
 	}
 
+	// First, generate the schema file that the NodeJS generator expects
+	// Use a separate in-memory filesystem for schema generation to avoid conflicts
+	schemaRoot := afero.NewMemMapFs()
```
blampe, line 157: 🔥

pkg/tests/schema_generation_test.go
```diff
@@ -153,6 +153,43 @@ func Test_Generate(t *testing.T) {
...
+	// since the Generate() function uses os.ReadFile() which reads from the current working directory
+	schemaDir := filepath.Join("provider", "cmd", "pulumi-resource-prov")
+	require.NoError(t, os.MkdirAll(schemaDir, 0o755))
+	schemaPath := filepath.Join(schemaDir, "schema.json")
+	require.NoError(t, os.WriteFile(schemaPath, schemaBytes, 0o600))
+
+	// Clean up the provider directory after the test
+	t.Cleanup(func() {
+		os.RemoveAll("provider")
+	})
```
blampe: Is this all for a temp dir? If so maybe `t.TempDir()`?

pkg/tfgen/docs.go
```diff
@@ -2183,39 +2201,67 @@ func (c infoContext) fixupPropertyReference(text string) string {
...
-				// Use `ec2.Instance` format
-				return open + modname + resname.String() + close
-			default:
-				// Use `aws.ec2.Instance` format
-				return open + c.pkg.String() + "." + modname + resname.String() + close
+
+			// Use `ec2.Instance` format for Go and Python
+			goAndPyFormat := open + modname + resname.String() + close
+			// Use `aws.ec2.Instance` format for all other languages
+			allOtherLangs := open + c.pkg.String() + "." + modname + resname.String() + close
```
blampe, line 2208: So the text we get is already appropriately snake/pascal-cased, we don't need to handle that?

pkg/tfgen/docs_test.go
```diff
@@ -44,40 +44,48 @@ import (
...
-	Expected string
+	name   string
+	input  string
+	expect autogold.Value
 }
 
 func TestReformatText(t *testing.T) {
 	t.Parallel()
-	tests := []testcase{
+	testCases := []testcase{
```
blampe: Nit: `tests` is more idiomatic.

pkg/tfgen/docs_test.go
```diff
@@ -44,40 +44,48 @@ import (
 var accept = cmdutil.IsTruthy(os.Getenv("PULUMI_ACCEPT"))
 
 type testcase struct {
-	Input    string
-	Expected string
+	name   string
+	input  string
+	expect autogold.Value
```
blampe: I think `autogold.ExpectFile(t, autogold.Raw(text))` might address the failures you were seeing on windows.

pkg/tfgen/docs_test.go
```diff
@@ -92,11 +100,15 @@ func TestReformatText(t *testing.T) {
 		},
 	}
 
-	for _, test := range tests {
-		text, elided := reformatText(infoCtx, test.Input, nil)
-		assert.Equal(t, test.Expected, text)
-		assert.Equalf(t, text == "", elided,
-			"We should only see an empty result for non-empty inputs if we have elided text")
+	for _, tc := range testCases {
+		tc := tc
```
blampe: Not needed since Go 1.22.

pkg/tfgen/docs_test.go
```diff
@@ -2745,3 +2757,113 @@ resource "aws_ami" "example" {
...
+						"random_pet": {Tok: "random:index/randomPet:RandomPet"},
+					},
+				},
+				language: RegistryDocs,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
blampe: Nit: not needed.

pkg/tfgen/generate_test.go
```diff
@@ -915,3 +917,90 @@ func TestExtraMappingError(t *testing.T) {
...
+			inputSchema: []byte(readfile(t, "testdata/TestFilterSchemaByLanguage/schema-backticks-and-quotes.json")),
+			generator: &Generator{
+				version:  "1.2.3-test",
+				language: "nodejs",
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
```
blampe:
```suggestion
```

pkg/tfgen/docs.go
guineveresaenger: Yeah! The resource name is already Pulumified.

go.mod
corymhall: I think the policy is to only support the _last 2_ releases which is now 1.24 and 1.25. We also had to updated it already in https://github.com/pulumi/pulumi-terraform-bridge/pull/3191

https://github.com/pulumi/home/blob/master/docs/playbooks/language-support.md#go

tgummerer: Yeah, in pulumi/pulumi we keep the older version as long as we don't see a good reason to update (as long as we don't want to use any new features from the newer release). Imo providers should not be bound by `pulumi/pulumi` Go version upgrades here, and just support the last 2 releases as is convenient.

guineveresaenger: I think this was resolved via rebase onto the changes from #3191.

pkg/tests/schema_generation_test.go
guineveresaenger: done!

PR #3191, opened 2025-09-25
Update terraform-plugin-sdk dependency to `v2.0.0-20250923233607-7f1981c8674a`

EronWright:
Update `github.com/pulumi/terraform-plugin-sdk` to `v2.0.0-20250923233607-7f1981c8674a`.


go.mod
```diff
@@ -1,27 +1,27 @@
 module github.com/pulumi/pulumi-terraform-bridge/v3
 
-go 1.23.11
+go 1.24.0
```
corymhall, line 3: Had to check on this, but it looks like this should be ok to update now since we only support the latest two versions.

https://github.com/pulumi/home/blob/master/docs/playbooks/language-support.md#go

PR #3192, opened 2025-09-25
Add CLAUDE.md for Claude Code assistance

EronWright:
## Summary
• Add CLAUDE.md file to provide guidance for Claude Code when working with this repository
• Documents common commands (build, test, lint) using the project's Makefile
• Includes architecture overview of the Pulumi Terraform Bridge
• Details key environment variables and testing structure

## Test plan
- [x] File is properly formatted and contains accurate project information
- [ ] Review content for completeness and accuracy

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,71 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Commands
+
+This is a Go-based project using a Makefile for common tasks:
+
+### Building
+- `make build` - Build all Go packages
+- `go mod tidy` - Update dependencies
+
+### Testing
+- `make test` - Run full test suite (includes installing Pulumi plugins, takes ~2 hours)
+- `make test_accept` - Run tests accepting current output as golden files
+- `PULUMI_ACCEPT=1 go test ./...` - Alternative way to run acceptance tests
+- Run specific test: `go test -run TestSpecificTest ./path/to/package`
+
+### Linting and Formatting
+- `make lint` - Run linting across all Go modules
+- `make lint_fix` - Run linting with auto-fix
+- `make fmt` - Format Go code with gofmt
+
+### Plugin Management
+- `make install_plugins` - Install required Pulumi plugins for testing
+
+### Other Commands
+- `make tidy` - Run `go mod tidy` on all Go modules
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,71 @@
+# CLAUDE.md
```
corymhall, line 1: What about AGENTS.md instead? So we are not tied to claude?

EronWright: My specific goal is to enable Claude's "memory" for this repository.  There's a whole UX within Claude Code to read and write to these CLAUDE.md files over time and in various locations, see:
https://docs.claude.com/en/docs/claude-code/memory

Of course I agree with facilitating experimentation with other tools and with avoiding lock-in.  The "import" strategy (e.g. to source an AGENTS.md) would probably help here, and hopefully not break the UX too much.

CLAUDE.md
```diff
@@ -0,0 +1,71 @@
...
+
+This is a Go-based project using a Makefile for common tasks:
+
+### Building
+- `make build` - Build all Go packages
+- `go mod tidy` - Update dependencies
+
+### Testing
+- `make test` - Run full test suite (includes installing Pulumi plugins, takes ~2 hours)
+- `make test_accept` - Run tests accepting current output as golden files
```
EronWright, line 15: @guineveresaenger you called this out so I checked the Makefile.  Seems intentional:

https://github.com/pulumi/pulumi-terraform-bridge/blob/71393629a91961afb82aa674c088605df1d927e4/Makefile#L40-L44

guineveresaenger: Yeah! That part is _somewhat_ correct - it's the line below that talks about "acceptance tests" which...is not what PULUMI_ACCEPT=1 means.

guineveresaenger: We have both `.golden` files (managed via hexops/autogold) and hand-rolled golden files (managed via PULUMI_ACCEPT). It's not standardized.

CLAUDE.md
```diff
@@ -0,0 +1,71 @@
...
+- `PULUMI_ACCEPT=1 go test ./...` - Alternative way to run acceptance tests
```
guineveresaenger, line 16: What are "acceptance tests"? There's no such thing. PULUMI_ACCEPT=1 is The Other Way in which we update test files (the ones not using hexops/autogold).

PR #3193, opened 2025-09-25
Automatically infer an id for pf resources

### LINKED ISSUE #3182: Infer ids for resource that lack them
Most new plugin framework TF resources do not come with an ID field. We currently require an ID field in pulumi (in fact this is enforced by the engine), but it's not actually used for anything significant. Many resources try to make this field the import id for the resource, but that's an informal convention at best. 

Unfortunately, since TF providers are creating more and more PF resources, the lack of ID fields on these resources is breaking our automated updates frequently. Since it's entirely safe to just insert a placeholder value for the ID field, we should convert these failures to just warnings and just return a "dummy value" as the id for any resource that doesn't have one defined.

corymhall:
This applies the logic from `dynamic` to `pf` based resources. As more
and more providers are moved to creating new resources with the plugin
framework, more and more of our upgrades are failing due to the missing
id error. This requires manual work for the team to create a `ComputeID`
mapping.

Since this `id` is not actually used for anything in Pulumi, we have
decided to fallback to hardcoding a "missing ID" value. This will reduce
the number of provider upgrades that require manual intervention by the
team and decrease the delay in releasing upgrades.

closes #3182

### DIFF
--- a/docs/guides/resource-ids.md
+++ b/docs/guides/resource-ids.md
@@ -60,8 +60,14 @@ If the type of the `"id"` attribute is not coercible to a string, you must set `
 error: Resource test_res has a problem: no "id" attribute. To map this resource consider specifying ResourceInfo.ComputeID
 ```
 
-If the resource simply doesn't have an `"id"` attribute, you will need to set `ResourceInfo.ComputeID`.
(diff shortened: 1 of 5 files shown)


Graham-Pedersen (approved):

Graham-Pedersen:
I would love to check back in a month after this has been merged and released, to see how different our average upgrade provider success rate is!

PR #3197, opened 2025-10-02
Add metadata-aware case-insensitive token deduplication

### LINKED ISSUE #2904: Tokens identical up to `_` conflict
### What happened?

When the automatic token mapping is applied to two TF tokens that are identical up to `_`, we generate 2 pulumi tokens that are identical up to capitalization:

| TF Token | Pulumi Token |
|-------|--------|
| [`fortios_firewall_securitypolicy`](https://registry.terraform.io/providers/fortinetdev/fortios/latest/docs/resources/fortios_firewall_securitypolicy) | `fortios:index/firewallSecuritypolicy:FirewallSecuritypolicy` |
| [`fortios_firewall_security_policy`](https://registry.terraform.io/providers/fortinetdev/fortios/latest/docs/resources/fortios_firewall_security_policy) | `fortios:index/firewallSecurityPolicy:FirewallSecurityPolicy` |

This schema then gets (correctly) rejected by the Pulumi CLI. We should have a mode where token generation will de-duplicate the tokens. This is especially relevant for Any TF Provider, where provider authors don't have the ability to manually map tokens.

### Example

https://github.com/pulumi/pulumi-terraform-provider/issues/67?email_source=slack

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

corymhall:
**Why we’re doing this**

Terraform providers frequently export resources whose names differ only by underscores or case (`foo_bar`, `foo-bar`, `fooBar`). Our standard token strategy collapses them to the same Pulumi token (`pkg:index/fooBar:FooBar`), forcing provider authors to hand-maintain overrides and often pushing breaking changes downstream. We wanted a turnkey wrapper that (a) keeps automatically generated tokens unique, (b) leaves manual overrides untouched, and (c) stays stable across successive tfgen runs so the “winner” of the unsuffixed name never flips.

**Key ideas**

- Wrap any existing `tokens.Strategy` with `WithCaseInsensitiveDedup`. The wrapper runs the base strategy first, then checks the normalized (module, PascalName) key against a small state machine to decide whether to reuse the original token or append a suffix.
- Persistence comes for free: the deduper consults the provider’s bridge-metadata.json (specifically the auto-aliasing history) to preload which token owned the unsuffixed name last time and what suffix counter we reached. That means the original resource keeps `Foo` even if a `foo_` resource is added later and happens to sort first.
- Suffixes use the compact `FooV2` format.

**How it works (by example)**

Suppose last release shipped a single TF
resource `foo_widget` -> `pkg:index/fooWidget:FooWidget`. After running tfgen, the metadata history records `{ "foo_widget": { "current": "pkg:index/fooWidget:FooWidget" } }`.

_A week later:_

1. We create the deduper: `seedFromMetadata` reads the auto-aliasing blob, records that `foo_widget` already owns `FooWidget`, and registers the normalized key `index:foowidget` so the suffix counter starts at 2.
2. When `ComputeTokens` walks the existing TF resource, `wrapResource` sees `elem.Tok` is empty, replays the metadata token, registers it (keeping the counter at 2), and exits because it’s a manual override.
3. A new TF resource `foo_Widget` arrives; the base strategy proposes `FooWidget`. `ensureUnique` notices the normalized key already exists with counter 2, so it returns `FooWidgetV2`, registers both `index:foowidget` (now bumped to 3) and `index:foowidgetv2`, and we assign the suffixed token. No resource renamed.
4. Future collisions get `FooWidgetV3`, etc., because we never reset the counter.

If metadata is missing (dynamic provider CLI, first run, metadata not enabled yet) the deduper still works—but only within that run. The doc calls this out so authors know stability requires metadata.

**Major components**

- `WithCaseInsensitiveDedup` signature gained a `info.ProviderMetadata` parameter (nullable). `Strategy` wrappers call it with either the provider’s metadata blob or nil.
- `standardTokenDeduper` tracks seen normalized keys, prior tokens, metadata, and any initialization error. register/unregister manage the per-key counter, while priorToken allows the wrapper to prefill existing resources.
- `seedFromMetadata` reads auto-aliasing history, caches the previously published token per TF token, and registers keys so `ensureUnique` starts from the last used suffix.

**Trade-offs / limitations**

- Manual overrides remain authoritative: if a provider author sets `info.Resources["foo_widget"].Tok` before calling `ComputeTokens`, we do not rewrite it, but we still update the counter so future collisions keep incrementing.
- Dynamic providers at runtime pass metadata with Path: "", so behaviour is per-run. That is documented; if we ever want runtime persistence we’ll need a different storage mechanism.

fixes #2904

### DIFF
--- a/pkg/tfbridge/tokens/case_insensitive_dedup.go
+++ b/pkg/tfbridge/tokens/case_insensitive_dedup.go
@@ -1,0 +1,296 @@
+package tokens
+
+import (
+	"fmt"
+	"strings"
+
+	ptokens "github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/info"
+	md "github.com/pulumi/pulumi-terraform-bridge/v3/unstable/metadata"
+)
+
+// WithCaseInsensitiveDedup wraps an existing strategy and rewrites automatically generated
+// tokens whose names collide when compared case-insensitively. The first occurrence of a
+// token is preserved; subsequent collisions receive a deterministic V suffix. The
+// provided finalize function must be the same one used by the underlying strategy (e.g.
+// tokens.MakeStandard). When provider metadata is supplied, previously emitted tokens are
+// replayed from the auto-alias history so that the same resource keeps the unsuffixed name
+// across multiple tfgen runs—a necessity for stable published providers.
+func WithCaseInsensitiveDedup(base Strategy, finalize Make, metadata info.ProviderMetadata) Strategy {
+	if base.Resource == nil && base.DataSource == nil {
+		return base
+	}
+
+	resourceDeduper := newStandardTokenDeduper(finalize, metadata, deduperKindResource)
+	dataSourceDeduper := newStandardTokenDeduper(finalize, metadata, deduperKindDataSource)
+
+	wrapResource := func(tfToken string, elem *info.Resource) error {
+		if err := resourceDeduper.Err(); err != nil {
+			return err
+		}
+		if elem != nil && elem.Tok == "" {
+			if prior, ok := resourceDeduper.priorToken(tfToken); ok {
+				elem.Tok = ptokens.Type(prior)
+			}
+		}
+		var userOverride string
+		// Remember user-specified tokens (and prior tokens) so we don't rewrite them
+		if elem != nil && elem.Tok != "" {
+			userOverride = string(elem.Tok)
+			resourceDeduper.register(userOverride)
+		}
+		// If there is no Strategy then we are done
+		if base.Resource == nil {
+			return nil
+		}
+		if err := base.Resource(tfToken, elem); err != nil {
+			return err
+		}
+		// If after running the Strategy we still don't have a token, then
+		// we are done
+		if elem == nil || elem.Tok == "" {
+			return nil
+		}
+		if userOverride != "" {
+			current := string(elem.Tok)
+			// None of the current built-in strategies will override a user provided token, but
+			// we should still handle this in case users provide one that does
+			if userOverride != current {
+				resourceDeduper.unregister(userOverride)
+				resourceDeduper.register(current)
+			}
+			return nil
+		}
+		tok, err := resourceDeduper.ensureUnique(string(elem.Tok))
+		if err != nil {
+			return fmt.Errorf("resource %q: %w", tfToken, err)
+		}
+		elem.Tok = ptokens.Type(tok)
+		return nil
+	}
+
+	wrapDataSource := func(tfToken string, elem *info.DataSource) error {
+		if err := dataSourceDeduper.Err(); err != nil {
+			return err
+		}
+		if elem != nil && elem.Tok == "" {
+			if prior, ok := dataSourceDeduper.priorToken(tfToken); ok {
+				elem.Tok = ptokens.ModuleMember(prior)
+			}
+		}
+		var userOverride string
+		if elem != nil && elem.Tok != "" {
+			userOverride = string(elem.Tok)
+			dataSourceDeduper.register(userOverride)
+		}
+		if base.DataSource == nil {
+			return nil
+		}
+		if err := base.DataSource(tfToken, elem); err != nil {
+			return err
+		}
+		if elem == nil || elem.Tok == "" {
+			return nil
+		}
+		if userOverride != "" {
+			current := string(elem.Tok)
+			if userOverride != current {
+				dataSourceDeduper.unregister(userOverride)
+				dataSourceDeduper.register(current)
+			}
+			return nil
+		}
+		tok, err := dataSourceDeduper.ensureUnique(string(elem.Tok))
+		if err != nil {
+			return fmt.Errorf("datasource %q: %w", tfToken, err)
+		}
+		elem.Tok = ptokens.ModuleMember(tok)
+		return nil
+	}
+
+	return Strategy{
+		Resource:   wrapResource,
+		DataSource: wrapDataSource,
+	}
+}
+
+type standardTokenDeduper struct {
+	finalize Make
+	seen     map[string]int
+	metadata info.ProviderMetadata
+	prior    map[string]string
+	err      error
+	kind     deduperKind
+}
+
+const aliasMetadataKey = "auto-aliasing"
+
+type deduperKind int
+
+const (
+	deduperKindResource deduperKind = iota
+	deduperKindDataSource
+)
+
+type aliasHistorySnapshot struct {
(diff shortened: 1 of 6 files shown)


pkg/tfbridge/tokens/case_insensitive_dedup.go
```diff
@@ -0,0 +1,296 @@
...
+				elem.Tok = ptokens.Type(prior)
+			}
+		}
+		var userOverride string
+		// Remember user-specified tokens (and prior tokens) so we don't rewrite them
+		if elem != nil && elem.Tok != "" {
+			userOverride = string(elem.Tok)
+			resourceDeduper.register(userOverride)
+		}
+		// If there is no Strategy then we are done
```
guineveresaenger: What are scenarios where there is no Strategy? above, we already check if _both_ strategies are nil; should we not check this earlier?

pkg/tfbridge/tokens/case_insensitive_dedup.go
```diff
@@ -0,0 +1,296 @@
...
+		return "", d.err
+	}
+	parts, ok := parseStandardToken(token)
+	if !ok {
+		// Token format is not recognized; leave it unchanged.
+		return token, nil
+	}
+	key := parts.normalizedKey()
+	next, exists := d.seen[key]
+	if !exists {
```
guineveresaenger, line 243: Isn't this the same logic as in `register`?

pkg/tfbridge/tokens/case_insensitive_dedup.go
```diff
@@ -0,0 +1,296 @@
...
+			return "", err
+		}
+		nextParts, ok := parseStandardToken(newToken)
+		if !ok {
+			return "", fmt.Errorf("unexpected token format from finalize: %q", newToken)
+		}
+		candidateKey := nextParts.normalizedKey()
+		if _, exists := d.seen[candidateKey]; !exists {
+			d.seen[key] = variant + 1
+			d.seen[candidateKey] = 2
```
guineveresaenger: I'm not quite clear on why we need to do _both_ increasing the variant count for our unversioned token `foo` but _also_ registering the variant count for the `fooV2` token as the starter `2`.

pkg/tfbridge/tokens/case_insensitive_dedup.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// newStandardTokenDeduper constructs a deduper for either resources or data sources and
+// immediately seeds it from persisted metadata when available.
+func newStandardTokenDeduper(finalize Make, metadata info.ProviderMetadata, kind deduperKind) *standardTokenDeduper {
+	d := &standardTokenDeduper{
+		finalize: finalize,
+		seen:     map[string]int{},
+		metadata: metadata,
+		prior:    map[string]string{},
```
guineveresaenger:
```suggestion
		lastPublishedToken:    map[string]string{},
```

pkg/tfbridge/tokens/case_insensitive_dedup.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	return Strategy{
+		Resource:   wrapResource,
+		DataSource: wrapDataSource,
+	}
+}
+
+type standardTokenDeduper struct {
+	finalize Make
```
guineveresaenger: Is there prior art to calling this `finalize`? If so, please disregard.
```suggestion
	makeToken Make
```

pkg/tfbridge/tokens/case_insensitive_dedup.go
```diff
@@ -0,0 +1,296 @@
...
+	for tfToken, snapshot := range entries {
+		if snapshot.Current == "" {
+			continue
+		}
+		// Record the last published token so that when we encounter the TF token again on
+		// a future run we can reuse that identity rather than giving the unsuffixed name to
+		// a newly added resource that happens to sort earlier.
+		d.prior[tfToken] = snapshot.Current
+		// Seeding the counter ensures subsequent collisions pick the next available suffix
+		// instead of reusing V2 and producing churn in generated SDKs.
```
guineveresaenger, line 185: What's the scenario in which that would happen? Why would V2 in particular be reused without this? Don't we _want_ to reuse it if it was the last published token?

pkg/tfbridge/tokens/case_insensitive_dedup.go
```diff
@@ -0,0 +1,296 @@
...
+// register notes that we have already seen a token at least once and should begin handing
+// out V2 on the next collision. Safe to call multiple times on the same token.
+func (d *standardTokenDeduper) register(token string) {
+	parts, ok := parseStandardToken(token)
+	if !ok {
+		return
+	}
+	key := parts.normalizedKey()
+	if _, exists := d.seen[key]; !exists {
+		d.seen[key] = 2
```
guineveresaenger: I'm not sure I love starting the counter at 2 - it's really confusing to think that d.seen[foo] would have a variant of `2` to start out with.

pkg/tfbridge/tokens/case_insensitive_dedup.go
corymhall: It's probably a rare case, but it would be a case where we have:

`pkg_example_a` => `Example`
`pkg_example_b` => `ExampleV2`
`pkg_example_v2` => `ExampleV2` <- this is what we would get if we don't also register `V2`.

I've added a new test case `TestCaseInsensitiveDedupMonotonicSuffixes` to illustrate.

pkg/tfbridge/tokens/case_insensitive_dedup.go
corymhall: Good point. Refactored.

pkg/tfbridge/tokens/case_insensitive_dedup.go
corymhall: Same response as https://github.com/pulumi/pulumi-terraform-bridge/pull/3197#discussion_r2410739612

pkg/tfbridge/tokens/case_insensitive_dedup.go
corymhall: Yeah `finalize` is what is used elsewhere for this concept.

pkg/tfbridge/tokens/case_insensitive_dedup.go
corymhall: I've added a `const` and some comments to hopefully make this clearer.

corymhall:
> What happens if we have a token already that is called fooV2? would it dedupe into fooV2V2?

Yes it would. I added a test to illustrate. Overall I think this whole deduplication logic isn't great, but I can't think of a good alternative for dynamic providers. For normal bridged providers we would recommend to just add a manual mapping, but for dynamic providers there is no way of doing so.

PR #3206, opened 2025-10-09
Update references to info types

blampe:
Continuation of #1879.

This updates our usage of `info` types to use the types directly instead of
their aliases. This is a prerequisite to being able to mark the aliases as
deprecated.

Moving to the `info` package's shared API types allows consumers to link
against a significantly smaller surface area.


internal/logging/logging.go
```diff
@@ -254,21 +254,21 @@ var _ io.Writer = &logSinkWriter{}
...
-		return
+		return n, err
 	}
 
 	urn, raw := parseUrnFromRawString(raw)
 	severity := logLevelToSeverity(level)
 
 	err = w.sink.Log(w.ctx, severity, urn, raw)
-	return
+	return n, err
```
blampe, line 271: There are a few other changes like this to fix local lint errors. CI must be using a less strict lint config?

tgummerer: We've seen this in `pulumi/pulumi` upgrading to `golangci-lint` 2.5.0. This is because of a newer version of `gofumpt`, which requires these return values now.

Issue #2323, opened 2024-08-19
PULUMI_CONVERT_EXAMPLES_CACHE_DIR should be emptied on major version upgrades


guineveresaenger:
https://github.com/pulumi/pulumi-confluentcloud/pull/559 failed because the upgrade process was using an old examples cache locally. 
As a result, the Go SDK was generated with old version paths in the examples code, which failed the Clean Worktree check.

`PULUMI_CONVERT_EXAMPLES_CACHE_DIR` and/or surrounding logic should be sensitive to whether a major upgrade is underway and not apply the cache for such upgrades

t0yv0:
Consider adding the major version into the cache key instead if feasible.

t0yv0:
E.g. alongside BuildFileHashes https://github.com/pulumi/pulumi-terraform-bridge/blob/t0yv0%2Fautofill/pkg/tfgen/examples_cache.go#L49 add something pertaining to the version.

PR #3214, opened 2025-10-14
Invalidate example cache keys more aggressively

blampe:
The cache key we use for examples is currently derived from a hash of the HCL and the target language. This causes us to incorrectly skip generating examples when the converter is upgraded, when the provider's version changes (https://github.com/pulumi/pulumi-terraform-bridge/issues/2323), or when the provider's flags are changed (https://github.com/pulumi/ci-mgmt/issues/1025).

This changes updates our key to derive from a couple more inputs:
* The version of Pulumi and the bridge we're linked against.
* The provider's info, including language flags.

Fixes https://github.com/pulumi/ci-mgmt/issues/1025.

### DIFF
--- a/pkg/tfgen/examples_cache.go
+++ b/pkg/tfgen/examples_cache.go
@@ -25,6 +25,7 @@ import (
 	"crypto/md5" //nolint:gosec
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"maps"
 	"os"
 	"os/exec"
@@ -123,15 +124,21 @@ func (*examplesCache) checksum(bytes []byte) string {
 }
 
 // exampleKey determines the cache key to use for the given HCL. It's computed
-// from the HCL, target language, and our package versions.
+// from the HCL, target language, our Pulumi & Bridge versions, and the
+// provider's info.
 func (ec *examplesCache) exampleKey(originalHCL, language string) string {
 	var buf bytes.Buffer
-	_, _ = buf.WriteString(originalHCL)
-	_, _ = buf.WriteString(language)
+
+	fmt.Fprint(&buf, originalHCL)
+	fmt.Fprint(&buf, language)
+	fmt.Fprint(&buf, ec.ProviderInfoHash)
+
+	// Sorted for stable keys.
 	for _, key := range slices.Sorted(maps.Keys(ec.SoftwareVersions)) {
-		_, _ = buf.WriteString(key)
-		_, _ = buf.WriteString(ec.SoftwareVersions[key])
+		fmt.Fprint(&buf, key)
+		fmt.Fprint(&buf, ec.SoftwareVersions[key])
 	}
+
 	return ec.checksum(buf.Bytes())
 }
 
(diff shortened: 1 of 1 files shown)


pkg/tfgen/examples_cache.go
```diff
@@ -167,55 +179,30 @@ func (*examplesCache) inferPlugins() map[string]map[string]string {
...
-}
-
 func (ec *examplesCache) inferSoftwareVersions() map[string]string {
-	used := []string{
-		"github.com/pulumi/pulumi/pkg/v3",
-		"github.com/pulumi/pulumi-terraform-bridge/v3",
-		"github.com/pulumi/pulumi-terraform-bridge/v3/pf",
+	v := map[string]string{}
+
+	bi, ok := debug.ReadBuildInfo()
```
guineveresaenger, line 185:
```suggestion
	buildInfo, ok := debug.ReadBuildInfo()
```

This is honestly more because "bi" reminds me of the prefix from "binary" and could just be a personal preference.

pkg/tfgen/examples_cache.go
```diff
@@ -167,55 +179,30 @@ func (*examplesCache) inferPlugins() map[string]map[string]string {
...
-	}
-	return rr
-}
-
 func (ec *examplesCache) inferSoftwareVersions() map[string]string {
-	used := []string{
-		"github.com/pulumi/pulumi/pkg/v3",
-		"github.com/pulumi/pulumi-terraform-bridge/v3",
-		"github.com/pulumi/pulumi-terraform-bridge/v3/pf",
+	v := map[string]string{}
```
guineveresaenger, line 183: Because we're using it across the entire method, can we give `v` a descriptive name?
```suggestion
	versions := map[string]string{}
```

pkg/tfgen/examples_cache.go
```diff
@@ -167,55 +179,30 @@ func (*examplesCache) inferPlugins() map[string]map[string]string {
...
-		if r.Replace.Version != "" {
-			p[u] = r.Replace.Version
+
+		switch mod.Path {
+		case "github.com/pulumi/pulumi/pkg/v3",
+			"github.com/pulumi/pulumi/sdk/v3",
+			"github.com/pulumi/pulumi-terraform-bridge/v3":
+
+			v[mod.Path] = mod.Version
+		default:
```
guineveresaenger, line 201: should we return `versions` directly here?

pkg/tfgen/examples_cache.go
```diff
@@ -46,7 +49,6 @@ type examplesCache struct {
 	ProviderName        string                       `json:"providerName"`
 	PulumiVersion       string                       `json:"pulumiVersion"`
 	SoftwareVersions    map[string]string            `json:"softwareVersions"`
-	BuildFileHashes     map[string]string            `json:"buildFileHashes"`
 	Plugins             map[string]map[string]string `json:"plugins"`
```
corymhall, line 52: I noticed that the way we calculate the plugins is via [pulumi plugin ls](https://github.com/pulumi/pulumi-terraform-bridge/blob/2a8b877dbe500f33980524c68855f0e732c3d6aa/pkg/tfgen/examples_cache.go#L146-L146) which apparently does not take into account PATH manipulation.

The way we currently do it (using the `.pulumi` local folder) and the way we plan to do it (with `mise`) work by putting the plugins on the PATH. This might be something that we have to fix with `pulumi plugin ls`

Issue #3217, opened 2025-10-15
Docs: Warn about write-only fields in description for such fields


guineveresaenger:
Write-only fields are not necessarily recognizable as such in the provider docs. We should add a note in the docs that tells the user that a field is write-only.

We can also link it to any documentation we might create about write-only fields - see https://github.com/pulumi/home/issues/4350.

Implementation plan:
Write-only fields are not pulumi-schematized (as of yet). This means the current path for implementation is to add to the write-only fields' Descriptions in the bridge. 
If we were to add this functionality to Pulumi core in the future, then we could do something similar as we do for Required fields - mark them in documentation with an asterisk. This would need changes to the core pulumi schema, as well as docsgen.

PR #3218, opened 2025-10-15
Add docs note to write-only fields

guineveresaenger:
This pull request inserts a note into the write-only field's docs description to inform the user that this is a write-only field.
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/3217.

Preview of how this renders:
Screenshot 2025-10-15 at 1 09 08 PM

### DIFF
--- a/pkg/pf/tfgen/testdata/TestWriteOnlyAttributesGenerateToSchema.golden
+++ b/pkg/pf/tfgen/testdata/TestWriteOnlyAttributesGenerateToSchema.golden
@@ -15,14 +15,14 @@
             "properties": {
                 "a1": {
                     "type": "string",
-                    "description": "**NOTE:** This field is write-only and will not be read from the API.\n",
+                    "description": "**NOTE:** This field is write-only and its value will not be updated in state as part of read operations.\n",
                     "secret": true
                 }
             },
             "inputProperties": {
                 "a1": {
                     "type": "string",
-                    "description": "**NOTE:** This field is write-only and will not be read from the API.\n",
+                    "description": "**NOTE:** This field is write-only and its value will not be updated in state as part of read operations.\n",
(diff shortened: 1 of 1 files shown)


pkg/tfgen/generate.go
```diff
@@ -1981,6 +1981,19 @@ func (g *Generator) propertyVariable(parentPath paths.TypePath, key string,
 			}
 		}
 
+		if shimSchema.WriteOnly() {
```
corymhall: Should we move this up to be part of the previous `if shimSchema.WriteOnly()` block?

pkg/tfgen/generate.go
```diff
@@ -1981,6 +1981,19 @@ func (g *Generator) propertyVariable(parentPath paths.TypePath, key string,
...
+			writeOnlyNote := "**NOTE:** This field is write-only and will not be read from the API.\n"
```
corymhall:
```suggestion
			writeOnlyNote := "**NOTE:** This field is write-only and will not updated as part of read operations.\n"
```

Maybe something like this instead? I think it might be unclear to some users what "read from the API" means. My suggestion might also be too vague though so feel free to disregard or come up with something else. Other options:

- "This field is write-only and will not be updated in state as part of read operations"
- "This field is write-only and will not have it's state refreshed as part of read operations"
- "This field is write-only and will not have it's state updated during refresh"

pkg/tfgen/generate.go
guineveresaenger: d'oh - got caught in some refactoring!

pkg/tfgen/generate.go
guineveresaenger: I've reworded it to focus on the no-read part. Thank you for the ideas!

Issue #3240, opened 2025-11-14
Go SDK Truncates Arrays to First Element During Read Operations


twingate-blee:
## Summary

The Pulumi Terraform Bridge's Go SDK generation has a bug where array/list outputs are truncated to only the first element during Read operations (triggered by resource updates). This bug **only affects the Go SDK** - Python, TypeScript, and .NET SDKs all work correctly with the same provider and schema.

## Environment

- **Pulumi Terraform Bridge Version**: v3.116.0
- **Terraform Provider Framework**: terraform-plugin-framework (modern framework, not SDK v2)
- **Affected SDK**: Go only
- **Working SDKs**: Python, TypeScript, .NET all work correctly
- **Terraform Provider**: [terraform-provider-twingate](https://github.com/Twingate/terraform-provider-twingate) v3.x
- **Schema Type**: ListNestedBlock (also reproduced with SetNestedBlock)

## Impact

This affects all Go users of bridged Terraform providers that use array/list types. Users lose visibility into all but the first element of arrays, making the Go SDK unusable for resources with multiple list items.

### Example

1. Clone the reproduction repository:
   ```bash
   git clone https://github.com/Twingate/pulumi-twingate
   cd pulumi-twingate/examples/go
   ```

2. Create a resource with 2 items in each array:
   ```go
   resource, err := twingate.NewTwingateResource(ctx, "example", &twingate.TwingateResourceArgs{
       Name:            pulumi.StringPtr("Example Resource"),
       Address:         pulumi.String("example.com"),
       RemoteNetworkId: remoteNetwork.ID(),
       AccessGroups: &twingate.TwingateResourceAccessGroupArray{
           &twingate.TwingateResourceAccessGroupArgs{GroupId: group1.ID()},
           &twingate.TwingateResourceAccessGroupArgs{GroupId: group2.ID()},
       },
       AccessServices: &twingate.TwingateResourceAccessServiceArray{
           &twingate.TwingateResourceAccessServiceArgs{ServiceAccountId: svc1.ID()},
           &twingate.TwingateResourceAccessServiceArgs{ServiceAccountId: svc2.ID()},
       },
   })
   ```

3. Initial create:
   ```bash
   pulumi up --yes
   ```

4. Check state after create:
   ```bash
   pulumi stack export | jq '.deployment.resources[] |
     select(.type == "twingate:index/twingateResource:TwingateResource") |
     {inputServices: (.inputs.accessServices | length),
[...]

PR #3241, opened 2025-11-14
Handle multi-element arrays in flattenedEncoder and flattenedDecoder

twingate-blee:
## Summary

Fixes a bug where arrays/lists were truncated to only the first element in the Go SDK during Read operations. This **only affected the Go SDK** - Python, TypeScript, and .NET SDKs worked correctly.

Fixes: https://github.com/pulumi/pulumi-terraform-bridge/issues/3240

## Problem

When using Plugin Framework providers with `ListNestedBlock` or `SetNestedBlock` types, the Go SDK would:
- ✅ Show all array elements after initial resource creation
- ❌ Truncate arrays to 1 element after any update (which triggers a Read operation)
- ✅ Work correctly in Python, TypeScript, and .NET SDKs

### Example (Twingate provider)
**Before fix:**
```json
{
  "inputs": {
    "accessServices": [2 elements],
    "accessGroups": [2 elements]
  },
  "outputs": {
    "accessServices": [1 element],  // ❌ Truncated!
    "accessGroups": [1 element]     // ❌ Truncated!
  }
}
```

**After fix:**
```json
{
  "inputs": {
    "accessServices": [2 elements],
    "accessGroups": [2 elements]
  },
  "outputs": {
    "accessServices": [2 elements],  // ✅ Fixed!
    "accessGroups": [2 elements]     // ✅ Fixed!
  }
}
```

## Root Cause

The bug occurred in two places in `pkg/convert/flattened.go`:

1. **`flattenedEncoder.fromPropertyValue()`** (lines 29-58): Only handled single elements, not arrays. When encoding multi-element arrays, only the first element was sent to the Terraform provider.

2. **`flattenedDecoder.toPropertyValue()`** (lines 58-74): Returned an error when receiving >1 elements. This error was swallowed somewhere in the chain, causing only the first element to be preserved in Go SDK outputs.

These encoders/decoders are selected when `IsMaxItemsOne()` incorrectly identifies a multi-item list as single-item. This commonly occurs with Plugin Framework providers using `ListNestedBlock` or `SetNestedBlock`.

### Encoder Fix
- Checks if input is an array using `v.IsArray()`
- If so, encodes all elements and returns them as a list
- Maintains backward compatibility for legitimate single-item inputs

### Decoder Fix
- When list has >1 elements, decodes all into an array
- Returns as `resource.NewArrayProperty(values)` instead of erroring
- Maintains backward compatibility for 0 or 1 element cases

## Testing

Tested with the [Twingate Pulumi provider](https://github.com/Twingate/pulumi-twingate), which uses Plugin Framework `ListNestedBlock` types:

### Test Results

**Before fix:**
- ✅ Initial create: Arrays show all elements (2)
- ❌ After update (triggers Read): Arrays truncated to 1 element
- ✅ Only affected Go SDK; Python/TypeScript/.NET worked correctly

**After fix:**
- ✅ Create operations: All elements preserved
- ✅ Update operations: All elements preserved
- ✅ Read/refresh operations: All elements preserved
- ✅ Works for both `types.List` and `types.Set`
- ✅ Works for both `ListNestedBlock` and `SetNestedBlock`
- ✅ Independent arrays remain independent


guineveresaenger:
Awaiting feedback on [linked issue](https://github.com/pulumi/pulumi-terraform-bridge/issues/3240#issuecomment-3555248612).

PR #3243, opened 2025-11-19
[doc gen] Add support for `console` code blocks

### LINKED ISSUE #3233: Import section is broken for some resources
Apologies if this is not the right repo to report the issue in.

Some resource pages like [aws.cloudwatch.LogGroup](https://www.pulumi.com/registry/packages/aws/api-docs/cloudwatch/loggroup/#import) and [aws.alb.Listener](https://www.pulumi.com/registry/packages/aws/api-docs/alb/listener/#import) have a broken snippet rendering in the "Import" section:

Image

It works fine on other resources like [aws.alb.ListenerCertificate](https://www.pulumi.com/registry/packages/aws/api-docs/alb/listenercertificate/#import).

pose:
Fixes #3233

`tfgen/docs.go` has logic to detect the different types of `shell` code blocks when we are generating docs. A new type codeblock labeled `console` was added which meant that broke the rendering. This PR adds logic to remove those new `console` codeblocks.

### DIFF
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1230,6 +1230,7 @@ func (p *tfMarkdownParser) parseImports(body string) {
 		// There are multiple variations of codeblocks for import syntax
 		line = strings.ReplaceAll(line, "```shell", "")
 		line = strings.ReplaceAll(line, "```sh", "")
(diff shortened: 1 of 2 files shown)


Graham-Pedersen (approved):
LGTM Lets ship it!

PR #3287, opened 2025-12-23
Drop schema-tools dependency

blampe:
The bridge is stuck on a very old version of schema-tools due to https://github.com/pulumi/schema-tools/pull/36 which made the project only consumable as a CLI. This causes problems when trying to bump dependencies (e.g. https://github.com/pulumi/pulumi-azuread/pull/2155).

We only use a couple types related to stat tracking. Rather than expose this as a public package on stats-tools, it's better to simply vendor what we need.

### DIFF
--- a/pkg/tfgen/stats.go
+++ b/pkg/tfgen/stats.go
@@ -19,7 +19,6 @@ import (
 	"strings"
 
 	mapset "github.com/deckarep/golang-set/v2"
-
 	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/tfgen/generate.go
```diff
@@ -1110,7 +1109,7 @@ func (g *Generator) generateSchemaResult(ctx context.Context) (*GenerateSchemaRe
 // API.
 func (g *Generator) UnstableGenerateFromSchema(genSchemaResult *GenerateSchemaResult) (*GenerateSchemaResult, error) {
 	pulumiPackageSpec := genSchemaResult.PackageSpec
-	schemaStats = schemaTools.CountStats(pulumiPackageSpec)
+	schemaStats = countStats(pulumiPackageSpec)
```
t0yv0, line 1112: Wonder if we rely on those stats anymore in some way. For now this PR seems good as it maintains functionality just in case.

corymhall (approved):

t0yv0 (approved):

PR #3295, opened 2026-01-02
chore: update pulumi to v3.214.0

pgavlin:
Done manually to absorb breaking changes to the Go provider interface.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -221,8 +221,8 @@ require (
 	github.com/pkg/term v1.1.0 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/posener/complete v1.2.3 // indirect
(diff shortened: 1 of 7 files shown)


corymhall (approved):

pgavlin:
> Have you run a schemagen on a big provider with these changes? [we don't currently have that as a downstream test yet](https://docs.google.com/document/d/1lyz6W0c3JPcjOJDIOoXPpeLHuW-GZ2VK0E-OUGnT1KQ/edit?tab=t.0#heading=h.uiqmssdz3phn) so I'd feel more comfortable verifying at least p-gcp schemabuild locally.

Good suggestion; done. Here are the changes: https://github.com/pulumi/pulumi-gcp/compare/pgavlin/probe-tfbridge-update. Looks like a few new examples and a fix or two in example codegen.

PR #3297, opened 2026-01-07
Fix #3080: Blocks in TF Plugin Framework are always optional

### LINKED ISSUE #3080: Optional input properties marked as required for ListAttributes with a `CustomType`
### What happened?

An example schema for `aws_bedrockagent_agent_alias.routing_configuration` https://github.com/hashicorp/terraform-provider-aws/blob/9fd0a6e0c3329ee1e2c1bf7b34f0dca00f462f84/internal/service/bedrockagent/agent_alias.go#L81-L94

In Terraform this can be expressed two different ways:

1. with block syntax

```hcl
resource "aws_bedrockagent_agent_alias" "example" {
  agent_alias_name = "example"
  agent_id         = "example"
  routing_configuration {
    agent_version = "example" # <-- this is optional
  }
}
```

2. With attribute syntax
```hcl
resource "aws_bedrockagent_agent_alias" "example" {
  agent_alias_name = "example"
  agent_id         = "example"
  routing_configuration = [
     {
        agent_version = "example" # <-- this is required
        provisioned_throughput = "example" # <-- this is required
     }
  ]
}
```

When using block syntax, the properties of routing_configuration are optional and when using attribute syntax the properties are required. For this resource specifically, the input properties are [actually optional](https://registry.terraform.io/providers/hashicorp/aws/5.98.0/docs/resources/bedrockagent_agent_alias#routing_configuration-block).

When the SDKs are generated for this type, they get marked as required. https://raw.githubusercontent.com/pulumi/pulumi-aws/refs/heads/master/sdk/nodejs/types/input.ts (search for `AgentAgentAliasRoutingConfiguration`)

### Example

See above.

Here is a TS version

```ts
import * as aws from '@pulumi/aws';

new aws.bedrock.AgentAgentAlias('alias', {
  agentAliasName: 'example',
  agentId: 'someid',
  routingConfigurations: [{
    agentVersion: 'version', // <-- required
    provisionedThroughput: 'throughput' // <-- required
  }]
});
```

### Output of `pulumi about`

NA

### Additional context

re https://github.com/pulumi/pulumi-aws/issues/5544

### Contributing

joeduffy:
[Coded in partnership with Claude.]

Blocks in Terraform Plugin Framework don't have Required/Optional fields like attributes do. They are always optional by default. Validators like SizeBetween(1, 2) only constrain the size when the block IS provided, not whether the block is required.

This fixes pulumi-aws#6101 where aws.bedrock.AgentcoreGateway's interceptorConfigurations property was incorrectly marked as required.

Changes:
- Fix blockSchema.Required() to always return false for Plugin Framework
- Fix blockSchema.Optional() to always return true for clarity
- Add test cases for block optionality with and without validators

Fixes #3080

### DIFF
--- a/pkg/pf/internal/schemashim/block_schema_test.go
+++ b/pkg/pf/internal/schemashim/block_schema_test.go
@@ -17,9 +17,9 @@ package schemashim
 import (
 	"testing"
 
+	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	rschema "github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
-	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/pf/internal/pfutils"
@@ -29,6 +29,7 @@ import (
 // are still marked as optional, not required.
 // This test addresses: https://github.com/pulumi/pulumi-terraform-bridge/issues/3080
 func TestBlockOptionalityWithValidator(t *testing.T) {
+	t.Parallel()
 	// Create a ListNestedBlock with a SizeBetween(1, 2) validator
 	// This mimics aws_bedrockagentcore_gateway's interceptor_configuration block
 	block := rschema.ListNestedBlock{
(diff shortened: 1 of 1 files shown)


pgavlin:
I don't think we want to take this; the hypothesis in the commit message / description is not correct. I have a fix for this here: https://github.com/pulumi/pulumi-terraform-bridge/pull/3298

PR #3298, opened 2026-01-08
[pf] Fix Required() for pf-bridged providers

### LINKED ISSUE #3080: Optional input properties marked as required for ListAttributes with a `CustomType`

pgavlin:
PF blocks use validators to decide IsRequired. This differs from attributes, which have a first-class field for indicating whether or not the attribute is required. These changes inspect a block's validators to determine whether or not the block is required.

Fixes #3080.

### DIFF
--- a/pkg/pf/internal/schemashim/block_schema_test.go
+++ b/pkg/pf/internal/schemashim/block_schema_test.go
@@ -1,0 +1,205 @@
+// Copyright 2016-2025, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schemashim
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/objectvalidator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/setvalidator"
+	rschema "github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/pf/internal/pfutils"
+)
+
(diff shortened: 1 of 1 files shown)


pgavlin:
Note that this is a very broad change to schema generation.

Some properties will flip from required to optional--which is not breaking, and which fixes the issue reported in #3080--but other properties that are currently optional will become required. At a source/SDK level this is a breaking change, but in practice this should not break any existing programs. Though the set of properties that will flip from optional to required are will not currently cause compile errors, attempting to run programs that do not specify these properties will result in runtime failures. Moving the validation from runtime to compile time should be a better experience overall, as it will move these failures earlier in the user journey. If the breaking changes seem too risky, we could choose to take the approach from https://github.com/pulumi/pulumi-terraform-bridge/pull/3297 and treat all blocks as optional.

corymhall (approved):
> Moving the validation from runtime to compile time should be a better experience overall, as it will move these failures earlier in the user journey. If the breaking changes seem too risky, we could choose to take the approach from https://github.com/pulumi/pulumi-terraform-bridge/pull/3297 and treat all blocks as optional.

I think this is fine. I would not consider this a breaking change.

guineveresaenger:
I agree with Cory - I really like this for its correctness.

PR #3336, opened 2026-02-11
Better example cache key invalidation

blampe:
Currently the examples cache has two core calculations:
1. A key for each example, derived from { HCL, language, provider version }.
2. A key for the cache itself, derived from {`go.work`, Pulumi version, bridge version, __plugin versions__ }.

There was some earlier work in https://github.com/pulumi/pulumi-terraform-bridge/pull/3214 which stalled because the changes there dropped plugin versions, and we need to continue tracking those. I'm making a separate change to have `pulumi plugin ls` report ambient plugins (it's a good idea in general), but that doesn't need to block us here. (I should also add that detecting version information from ambient plugins is... not straightforward. And that version info is what we need for invalidation.)

In particular, we're now tracking plugins explicitly in `mise.toml` and we can treat this as a build file responsible for invalidating our cache. More generally we can also treat other dependency changes in `go.mod` as cache-invalidating; I think we were essentially already doing this if you happened to be using `go.work`.

The impact of this will be a no-op on CI: examples will continue to be cached correctly over the life of the workflow. There will be a greater impact locally since we'll be invalidating the cache much more often. My cache for AWS is ~80MB, which isn't too bad if it's duplicated a few times. If it becomes a problem there is `make clean` and if it's _really_ a problem we can look into automatically cleaning up old caches.

Making this change will address part of https://github.com/pulumi/ci-mgmt/issues/2052 because example generation will no longer depend on these plugins being visible to `pulumi plugin ls`.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/2323.

### DIFF
--- a/pkg/tfgen/examples_cache.go
+++ b/pkg/tfgen/examples_cache.go
@@ -81,8 +81,8 @@ func newExamplesCache(info *tfbridge.ProviderInfo, cacheDir string) *examplesCac
 		dir:          dir,
 		ProviderName: providerName,
 	}
-	ec.inferToolingVersions()
 	ec.computeProviderInfoHash(info)
+	ec.inferToolingVersions()
 	ec.prepareDir()
 	return ec
 }


pkg/tfgen/examples_cache.go
```diff
@@ -43,13 +43,11 @@ const (
 type examplesCache struct {
 	enabled             bool
 	dir                 string
-	ProviderName        string                       `json:"providerName"`
-	PulumiVersion       string                       `json:"pulumiVersion"`
-	SoftwareVersions    map[string]string            `json:"softwareVersions"`
-	BuildFileHashes     map[string]string            `json:"buildFileHashes"`
-	Plugins             map[string]map[string]string `json:"plugins"`
```
blampe, line 50: `SoftwareVersions` and `Plugins` are no longer tracked because they're now subsumed by `go.mod` and `mise.toml`.

pkg/tfgen/examples_cache.go
```diff
@@ -83,8 +81,8 @@ func newExamplesCache(info *tfbridge.ProviderInfo, cacheDir string) *examplesCac
 		dir:          dir,
 		ProviderName: providerName,
 	}
-	ec.computeProviderInfoHash(info)
 	ec.inferToolingVersions()
+	ec.computeProviderInfoHash(info)
```
guineveresaenger, line 84: is this noise or does the order matter?

guineveresaenger (approved):

pkg/tfgen/examples_cache.go
blampe: I had these things depending on each other but then scoped it down so this is just noise now, reverted.

PR #3363, opened 2026-03-04
Translate Terraform property names in diagnostics

### LINKED ISSUE #6: Terraform deprecation warning in Pulumi output
Terraform AzureRM provider has deprecated some of the properties. When we run `pulumi up`, the deprecation warnings are shown as-is:

```
warning: urn:pulumi:dev::azure-http-multi::azure:appservice:MultiCallbackFunctionApp
$azure:storage/container:Container::http-multi verification warning: 
"resource_group_name": [DEPRECATED] This field has been deprecated and is no longer 
used - will be removed in 2.0 of the Azure Provider
```

This might be confusing:

1. What is `resource_group_name` property? (wrong format)
2. What is 2.0 of the Azure Provider? (well, it might mean `pulumi-azure 2.0.0`, but probably not a fully valid statement)
3. There is no explicit warning in the code itself. Although, there is a comment `This field is no longer used and will be removed in 2.0.`.

pgavlin:
These changes translate Terraform names in warning and error diagnostics into the equivalent Pulumi names.

One of the key changes here is a new shared type for expressing structured diagnostics: `ValidationWarning`. This type includes an attribute path alongside summary and detail message (essentially mirroring the underlying TF types). Provider shims now produce this type instead of a plain string. The v1 SDK polyfills with an empty attribute path.

The replacement pipeline then builds on top of this foundation by applying the following rules to each diagnostic:

1. Use the existing support for path translation to convert `AttributePath` from TF to Pulumi
2. Match known properties in the summary and details and convert them with the existing name mapping
3. Remove explicit references to Terraform and specific upstream provider versions

This pipeline is applied to both warnings and errors, and should Just Work™ for all providers.

In addition to the unit tests, I manually validated this with `pulumi-azure` (v2 SDK) and `pulumi-random` (PF).

Fixes #6.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/pkg/tfbridge/format_warnings.go
+++ b/pkg/tfbridge/format_warnings.go
@@ -155,9 +155,12 @@ func CleanTerraformLanguage(msg string) string {
 
 // Matches patterns like "will be removed in X.Y.Z of the <Name> Provider" or
 // "will be removed in a future version of the <Name> Provider" or
-// "will be removed in version X.Y of the <Name> provider".
+// "will be removed in version X.Y of the <Name> provider" or
+// "will be removed in version X.Y" (no provider name).
+// Provider names may be multi-word (e.g. "Google Cloud", "Palo Alto Networks").
 var tfVersionRemovalRegex = regexp.MustCompile(
-	`will be removed in (?:version )?[v]?[\d]+\.[\d]+(?:\.[\d]+)? of the \w+ [Pp]rovider`)
+	`will be removed in (?:version )?[v]?[\d]+\.[\d]+(?:\.[\d]+)?` +
+		`(?: of the [\w][\w ]* [Pp]rovider)?`)
 
 // Matches "Terraform" when used to refer to the provider system (not as part of a compound word).
 var terraformRefRegex = regexp.MustCompile(`(^|\W)Terraform(\W|$)`)
--- a/pkg/tfbridge/format_warnings_test.go
+++ b/pkg/tfbridge/format_warnings_test.go
@@ -448,6 +448,16 @@ func TestCleanTerraformLanguage(t *testing.T) {
 			expected: "will be removed in a future version",
 		},
 		{
+			name:     "version removal with multi-word provider name",
+			input:    "will be removed in 4.0 of the Google Cloud Provider",
+			expected: "will be removed in a future version",
+		},
+		{
+			name:     "version removal without provider name",
+			input:    "will be removed in version 3.0",
+			expected: "will be removed in a future version",
+		},
+		{
 			name:     "terraform as part of compound word is not replaced",
 			input:    "Use the TerraformCloud integration",
 			expected: "Use the TerraformCloud integration",


pkg/tfbridge/format_warnings.go
```diff
@@ -0,0 +1,196 @@
...
+	msg = terraformRefRegex.ReplaceAllString(msg, "${1}the upstream provider${2}")
+
+	return msg
+}
+
+// Matches patterns like "will be removed in X.Y.Z of the <Name> Provider" or
+// "will be removed in a future version of the <Name> Provider" or
+// "will be removed in version X.Y of the <Name> provider".
+var tfVersionRemovalRegex = regexp.MustCompile(
+	`will be removed in (?:version )?[v]?[\d]+\.[\d]+(?:\.[\d]+)? of the \w+ [Pp]rovider`)
```
corymhall: Suggestion: `\w+` only matches single-word provider names; consider allowing multi-word names to improve robustness (fallback is still safe if unmatched).

pkg/pf/tfbridge/provider_diagnostics.go
```diff
@@ -46,22 +166,41 @@ func (p *provider) processDiagnostics(ctx context.Context, diagnostics []*tfprot
...
+	if sc != nil && d.Attribute != nil && len(d.Attribute.Steps()) > 0 {
+		pp, err := formatAttributePathAsPropertyPath(sc.schemaMap, sc.schemaInfos, d.Attribute)
+		if err == nil {
+			isDeprecation := d.Severity == tfprotov6.DiagnosticSeverityWarning &&
+				tfbridge.IsDeprecationMessage(d.Summary)
+			msg := tfbridge.FormatPropertyDiagnostic(pp, d.Summary, d.Detail,
+				isDeprecation, sc.schemaMap, sc.schemaInfos)
+			return fmt.Sprintf("[%s] %s", d.Severity.String(), msg)
+		}
+		glog.V(9).Infof("Failed to translate diagnostic attribute path: %v", err)
```
corymhall, line 194: Question: should this debug path use the context-based logger for consistency with the rest of this method?

pkg/pf/tfbridge/provider_diagnostics.go
```diff
@@ -46,22 +166,41 @@ func (p *provider) processDiagnostics(ctx context.Context, diagnostics []*tfprot
...
+	}
+
+	// Default formatting.
 	msg := fmt.Sprintf("[%s] %s", d.Severity.String(), d.Detail)
 	if d.Summary != "" {
 		msg += fmt.Sprintf(": %s", d.Summary)
 	}
 	if d.Attribute != nil {
 		msg += fmt.Sprintf(" at attribute %s", d.Attribute.String())
```
corymhall, line 203: Suggestion: consider guarding this append with `len(d.Attribute.Steps()) > 0` to avoid the trailing `at attribute ` output for empty paths.

pkg/tfbridge/format_warnings.go
pgavlin: adjusted this to handle multi-word provider names and missing provider names.

pkg/pf/tfbridge/provider_diagnostics.go
pgavlin: no--this is debug output that should not be user-visible

pkg/pf/tfbridge/provider_diagnostics.go
pgavlin: this is existing code; I'll leave it as-is

PR #3410, opened 2026-04-13
Fix TypeSet reordering in Update/Diff to prevent oneof field leakage

### LINKED ISSUE #3383: extractInputs uses positional matching for TypeSet arrays during Refresh, corrupting state
## Summary

`extractInputs` in `pkg/tfbridge/schema.go` matches array elements by position index during Refresh. This is correct for `TypeList` (ordered) but wrong for `TypeSet` (unordered). When the cloud provider returns set elements in a different order than the user specified, fields from one element leak into or are dropped from another.

Reported downstream: https://github.com/pulumi/pulumi-gcp/issues/3657

## Root Cause

`extractInputs` (schema.go ~L1627) unconditionally pairs old and new array elements by index:

```go
for i := range oldArray {
    oldArray[i], _ = extractInputs(oldArray[i], newArray[i], etfs, eps)
}
```

The inner `extractInputsObject` then iterates **only old keys** — keys present in `new` but absent from `old` are silently dropped; keys absent from `new` but present in `old` are deleted.

When a TypeSet is reordered by the cloud provider, this pairs the wrong elements together, causing fields unique to one element (like `valueSource`) to be dropped because the mismatched old element never had that key.

## Concrete Example

Cloud Run v2 env vars are a TypeSet. Each env var has either `value` (plain string) or `valueSource` (secret reference) — a protobuf oneof.

User defines:
```
[0] ZOO_MODE     → value: "production"
[1] APP_SECRET   → valueSource: {secretKeyRef: ...}
```

GCP returns alphabetically:
```
[0] APP_SECRET   → value: "", valueSource: {secretKeyRef: ...}
[1] ZOO_MODE     → value: "production"
```

Refresh pairs old[1] (APP_SECRET, has `valueSource`) with new[1] (ZOO_MODE, no `valueSource`). The merge deletes `valueSource` because it's not in the new element. APP_SECRET's `valueSource` is silently lost.

On the next `pulumi up`, the user re-specifies `valueSource`, TF applies `Default: ""` for the `value` field, and the GCP API rejects the request because both oneof fields are set.

## Compounding Factor

The upstream TF schema for `value` has `Default: ""`, so the GCP response always includes `value` on every env var. This ensures the `value` key survives the mismatched merge (it exists in both old and new), while `valueSource` (only in one side) gets dropped.

## Suggested Fix

In the `case oldInput.IsArray() && newState.IsArray()` branch of `extractInputs`, when `tfs.Type() == shim.TypeSet`, match elements by content (e.g. using sub-field comparison or the TF `HashSchema` function) instead of by position. The bridge already has content-based matching in `detailed_diff.go:matchPlanElementsToInputs` that could be reused.

## Scope

This affects any `TypeSet` property where the cloud provider returns elements in a different order than the user specified and elements have optional/mutually-exclusive fields. Cloud Run v2 env vars are the reported case but there are likely others.

mjvankampen:
## Summary
This PR fixes an issue where TypeSet elements are matched by position rather than content during Update and Diff operations, causing oneof field violations when cloud providers return set elements in a different order than the user's code.

## Changes
- **New function `makeTerraformConfigWithOlds`**: Created a variant of `MakeTerraformConfig` that accepts old state, enabling content-based TypeSet matching via `matchSetElements` during Update and Diff operations.

- **Updated `Diff` and `Update` operations**: Modified both operations in `provider.go` to use `makeTerraformConfigWithOlds` instead of `MakeTerraformConfig`, passing old state to enable proper TypeSet element matching.

- **Fixed old value passing logic**: Changed the condition in `makeObjectTerraformInputs` to always pass old values when available (not just when `ApplyDefaults` is true), allowing `makeTerraformInput` to use old values for matching context without affecting default application.

- **Comprehensive test coverage**: Added `TestUpdateMakeTerraformConfigTypeSetReorder` with three test cases:
  - `cloud_run_oneof_no_corruption_on_update`: Verifies no oneof violations when elements are reordered
  - `three_element_reorder_with_mixed_value_types`: Tests complex reordering with mixed value/valueSource fields
  - `typelist_still_positional_on_update`: Confirms TypeList still uses positional matching (expected behavior)

## Implementation Details
The fix leverages the existing `matchSetElements` function to match TypeSet elements by their content (e.g., the "name" field) rather than position. This prevents scenarios where a cloud provider returns elements in alphabetical order while the user's code specifies them in a different order, which could cause fields from different elements to be incorrectly merged (e.g., both "value" and "value_source" on the same element).

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/3383

https://claude.ai/code/session_01LdR1vGmSc9SnfshbU9ACpY

### DIFF
--- a/pkg/tfbridge/diff_test.go
+++ b/pkg/tfbridge/diff_test.go
@@ -94,7 +94,7 @@ func TestCustomizeDiff(t *testing.T) {
 			makeTerraformStateOptions{defaultZeroSchemaVersion: true})
 		assert.NoError(t, err)
 
-		config, _, err := MakeTerraformConfig(ctx, &Provider{tf: provider}, inputsMap, sch, info)
+		config, _, err := MakeTerraformConfig(ctx, &Provider{tf: provider}, nil, inputsMap, sch, info)
 		assert.NoError(t, err)
 
 		tfDiff, err := provider.Diff(ctx, "resource", tfState, config, shim.DiffOptions{
@@ -138,7 +138,7 @@ func TestCustomizeDiff(t *testing.T) {
 			makeTerraformStateOptions{defaultZeroSchemaVersion: true})
 		assert.NoError(t, err)
 
-		config, _, err := MakeTerraformConfig(ctx, &Provider{tf: provider}, inputsMap, sch, info)
+		config, _, err := MakeTerraformConfig(ctx, &Provider{tf: provider}, nil, inputsMap, sch, info)
 		assert.NoError(t, err)
 
 		tfDiff, err := provider.Diff(ctx, "resource", tfState, config, shim.DiffOptions{
@@ -193,7 +193,7 @@ func TestCustomizeDiff(t *testing.T) {
 					makeTerraformStateOptions{defaultZeroSchemaVersion: true})
 				assert.NoError(t, err)
 
-				config, _, err := MakeTerraformConfig(ctx, &Provider{tf: provider}, inputsMap, sch, info)
+				config, _, err := MakeTerraformConfig(ctx, &Provider{tf: provider}, nil, inputsMap, sch, info)
 				assert.NoError(t, err)
 
 				// Calling Diff with the given CustomizeDiff used to panic, no more asserts needed.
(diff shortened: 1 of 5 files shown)


guineveresaenger (approved):
Thank you for this pull request, @mjvankampen! 🎉

PR #3411, opened 2026-04-14
feat(bridge): Forward Terraform provider warnings to Pulumi CLI

BlackDark:
## Summary

Terraform providers emit diagnostic warnings (deprecation notices, validation warnings, etc.) during Check, Diff, Create, Read, and Update operations. Terraform surfaces these to the user; Pulumi currently drops them silently.

This PR makes two sets of changes so that bridged Terraform providers forward warnings through to the Pulumi engine:

### 1. PF bridge — extract warnings from TF diagnostics

`processDiagnostics` / `processDiagnosticsWithContext` now return `([]string, error)` instead of just `error`. TF diagnostics with severity `Warning` are collected as string messages and returned alongside error handling. These warnings are threaded through:

- `provider_create.go` — `CreateWithContext` collects plan + apply warnings
- `provider_update.go` — `UpdateWithContext` collects plan + apply warnings
- `provider_read.go` — `ReadWithContext` collects read warnings
- `provider_check.go` — `CheckWithContext` collects validation warnings
- `provider_diff.go` — `DiffWithContext` collects plan warnings

### 2. Dynamic bridge — stop discarding warnings at gRPC layer

`pkg/pf/internal/plugin/provider_server.go` was discarding the `[]string` warnings returned by `CheckWithContext`, `CreateWithContext`, `ReadWithContext`, `UpdateWithContext` (using `_`) and not populating `Warnings` in gRPC responses. Fixed for all 5 operations (Check, Diff, Create, Read, Update).

### 3. SDKv2 bridge — forward Check/CheckConfig warnings

`pkg/tfbridge/provider.go`:
- `Check()` collects `formatValidationWarning` messages and populates `pulumirpc.CheckResponse.Warnings`
- `validateProviderConfig` returns `([]*pulumirpc.CheckFailure, []string)` — the second value is warning messages
- `CheckConfig` forwards config validation warnings

### Before vs After

**Before (Pulumi):** No warnings shown during preview/update, even when the underlying TF provider emits them.

**After (Pulumi + engine PR):**
```
warning: [WARNING] Use "proxmox_download_file" instead. This resource / data source will be removed in v1.0.: Deprecated
warning: [WARNING] property "networkDevices[0].enabled" is deprecated: The `enabled` attribute is deprecated...
```

**Terraform (reference):**
```
│ Warning: Deprecated
│   Use "proxmox_download_file" instead. This resource / data source will be removed in v1.0.
```

### Required companion change (engine)

> This PR requires the Pulumi engine to support the `warnings` field in provider gRPC responses:
> **pulumi/pulumi#22599** — adds `repeated string warnings` to proto messages and forwards them via the diagnostic sink.

### Related

- Engine PR: https://github.com/pulumi/pulumi/pull/22599
- Issue: https://github.com/pulumi/pulumi-terraform-bridge/issues/2659

## Testing

- Unit tests added for `processDiagnostics` warning extraction (`provider_diagnostics_test.go`)
- All existing bridge tests pass (`go test ./...` on modified packages)
- End-to-end validated against a real proxmox provider project with deprecation warnings

## Risk

**Low–Medium.** Warning extraction is additive — providers that emit no warnings are unaffected. The `processDiagnostics` signature change touches many files but is mechanical (adding a return value). The dynamic bridge fix is the most impactful since it affects all dynamically bridged providers.

### DIFF
--- a/pkg/pf/internal/plugin/provider_context.go
+++ b/pkg/pf/internal/plugin/provider_context.go
@@ -45,20 +45,20 @@ type ProviderWithContext interface {
 
 	CheckWithContext(ctx context.Context, urn resource.URN, olds, news resource.PropertyMap,
 		allowUnknowns bool, randomSeed []byte, autonaming *info.ComputeDefaultAutonamingOptions,
-	) (resource.PropertyMap, []plugin.CheckFailure, error)
+	) (resource.PropertyMap, []plugin.CheckFailure, []string, error)
 
 	DiffWithContext(ctx context.Context, urn resource.URN, id resource.ID, olds resource.PropertyMap,
 		news resource.PropertyMap, allowUnknowns bool, ignoreChanges []string) (plugin.DiffResult, error)
 
 	CreateWithContext(ctx context.Context, urn resource.URN, news resource.PropertyMap, timeout float64,
-		preview bool) (resource.ID, resource.PropertyMap, resource.Status, error)
+		preview bool) (resource.ID, resource.PropertyMap, resource.Status, []string, error)
 
 	ReadWithContext(ctx context.Context, urn resource.URN, id resource.ID,
-		inputs, state resource.PropertyMap) (plugin.ReadResult, resource.Status, error)
+		inputs, state resource.PropertyMap) (plugin.ReadResult, resource.Status, []string, error)
 
 	UpdateWithContext(ctx context.Context, urn resource.URN, id resource.ID,
 		olds resource.PropertyMap, news resource.PropertyMap, timeout float64,
-		ignoreChanges []string, preview bool) (resource.PropertyMap, resource.Status, error)
+		ignoreChanges []string, preview bool) (resource.PropertyMap, resource.Status, []string, error)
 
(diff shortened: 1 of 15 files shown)


Frassle:
I commented on the engine side PR for this, but just for visibility commenting here as well.

I do not think this needs protocol changes to support. The bridge already has access to the engine logging service and can just call log.Warning directly from inside it's check/update/etc methods. There's no need to pass these back to the engine just to call log.Warning on them from that side.

BlackDark:
After digging around I found multiple things:
- had still the `0.3.0` terraform-plugin provider in my setup (not autoupdated so that was an issue) Something for renovate to tackle
- this commit should solve a lot of things I noticed: https://github.com/pulumi/pulumi-terraform-bridge/commit/895cea268c2fd97f02e52bb45f700822f398a279
- but a new release of the plugin is needed: https://github.com/pulumi/pulumi-terraform-provider. No new release since over a month. Could trigger a new release? @Frassle

Frassle:
cc @pulumi/providers for when pulumi-terraform-provider can be released?

PR #3417, opened 2026-04-20
Narrow computed ID default fixups to explicit overrides

corymhall:
## Summary
- narrow computed non-string top-level `id` fixups to cases where the provider already made an explicit manual ID decision
- preserve the existing rename-and-alias behavior for dynamic and for providers without explicit overrides
- add targeted bridge and PF coverage for the explicit-override cases and keep the old behavior covered everywhere else

## Why
After `36325624e685bdf27dad62b0ce641cbbbdd588e1`, default fixups started running during token computation. That exposed two provider-side regressions:

- providers like Cloudflare that already set `ComputeID` for computed non-string `id` still failed validation unless `Fields[\"id\"].Type` was also set to `\"string\"`
- providers like Linode that already set `Fields[\"id\"].Type = \"string\"` for computed non-string `id` could still end up on the rename path

The first version of this PR fixed that automatically for all computed scalar non-string IDs, but that also changed the dynamic bridge schema contract by removing long-standing alias properties like `computeScriptId` and `dnsZoneId`.

This revision takes the narrower approach instead:
- if a provider already set `Fields[\"id\"].Type = \"string\"` for a computed-only non-string `id`, we preserve that and do not rename the field
- if a provider already set `ComputeID` for a computed-only non-string `id`, we preserve that and add `Fields[\"id\"].Type = \"string\"` for scalar/coercible upstream types
- otherwise, computed non-string IDs keep following the old rename path

That fixes the explicit-provider cases without changing dynamic's existing schema shape.

## Behavior
- computed-only string `id`:
  - unchanged
  - no rename
- computed-only non-string `id` with manual `Fields[\"id\"].Type = \"string\"`:
  - no rename
  - no synthesized alias field
- computed-only non-string `id` with manual `ComputeID`:
  - no rename
  - auto `Fields[\"id\"].Type = \"string\"` for scalar/coercible types
- computed-only non-string `id` without explicit overrides:
  - existing rename behavior remains in place
- input-shaped `id`:
  - existing rename + `ComputeID` behavior unchanged

## Dynamic Compatibility
This intentionally preserves the current dynamic bridge schema contract. Dynamic providers continue to expose the existing alias fields produced by the rename path for computed non-string IDs.

### DIFF
--- a/pkg/pf/internal/check/check_test.go
+++ b/pkg/pf/internal/check/check_test.go
@@ -23,12 +23,14 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/provider"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 	sdkschema "github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hexops/autogold/v2"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
 	property "github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	pfbridge "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/pf/tfbridge"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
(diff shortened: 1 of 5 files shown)


pkg/pf/internal/check/check_test.go
```diff
@@ -142,6 +145,147 @@ func TestSensitiveID(t *testing.T) {
 	assert.ErrorContains(t, err, "There were 1 unresolved ID mapping errors")
 }
 
+func TestComputedMistypedIDWithoutOverridesStillFailsValidation(t *testing.T) {
```
guineveresaenger, line 148: this is probably not surprising at all but I _love_ these test names. 💟

guineveresaenger (approved):

Issue #3428, opened 2026-05-06
Avoid eager full Plugin Framework schema materialization for muxed providers


corymhall:
## Summary

Muxed providers that include Terraform Plugin Framework resources can pay a large runtime memory cost from eagerly materializing the full PF provider schema during provider startup.

The bridge should avoid forcing full PF `GetProviderSchema` work in the long-lived Pulumi provider process when normal provider operations only require a subset of resource/data source schemas.

## Background

Terraform protocol v6 providers can advertise `GetProviderSchemaOptional`, which indicates that callers may use a cached schema and do not need to call `GetProviderSchema` on every live provider instance for normal operation.

Other Terraform-compatible runtimes can separate schema discovery from the provider instance used for actual resource operations. In contrast, the bridge currently has paths where PF schema initialization happens inside the long-lived Pulumi provider process. For large providers, this can retain substantial schema-related memory for the duration of preview/update/destroy, even when the program uses only a small subset of resources.

## Observed Impact

In local experiments with a large muxed SDKv2/PF provider, we tested a prototype that avoided eager full-provider PF schema materialization and instead allowed PF schemas to be reached lazily.

The results were consistent across preview, update, and destroy paths:

- Avoiding eager PF schema priming reduced provider-process RSS by roughly `120-180 MiB` in representative runs.
- Programs using only SDKv2 resources still benefited, because they no longer paid the full PF schema startup cost.
- Programs using PF resources also benefited. Creating one or more PF-backed resources did not cause memory to return to the eager full-schema baseline.
- Increasing the number of resources added normal operation memory, but the process still avoided the large up-front full-provider schema cost.

This suggests the current cost is primarily from provider-wide PF schema materialization during startup, not from the unavoidable cost of operating individual PF resources.

## Problem

For providers that mux SDKv2 and Plugin Framework implementations, bridge startup currently needs some PF server initialization metadata, but it appears to obtain that by forcing full PF provider schema materialization.

That has a few downsides:

- Provider memory scales with the full PF surface area, not with the resources actually used by the Pulumi program.
- Programs using only SDKv2 resources may still pay PF schema cost if the provider contains PF resources.
- Programs using a small number of PF resources may pay for all PF resource/data source schemas up front.
- The cost is retained in the long-lived Pulumi provider process rather than being isolated to a short-lived schema discovery/cache boundary.

## Desired Behavior

[...]

PR #3435, opened 2026-05-07
Lazily load PF schemas at runtime

### LINKED ISSUE #3430: Avoid broad schema loading from default token fixups
## Summary

Bridge default fixups can force every resource schema to load during provider startup.

In muxed SDKv2/PF providers, this defeats lazy PF schema loading: even an SDKv2-only program can trigger PF resource schema materialization through token/default-fixup initialization.

Parent: pulumi/pulumi-terraform-bridge#3428

## Problem

During local tracing, broad PF schema loading still happened after eager PF provider-schema priming was avoided. The remaining startup path was:

```text
MustComputeTokens
applyDefaultFixups
fixPropertyConflict
walkResources
resource.Schema()
```

These fixups handle compatibility cases like reserved property names and ID handling, but they currently do so by walking all resources and inspecting schemas.

## Desired behavior

Token/resource metadata initialization should not require loading every resource schema.

Schema-dependent compatibility fixups should be made narrower, deferred, or backed by cheaper metadata where possible.

## Acceptance criteria

- Default token/fixup initialization no longer forces all PF resource schemas to load during ordinary provider startup.
- Existing `id`, missing-ID, `urn`, and `pulumi` compatibility behavior remains covered.
- Add regression coverage for muxed SDKv2/PF providers where an SDKv2-only program does not trigger broad PF schema loading through default fixups.

corymhall:
## Context

This PR addresses the memory cost tracked in #3428 for muxed providers that include Terraform Plugin Framework resources. Before this change, runtime startup materialized the full Framework provider schema in the long-lived Pulumi provider process. That meant an SDKv2-only Pulumi program could still pay the memory cost of every PF resource, data source, list resource, and related Framework schema surface in the muxed provider.

Local experiments against the large muxed provider that motivated this work showed roughly 120-180 MiB lower provider-process RSS when startup avoided eager full-provider PF schema materialization. The goal here is to preserve the generated Pulumi schema and runtime behavior for used resources, while avoiding schema work for PF entities that a program never touches.

## What changed

Runtime startup now gathers cheap PF metadata for dispatch and type-name resolution, but defers individual PF resource, data source, and list resource schemas behind cached lazy wrappers. SDKv2-only operations in muxed providers should not load PF schemas. PF operations still load the selected schema when needed, so this does not try to avoid the cost of a PF resource, data source, or list resource that is actually used.

Because runtime startup no longer calls full Framework `GetProviderSchema`, this PR also moves Framework schema implementation validation into the build-time/tfgen path for generated PF entries. That keeps invalid generated PF schemas failing before release, rather than depending on provider startup to discover them. The validation boundary is intentionally the generated Pulumi schema surface: unmapped or ignored upstream PF entries remain skipped.

List resources are included in the lazy-loading model as first-class schema users. PF List now uses the selected CRUD resource schema for result decoding and the selected list-resource schema for query encoding, without forcing unrelated PF schemas or full provider schema materialization.

The PR also adds a design note, `docs/specs/gh-3428-lazy-pf-schema-runtime-semantics.md`, and comments around the main validation and lazy-loading hooks so the build-time/runtime boundary is easier to maintain.

## Risks and compatibility

The main intentional behavior change is validation timing. Invalid generated PF schema implementations now fail during tfgen/build time instead of being discovered as a side effect of runtime full-provider schema loading. That should be better for static generated providers, but provider upgrades may now surface upstream Framework schema issues earlier.

Runtime schema-load failures can still occur when schema construction depends on runtime-only state, parameterization, dynamic-provider behavior, or code paths that differ from tfgen. The lazy schema interface cannot return errors from every accessor today, so stable schema-load failures continue to panic through the existing provider-server recovery path. Cancellation-triggered load failures are not cached, so a later uncanceled operation can retry.

The generated Pulumi schema, bridge metadata, and dispatch metadata are intended to remain unchanged except for intentional validation failures. Dynamic bridge behavior is not being generalized under the same assumptions.

## Validation

Targeted test coverage was added for deferring resource, data source, and list resource schemas in the schema-only shim; concurrent lazy schema access and cancellation behavior; startup without full `GetProviderSchema` while preserving resource type-name resolution; build-time Framework validation for provider, resource, data source, and list resource schemas; tokenless generated mappings and unmapped invalid PF resources; and muxed SDKv2-only operations, alias dispatch, selected PF resource/data-source loads, and PF List query encoding.

Commands run:

- `mise exec -- go test ./pkg/pf/internal/pfutils ./pkg/pf/internal/schemashim ./pkg/pf/internal/check ./pkg/pf/tfgen -count=1`
- `mise exec -- go test ./pkg/pf/tests -count=1`
- `golangci-lint fmt --diff`
- `golangci-lint run`

closes #3430


pgavlin:
related to https://github.com/pulumi/pulumi-aws/issues/5860

pkg/pf/internal/schemashim/schemashim.go
```diff
@@ -23,7 +23,16 @@ import (
...
+// ShimSchemaOnlyProvider gathers cheap PF metadata immediately and defers
+// resource, data source, and list resource schema loading until the selected
+// schema is used. The stored construction context is detached from cancellation
+// so a canceled startup or gather context cannot poison future lazy schema
+// loads.
 func ShimSchemaOnlyProvider(ctx context.Context, provider pfprovider.Provider) shim.Provider {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	ctx = context.WithoutCancel(ctx)
```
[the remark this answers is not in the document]
corymhall: Fixed in c6c4e40b. `ShimSchemaOnlyProvider` now uses the caller context for eager `GatherResources`, `GatherDatasources`, and `GatherListResources`, and only stores a cancellation-detached context for future lazy schema loads. Added `TestShimSchemaOnlyProviderKeepsCancellationForEagerGathering` to pin this.

corymhall:
Review follow-up:

- Addressed the Codex inline cancellation finding in c6c4e40b. Eager PF metadata gathering now receives the caller context; only the stored lazy-load context is detached from cancellation.
- I checked the automated race finding in `lazy_schema.go`; the current code writes the nil-schema error while holding `s.mu`, and `mise exec -- go test -race ./pkg/pf/internal/schemashim -run TestLazySchemaLoadsOnceUnderConcurrentAccess -count=1` passes. I am treating that one as a stale/false-positive review note unless a newer review identifies a different race.

Additional focused validation:

- `mise exec -- go test ./pkg/pf/internal/schemashim -run 'TestShimSchemaOnlyProviderKeepsCancellationForEagerGathering|TestLazySchemaDoesNotUseCanceledConstructionContext|TestLazySchemaUsesCallerContextWithoutCachingCancellationFailure|TestLazySchemaLoadsOnceUnderConcurrentAccess' -count=1`
- `mise exec -- go test -race ./pkg/pf/internal/schemashim -run TestLazySchemaLoadsOnceUnderConcurrentAccess -count=1`

corymhall:
Follow-up on the two missing inline nits from the re-review run:

- Added the `(nil, nil)` lazy schema loader regression test in `2b2fa5fb`. It asserts the synthesized `schema loader returned nil schema without error` panic and that the failure is cached consistently.
- I intentionally did not apply the `Config(ctx)` suggestion. `Config` is an eager startup/config-schema path, so it should continue to honor the caller context. Using the stored cancellation-detached `p.ctx` there would make config schema loading ignore startup cancellation, which is the same boundary we just fixed for eager metadata gathering.

Validation for the test follow-up:

- `mise exec -- go test ./pkg/pf/internal/pfutils ./pkg/pf/internal/schemashim -run 'TestLazySchemaPanicsWhenLoaderReturnsNilSchemaWithoutError|TestLazySchemaLoadsOnceUnderConcurrentAccess|TestLazySchemaDoesNotUseCanceledConstructionContext|TestLazySchemaUsesCallerContextWithoutCachingCancellationFailure|TestShimSchemaOnlyProviderKeepsCancellationForEagerGathering' -count=1`
- `golangci-lint run`

docs/specs/gh-3428-lazy-pf-schema-runtime-semantics.md
guineveresaenger, line 1: I feel conflicted about this writeup.

On the one hand, I appreciate having a longform detailed writeup of the design and execution because future bots can read it and use it, but on the other hand, this is a pretty long read for a maintainer.

Additionally, the docs folder is specifically for the Sphinx docs for bridge developers; this seems like the wrong place for this, but I don't know where else to put it either.

We also have not the best `CHANGELOG` hygiene, but we do link to the PR on autogenerated release notes, which has a pretty extensive description already as well.

pkg/pf/internal/schemashim/schemashim.go
```diff
@@ -23,7 +23,16 @@ import (
 	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 )
 
+// ShimSchemaOnlyProvider gathers cheap PF metadata immediately and defers
+// resource, data source, and list resource schema loading until the selected
+// schema is used. Eager metadata gathering uses the caller's context. The
+// stored construction context is detached from cancellation so a canceled
+// startup or gather context cannot poison future lazy schema loads.
 func ShimSchemaOnlyProvider(ctx context.Context, provider pfprovider.Provider) shim.Provider {
+	if ctx == nil {
```
guineveresaenger, line 32: Should we use the `withoutCancel` helper from https://github.com/pulumi/pulumi-terraform-bridge/pull/3435/changes#diff-b24fd0e9a2754bffbd7e44ebf3694673238af80771da62da21b9fa3e4c04d973R52? 

Or, I'm okay with having this be explicitly inline in both cases too, it's really not that much duplication.

guineveresaenger (approved):
The approach looks reasonable to me.

I looked in the history with Claude's help and it looks like the runtime loading was added as a repression mechanism for other errors in https://github.com/pulumi/pulumi-terraform-bridge/pull/745/changes#diff-09ce7b8baa28e96ee8888104a609ff16c5bd74997aa6fcd0376c98b34ff07a6cR211.

docs/specs/gh-3428-lazy-pf-schema-runtime-semantics.md
corymhall: Moved this to `docs/specs/gh-3428-lazy-pf-schema-runtime-semantics.md` so it lives with long-form specs rather than the main docs tree. I'll keep it for now, we can always purge them later if we want to.

PR #3449, opened 2026-05-19
fix: sort HCL keys in bulkConvert to ensure deterministic example ordering

pierskarsenbarg:
## Summary

Go map iteration is randomized per-iteration, causing non-deterministic filename assignment (e0, e1, ...) to HCL snippets in `bulkConvert()`. Since `pulumi-converter-terraform v1.3.0` is sensitive to processing order, this produced non-deterministic schema output across runs. Sorting HCL keys before assignment ensures stable ordering and reproducible output.

## Changes

- `pkg/tfgen/convert_cli.go`: Sort HCL keys before assigning filenames in `bulkConvert()`
- `pkg/tfgen/convert_cli_test.go`: Add `deterministic-ordering` test that runs generation 5 times and asserts identical schema output

## Test Plan

- [x] New test `TestConvertViaPulumiCLI/deterministic-ordering` passes (generates schema 5 times, asserts identical output each time)
- [x] Existing test suite `TestConvertViaPulumiCLI` passes (all other sub-tests unaffected)

### DIFF
--- a/pkg/tfgen/convert_cli.go
+++ b/pkg/tfgen/convert_cli.go
@@ -18,10 +18,12 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"maps"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/tfgen/convert_cli.go
```diff
@@ -213,8 +213,13 @@ func (cc *cliConverter) bulkConvert() error {
 		return nil
 	}
 	examples := map[string]string{}
-	n := 0
+	hclKeys := make([]string, 0, len(cc.hcls))
 	for hcl := range cc.hcls {
+		hclKeys = append(hclKeys, hcl)
```
corymhall: You might be able to simplify this with `hclKeys := slices.Sorted(maps.Keys(cc.hcls))`

corymhall (approved):

pkg/tfgen/convert_cli.go
pierskarsenbarg: I'm on my phone - do you mind adding that as a "suggestion" so I can commit that?

PR #3457, opened 2026-05-26
fix: skip auto-named defaults that trigger upstream ConflictsWith

### LINKED ISSUE #1095: Validation-aware defaults application for Plugin Framework based providers
## Hello!

## Issue details

The v3/tfbrdige supports awareness of ExactlyOneOf and ConflictsWith properties in defaults application. That is it will not apply a default value even if it is specified if that would cause validation to fail.

Unfortunately with migration to Plugin Framework there is no easy-to-parse information on the presence of ExactlyOneOf and ConflictsWith. Instead there are validators, something like this:

```
                                boolvalidator.ConflictsWith(
                                    path.MatchRelative().AtParent().AtName("example_attribute_one"),
                                    path.MatchRelative().AtParent().AtName("example_attribute_two"),
                                    path.MatchRelative().AtParent().AtName("example_attribute_three"),
                                ),
```

The bridge code is not well positioned to infer these validators, but perhaps there could still be a way forward in iteratively unapplying defaults until all the validators pass, to respect these validators.

Note that as of now the scope of default value application for plugin framework code is smaller than the v3 tfbridge code - it only applies DefaultInfo derived defaults and does not try to replicate or mimic or interact with the way TF-defined default values are applied (plan modifiers or other mechanisms).

iwahbe:
## Summary

The dynamic bridge runs `SetAutonaming` against every wrapped provider, which attaches an `AutoName` default to any property literally called `name`. During `Check`, `defaults.ApplyDefaultInfoValues` fires that default whenever the user omits `name`, populating it with a generated value before forwarding to the upstream provider's `ValidateResourceConfig`.

For providers like `aws_cloudwatch_log_group` whose `name` and `name_prefix` are `Optional + Computed + ConflictsWith: each other`, the injected `name` collides with the user-supplied `name_prefix` and SDKv2 returns two "Conflicting configuration arguments" diagnostics. The protocol-level schema doesn't surface `ConflictsWith`, so the bridge can't anticipate the conflict.

This change detects the case after the first `ValidateResourceConfig` call: if a `ConflictsWith` diagnostic targets an attribute the bridge auto-named, strip the auto-default from the inputs and re-validate once. The cleaned inputs are also returned from `Check` so the property doesn't leak into state — the upstream SDK will compute the final `name` from `name_prefix` during plan/apply, matching tofu's behavior.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/1095

## Test plan

- [x] New `TestConflictsWithNamePrefixOnly` in `dynamic/conflicting_test.go` reproduces the failure and now passes (check + create subtests).
- [x] Existing `pkg/pf/tfbridge` unit tests still pass.
- [ ] CI green on this branch.

### DIFF
--- a/pkg/pf/tfbridge/provider_check.go
+++ b/pkg/pf/tfbridge/provider_check.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/pf/internal/defaults"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/info"
+	shim "github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfshim"
 )
 
 // Check validates the given resource inputs from the user program and computes checked inputs that fill out default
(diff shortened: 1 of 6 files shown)


pgavlin (approved):

pkg/pf/tfbridge/provider_check.go
```diff
@@ -133,8 +153,76 @@ func (p *provider) validateResourceConfig(
...
+	schemaMap shim.SchemaMap,
+	schemaInfos map[string]*tfbridge.SchemaInfo,
+) map[resource.PropertyKey]bool {
+	out := map[resource.PropertyKey]bool{}
+	for tfKey, info := range schemaInfos {
+		if info == nil || info.Default == nil || !info.Default.AutoNamed {
+			continue
+		}
+		pk := resource.PropertyKey(tfbridge.TerraformToPulumiNameV2(tfKey, schemaMap, schemaInfos))
+		if _, inUser := inputs[pk]; inUser {
```
guineveresaenger, line 175: This doesn't matter for dynamically bridged providers, as they don't have preCheck callbacks, but on regular bridged providers shouldn't we use the checked inputs?

iwahbe: yes

PR #3466, opened 2026-06-02
Default nested fields to optional when wire schema lacks optionality (SchemaConfigModeAttr)

### LINKED ISSUE #204: `aws_security_group` inline `ingress`/`egress` rule: optional fields (`prefix_list_ids`, `security_groups`, `self`, ...) reported as required
## What happens

Pulumi HCL rejects an inline `ingress`/`egress` rule on `aws_security_group`
that omits the rule's optional fields, reporting them as **required**. The
canonical minimal rule — `from_port` / `to_port` / `protocol` / `cidr_blocks` —
fails because Pulumi HCL also demands `description`, `ipv6_cidr_blocks`,
`prefix_list_ids`, `security_groups`, and `self`, all of which are **optional**
in the Terraform/OpenTofu provider schema.

The failure is at resource registration (schema validation), during `preview`,
after the provider is configured — so reproducing it needs AWS credentials, but
nothing is created (it errors before any apply).

```
error: an unhandled error occurred: registering aws_security_group.this:
main.tf:25,11-11: Missing required argument; The argument "description" is
required, but no definition was found., and 4 other diagnostic(s)
```

## Repro

`main.tf`:

```hcl
terraform {
  required_version = ">= 1.0"
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = ">= 5.0"
    }
  }
}

provider "aws" {
  region = "us-west-2"
}

resource "aws_vpc" "this" {
  cidr_block = "10.0.0.0/16"
}

resource "aws_security_group" "this" {
  name   = "sg-inline-rule-repro"
  vpc_id = aws_vpc.this.id

  ingress {
    from_port   = 0
    to_port     = 0
    protocol    = "-1"
    cidr_blocks = ["10.0.0.0/16"]
  }
}

output "sg_id_known" {
  value = startswith(aws_security_group.this.id, "sg-")
}
```

`Pulumi.yaml`:

```yaml
name: sg-inline-rule-repro
runtime: hcl
description: aws_security_group inline rule optional-field repro
```

## Works in OpenTofu, fails in Pulumi HCL

OpenTofu plans the group with no complaint about the omitted optional fields:

```console
$ tofu init && tofu plan
...
Plan: 2 to add, 0 to change, 0 to destroy.
```

Pulumi HCL fails during `preview`, treating the optional inline-rule fields as
required:

```console
$ pulumi preview
 +  aws:index:Vpc this create
 +  pulumi:pulumi:Stack sg-inline-rule-repro-test create error: an unhandled
    error occurred: registering aws_security_group.this: main.tf:25,11-11:
    Missing required argument; The argument "description" is required, but no
    definition was found., and 4 other diagnostic(s)
```

The "4 other diagnostic(s)" are the remaining omitted optional fields
(`ipv6_cidr_blocks`, `prefix_list_ids`, `security_groups`, `self`).

## Expected

The inline `ingress`/`egress` rule fields `description`, `ipv6_cidr_blocks`,
`prefix_list_ids`, `security_groups`, and `self` are optional in the Terraform
AWS provider schema. Pulumi HCL should plan and apply the group, matching
OpenTofu/Terraform, instead of requiring them.

## Environment

- OpenTofu v1.12.1
- Pulumi v3.245.0-dev.0
- pulumi-language-hcl built from `62728e7`

### LINKED ISSUE #3467: Nested fields are marked as required when SDKv2 uses `SchemaConfigModeAttr`
When a Terraform provider is **dynamically bridged** (schema sourced from the provider's plugin-protocol wire schema rather than tfgen'd from SDKv2 Go source), a `ConfigMode: schema.SchemaConfigModeAttr` block — the "attribute-as-blocks" shape used by e.g. `aws_security_group`'s `ingress`/`egress` — had **every** nested field marked required in the generated Pulumi schema. Genuinely-optional fields (`description`, `ipv6_cidr_blocks`, `prefix_list_ids`, `security_groups`, `self`) were reported as required.

This surfaced downstream in [pulumi-hcl#204](https://github.com/pulumi-labs/pulumi-hcl/issues/204), where an inline rule like:

```hcl
ingress {
  from_port   = 0
  to_port     = 0
  protocol    = "-1"
  cidr_blocks = ["10.0.0.0/16"]
}
```

was rejected with `Missing required argument; The argument "description" is required` even though it plans cleanly under OpenTofu. The required-set was baked into the bridged schema itself, so the problem was in the bridge, not the consumer.

Repro against `hashicorp/aws`:

```console
$ pulumi package get-schema terraform-provider hashicorp/aws \
    | jq '.types["aws:index/SecurityGroupIngress:SecurityGroupIngress"].required'
[ "cidrBlocks", "description", "fromPort", "ipv6CidrBlocks", "prefixListIds",
  "protocol", "securityGroups", "self", "toPort" ]
```

iwahbe:
## Problem

When a Terraform provider is **dynamically bridged** (schema sourced from the provider's plugin-protocol wire schema rather than tfgen'd from SDKv2 Go source), a `ConfigMode: schema.SchemaConfigModeAttr` block has **every** nested field marked required in the generated Pulumi schema. Genuinely-optional fields aree reported as required.

This surfaced downstream in [pulumi-hcl#204](https://github.com/pulumi-labs/pulumi-hcl/issues/204), where an inline rule like:

```hcl
ingress {
  from_port   = 0
  to_port     = 0
  protocol    = "-1"
  cidr_blocks = ["10.0.0.0/16"]
}
```

was rejected with `Missing required argument; The argument "description" is required` even though it plans cleanly under OpenTofu. The required-set was baked into the bridged schema itself, so the problem was in the bridge, not the consumer.

Repro against `hashicorp/aws`:

```console
$ pulumi package get-schema terraform-provider hashicorp/aws \
    | jq '.types["aws:index/SecurityGroupIngress:SecurityGroupIngress"].required'
[ "cidrBlocks", "description", "fromPort", "ipv6CidrBlocks", "prefixListIds",
  "protocol", "securityGroups", "self", "toPort" ]
```

## Root cause

An SDKv2 block declared with `ConfigMode: SchemaConfigModeAttr` is **not** serialized as a `NestedBlock` over the Terraform plugin wire protocol. It becomes an `Attribute` whose type is a `List<Object>`. The per-field `Optional`/`Required` information that exists in the SDKv2 Go source is **not** encoded into that object type — the wire `tftypes.Object` carries no `OptionalAttributes`.

The dynamic bridge consumes only the wire schema (`pkg/pf/proto`). In `elementObjectMap`, optionality was derived solely from `tftypes.Object.OptionalAttributes`, so a field absent from that (always-empty) set was treated as **required**.

The information is genuinely absent from the wire schema, so it cannot be recovered.

<details>

<summary> Why not fix this upstream in SDKv2? </summary>

This loss is **intentional, long-standing Terraform behavior, not an SDKv2 bug.** The chain is:

1. SDKv2 lowers a `ConfigMode: SchemaConfigModeAttr` block to a typed attribute whose cty type is computed by `coreConfigSchemaType` → `set.coreConfigSchema().ImpliedType()` (`helper/schema/core_schema.go`).
2. `configschema.Block.ImpliedType()` builds a plain `cty.Object` from each nested attribute's `Type` and **discards its `Optional`/`Required` flag** — it never emits `cty.ObjectWithOptionalAttrs`:

   ```go
   for name, attrS := range b.Attributes {
       atys[name] = attrS.Type   // only the type; optionality dropped
   }
   return cty.Object(atys)        // => every field required
   ```

This is by design. Terraform/OpenTofu's own `Block.ImpliedType` doc comment states the returned type has "no objects with optional attributes, since this value is not to be used for the decoding of configuration." Optionality instead lives in the block's *decoder spec* (`hcldec.Spec`), which is used to decode HCL bodies into cty values.

But the decoder spec is **never serialized over the wire**. The plugin protocol's `Schema` message (see `terraform-plugin-go`'s `SchemaBlock`/`SchemaAttribute`) carries only the cty `Type`, the PF-style `NestedType`, and flat `Optional`/`Required`/`Computed` flags on the attribute itself — nothing that re-encodes the per-field optionality of a plain `cty.Object` attribute type. The decoder spec is recomputed locally from a `configschema.Block` on whichever side needs it. So Terraform core hits the *same* loss: when it rebuilds a `configschema.Block` from the wire schema via `ProtoToConfigSchema` and calls `DecoderSpec()`, the nested object is all-required there too.

Terraform avoids the symptom not by recovering optionality, but via its `lang/blocktoattr` fixup. When decoding block syntax for an attribute-as-blocks, it synthesizes an ephemeral schema from the attribute's cty element type in which **every nested attribute is marked optional** (`SchemaForCtyElementType`):

```go
// terraform: internal/lang/blocktoattr/schema.go
for name, aty := range atys {
    ret.Attributes[name] = &configschema.Attribute{
        Type:     aty,
        Optional: true,   // every nested field treated as optional
    }
}
```

So Terraform's own behavior for these is "treat all nested fields as optional" — which is exactly what this PR does in the bridge. Defaulting to optional is therefore the faithful behavior, not a workaround.

cty *can* express per-field optionality (`cty.ObjectWithOptionalAttrs`), and the modern path does: Plugin Framework nested objects serialize as `NestedType` with per-field `Optional` flags over the wire (work tracked in [terraform-plugin-go#48](https://github.com/hashicorp/terraform-plugin-go/issues/48)). That's why this bug is specific to SDKv2 attribute-as-blocks and never affects PF providers — the bridge's `object`/`attrMap` path reads those flags correctly. SDKv2's legacy attribute-as-blocks predates optional object attributes and was never migrated.

So there's no upstream fix to wait on, and changing `ImpliedType` upstream would break a deliberately-stable representation.

</details>

## Solution

In `pkg/pf/proto/element.go`, when an object type's `OptionalAttributes` set is empty we cannot distinguish "every field is required" from "optionality is unknown", so we default to **optional** rather than incorrectly marking every field required. When `OptionalAttributes` *is* populated we honor it exactly, preserving genuine cty optional-attribute semantics.

This means that genuinely-required fields (`fromPort`, `protocol`, `toPort`) also become optional under the dynamic bridge, since their requiredness isn't recoverable from the wire schema. Defaulting to optional is a better trade-off then defaulting to required, since it allows users to express all valid input shapes.

Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/3467
Fixes https://github.com/pulumi-labs/pulumi-hcl/issues/204

### DIFF
--- a/dynamic/testdata/TestSchemaGeneration/Backblaze/b2-0.8.9.golden
+++ b/dynamic/testdata/TestSchemaGeneration/Backblaze/b2-0.8.9.golden
@@ -238,18 +238,7 @@
                     "type": "string"
                 }
             },
-            "type": "object",
-            "required": [
-                "bucketId",
-                "bucketName",
-                "capabilities",
-                "namePrefix"
-            ],
-            "language": {
-                "nodejs": {
-                    "requiredInputs": []
-                }
-            }
+            "type": "object"
         },
         "b2:index/getBucketCorsRule:getBucketCorsRule": {
             "properties": {
@@ -284,20 +273,7 @@
                     "type": "number"
                 }
             },
-            "type": "object",
-            "required": [
-                "allowedHeaders",
-                "allowedOperations",
-                "allowedOrigins",
-                "corsRuleName",
-                "exposeHeaders",
-                "maxAgeSeconds"
-            ],
-            "language": {
-                "nodejs": {
-                    "requiredInputs": []
-                }
-            }
+            "type": "object"
         },
         "b2:index/getBucketDefaultServerSideEncryption:getBucketDefaultServerSideEncryption": {
             "properties": {
@@ -308,16 +284,7 @@
                     "type": "string"
                 }
             },
-            "type": "object",
-            "required": [
-                "algorithm",
-                "mode"
-            ],
-            "language": {
-                "nodejs": {
-                    "requiredInputs": []
-                }
-            }
+            "type": "object"
         },
         "b2:index/getBucketFileFileVersion:getBucketFileFileVersion": {
             "properties": {
@@ -361,25 +328,7 @@
                     "type": "number"
                 }
             },
-            "type": "object",
-            "required": [
-                "action",
-                "bucketId",
-                "contentMd5",
-                "contentSha1",
-                "contentType",
-                "fileId",
-                "fileInfo",
-                "fileName",
-                "serverSideEncryptions",
-                "size",
-                "uploadTimestamp"
-            ],
-            "language": {
-                "nodejs": {
-                    "requiredInputs": []
-                }
-            }
+            "type": "object"
         },
         "b2:index/getBucketFileFileVersionServerSideEncryption:getBucketFileFileVersionServerSideEncryption": {
             "properties": {
@@ -390,16 +339,7 @@
                     "type": "string"
                 }
             },
-            "type": "object",
-            "required": [
-                "algorithm",
-                "mode"
-            ],
-            "language": {
-                "nodejs": {
-                    "requiredInputs": []
-                }
-            }
+            "type": "object"
         },
         "b2:index/getBucketFileLockConfiguration:getBucketFileLockConfiguration": {
             "properties": {
@@ -413,16 +353,7 @@
                     "type": "boolean"
                 }
             },
-            "type": "object",
-            "required": [
-                "defaultRetentions",
-                "isFileLockEnabled"
-            ],
-            "language": {
-                "nodejs": {
-                    "requiredInputs": []
-                }
-            }
+            "type": "object"
         },
         "b2:index/getBucketFileLockConfigurationDefaultRetention:getBucketFileLockConfigurationDefaultRetention": {
             "properties": {
@@ -436,16 +367,7 @@
                     }
                 }
             },
-            "type": "object",
-            "required": [
-                "mode",
-                "periods"
-            ],
-            "language": {
-                "nodejs": {
-                    "requiredInputs": []
-                }
-            }
+            "type": "object"
         },
         "b2:index/getBucketFileLockConfigurationDefaultRetentionPeriod:getBucketFileLockConfigurationDefaultRetentionPeriod": {
             "properties": {
@@ -456,16 +378,7 @@
                     "type": "string"
                 }
             },
-            "type": "object",
-            "required": [
-                "duration",
-                "unit"
-            ],
-            "language": {
-                "nodejs": {
-                    "requiredInputs": []
-                }
-            }
+            "type": "object"
         },
         "b2:index/getBucketFilesFileVersion:getBucketFilesFileVersion": {
             "properties": {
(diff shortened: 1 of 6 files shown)


guineveresaenger (approved):
Sounds like a very defensible default.

PR #3471, opened 2026-06-05
Replay default fixups from runtime metadata

corymhall:
## Context

This PR is stacked on top of #3435 and addresses #3430.

#3435 makes Plugin Framework resource schemas lazy at runtime so muxed providers do not have to materialize every PF resource schema during provider startup. That exposes one remaining eager path in the bridge: `ComputeTokens` runs the bridge's default ResourceInfo fixups, and those fixups historically inspect each Terraform resource schema to decide whether to rename reserved fields like `id`, `urn`, and `pulumi`, or to synthesize a default `ComputeID`.

That schema inspection is fine during build-time schema generation, where provider schemas are expected to be available. It is not fine as a runtime startup requirement for PF-owned resources, because it defeats the lazy schema boundary from the lower PR. The important compatibility constraint is that runtime paths such as `GetMapping`, encoders/decoders, defaults, diagnostics, and schema metadata still need to see the same final `ResourceInfo.Fields` decisions that build time would have produced.

## What changed

The bridge now treats the default schema-dependent fixups as build-time decisions that can be recorded into provider metadata and replayed at runtime. During build-time/schema-generation paths, the existing eager behavior remains in place: the fixups inspect schemas, mutate ResourceInfo as before, and record the replayable facts under `default-resource-schema-fixups`.

At runtime, when the provider metadata is explicitly marked as runtime metadata and the PF muxer says a resource's fixups may be precomputed, `ComputeTokens` can replay the recorded field metadata and default `ComputeID` decision without calling the PF resource's `Schema()` method. A resource entry may be empty; that is still meaningful because it means build time inspected the resource and found no default fixup to apply. If the runtime marker, the generated fixup section, or the per-resource entry is missing, the bridge falls back to the existing schema-inspecting behavior rather than guessing.

The implementation also moves the runtime/build-time distinction away from `MetadataInfo.Path`. Real providers can and do pass embedded runtime metadata through `NewProviderMetadata`, so path-based detection would only prove the prototype path. Runtime metadata is now marked centrally when it is extracted, and `ComputeTokens` keys off that marker.

## Design decisions and compatibility

The main decision here is to keep schema-dependent fixups as a build-time concern first. Runtime startup consumes precomputed metadata; it does not broadly mutate PF ResourceInfo late after other runtime systems may already have observed it. Selected-resource schema inspection remains the fallback for compatibility when metadata is unavailable or stale.

SDKv2 behavior remains eager. SDKv2 providers are not the target of the PF lazy-schema optimization, and the new provider capability only marks resources whose default fixup decisions may be precomputed. Build-time schema generation also remains eager so generated schema output continues to be derived from the real schemas.

The metadata is rewritten from the current build-time resource set rather than merged forever. That matters because PF ownership can change and resources can disappear; stale fixup entries should not silently keep runtime skipping schema inspection for resources that build time no longer proves.

## Validation

Focused validation run locally:

- `make test RUN_TEST_CMD='./pkg/tfbridge/tokens -run TestDefault'`
- `make test RUN_TEST_CMD='./pkg/tfbridge/tokens -run TestPrecomputedDefaultFixupsAvoidResourceSchema'`
- `make test RUN_TEST_CMD='./pkg/pf/tests -run TestMuxedRuntimeComputeTokensDoesNotLoadPFResourceSchemas'`

I also ran `gofmt` and `git diff --check` after the final documentation pass.

### DIFF
--- a/docs/specs/gh-3430-default-token-fixup-laziness.md
+++ b/docs/specs/gh-3430-default-token-fixup-laziness.md
@@ -1,0 +1,237 @@
+# GH-3430 Default Token Fixup Laziness
+
+## Summary
+
+Issue [#3430](https://github.com/pulumi/pulumi-terraform-bridge/issues/3430)
+tracks a remaining broad-schema-load path after lazy Terraform Plugin Framework
+schema loading. In muxed SDKv2/PF providers, provider startup can still load
+every PF resource schema while computing tokens and default compatibility
+fixups, even when the Pulumi program uses only SDKv2 resources.
+
+The desired behavior is:
+
+1. Token/resource metadata initialization may enumerate resource names.
+2. Startup must not call `Schema()` for every PF resource.
+3. Existing default compatibility behavior for `id`, missing IDs, `urn`, and
+   `pulumi` remains available for resources that need it.
+4. A selected PF resource may load its own schema as part of the normal PF RPC
+   path; unused PF resources must remain unloaded.
+
+This is a runtime memory behavior change. It should not change generated
+Pulumi schema output.
+
+## Current Flow
+
+Downstream providers commonly call `ProviderInfo.MustComputeTokens` during
+provider initialization. That enters `pkg/tfbridge/tokens.ComputeTokens`, which
+calls `applyDefaultFixups` before default resource and data source token
+computation.
+
+The broad load is in the default fixup pass:
+
+1. `ComputeTokens` calls `applyDefaultFixups` unless `SkipDefaultFixups` is set.
+2. `applyDefaultFixups` calls `fixPropertyConflict` and `fixMissingIDs`.
+3. Both fixups call `walkResources`.
+4. `walkResources` enumerates `p.P.ResourcesMap()`.
+5. `fixPropertyConflict` calls `r.TF.Schema().Range(...)`.
+6. `fixMissingIDs` calls `r.TF.Schema().GetOk("id")`.
+
+On a PF schema-only resource, `schemaOnlyResource.Schema()` delegates to the
+lazy PF schema adapter. Calling it forces the selected resource schema to load.
+Because the current fixup pass walks every resource, a muxed provider pays for
+all PF resource schemas during startup.
+
+Relevant code:
+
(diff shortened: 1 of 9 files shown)


corymhall:

- #3471 ◀

pkg/tfbridge/tokens/fixups.go
```diff
@@ -26,79 +26,401 @@ import (
...
+	}
+	return metadata.Set(ctx.metadata, metadatakeys.DefaultResourceSchemaFixups, ctx.defaultFixups)
+}
+
+// captureResourceFixup extracts the narrow set of eager default-fixup decisions
+// that runtime needs before any schema load: synthesized field metadata for
+// reserved Terraform property names, plus the default ComputeID decision.
+func captureResourceFixup(r resourceInfo) defaultResourceSchemaFixup {
+	var fixup defaultResourceSchemaFixup
+	for _, key := range []string{"id", "urn", "pulumi"} {
```
guineveresaenger: should we programmatically ensure that this uses the same strings as `badPropertyName`? I'm assuming the list is the same.

pkg/tfbridge/tokens/fixups.go
```diff
@@ -26,79 +26,401 @@ import (
...
+			if res.ComputeID == nil {
+				res.ComputeID = missingIDComputeID()
+			}
+		case defaultComputeIDDelegate:
+			if fixup.ComputeID.Field == "" {
+				return false, fmt.Errorf("precomputed default fixup for %s has delegate ComputeID without a field", tfToken)
+			}
+			if res.ComputeID == nil {
+				res.ComputeID = delegateIDField(resource.PropertyKey(fixup.ComputeID.Field), p.Name,
+					"https://github.com/pulumi/pulumi-terraform-provider")
```
guineveresaenger: should that URL be a const?

docs/specs/gh-3430-default-token-fixup-laziness.md
guineveresaenger, line 1: should this live in `docs/specs` maybe?

guineveresaenger (approved):
Couple nits but otherwise LGTM

PR #3472, opened 2026-06-05
Migrate tests to providertest v0.7.0 per-operation provider model

iwahbe:
## Why

`providertest` v0.7.0 changed how `PULUMI_DEBUG_PROVIDERS` is wired: instead of persisting it on the workspace at stack creation (v0.6.0), it scopes a freshly-started provider to each operation via pulumitest's own wrappers (`pt.Up`/`pt.Preview`/`pt.Refresh`/`Destroy`). The new behavior is correct — tests should not assume a provider stays attached across operations — but it breaks tests that drive the engine through the raw automation API (`pt.CurrentStack().Up()` etc.), which then run with no attached provider and fail with `no resource plugin 'pulumi-resource-prov' found`.

#3447 pinned `providertest` to v0.6.0 to unblock the dependency bump. This PR does the actual migration so we can move back to v0.7.0.

## What

- Un-pin `providertest`: v0.6.0 → **v0.7.0**.
- Replace raw `pt.CurrentStack().Up/Preview/Refresh(pt.Context(), ...)` calls with the `pt.UpErr/PreviewErr/RefreshErr(t, ...)` wrappers. These attach a fresh provider per operation (via `withProviders`) and return the error, so the error-expecting tests still work.
- Rework the cross-test upgrade harness (`upgrade_state_check_test.go`) to register the post-upgrade provider as the default and run the initial create against the original provider via `pt.Run(..., optrun.WithOpts(opttest.AttachProvider(...)))`, matching the existing `diff_check.go` pattern. This removes the manual `StartPulumiProvider` + `SetEnvVar("PULUMI_DEBUG_PROVIDERS", ...)` swap that v0.7.0 clobbers.

## Testing

Locally under v0.7.0: `TestResourceInitFailure`, `TestCollectionsNullEmptyRefreshClean`, and all `TestUpgrade_*` pass; `go vet` clean on the affected packages.

Draft pending full CI.


VenelinMartinov (approved):
makes sense

PR #3478, opened 2026-06-11
Round-trip non-UTF-8 strings through the PF bridge

iwahbe:
Observed via pulumi-terraform-provider: https://github.com/pulumi-labs/pulumi-hcl/issues/268

## Problem

Terraform's wire protocol is msgpack-encoded cty, whose strings may carry arbitrary bytes, but Pulumi's property protocol uses proto3 string fields that require valid UTF-8. When a Plugin Framework or dynamically bridged provider returns a string containing invalid UTF-8 in its post-create state, the bridge converts it into Pulumi properties unchanged and the gRPC `CreateResponse` fails to marshal:

```
error: grpc: error while marshaling: string field contains invalid UTF-8
```

The resource has already been created at that point, so every failed update leaks an unmanaged resource that is never recorded in the Pulumi state. Hit in the wild by `aws_instance.user_data` fed from `data.cloudinit_config` with `gzip = true, base64_encode = true` (e.g. the RaJiska/fck-nat module): the AWS provider base64-decodes `user_data` on read-back, leaving raw gzip bytes as a "string" in state.

## Fix

In the PF value conversion layer (`pkg/convert`), strings that are not valid UTF-8 — or that collide with the marker — cross into Pulumi as base64 with a marker prefix, and are decoded back to the original bytes on the way into Terraform. Applied in both the typed string converters and the `DynamicPseudoType` converters, covering all state-carrying directions for the PF bridge and the dynamic provider.

Properties of the encoding:

- The provider always observes the exact bytes it produced; encode∘decode is the identity for every string (valid strings that happen to start with the marker are escaped too, so decoding is unambiguous).
- Deterministic, so repeated `pulumi up` sees no diff.
- The Pulumi-side value stays a plain string, so no schema or SDK changes are needed.

The SDKv2 shim path is unaffected: go-cty normalizes invalid bytes to U+FFFD before they reach the bridge, matching Terraform core's own behavior, so it neither fails nor can carry the raw bytes.

## Tests

- `TestNonUTF8StringRoundTrip` (`pkg/convert`): rapid property test over arbitrary byte strings (and marker-prefixed variants) through both converter pairs, asserting the Pulumi side is always valid UTF-8 and the round trip is byte-identical.
- `TestNonUTF8StringRepresentation`: pins the exact escaped forms, since they are persisted in state files.
- `TestNonUTF8StringInvalidBase64`: marker-prefixed strings that are not valid base64 are rejected.
- `TestCreateNonUTF8StringInState` (`pkg/pf/tests`): end-to-end repro mirroring the `aws_instance.user_data` scenario over a real gRPC connection. Without the fix it fails with the marshal error above; with it, the create succeeds, the resource is recorded, and a second `pulumi up` reports no changes.

Full `pkg/convert` and `pkg/pf/tests` suites pass.

### DIFF
--- a/pkg/convert/dynamic.go
+++ b/pkg/convert/dynamic.go
@@ -41,7 +41,11 @@ func (enc *dynamicEncoder) fromPropertyValue(p resource.PropertyValue) (tftypes.
 	case p.IsNumber():
 		return tftypes.NewValue(tftypes.Number, p.NumberValue()), nil
 	case p.IsString():
-		return tftypes.NewValue(tftypes.String, p.StringValue()), nil
+		s, err := unescapeNonUTF8String(p.StringValue())
+		if err != nil {
+			return tftypes.Value{}, err
+		}
+		return tftypes.NewValue(tftypes.String, s), nil
 	case p.IsArray():
 		var result []tftypes.Value
 		for _, e := range p.ArrayValue() {
@@ -130,7 +134,7 @@ func (dec *dynamicDecoder) toPropertyValue(v tftypes.Value) (resource.PropertyVa
 		if err != nil {
 			return resource.PropertyValue{}, err
 		}
-		return resource.NewStringProperty(s), nil
+		return resource.NewStringProperty(escapeNonUTF8String(s)), nil
 	case v.Type().Is(tftypes.List{}):
 		var elements []tftypes.Value
 		err := v.As(&elements)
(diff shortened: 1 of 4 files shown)


iwahbe:
We are going to do better here: https://app.notion.com/p/Pulumi-String-Arbitrary-bytes-382fdbdf1cce809c899df6caae54f662?source=copy_link

PR #3496, opened 2026-06-24
Target upstream terraform-plugin-sdk/v2 instead of the Pulumi fork

iwahbe:
This PR removes the fork the bridge carried of `terraform-plugin-sdk/v2` without a change of behavior. This allows downstream providers to avoid `replace` statements in their `go.mod` and participate fully in Go's normal provider logic. It also means we don't need to maintain a fork. This is all for the good.

The downside is that we are using linknames instead of the fork, tying us to `terraform-plugin-sdk/v2` internals.

## What the fork provided

The fork carried three patches over its upstream base (v2.40.1):

1. **`PlanResourceChangeExtra`**: a variant of `PlanResourceChange` that surfaces the computed `*terraform.InstanceDiff` (upstream discards it) and adds a `TransformInstanceDiff` hook to mutate the diff before it becomes planned state. The bridge needs the `InstanceDiff` to drive its own diff, and uses the hook to apply `ignoreChanges`.
2. **`exports.go`**: re-exported `HCL2ValueFromConfigValue` (and `NormalizeObjectFromLegacySDK`, unused by the bridge) from the SDK's `internal/` packages.
3. **`RunProviderInternalValidation`**: gated `Provider.Validate`'s call to the slow `InternalValidate`, defaulting to skip it at runtime.

## How this PR replaces them

- **`plan_resource_change.go`** reimplements the planning half of `PlanResourceChange`. It obtains the `InstanceDiff` through the public `Resource.SimpleDiff`, applies `ignoreChanges` to that diff, then turns it into planned state with the same SDK helpers upstream uses. A null proposed state plans to the null state, as upstream does. Branches the bridge never exercises (provider deferral, resource identity, `RequiresReplace`) are omitted.
- **`linkname.go`** recovers the SDK's unexported and internal-only state-shaping helpers (`HCL2ValueFromConfigValue`, `HCL2ValueFromFlatmap`, `ValuesSDKEquivalent`, `normalizeNullValues`, `copyTimeoutValues`, `setWriteOnlyNullValues`, `validateConfigNulls`) via `//go:linkname`. `setWriteOnlyNullValues` takes an internal `*configschema.Block`, passed as `unsafe.Pointer`.
- **`Provider.Validate`** calls `schema.InternalMap(p.tf.Schema).Validate(...)` directly (`InternalMap` is a type alias for the unexported `schemaMap`), which runs the same validation without `InternalValidate`. The toggle and its test are removed.
- The `pin_upstream_sdk` Makefile target is removed.

Dropping the fork also prunes the module graph, removing about 4,600 stale transitive `go.sum` entries (mostly `cloud.google.com/go/*`).

**Linkname guard.** `linkname_test.go` exercises every linked symbol, so an SDK upgrade that renames one fails as a test rather than at runtime.


iwahbe:

- #3615
    - #3616
        - #3617
            - #3618
                - #3619
                    - #3620
                        - #3621
                            - #3622
                                - #3496 ◀

corymhall (approved):
🎉

PR #3500, opened 2026-06-28
Fix detailed diff matching for reordered nested sets

### LINKED ISSUE #3495: TypeSet detailed diff falls back to whole-set UPDATE on deeply nested set blocks (WAFv2 WebACL rules)
## Summary

After `pulumi import` of WAFv2 WebACL resources, `pulumi preview` shows phantom diffs on the `rules` property — the entire set is shown as `UPDATE` rather than element-level diffs, even when no changes exist. This is caused by the bridge's detailed diff algorithm failing to match nested TypeSet elements.

Discovered during a Pulumi Professional Services migration of ~700 resources from Terraform.

## Root Cause

The bridge's `validInputsFromPlan` function explicitly documents this limitation:

[`detailed_diff.go:166`](https://github.com/pulumi/pulumi-terraform-bridge/blob/79b2ab42273110be19bfc06997abfacae7772604/pkg/tfbridge/detailed_diff.go#L166):
> "nested sets will only get matched if they are exactly the same"

[`detailed_diff.go:197-198`](https://github.com/pulumi/pulumi-terraform-bridge/blob/79b2ab42273110be19bfc06997abfacae7772604/pkg/tfbridge/detailed_diff.go#L197-L198):
> "nested sets will likely get their elements reordered. This means that nested sets will not be matched correctly but that should be a rare case."

When `matchPlanElementsToInputs` fails to match all elements, [`makeSetDiff` falls back to a whole-set UPDATE](https://github.com/pulumi/pulumi-terraform-bridge/blob/79b2ab42273110be19bfc06997abfacae7772604/pkg/tfbridge/detailed_diff.go#L708-L715).

WAFv2 WebACL `rules` is a real-world case where this "rare case" is the common case, because the schema has multiple levels of nested TypeSet blocks:

```
rules (TypeSet)
  ├─ action (TypeSet)
  ├─ visibility_config (TypeSet)
  └─ statement (TypeSet)
       ├─ byte_match_statement (TypeSet)
       │    ├─ field_to_match (TypeSet)
       │    └─ text_transformation (TypeSet)
       ├─ and_statement (TypeSet)
       │    └─ statement (TypeSet)  ← recursive nesting
       ├─ or_statement (TypeSet)
       ├─ not_statement (TypeSet)
       └─ ... (10+ statement types, all TypeSet)
```

The nested TypeSet elements get reordered during planning, `validInputsFromPlan` returns false for the mismatched elements, and the fallback triggers.

## Observed Behavior

After `pulumi import` of a WAFv2 WebACL:

```
pulumi preview
...
  ~ rules: [
      ... (entire set shown as UPDATE)
    ]
```

## Expected Behavior

Either element-level diffs (if actual changes exist) or no diff (if the rules are unchanged).

goutamadwant:
## Summary

Resolves #3495

This updates detailed diff input-to-plan validation so `TypeSet` values are matched by set membership instead of array position.

The reported WAFv2 case can reorder nested `TypeSet` blocks during planning. Before this change, `validInputsFromPlan` compared those nested sets positionally, failed to match the planned set element back to the user input, and `makeSetDiff` fell back to a whole-set `UPDATE`.

With this change, set values are matched against any compatible planned element while lists continue to use positional matching. That lets detailed diff keep reporting the real changed path instead of falling back to the whole set.

## Tests

- `go test ./pkg/tfbridge -run 'TestValidInputsFromPlan|TestDetailedDiffMatchesSetElementsWithReorderedNestedSets' -count=1`
- `go test ./pkg/tfbridge -count=1`

Not run locally:

- `go test ./pkg/tests/diff_test -run 'TestSDKv2DetailedDiffNestedSets' -count=1` requires the Terraform CLI in `PATH`.

## Breaking changes

None.

### DIFF
--- a/pkg/tfbridge/detailed_diff.go
+++ b/pkg/tfbridge/detailed_diff.go
@@ -163,7 +163,6 @@ func makeBaseDiff(old, new resource.PropertyValue) baseDiff {
 // Under the hood, it walks the plan and the inputs and checks that all differences stem from computed properties.
 // Any differences coming from properties which are not computed will be rejected.
 // Note that we are relying on the fact that the inputs will have defaults already applied.
-// Also note that nested sets will only get matched if they are exactly the same.
 func validInputsFromPlan(
 	path propertyPath,
 	inputs resource.PropertyValue,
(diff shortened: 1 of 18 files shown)


pkg/tfbridge/detailed_diff.go
```diff
@@ -163,14 +163,66 @@ func makeBaseDiff(old, new resource.PropertyValue) baseDiff {
...
 // Note that we are relying on the fact that the inputs will have defaults already applied.
-// Also note that nested sets will only get matched if they are exactly the same.
 func validInputsFromPlan(
 	path propertyPath,
 	inputs resource.PropertyValue,
 	plan resource.PropertyValue,
 	tfs shim.SchemaMap,
 	ps map[string]*info.Schema,
 ) bool {
+	validSetInputsFromPlan := func(
```
guineveresaenger: nit: I think this should be a package-level function.

guineveresaenger (commented):
I want to give @VenelinMartinov a chance to weigh in because he's more familiar with this section of the codebase.

In the meantime I'll kick off tests.

Thank you very much, @goutamadwant.

guineveresaenger:
OK, so it looks like we need to re-record some test baselines. Please let us know if you'd like help with that.

Looking at the unit tests again, it seems like the code only gets tested on the easy cases where each item has one obvious match. 
Could you add a few cases to `TestValidInputsFromPlan` that would fail without the fix?
- Items where more than one pairing is possible and the code has to pick the right one so everything matches
- A case where two items compete for the same single match, so not everything can be paired
- A reordered set where one field is also auto-filled by the system, like in the actual WAFv2 case, not just the same values shuffled

VenelinMartinov:
Hey, thanks for tagging me here! I had a look at the changes and I don't think this addresses the WAFV2 example from https://github.com/pulumi/pulumi-terraform-bridge/issues/3495

@goutamadwant which provider did you hit this in? Did you test with the provider?

goutamadwant:
thanks @VenelinMartinov I reproduced the WAFv2 shape using the copied AWS WebACL provider schema fixture in this repo.
I pushed a follow-up commit that handles flattened `MaxItemsOne` values before list/set matching.
Let me know if you have any additional thoughts to handle this. thanks!

PR #3504, opened 2026-07-01
Skip redundant schema validation during schema generation

iwahbe:
`genPackageSpec` bound the generated schema purely to validate it and then discarded the bound package. That `BindSpec` call can dominate `pulumi package add` on large providers. For dynamic AWS, it takes roughly 8.7s. The bridge should generate valid schemas, and that should be enforced with tests, not extremely expensive runtime checks.

The SDK-emission path in `GenerateFromSchema` still binds and validates the spec before emitting, so dropping the check here only affects the schema-only path, which marshals the spec directly. This removes the `BindSpec` call along with the loader field and diagnostic-sink plumbing that existed solely to support it.

This does weaken `NoDanglingReferences`, limiting it to cases where we already pay the expensive binding cost: language specific SDK generation. If authors want the surety back, they should do the check out of the hot path via `pulumi schema check ./bin/pulumi-resource-<your-provider>`.

### DIFF
--- a/pkg/tfbridge/info/info.go
+++ b/pkg/tfbridge/info/info.go
@@ -180,6 +180,10 @@ type Provider struct {
 	DisableRequiredWithDefaultTurningOptional bool
 
 	// Check generated schema for dangling references. Newer providers should opt into this.
+	//
+	// This is not consistently enforced. Providers authors should prefer to run:
+	//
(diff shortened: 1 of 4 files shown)


iwahbe:

- #3504 ◀

corymhall (approved):

PR #3505, opened 2026-07-01
defer expensive operations until they are needed

tgummerer:
Currently providers using tfbridge can take a long time to start up. For example `pulumi-aws` takes 1.7 seconds to start up.

There are two expensive operations that can easily be deferred until they are needed.  Namely decoding the pulumi schema, and loading the resources map when creating the importer.

Both of these can be deferred until later, so e.g. just retrieving the schema, which we do in `pulumi do <resource> --help` doesn't need to wait for these expensive operations to finish.  These changes combined reduce the startup time of the `pulumi-aws` by ~1 second, which given `pulumi do <package> --help>` takes 2.5 seconds right now is significant.

### DIFF
--- a/pkg/tfbridge/provider.go
+++ b/pkg/tfbridge/provider.go
@@ -98,8 +98,7 @@ type Provider struct {
 	dataSources      map[tokens.ModuleMember]DataSource // a map of Pulumi module tokens to data sources.
 	supportsSecrets  bool                               // true if the engine supports secret property values
 	pulumiSchema     []byte                             // the JSON-encoded Pulumi schema.
-	pulumiSchemaSpec *pschema.PackageSpec
-	pulumiSchemaOnce sync.Once // guards lazy decoding of pulumiSchema into pulumiSchemaSpec.
+	pulumiSchemaSpec func() *pschema.PackageSpec        // lazily decodes pulumiSchema exactly once.
 	memStats         memStatCollector
 	hasTypeErrors    map[resource.URN]struct{}
 	providerOpts     []providerOption
@@ -285,6 +284,9 @@ func newProvider(ctx context.Context, host *provider.HostClient,
 		providerOpts:  opts,
 	}
 	ctx = p.loggingContext(ctx, "")
+	p.pulumiSchemaSpec = sync.OnceValue(func() *pschema.PackageSpec {
+		return decodePulumiSchema(ctx, pulumiSchema)
+	})
 	p.initResourceMaps()
 	return providerserver.NewPanicRecoveringProviderServer(&providerserver.PanicRecoveringProviderServerOptions{
 		Logger:                 host,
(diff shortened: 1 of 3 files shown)


tgummerer:
> Pull Request / Test and Lint / lint (pull_request)

This test needs https://github.com/pulumi/pulumi-terraform-bridge/pull/3506 to get green.

pkg/tfbridge/provider.go
```diff
@@ -98,6 +99,7 @@ type Provider struct {
 	supportsSecrets  bool                               // true if the engine supports secret property values
 	pulumiSchema     []byte                             // the JSON-encoded Pulumi schema.
 	pulumiSchemaSpec *pschema.PackageSpec
+	pulumiSchemaOnce sync.Once // guards lazy decoding of pulumiSchema into pulumiSchemaSpec.
```
iwahbe: I'd replace both of these with a [`sync.OnceValue`](https://pkg.go.dev/sync#OnceValue). 
```suggestion
pulumiSchemaSpec func() *pschema.PackageSpec
```

pkg/tfbridge/provider.go
```diff
@@ -284,13 +286,6 @@ func newProvider(ctx context.Context, host *provider.HostClient,
 	}
 	ctx = p.loggingContext(ctx, "")
 	p.initResourceMaps()
-	if pulumiSchema != nil || len(pulumiSchema) != 0 {
-		var schema pschema.PackageSpec
-		if err := json.Unmarshal(pulumiSchema, &schema); err != nil {
-			GetLogger(ctx).Debug(fmt.Sprintf("unable to unmarshal pulumi package spec: %s", err.Error()))
```
iwahbe, line 290: This should be error, surely?

iwahbe (commented):
I think this could be better expressed with `sync.OnceValue`.

pkg/tfbridge/provider.go
tgummerer: Looking at the code where it's used, it looks like we'd rather skip type checking the config rather than erroring out here.  But I don't know enough about the bridge to say one way or another really.

pkg/tfbridge/provider.go
tgummerer: Yeah, that seems like a good idea, done.

pkg/tfbridge/provider.go
```diff
@@ -299,6 +296,30 @@ func newProvider(ctx context.Context, host *provider.HostClient,
...
+}
+
+// schemaSpec returns the decoded Pulumi schema, decoding it on first use. Returns nil if no schema
+// was provided or if it fails to decode.
+func (p *Provider) schemaSpec() *pschema.PackageSpec {
+	if p.pulumiSchemaSpec == nil {
+		return nil
+	}
+	return p.pulumiSchemaSpec()
+}
```
iwahbe, line 321: Feels like we shouldn't need this anymore, since `p.pulumiSchemaSpec` should always be non-nil, even if `p.pulumiSchemaSpec()` is nil.

iwahbe (approved):

pkg/tfbridge/provider.go
tgummerer: This is still needed for testing I believe.

iwahbe: 😦 OK. We can merge with it.

guineveresaenger: 😭

guineveresaenger (approved):

Issue #1955, opened 2024-05-10
Support Provider-Defined Functions (Plugin Framework)


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

See https://developer.hashicorp.com/terraform/plugin/framework/functions/concepts

iwahbe:
## Acceptance criteria

Provider-defined functions are a Plugin Framework / protocol 5.5+/6.5+ feature (Terraform 1.8+). In Terraform they live in a namespace fully separate from resources and data sources: names are registered unprefixed (e.g. `parse_arn`, not `aws_parse_arn`) and called as `provider::<local-name>::<function-name>(...)`. They are pure, offline computations — no side effects, no access to provider configuration. The bridge maps them to Pulumi invokes.

### Build-time (tfgen)

- [ ] Functions reported by a PF provider's `GetFunctions`/`GetProviderSchema` appear in the generated Pulumi schema as entries in the `functions` map, with SDKs generated for all supported languages.
- [ ] `ProviderInfo` gains a mapping surface for functions, parallel to `DataSources`, but keyed by the unprefixed TF function name. Automatic token computation (`MustComputeTokens`) covers functions, and a token strategy guarantees no collision with data-source invoke tokens (a TF function `foo` and a data source mapped to `getFoo` must coexist).
- [ ] Missing/extra mapping validation applies to functions with the same semantics as resources and data sources (`PULUMI_SKIP_MISSING_MAPPING_ERROR`, `PULUMI_SKIP_EXTRA_MAPPING_ERROR`).
- [ ] Generated functions use the modern `FunctionSpec` shape: ordered TF positional parameters project via `multiArgumentInputs` (TF parameter `Name` is documentation-only, so duplicate/empty names must be handled deterministically), and the TF `Return` type projects via `returnType` as a direct (non-object) return. A trailing TF variadic parameter projects to a final list-typed argument, since the Pulumi schema has no variadic concept.
[...]

PR #3507, opened 2026-07-02
Add provider-defined functions to the shim, info, and token layers

iwahbe:
## Summary

Terraform providers built with the Plugin Framework can define provider functions: pure computations called as `provider::<name>::<fn>(...)` in Terraform. This change adds the plumbing needed to represent them in the bridge, ahead of schema generation (#3508) and runtime support (#3509).

- `pkg/tfshim`: new `shim.Function` and `shim.FunctionParameter` types describe a function signature with tftypes type constraints, including a trailing variadic parameter and `AllowNullValue`. The `shim.Provider` interface gains a `Functions()` accessor, keyed by the unprefixed Terraform function name. SDK-based providers return nil; the schema-only, filtering, and unimplemented providers are updated accordingly.
- `pkg/pf`: the Plugin Framework shim gathers functions from providers that implement `provider.ProviderWithFunctions` (new `pfutils.GatherFunctions`), and the proto shim projects `GetProviderSchemaResponse.Functions` for the dynamic bridge. The internal muxer shim layers functions across sub-providers like resources and data sources: the base shim dominates on overlap, and only `AugmentShimWithDisjointPF` asserts disjointness. Mapped function tokens dispatch to the sub-provider that defines them.
- `pkg/tfbridge/info`: `ProviderInfo` gains a `Functions` section, keyed by the unprefixed Terraform name, with a minimal `info.Function{Tok, Docs, DeprecationMessage}`. Functions marshal into the GetMapping payload so converters can translate provider function calls.
- `pkg/tfbridge/tokens`: every token strategy maps functions into the top-level (index) module, camelCasing the Terraform name with no `get` prefix. `ComputeTokens` fills missing mappings, honoring `IgnoreMappings` and existing entries. `WithCaseInsensitiveDedup` passes function strategies through.

Part of #1955.

## Testing

- New token strategy tests cover default mapping, existing-mapping overrides, `IgnoreMappings`, and module-mapping strategies (functions stay top-level).
- `go build ./...`, `go vet ./...`, and the `pkg/tfbridge`, `pkg/tfshim`, `pkg/pf/internal/muxer`, `pkg/pf/internal/schemashim`, and `pkg/pf/proto` test suites pass at this commit.

### DIFF
--- a/pkg/pf/internal/muxer/muxer.go
+++ b/pkg/pf/internal/muxer/muxer.go
@@ -17,6 +17,7 @@ package muxer
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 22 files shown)


iwahbe:

- #3507 ◀
    - #3508
        - #3509
            - #3510

corymhall (approved):

PR #3508, opened 2026-07-02
Generate Pulumi schema for provider-defined functions

iwahbe:
## Summary

Provider-defined functions reported by the shim (#3507) now appear in the generated Pulumi schema as invokes using the modern `FunctionSpec` shape.

- tfgen gathers provider functions next to data sources, with the same missing and extra mapping validation semantics (`PULUMI_SKIP_MISSING_MAPPING_ERROR`, `PULUMI_SKIP_EXTRA_MAPPING_ERROR`, and `ProviderInfo.IgnoreMappings`). A function whose token collides with a data source invoke token fails schema generation, since both share the schema's `functions` section.
- Ordered Terraform parameters project positionally via `multiArgumentInputs`. Terraform parameter names are documentation-only, so empty and duplicate names resolve deterministically. A trailing variadic parameter projects to a final optional list-typed argument. The Terraform return type projects via `returnType`, directly (non-object) for scalar returns. A parameter that allows null projects as optional unless a later parameter is required.
- Function signatures reuse the bridge's existing schema-driven machinery: the new `pkg/internal/functions` package synthesizes shim schemas from the tftypes type constraints, so object attribute naming goes through `TerraformToPulumiNameV2` (including standard pluralization) and value conversion at runtime (#3509) goes through the `pkg/convert` encoders and decoders. Top-level argument names are pinned exactly, since `multiArgumentInputs` refers to them positionally.
- Objects become named auxiliary types; dynamic maps to `Any`. Tuple types are rejected; a survey of all published Terraform providers (515 functions across 108 providers) found no function using them.
- Function documentation is discovered under `website/docs/functions/` and `docs/functions/`, falling back to the `Summary` and `Description` the provider ships in its schema. `Summary`, `Description`, and `DeprecationMessage` flow into the schema.

Part of #1955.

## Testing

- New schema-generation tests cover positional parameters (incl. `AllowNullValue` optionality rules), variadic functions, empty/duplicate parameter names, object parameters producing auxiliary named types, scalar/object/dynamic returns, the tuple hard error, the data-source token collision error, and missing/extra/ignored mapping validation.
- Unit tests for `pkg/internal/functions` cover argument naming, schema synthesis, naming consistency (incl. pluralization), and round trips through the `pkg/convert` encoders.

### DIFF
--- a/pkg/internal/functions/functions.go
+++ b/pkg/internal/functions/functions.go
@@ -26,7 +26,6 @@ package functions
 
 import (
 	"fmt"
-	"sort"
 
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 
@@ -67,16 +66,6 @@ func ArgumentNames(fn shim.Function) []string {
 		assign(v.Name, len(fn.Parameters))
 	}
 	return names
-}
-
-// SortedAttributeNames returns the attribute names of t in a stable order.
-func SortedAttributeNames(t tftypes.Object) []string {
-	attrs := make([]string, 0, len(t.AttributeTypes))
-	for attr := range t.AttributeTypes {
-		attrs = append(attrs, attr)
-	}
-	sort.Strings(attrs)
-	return attrs
 }
 
 // SchemaFromType synthesizes a shim schema mirroring the given type constraint, so that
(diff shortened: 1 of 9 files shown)


iwahbe:

- #3508 ◀
    - #3509
        - #3510

corymhall (approved):

PR #3509, opened 2026-07-02
Route function invokes through the Terraform CallFunction RPC

iwahbe:
## Summary

Calling a generated function invoke on a Plugin Framework based provider now routes through the Terraform `CallFunction` RPC instead of `ReadDataSource`.

- The PF provider resolves function tokens before data source tokens in `Invoke`. Arguments encode through the existing `pkg/convert` encoders, driven by schemas synthesized from the function signature (#3508); the final list-typed Pulumi argument unpacks into individual call arguments for a variadic parameter.
- Unknown (computed) argument values do not error and do not call the function: functions are pure, so an unknown input makes the whole result unknown, and the invoke resolves once inputs are known.
- Functions are pure by the Terraform contract, so the call does not require the provider to be configured.
- Null handling honors `AllowNullValue`: a missing or null argument for a non-nullable parameter returns a check failure naming the argument. Unexpected arguments are rejected the same way.
- A `FunctionError` with an argument position maps back to the Pulumi argument name, including expanded variadic positions, and surfaces as a check failure; positions past the declared parameters of a non-variadic function surface as a function-level error rather than blaming the wrong argument.
- Non-object results return as a single-property map (`result`), matching the wire convention SDKs use to unwrap direct returns (for example, `invokeSingle` in Node.js). Object results return their renamed properties directly.

Part of #1955.

## Testing

The synthetic testbridge provider gains three functions covering variadic parameters, object returns with argument-scoped errors, and nullable parameters; its golden schema is regenerated (reviewable in this diff). gRPC replay tests exercise success, failure, and unknown-input paths, all without configuring the provider. `TestSchemaGenInSync` verifies the regenerated golden.

### DIFF
--- a/pkg/pf/internal/providerbuilder/build_provider.go
+++ b/pkg/pf/internal/providerbuilder/build_provider.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
 	dschema "github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/function"
 	"github.com/hashicorp/terraform-plugin-framework/provider"
 	"github.com/hashicorp/terraform-plugin-framework/provider/schema"
 	"github.com/hashicorp/terraform-plugin-framework/providerserver"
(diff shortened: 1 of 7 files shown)


iwahbe:

- #3509 ◀
    - #3510

corymhall (approved):

PR #3520, opened 2026-07-09
Add a cross-test for provider-defined function parity

iwahbe:
crosstests.Function evaluates a provider-defined function on both sides
of the bridge — Terraform computes an output from a provider::<name>::<fn>
call against the reattached provider, and the bridged Pulumi provider
serves an Invoke of the same call — and returns both values for parity
assertions.

TestCrossFunctionParity covers the testbridge functions: a variadic call
(with and without trailing arguments), an object return, and a nullable
parameter passed as null versus omitted. tfcheck gains GetOutputJSON so
structured output values survive the readback intact.


iwahbe:

- #3520 ◀

corymhall (approved):

PR #3536, opened 2026-07-16
fix(pf): send a compatible Terraform version

### LINKED ISSUE #3514: Dynamic bridge sends non-semver TerraformVersion during ConfigureProvider
## Summary

The dynamic Terraform provider bridge hardcodes a non-semver TerraformVersion string during provider configure:

```go
req := &tfprotov6.ConfigureProviderRequest{
    Config:           config,
    TerraformVersion: "pulumi-terraform-bridge",
}
```

Source: `pkg/pf/tfbridge/provider_configure.go`.

This can crash Terraform providers that parse `TerraformVersion` as a semantic version. Terraform/OpenTofu normally send a version string such as `1.12.3`; some providers use that field for compatibility checks.

Concrete affected provider: Rootly `rootlyhq/rootly` `v5.16.1`.

Related Rootly issue: https://github.com/rootlyhq/terraform-provider-rootly/issues/379

## Affected versions

Confirmed with:

- Pulumi CLI `v3.242.0`
- `pulumi-terraform-provider` bridge plugin `v1.1.4`
- `pulumi-terraform-bridge` master still appears to contain the hardcoded value
- Rootly Terraform provider `v5.16.1`
- OpenTofu `v1.12.3` for contrast testing

## User-visible failure

When importing or previewing a Rootly resource via the Pulumi dynamic bridge, configure fails with:

```text
error calling ConfigureProvider: rpc error: code = Unavailable desc = error reading from server: EOF
```

The EOF makes this look like an auth/network/plugin transport problem. In reality, the inner Terraform provider exits during ConfigureProvider because it tries to parse the bridge-provided TerraformVersion as semver.

## Root cause

Rootly's provider has a Terraform >= 1.0 guard similar to:

```go
if goversion.Must(goversion.NewVersion(p.TerraformVersion)).LessThan(goversion.Must(goversion.NewVersion("1.0"))) {
    // return diagnostic
}
```

The bridge sends:

```text
pulumi-terraform-bridge
```

That is not semver, so `goversion.Must(...)` panics, the provider process exits, and Pulumi reports only EOF.

Rootly should also avoid panicking on an unparseable version, but the bridge's value is surprising because Terraform/OpenTofu callers normally provide a parseable Terraform version.

## Reproduction outline

1. Add Rootly through the dynamic Terraform bridge:

   ```bash
   pulumi package add terraform-provider registry.opentofu.org/rootlyhq/rootly 5.16.1
   ```

2. Configure the Rootly provider and run `pulumi preview` or `pulumi import`.

3. Observe:

   ```text
   error calling ConfigureProvider: rpc error: code = Unavailable desc = error reading from server: EOF
   ```

I also reproduced this by driving the bridge directly over gRPC: `Parameterize` succeeds, then `Configure` causes the inner Rootly provider process to exit. When I locally patched the bridge to send `TerraformVersion: "1.5.0"`, Rootly `Configure`, authenticated `Read`, `pulumi import`, and `pulumi preview` all succeeded.

## OpenTofu contrast test

Using OpenTofu `v1.12.3` with the same Rootly provider and an intentionally invalid token does not panic; it fails cleanly with a 401 from the Rootly API:

```text
data.rootly_severity.test: Reading...

Planning failed. OpenTofu encountered an error while generating this plan.

Error: Failed to make request: (401) {"errors":[{"title":"Invalid token","status":"401"}]}
```

This confirms the provider can configure normally when TerraformVersion is parseable.

## Expected behavior

The bridge should avoid sending a non-semver value in the `TerraformVersion` protocol field, or should provide a supported way to override it for providers that parse that field.

Possible fixes:

1. Send a conservative semver-compatible default, e.g. `1.0.0` or another compatible Terraform/OpenTofu version.
2. Add an environment/config override such as `PULUMI_TERRAFORM_VERSION`, defaulting to the current behavior if compatibility requires it.
3. Omit the field if the protocol/provider framework permits that safely.

The local proof-of-fix I used was equivalent to option 2: bridge reads `PULUMI_TERRAFORM_VERSION`, and setting `PULUMI_TERRAFORM_VERSION=1.5.0` allows Rootly imports/previews to work.

## Why this matters

Any Terraform provider that parses `TerraformVersion` as semver can fail under the dynamic bridge even though it works under Terraform/OpenTofu. The failure mode is especially hard to diagnose because Pulumi only reports EOF from the provider process.

corymhall:
## Summary

- send providers the stable SemVer value `1.0.0+pulumi-terraform-bridge` during PF configuration
- preserve the bridge identity as SemVer build metadata without affecting version precedence
- cover the shared PF request and the dynamic protocol-v5 forwarding path
- update dynamic replay fixtures for the new configure request

## Rationale

`TerraformVersion` is intended for logging, analytics, and User-Agent purposes, but some providers parse it as SemVer or use it for compatibility checks. The previous `pulumi-terraform-bridge` value can therefore crash providers such as Rootly during configuration.

The `1.0.0` base is a conservative compatibility floor for the protocol versions the bridge uses. The `+pulumi-terraform-bridge` metadata retains the original caller identity and is ignored for SemVer precedence comparisons.

Fixes #3514.

## Testing

- `make test RUN_TEST_CMD='./pkg/pf/tests'`
- `make test RUN_TEST_CMD='./dynamic -run "^(TestSDKv1Provider|TestLogReplayProvider|TestLogReplayProviderWithProgram|TestConfigInDestroy)$"'`
- `make lint`

### DIFF
--- a/dynamic/test/sdkv1provider/main.go
+++ b/dynamic/test/sdkv1provider/main.go
@@ -1,16 +1,16 @@
 package main
 
 import (
+	"fmt"
+
+	goversion "github.com/hashicorp/go-version"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/plugin"
 	"github.com/hashicorp/terraform-plugin-sdk/terraform"
 )
 
 func Provider() *schema.Provider {
-	return &schema.Provider{
-		ConfigureFunc: func(*schema.ResourceData) (any, error) {
-			return nil, nil
-		},
+	p := &schema.Provider{
 		ResourcesMap: map[string]*schema.Resource{
 			"test_res": {
 				Schema: map[string]*schema.Schema{
@@ -26,6 +26,19 @@ func Provider() *schema.Provider {
 			},
 		},
 	}
+	p.ConfigureFunc = func(*schema.ResourceData) (any, error) {
+		terraformVersion, err := goversion.NewVersion(p.TerraformVersion)
+		if err != nil {
+			return nil, fmt.Errorf("invalid Terraform version %q: %w", p.TerraformVersion, err)
+		}
+		minimumVersion := goversion.Must(goversion.NewVersion("1.0.0"))
+		if terraformVersion.LessThan(minimumVersion) {
+			return nil, fmt.Errorf("terraform version %s is below minimum version %s",
+				terraformVersion, minimumVersion)
+		}
+		return nil, nil
+	}
+	return p
 }
 
 func main() {
--- a/dynamic/provider_test.go
+++ b/dynamic/provider_test.go
@@ -790,6 +790,9 @@ func TestSDKv1Provider(t *testing.T) {
 
 	server := parameterizedTestServer(t, sdkv1ProviderPath)
 
+	_, err := server.Configure(t.Context(), &pulumirpc.ConfigureRequest{})
+	require.NoError(t, err)
+
 	const typ = "sdkv1:index/res:Res"
 	urn := string(resource.NewURN(
 		"test", "test", "", typ, "res",
--- a/dynamic/testdata/TestConfigInDestroy/logs.json
+++ b/dynamic/testdata/TestConfigInDestroy/logs.json
@@ -26,7 +26,7 @@
   {
     "name": "github.com/hashicorp/terraform-plugin-mux/tf5to6server.v5tov6Server.ConfigureProvider",
     "call": "/home/runner/go/pkg/mod/github.com/hashicorp/terraform-plugin-mux@v0.16.0/tf5to6server/tf5to6server.go:60",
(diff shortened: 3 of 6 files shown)


dynamic/test/sdkv1provider/main.go
```diff
@@ -1,16 +1,16 @@
 package main
 
 import (
+	"fmt"
+
+	goversion "github.com/hashicorp/go-version"
```
pose, line 6: (minor) Not sure how much we care about this, but I noticed that `pulumi-java` uses https://github.com/blang/semver . Do you think we should standarize towards `go-version` instead?

pose (approved):

dynamic/test/sdkv1provider/main.go
corymhall: I don't think it matters that much. Each semver parser probably optimizes for certain things, in this case we want to match what upstream uses.

PR #3565, opened 2026-08-07
Filter underscore properties from schema

Frassle:
Codegen doesn't handle underscore prefixed keys well, they are unexported in Go, and "private" in Python (and name mangled if starting with a double underscore, which some TF providers do).

Core is looking at erroring in schema binding on properties starting with underscore, so in preparation for that we need to make sure tfbridge doesn't echo these properties to the Pulumi schema.

Looking at some examples of this across terraform providers these mostly appear to be used for "private" fields that aren't shown in documentation. So I think we can just filter these out from the schema and let state and `__meta` keep track of them.


iwahbe:
I think https://github.com/pulumi/pulumi-terraform-bridge/pull/3597 is the same logic with a cleaner implementation.

Frassle:
Superseded by https://github.com/pulumi/pulumi-terraform-bridge/pull/3597

PR #3568, opened 2026-08-08
tfbridge: stop stripping empty __defaults from refresh read inputs

### LINKED ISSUE #3119: Lots of `"SnapshotManager: mustWrite() true because of Inputs"` with`--refresh`, without any changes
### What happened?

We see a lot of requests to our state backend when doing a `pulumi up` without any changes.
Our large >3000 resource stacks are currently taking a very long time (>40min) to up even if there are no changes.  We are looking to speed those up.
Per recommendation of @Frassle on [Slack](https://pulumi-community.slack.com/archives/C84L4E3N1/p1748870029246159) we've enabled debug logging with ` --logtostderr -v9`.

When doing a `pulumi up -p 1 --logtostderr -v9 2> stderr.log` on a test stack I'm seeing:
```
mustWrite() false
```
on all resources when parsing the stderr logs.
When doing `pulumi up -p 1 --refresh --logtostderr -v9 2> stderr-with_refresh.log` I see:
```
<snip>
Resources:
    39 unchanged
```
in stdout
And in stderr, a lot of:
```
<snip>
SnapshotManager: mustWrite() true because of Inputs
<snip>
```

The logs are a bit hard to parse. I have added `-p 1` in the hopes log lines wouldn't interleave, but I'm still having trouble seeing where individual operations start and end. (Also seeing the same `.Diff()` lines multiple times)

I am looking at the `.Diff()` lines and am seeing that the `.oldInputs`  have 2 entries with `__defaults={[]}` and the `.newInputs` have 3.

Here is a sanitized chunk of the `.Diff()` logs. Note that I'm grepping so might be missing valuable information. Let me know if I need to provide more.

Here is a sanitized chunk of the `.Diff()` logs:
```
I0603 09:31:54.833557    4222 provider_plugin.go:1173] Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW): executing (#oldInputs=11#oldOutputs=12,#newInputs=11)
.oldInputs]: __defaults={[]}
.oldInputs]: credentials={&{{map[key:{&{{REDACTED}}} secret:{&{{REDACTED}}}]}}}
.oldInputs]: key={&{{REDACTED}}}
.oldInputs]: secret={&{{REDACTED}}}
.oldInputs]: host={*}
.oldInputs]: kafkaCluster={map[__defaults:{[]} id:{lkc-REDACTED}]}
.oldInputs]: __defaults={[]}
.oldInputs]: id={lkc-REDACTED}
.oldInputs]: operation={READ}
.oldInputs]: patternType={LITERAL}
.oldInputs]: permission={ALLOW}
.oldInputs]: principal={User:sa-REDACTED}
.oldInputs]: resourceName={*}
.oldInputs]: resourceType={GROUP}
.oldInputs]: restEndpoint={https://pkc-REDACTED.westeurope.azure.confluent.cloud:443}
.newInputs]: __defaults={[]}
.newInputs]: credentials={&{{map[__defaults:{[]} key:{&{{REDACTED}}} secret:{&{{REDACTED}}}]}}}
.newInputs]: __defaults={[]}
.newInputs]: key={&{{REDACTED}}}
.newInputs]: secret={&{{REDACTED}}}
.newInputs]: host={*}
.newInputs]: kafkaCluster={map[__defaults:{[]} id:{lkc-REDACTED}]}
.newInputs]: __defaults={[]}
.newInputs]: id={lkc-REDACTED}
.newInputs]: operation={READ}
.newInputs]: patternType={LITERAL}
.newInputs]: permission={ALLOW}
.newInputs]: principal={User:sa-REDACTED}
.newInputs]: resourceName={*}
.newInputs]: resourceType={GROUP}
.newInputs]: restEndpoint={https://pkc-REDACTED.westeurope.azure.confluent.cloud:443}
```

Here is the same, but redacted raw:
```
I0603 09:31:54.833557    4222 provider_plugin.go:1173] Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW): executing (#oldInputs=11#oldOutputs=12,#newInputs=11)
I0603 09:31:54.833598    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: __defaults={[]}
I0603 09:31:54.833646    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: credentials={&{{map[key:{&{{REDACTED}}} secret:{&{{REDACTED}}}]}}}
I0603 09:31:54.833715    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: key={&{{REDACTED}}}
I0603 09:31:54.833768    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: secret={&{{REDACTED}}}
I0603 09:31:54.833818    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: host={*}
I0603 09:31:54.833853    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: kafkaCluster={map[__defaults:{[]} id:{lkc-REDACTED}]}
I0603 09:31:54.833875    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: __defaults={[]}
I0603 09:31:54.833895    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: id={lkc-REDACTED}
I0603 09:31:54.833914    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: operation={READ}
I0603 09:31:54.833932    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: patternType={LITERAL}
I0603 09:31:54.833949    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: permission={ALLOW}
I0603 09:31:54.833968    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: principal={User:sa-REDACTED}
I0603 09:31:54.833987    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: resourceName={*}
I0603 09:31:54.834006    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: resourceType={GROUP}
I0603 09:31:54.834023    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).oldInputs]: restEndpoint={https://pkc-REDACTED.westeurope.azure.confluent.cloud:443}
I0603 09:31:54.834440    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: __defaults={[]}
I0603 09:31:54.834472    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: credentials={&{{map[__defaults:{[]} key:{&{{REDACTED}}} secret:{&{{REDACTED}}}]}}}
I0603 09:31:54.834513    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: __defaults={[]}
I0603 09:31:54.834534    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: key={&{{REDACTED}}}
I0603 09:31:54.834575    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: secret={&{{REDACTED}}}
I0603 09:31:54.834613    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: host={*}
I0603 09:31:54.834636    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: kafkaCluster={map[__defaults:{[]} id:{lkc-REDACTED}]}
I0603 09:31:54.834658    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: __defaults={[]}
I0603 09:31:54.834681    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: id={lkc-REDACTED}
I0603 09:31:54.834703    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: operation={READ}
I0603 09:31:54.834722    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: patternType={LITERAL}
I0603 09:31:54.834740    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: permission={ALLOW}
I0603 09:31:54.834759    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: principal={User:sa-REDACTED}
I0603 09:31:54.834776    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: resourceName={*}
I0603 09:31:54.834798    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: resourceType={GROUP}
I0603 09:31:54.834817    4222 rpc.go:77] Marshaling property for RPC[Provider[confluentcloud, 0x14002955450].Diff(urn:pulumi:stackname::project-name::confluentcloud:index/kafkaAcl:KafkaAcl::resource-name-lkc-REDACTED-group-*-read,lkc-REDACTED/GROUP#*#LITERAL#User:sa-REDACTED#*#READ#ALLOW).newInputs]: restEndpoint={https://pkc-REDACTED.westeurope.azure.confluent.cloud:443}
```

### Example

A working Confluent Cloud environment is required to repro.

Create a Pulumi program with:

```typescript
// index.ts
import * as confluentcloud from "@pulumi/confluentcloud";

export const topic = new confluentcloud.KafkaTopic("test", {
  topicName: "test",
  partitionsCount: 1,
  config: {
    "retention.ms": "3600",
  },
  kafkaCluster: {
    id: "lkc-qwerty",
  },
  restEndpoint: "https://pkc-asdfgh.westeurope.azure.confluent.cloud:443",
  credentials: {
    key: process.env.CONFLUENT_KEY!!,
    secret: process.env.CONFLUENT_SECRET!!,
  },
});
```

```javascript
// package.json
{
  "name": "test",
  "main": "index.ts",
  "scripts": {},
  "devDependencies": {
    "@swc/core": "1.9.2",
    "@swc/types": "0.1.16",
    "@types/node": "^20.12.12",
    "ts-node": "10.9.2",
    "typescript": "5.6.2"
  },
  "dependencies": {
    "@pulumi/confluentcloud": "2.29.0",
    "@pulumi/pulumi": "3.173.0"
  }
}
```

Install pulumi
```bash
curl -fsSL https://get.pulumi.com | sh -s -- --version 3.173.0
```

Initialize the stack
```bash
pulumi stack init test-stack
pulumi stack select test-stack
```

Deploy the resources
```bash
pulumi up
```

Run with `--refresh` and check logs
```bash
log_filename="$(date -u "+%Y-%m-%dT%H:%M:%S").log"
pulumi up -y -p 1 --refresh --stack test-stack --logtostderr -v9 2> $log_filename
cat $log_filename | grep "mustWrite() true" -B 1
```

### Output of `pulumi about`

```
CLI
Version      3.173.0
Go Version   go1.24.3
Go Compiler  gc

Plugins
KIND      NAME    VERSION
language  nodejs  3.173.0

Host
OS       darwin
Version  15.4.1
Arch     arm64
```

### Additional context

_No response_

### Contributing

### LINKED ISSUE #3567: Empty `__defaults` oscillate between `refresh` and `up`, forcing a checkpoint write for every affected resource on every run
Hello, human here. I am using Pulumi at my company to automate configuring Azure DevOps, which is normally great. Lately, we have had extreme slowness on tiny applies, like a 1 line change results in a 15 minute apply, which is obviously absurd. So, I had my Sol 5.6 investigate. It claims that the root cause of the slowness is a bug in the terraform bridge. The following is the LLM generated bug report:

### What happened?

`Read()` strips empty `__defaults: []` during refresh (added in #3356 to fix #3119), but `Check()`
still emits them. Because the two sides disagree, the value flips on every run and never settles:

1. `refresh` calls `Read()`, which strips empty `__defaults` and writes the stripped inputs to state.
2. The program's `Check()` returns inputs that still contain `__defaults: []`.
3. `sameSnapshotMutation.mustWrite` compares the two raw input bags, sees a difference, and cannot
   elide the checkpoint write - even though `Diff()` returned `DIFF_NONE` and the user sees no change.
4. `up` persists the checked inputs, so `__defaults` is back in state.
5. The next `refresh` strips it again.

The net effect is that a no-op `up` that follows a `refresh` is forced to write the checkpoint once
per affected resource. On our stack that is 787 of 1434 unchanged resources, on every single run.

#3356's behavior note anticipated a one-time normalization write and predicted that "subsequent
refreshes should stabilize". We measured that it does not stabilize.

### Example

Any bridged SDKv2 provider whose resources carry empty `__defaults` should reproduce this. Run
`refresh` followed by `up` twice and compare, for every `same` step, `oldState.inputs` against
`newState.inputs` from `pulumi preview --json --show-sames` - this is the same comparison
`mustWrite` performs.

We measured drift at three points in the cycle:

| Point in the cycle          | Resources whose inputs differ |
| --------------------------- | ----------------------------- |
| Immediately after `up`      | 0                             |
| Before refresh (next run)   | 0                             |
| Immediately after `refresh` | 787                           |

That count of 787 matches the engine exactly. Running `up --logtostderr -v=9` yields 787 occurrences
of `mustWrite() true because of Inputs`, and zero occurrences of `mustWrite() true` for any other
reason.

Representative differences (old = what refresh wrote to state, new = what the program produced):

```text
# 469 resources - nested, buildDefinition.ciTrigger
old: {"useYaml":true}
new: {"__defaults":[],"useYaml":true}

# ~246 resources - top level, many resource types
old: <absent>
new: []
```

Roughly 730 of the 787 are empty `__defaults`. The remainder are ordinary refresh normalization
differences (for example a list whose order the API does not preserve) and are not part of this report.

The relevant call site in v3.134.0, `pkg/tfbridge/provider.go`:

```go
cleanInputs := deconflict(ctx, res.TF.Schema(), res.Schema.Fields, inputs)
if isRefresh {
    deleteDefaultsKey(cleanInputs)
}
```

### Expected behavior

`Read()` and `Check()` should use the same convention for empty `__defaults`, so that a no-op run
after a refresh produces byte-identical inputs and the engine can elide the checkpoint write.

### Why this is expensive

The difference is cosmetic, but defeating elision is costly because every non-elided step
re-serializes and re-uploads the entire state file on a single goroutine. For us that is 787 x
19.5MB, roughly 15GB of blob traffic, to persist zero actual changes. The cost grows as O(N^2) in
resource count, which likely explains the 40+ minutes at 3000 resources reported in #3119.

Provider time is not the problem. From `PULUMI_DEBUG_GRPC` on an 837s `up`:

| Method                   | Calls | Cumulative |
| ------------------------ | ----- | ---------- |
| `LanguageRuntime/Run`    | 1     | 833.92s    |
| `ResourceProvider/Diff`  | 1314  | 4.54s      |
| `ResourceProvider/Check` | 1314  | 2.01s      |

`Diff` and `Check` together are under 7 seconds. Almost all remaining time is the engine writing
checkpoints.

Two controls confirm the attribution:

| Configuration                                        | `up` duration |
| ---------------------------------------------------- | ------------- |
| Refresh, then up (baseline)                          | 954s          |
| Refresh, then up with `PULUMI_SKIP_CHECKPOINTS=true` | 78s           |
| No refresh, then up                                  | 80s           |

Skipping the refresh reaches the same floor as disabling checkpoint writes entirely, which is what
you would expect if refresh-induced input churn is the only thing forcing the writes.

### Suggested fix

This is the symmetry problem @Frassle described on #3119:

> The cleaner fix here is probably to get the bridge fixed to always return `__defaults` in its
> input data so that we don't get these diffs.

#3356 established symmetry in the other direction, making `Read()` omit empty `__defaults` while
`Check()` continues to emit them. Either convention should work provided both sides use it:

1. Have `Read()` always return `__defaults`, as originally proposed, or
2. Also strip empty `__defaults` on the `Check()` path, so neither side ever emits them.

Option 2 is the smaller change relative to the current code, but it has to cover the program path as
well, otherwise the oscillation remains.

### Environment

- `pulumi-azuredevops` 3.16.0 (released 2026-07-17)
- `pulumi-terraform-bridge` v3.134.0, which contains #3356
- Pulumi CLI <FILL IN>
- Self-managed (DIY) state backend on Azure Blob Storage
- 1435 resources, 1434 of them `same` on a no-op run, 19,483,876 byte state file

### Additional context

Related: #3119 (closed by #3356), #3405, #3432, #3434.

The O(N^2) checkpoint amplification described above is an engine behavior rather than a bridge one,
and may deserve a separate issue against pulumi/pulumi. It is included here only to convey impact.

Happy to supply the full per-type breakdown or rerun with additional instrumentation.

### Output of `pulumi about`

```
$ pulumi about
CLI
Version      3.256.0
Go Version   go1.26.5
Go Compiler  gc

Host
OS       ubuntu
Version  26.04
Arch     x86_64
```

### Additional context

_No response_

### Contributing

Zamiell:
# Human:

This PR was generated by Claude Opus 5. It deletes a bunch of code that was introduced to fix #3119, but does not seem to actually fix it. So this PR removes it to fix the perf regression, and then I attempted to actually fix #3119 in #3569.

# LLM:

Fixes #3567.

This reverts the behavior change from #3356 (`tfbridge: strip empty __defaults from refresh read inputs`), which did not fix #3119 and introduced a semantic regression.

## Why the strip is not a normalization

`deleteDefaultsKey` treats `__defaults: []` as equivalent to no `__defaults` key at all. The bridge does not:

- `applyDefaults` (`pkg/tfbridge/schema.go`) uses the presence of the key to decide `useOldDefault`. When the key is **absent** it falls back to `func(key) bool { return true }`, meaning *every* old value is eligible to be reused as a default. When the key is present but **empty**, *no* old value is a default.
- `nameRequiresDeleteBeforeReplace` treats a missing `__defaults` list as evidence that the resource was autonamed, which changes delete-before-replace behavior.

So stripping the empty marker silently flips default tracking and replacement semantics for every resource, on every refresh.

## Why it does not fix #3119

#3356's behavior note predicted "a one-time normalization write ... After that write, subsequent refreshes should stabilize." That stabilization never happens, because the marker is not persisted state that can settle - it is regenerated by `Check` on every run:

1. `refresh` reads, and the strip removes `__defaults: []` from the refreshed inputs.
2. Those stripped inputs are written to state.
3. On the next `up`, the program's `Check` re-emits `__defaults: []`.
4. `sameSnapshotMutation.mustWrite` compares raw input bags with `DeepEquals`, sees a difference, and cannot elide the checkpoint write - even though `Diff` returns `DIFF_NONE`.
5. `up` persists `__defaults` back, so the next `refresh` strips it again.

The result is a permanent oscillation rather than a one-time write.

Because the engine serializes checkpoint writes on a single goroutine and rewrites the *whole* state file per non-elided step, this is quadratic in stack size. On a stack of 1434 resources with a 19.5 MB state file, measured against `pulumi-azuredevops` (which ships bridge v3.134.0, i.e. *with* #3356):

| Scenario | `up` wall time |
| --- | --- |
| Baseline (`refresh` then `up`, no user-visible changes) | 954s |
| Same, with `PULUMI_SKIP_CHECKPOINTS=true` | 78s |
| Same, with the `refresh` skipped | 80s |

Pulumi's own `-v=9` logging reported **787** `mustWrite() true because of Inputs` and zero writes for any other reason. Counting drifted resources independently via `pulumi preview --json --show-sames` at three points gives: **0** before refresh, **787** after refresh, **0** after up - and roughly 730 of those 787 are `__defaults`.

## The premise of #3356 was incorrect

#3356 states that empty `__defaults` is "reintroduced during refresh input reconstruction". The refresh path does not do this. `extractInputsObject` re-emits the key only `if hasOldDefaults` - that is, only when the old inputs already carried it - and recomputes the filtered list. It never invents the key.

The code paths that *do* add an unconditional `__defaults` are on the import/get side: `extractSchemaInputsObject`, and the map branch of `extractSchemaInputs` (which still carries the comment `// TODO: We should probably remove the extraneous reservedkeys.Defaults here.`). `ExtractInputsFromOutputs` routes `isRefresh=true` to `extractInputsObject` and `isRefresh=false` to `extractSchemaInputsObject`, so #3356's guard was applied to the one path that was already correct. This PR leaves the import/get path unchanged.

## Changes

- Remove the `if isRefresh { deleteDefaultsKey(cleanInputs) }` call from `Provider.Read` in `pkg/tfbridge/provider.go`.
- Remove the now-unused `deleteDefaultsKey` / `isEmptyDefaults` helpers and `defaults_test.go`. (`normalizeTFDefaultValue` stays; it is used by `schema.go`.)
- Restore the `"__defaults": []` expectation that #3356 deleted from `TestRegress923`. That existing test is direct evidence of the change: its refresh `Read` legitimately produces `__defaults: []` by narrowing an old `__defaults: ["name"]` after `name` changed, and #3356 discarded that result.
- Add `TestRegress3567`, asserting that a refresh `Read` round-trips `__defaults` at both the top level and inside a nested block. It fails on `main` with exactly `[#["inputs"]["__defaults"]] missing a required value` and `[#["inputs"]["settings"]["__defaults"]] missing a required value`.
- Add `TestRegress3567RefreshPreservesStateInputs`, an end-to-end test that runs a real `up` followed by a real `refresh` and asserts the inputs recorded in state are unchanged.

### A note on why the end-to-end test asserts against state

This bug is invisible to every diff-based assertion in this repo. The bridge returns `DIFF_NONE` for the drift, so both `optrefresh.ExpectNoChanges()` and `optpreview.ExpectNoChanges()` pass while the inputs stored in state have in fact changed. The engine compares raw input bags in `sameSnapshotMutation.mustWrite`, not the provider diff, so state is the only place the regression is observable - and is what determines whether a checkpoint write can be elided.

That is very likely why #3356 looked correct in CI. Any new coverage in this area needs to assert on `ExportStack` rather than on a preview.

## Testing

`go test ./pkg/tfbridge/...` is green. `go test ./pkg/tests/` shows the same single pre-existing failure before and after this change (`TestAccProviderConfig`, which needs `make install_plugins` to fetch `pulumi-resource-random`); I verified this against a clean checkout of `main` rather than assuming.

Both new tests were confirmed to actually exercise the bug: with the fix they pass, and with the strip temporarily reinstated they fail with `[#["inputs"]["__defaults"]] missing a required value` and `refresh must not change the inputs recorded in state` respectively.

## Notes for reviewers

- **Alternative considered and rejected:** making `Check` stop emitting `__defaults: []`. That would remove the marker from the *inputs* side too, flipping `useOldDefault` to the legacy "all old values are defaults" fallback globally. Removing the strip restores pre-#3356 behavior and is the safer direction.
- **Migration:** stacks refreshed on v3.123.0 or later have the marker missing from state. The first `up` after this fix restores it, which is a genuine one-time write, and it then stays stable.
- **PF path out of scope:** `pkg/pf/tfbridge/provider_read.go` has its own equivalent strip, which predates #3356 and was cited as the parity precedent. I have not touched it because I have no PF reproduction, but if the `__defaults` semantics above apply equally there, it likely has the same bug. Happy to follow up if maintainers confirm.
- **#3119 needs to be reopened, and I have identified its actual cause.** Reverting #3356 means #3119 is unfixed - but #3356 never fixed it either, since the strip removed `__defaults` from the refresh side while `Check` kept producing it on the other. The real cause is separate: `extractInputs` has no `IsSecret` case, so a secret-wrapped object fails the `oldInput.IsObject() && newState.IsObject()` check, falls through to `default: return newState`, and refresh substitutes the raw output for the input - discarding any nested `__defaults`. This matches the log in #3119 exactly, where the secret `credentials` object loses its nested `__defaults` while the plain `kafkaCluster` map keeps it. I have reproduced it in this harness:

  ```
  BEFORE refresh: credentials -> {"__defaults":[],"key":"k"}
  AFTER  refresh: credentials -> {"key":"k"}
  plain (non-secret) keeps __defaults in both
  ```

  Note this is broader than `__defaults`: for any secret object, refresh currently replaces the recorded *input* with the raw *output*. I am happy to send that fix as a follow-up PR; I kept it out of this one because it is a different root cause in a much hotter code path and deserves separate review.
- If you would prefer a narrower change - for example keeping the helper but gating it behind the import/get path, or leaving the strip in place and fixing the extraneous producers instead - I am glad to rework this.


corymhall (approved):
@Zamiell thanks for the great work on this!

Zamiell:
No problem, happy to help.

PR #3586, opened 2026-08-20
tfgen: fix Import sections for GCloud IAM-style resources

### LINKED ISSUE #3584: Import docs for GCloud IAM resources are wrong: parseImportCode can't handle space-delimited import IDs
### What happened?

The Import docs for IAM resources are wrong for `pulumi-gcp` — roughly 425 of its ~1438 resources with an Import section still emit a literal `terraform import ...` command, and the large majority of those are the IAM "triad" resources (`*IamBinding` / `*IamMember` / `*IamPolicy`). See pulumi/pulumi-gcp#1473, where the residual after the earlier import-rewriting work was traced to this repo.

The root cause is `parseImportCode` in `pkg/tfgen/docs.go`:

```go
var importCodePattern = regexp.MustCompile(
	`^\s*(?:[%$]\s+)?(?:pulumi|terraform) import[\\\s]+([^.]+)[.]([^\s]+)[\\\s]+([^\s]+)\s*$`)
```

The ID capture group is `([^\s]+)\s*$` — a single whitespace-free token. IAM import IDs are **space-delimited, quoted** identifiers (resource + role + member), so the pattern never matches and `rewriteImportLines` leaves the upstream `terraform import` line verbatim.

This produces three distinct defects on a single rendered page. From [`gcp:accesscontextmanager/accessPolicyIamBinding`](https://www.pulumi.com/registry/packages/gcp/api-docs/accesscontextmanager/accesspolicyiambinding/) (live today, gcp v9.34.0):

> IAM **member** imports use space-delimited identifiers: the resource in question, the role, and the member identity, e.g.
> `$ terraform import google_access_context_manager_access_policy_iam_member.editor "accessPolicies/{{access_policy}} roles/accesscontextmanager.policyAdmin user:jane@example.com"`
>
> IAM **binding** imports use space-delimited identifiers: the resource in question and the role, e.g.
> `$ terraform import google_access_context_manager_access_policy_iam_binding.editor "accessPolicies/{{access_policy}} roles/accesscontextmanager.policyAdmin"`
>
> IAM **policy** imports use the identifier of the resource in question, e.g.
> `$ pulumi import gcp:accesscontextmanager/accessPolicyIamBinding:AccessPolicyIamBinding editor accessPolicies/{{access_policy}}`

1. **Unrewritten `terraform import` lines.** The two space-delimited examples keep the Terraform CLI, the upstream `google_*` resource name, and no Pulumi token. A user following these docs runs the wrong binary.
2. **The one line that *does* rewrite gets the wrong token.** Upstream ships one shared doc page for all three triad members, so the *IamPolicy* example is the only one whose ID is a single token — and the rewriter stamps it with whatever page it is currently generating. On the **Binding** page above, the *policy* example is labeled `...accessPolicyIamBinding:AccessPolicyIamBinding`. The same happens on `storage/bucketIAMMember`, where the policy example renders as `pulumi import gcp:storage/bucketIAMMember:BucketIAMMember editor b/{{bucket}}`. So the only `pulumi import` line on the page is an incorrect command.
3. **Some pages lose the Import body entirely.** On [`gcp:projects/iAMMember`](https://www.pulumi.com/registry/packages/gcp/api-docs/projects/iammember/) the whole Import section collapses to two `->` notes — one of which itself contains a bare ``terraform import google_project_iam_binding.my_project "{{your-project-id}} roles/{{role_id}} condition-title"`` in prose. Every actual import instruction is gone; the page gives no `pulumi import` guidance at all.

### Expected behavior

`parseImportCode` / `rewriteImportLines` should handle space-delimited import IDs, so that

```
$ terraform import google_storage_bucket_iam_member.editor "b/{{bucket}} roles/storage.objectViewer user:jane@example.com"
```

becomes

```
$ pulumi import gcp:storage/bucketIAMMember:BucketIAMMember editor "b/{{bucket}} roles/storage.objectViewer user:jane@example.com"
```

Suggested scope:

- Extend the ID capture to accept a quoted, whitespace-containing ID (`"..."` / `'...'`), keeping the quotes in the emitted `pulumi import` command. This alone fixes the bulk of the ~425 gcp resources.
- Decide what to do with triad pages where the upstream example refers to a *sibling* resource (`_iam_policy` example on the `IamBinding` page). Substituting the current page's token, as happens today, is actively wrong. Options: map the upstream `google_*_iam_{member,binding,policy}` name to the corresponding Pulumi token instead of the page's own token, or drop examples that don't correspond to the resource being generated.
- Handle inline-code `terraform import` mentions in prose (outside fences) — that's what leaks on `projects/iAMMember`.
- Investigate why the `projects/iAMMember` Import body is elided entirely; that's a separate failure from the leak and leaves the resource undocumented for import.

### Reproduction

1. Upstream source: [`website/docs/r/google_project_iam.html.markdown`](https://github.com/hashicorp/terraform-provider-google/blob/main/website/docs/r/google_project_iam.html.markdown) and the `*_iam.html.markdown` pages generally.
2. `make tfgen` in `pulumi/pulumi-gcp` emits `parseImports "gcp:<...>" should not render the string 'terraform' in its emitted markdown.` for each affected resource.
3. Grep the generated `provider/cmd/pulumi-resource-gcp/schema.json` for `terraform import` — ~425 resources, ~800 lines, concentrated in `*IamBinding` / `*IamMember` / `*IamPolicy`.

This is not gcp-specific — any bridged provider whose upstream docs use quoted, space-delimited import IDs hits it (aws, azure, and others with IAM-style composite IDs). <- Not sure why Claude thinks this. Homie better verify this against an actual Registry page to prove it.

### Related

- pulumi/pulumi-gcp#1473 — the originating issue; the two examples cited there are fixed, this is the residual.
- #2251 — broader "replace `terraform` with `pulumi` globally".
- #2486 — multi-line import statements not rendered correctly (adjacent gap in the same rewriter).

### LINKED ISSUE #3585: Docs: `tf`-fenced HCL import blocks are not dropped, and their failed conversion deletes the whole Import subsection
### What happened?

`rewriteImportFence` drops Terraform-only HCL `import { ... }` blocks from the Import section, but only when the fence is tagged ```` ```terraform ````:

```go
// pkg/tfgen/docs.go
if info == "terraform" && looksLikeTerraformImportBlock(code) {
	return nil, false, false
}
```

Upstream overwhelmingly tags these fences ```` ```tf ````, not ```` ```terraform ````. Every `google_*_iam.html.markdown` page in `terraform-provider-google{,-beta}` uses `tf`:

```tf
import {
  id = "{{project_id}} roles/viewer user:foo@example.com"
  to = google_project_iam_member.default
}
```

A `tf` fence matches none of the branches in `rewriteImportFence`, so it falls through to the tail and is emitted verbatim. That has two consequences, and the second is the damaging one.

**1. Terraform-only HCL is rendered on Pulumi docs pages.** An `import {}` block is Terraform 1.5+ config syntax with no Pulumi analogue. It should not be on a Pulumi page at all — that is exactly the judgement `looksLikeTerraformImportBlock` already encodes.

**2. It silently deletes the surrounding docs subsection.** `isHCL` classifies `tf` as HCL:

```go
// pkg/tfgen/docs.go
func isHCL(fenceLanguage, code string) bool {
	switch fenceLanguage {
	case "terraform", "hcl", "tf":
		return true
```

so `convertExamplesInner` attempts an HCL→PCL conversion of the block. That conversion cannot succeed — PCL has no `import` block:

```
$ pulumi convert --from terraform --language typescript
error: main.tf:1,1-7: Unsupported block type; Blocks of type "import" are not expected here.
error: conversion failed
```

and on error `convertExamplesInner` does this:

```go
if err != nil {
	// We do not write this section, ever.
	//
	// We have to strip the entire section: any header, the code
	// block, and any surrounding text.
	stripSection = true
	stripSectionHeader = tfBlock.headerStart
}
```

The whole enclosing subsection — heading, prose, and every other code block under it — is discarded.

### Impact

This is the root cause of the fourth defect reported in #3584 ("Some pages lose the Import body entirely"), which I had filed there as unexplained.

On [`gcp:projects/iAMMember`](https://www.pulumi.com/registry/packages/gcp/api-docs/projects/iammember/#import) (live, gcp v9.34.0) the upstream Import section has five H3 subsections. **Every one of them opens with a `tf`-fenced `import {}` block**, so every one is stripped. What survives is the H2 preamble only — two `->` notes — and the page gives the reader no `pulumi import` guidance at all. Fetching the rendered page confirms it: `Importing IAM members`, `space-delimited`, `Import via resource identity`, and `roles/viewer` each appear **zero** times.

The elision is invisible in local repros that set `PULUMI_CONVERT=0`, which is why it did not show up when #3584 was investigated against the parser alone.

Blast radius is every bridged provider whose upstream docs use `tf`-fenced import blocks. That idiom is spreading: HashiCorp has been adding `import {}` examples alongside `terraform import` commands across the Google, AWS and Azure doc sets since Terraform 1.5.

### Expected behavior

A Terraform-only HCL `import { ... }` block should be dropped from the Import section regardless of whether the fence is tagged `terraform`, `tf`, or `hcl` — and dropping it in `rewriteImportFence` means it never reaches `convertExamplesInner`, so the surrounding subsection is preserved.

Suggested scope:

- Widen the `info == "terraform"` guard in `rewriteImportFence` to cover `tf` and `hcl`. This is the whole fix for the Import section.
- Add a regression test asserting that a `tf`-fenced `import {}` block is dropped while its heading and sibling `sh` fence survive.
- Separately, consider whether an unconvertible HCL block should be able to strip a subsection at all, or whether it should be dropped on its own. Silently deleting documentation on a conversion failure is a sharp edge well beyond import sections — but that is a larger behavioral change and does not need to block this one.

### Reproduction

1. Take the Import section of [`google_project_iam.html.markdown`](https://github.com/hashicorp/terraform-provider-google-beta/blob/main/website/docs/r/google_project_iam.html.markdown).
2. Run it through `parseImports` with example conversion **enabled** (do not set `PULUMI_CONVERT=0`).
3. Every `### Importing ...` subsection is absent from the output. With `PULUMI_CONVERT=0` they are all present — that difference is the bug.

### Related

- #3584 — space-delimited IAM import IDs; this issue is the root cause of that issue's fourth defect.
- #2251 — broader "replace `terraform` with `pulumi` globally".

jkodroff:
Fixes #3584. Fixes #3585.

Upstream documents a family of IAM resources on a single page, using quoted, space-delimited composite IDs. Four defects compounded there, so a page like [`gcp:projects/iAMMember`](https://www.pulumi.com/registry/packages/gcp/api-docs/projects/iammember/#import) renders no usable import guidance at all — the entire Import body is two `->` notes, one of which itself contains a bare `terraform import` command.

### What changed

**1. Quoted, space-delimited IDs.** `parseImportCode`'s ID capture was `([^\s+])` — one whitespace-free token. IAM import IDs are quoted and space-delimited, so the pattern never matched and `rewriteImportLines` left the upstream line verbatim. The capture now also accepts a quoted string, keeping the quotes in the emitted command.

**2. Sibling-resource tokens.** Fixing (1) alone would have made things *worse*. Upstream ships one page for the whole triad, so once every example parses, every example gets stamped with the token of the page being generated — the `IamMember` example on the `IamBinding` page would silently become a wrong `pulumi import`. Each example's own Terraform resource type is now resolved through `ProviderInfo.Resources`, falling back to the page's token when the type is not a resource this provider bridges. That fallback is what keeps every non-sibling case byte-identical to today.

**3. Import examples in prose.** Upstream embeds `terraform import` in backticked spans outside any fence. Those are rewritten now too.

**4. `tf`/`hcl`-fenced HCL `import {}` blocks** (#3585, found while verifying (1)–(3) against the live page). Only ` ```terraform ` was recognized for dropping, but upstream overwhelmingly writes ` ```tf `. Leaving one in did more than render stray HCL — `isHCL` treats `tf`/`hcl` as convertible, `pulumi convert` rejects import blocks (`Unsupported block type; Blocks of type "import" are not expected here`), and `convertExamplesInner` responds by stripping *the entire enclosing subsection*: heading, prose, and sibling examples. Every H3 on the gcp IAM page opens with such a block, which is why the whole body disappeared. This was invisible in local repros that set `PULUMI_CONVERT=0`.

### Result

Full parser path over upstream `google_project_iam.html.markdown` generating `IAMMember`. Before, one wrong `pulumi import` line and two unrewritten `terraform import` lines; after:

```
### Importing IAM members
$ pulumi import gcp:projects/iAMMember:IAMMember default "{{project_id}} roles/viewer user:foo@example.com"
### Importing IAM bindings
$ pulumi import gcp:projects/iAMBinding:IAMBinding default "{{project_id}} roles/viewer"
### Importing IAM policies
$ pulumi import gcp:projects/iAMPolicy:IAMPolicy default {{project_id}}
### Importing Audit Configs
$ pulumi import gcp:projects/iAMAuditConfig:IAMAuditConfig default "{{project_id}} foo.googleapis.com"
```

Each example carries its own resource's token, no `terraform import` survives, and no subsection is stripped.

### Tests

Written failing first, then fixed.

- `TestParseImportCode` — 30 cases pinning the regex. 24 pass on unfixed code and exist as the regression guard, including 10 "must NOT parse" cases (unquoted spaces, unterminated quote, two quoted args, trailing comment, HCL `import {}` block, prose mentions). 6 were red: the quoted space-delimited IDs.
- `TestParseImports_SpaceDelimitedIDs` — the real upstream page, added as testdata.
- `TestParseImports_UnmappedResourceFallsBackToPageToken` — pins the fallback so unbridged siblings keep today's behavior.
- `TestParseImports_DropsHCLImportBlocks` — one subtest per fence tag; `terraform` passed as the guard, `tf` and `hcl` were red.

One golden moved, and it is an improvement: `random-string-full-expected.md`, two prose lines from `terraform import random_string.test test` to `pulumi import random:index/string:String test test`.

`pkg/tfgen`'s 4 remaining failures are pre-existing — verified identical on a clean checkout of `main` in a throwaway worktree, and unrelated (`simple:resource` vs `simple:Resource` casing in example conversion).

### Notes for the reviewer

- **`dynamic` goldens are deliberately untouched.** `TestSchemaGenerationFullDocs` already fails on `main` for both random and fortimanager from unrelated drift. I regenerated both in a worktree and diffed: this change's only effect is the same two prose lines in random, and fortimanager comes out byte-identical. Committing them would drag that unrelated drift into this PR, so I left it to the accept-goldens workflow.
- **One cosmetic leftover.** Dropping a `tf` fence leaves its introducing sentence dangling — "An `import` block (Terraform v1.5.0 and later) can be used…" now leads into the `pulumi import` example. Fixing it needs heuristics for detecting a preamble to a dropped block, which felt like the wrong thing to fold in here. Happy to file it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pkg/tfgen/docs.go
```diff
@@ -1689,31 +1733,71 @@ func looksLikeTerraformImportBlock(code string) bool {
...
+	return ret{
+		Type: strings.Trim(matches[1], " \t\n\\"),
+		Name: matches[2],
+		ID:   matches[3],
+	}, true
+}
+
+// importTokens resolves the Pulumi type token to stamp onto a rewritten import example.
+//
+// Upstream frequently documents a family of resources on a single page - the Google IAM
```
jkodroff: This is overstated. It's just Google Cloud so far as I know.

pkg/tfgen/import_code_test.go
```diff
@@ -0,0 +1,406 @@
+// Copyright 2016-2025, Pulumi Corporation.
```
jkodroff: 2026.

pkg/tfgen/docs.go
jkodroff: Agreed - rescoped to Google Cloud only, in 89b37d67. Google Cloud is the only provider I actually verified shares one upstream page across an IAM triad.

pkg/tfgen/import_code_test.go
jkodroff: Fixed in 89b37d67 - matches the 2016-2026 header on the other files added this year.

pkg/tfgen/docs.go
```diff
@@ -1689,31 +1733,71 @@ func looksLikeTerraformImportBlock(code string) bool {
 //	$ terraform import \
 //	      some_resource.name \
 //	      <some-ID>
+//
+// The ID is either a single whitespace-free token or a quoted string. Quoting is what lets a
+// composite ID contain spaces, as IAM-style resources do:
```
guineveresaenger: This may happen to non-IAM resources also. Example is good; comment should be more generic.

pkg/tfgen/docs.go
```diff
@@ -1657,6 +1688,19 @@ func dropImportSyntaxLine(lines []string) []string {
 	return out
 }
 
+// isHCLFenceInfo reports whether a fence info string tags the block as HCL. Upstream tags
+// these fences `tf` far more often than `terraform`, and the distinction matters: an HCL
+// block left in the Import section is handed to convertExamples, which strips the entire
+// enclosing subsection when the conversion fails.
+func isHCLFenceInfo(info string) bool {
```
guineveresaenger: Any reason we can't use [`isHCL`](https://github.com/pulumi/pulumi-terraform-bridge/blob/main/pkg/tfgen/docs.go#L2697) here?

pkg/tfgen/docs.go
```diff
@@ -1689,31 +1733,71 @@ func looksLikeTerraformImportBlock(code string) bool {
...
+	typeToken string
+	// resources maps upstream Terraform resource names to their bridged Pulumi resources.
+	// It is nil when the provider's resource map is unavailable, in which case every
+	// example falls back to typeToken.
+	resources map[string]*tfbridge.ResourceInfo
+}
+
+// tokenFor returns the token to use for an example importing the given Terraform resource
+// type, falling back to the token of the resource being documented when that type is not a
+// resource this provider bridges.
```
guineveresaenger, line 1803: This would not give the right token when Docs.Source is set to override to another location. For example, this would give the wrong token for `aws_alb`.

pkg/tfgen/import_code_test.go
```diff
@@ -0,0 +1,406 @@
...
+			assert.Contains(t, actual,
+				"$ pulumi import gcp:projects/iAMMember:IAMMember default "+
+					`"{{project_id}} roles/viewer user:foo@example.com"`)
+		})
+	}
+}
+
+// TestParseImports_UnmappedResourceFallsBackToPageToken pins the fallback: when an example
+// names a Terraform resource this provider does not bridge, we keep stamping the token of the
+// page being generated rather than dropping the example.
```
guineveresaenger: I think this comment is unnecessary

pkg/tfgen/import_code_test.go
```diff
@@ -0,0 +1,406 @@
...
+}
+
+// TestParseImports_DropsHCLImportBlocks covers
+// https://github.com/pulumi/pulumi-terraform-bridge/issues/3585.
+//
+// A Terraform `import { ... }` block is Terraform-only config syntax with no Pulumi analogue,
+// so it has to be dropped from the Import section. Upstream tags these fences ```tf far more
+// often than ```terraform, and leaving one in does more than render stray HCL: isHCL treats
+// tf/hcl as convertible, the conversion cannot succeed, and convertExamples then strips the
+// entire enclosing subsection - heading, prose and sibling examples included.
```
guineveresaenger: Can the test name just be `TestParseImports_DropsHCLImportBlocksForAllTerraformLanguageIdentifiers`, and have no comment?

pkg/tfgen/docs.go
```diff
@@ -1516,6 +1518,35 @@ func rewriteImportLines(lines []string, typeToken string) ([]string, bool) {
 	return rewritten, updated
 }
 
+// inlineImportCodeSpan matches a backtick-delimited code span in prose.
+var inlineImportCodeSpan = regexp.MustCompile("`[^`]*`")
```
guineveresaenger: I think it would be great to find a name for this variable that doesn't conflate "code" and "CLI command". We're not looking for a thing of code; we're looking for a shell command.
I'm pointing this out because we _do_ look for real code blocks elsewhere in this file and it would be good to differentiate.

guineveresaenger (changes requested):
I think this is good, although perhaps a bit too large-brush for something that we assume affects only GCP at this point.

There's a few "more correct" tweaks in here, too, that I think other providers can benefit from; however we're trading correctness with a few other resources when we're trying to determine the correct token for each `pulumi import` invocation.

What's the practicality tradeoff using a docs override inside Pulumi GCP as an alternative?

pkg/tfgen/docs.go
jkodroff: Nope, other than Claude being dumb. This is why we cannot leave such matters solely to the machines. Should be fixed.

pkg/tfgen/docs.go
jkodroff: This should be fixed.

pkg/tfgen/docs.go
jkodroff: Agree - "IAM" would've made me think AWS first anyway. Should be fixed.

pkg/tfgen/import_code_test.go
jkodroff: Yes. Should be fixed.

pkg/tfgen/import_code_test.go
jkodroff: Removed.

pkg/tfgen/docs.go
jkodroff: Excellent point. Should be fixed.

jkodroff:
@guineveresaenger 

All your code feedback should be addressed. Regarding your other comments:

> perhaps a bit too large-brush for something that we assume affects only GCP at this point.
> What's the practicality tradeoff using a docs override inside Pulumi GCP as an alternative?

I've never seen this pattern anywhere but in the GCloud provider. If we did this as docs override, it would be hundreds of handwritten (or LLM-written if you like) pages, so I believe it _has_ to be fixed in the bridge if it is to be fixed unless there's another technical solution I'm not considering.

On the other hand, if we're confident that fixing this in the bridge is not something we want to do, I think we should close the upstream Registry issue "won't fix". That wouldn't be my preferred outcome, but it's clearly not worth breaking other providers either.

> There's a few "more correct" tweaks in here, too, that I think other providers can benefit from; however we're trading correctness with a few other resources when we're trying to determine the correct token for each pulumi import invocation.

I _think_ these should be addressed now (`aws_alb` was the only one you cited, but LMK if there are others), so this should be just an additive improvement. I am happy to run any additional tests against other providers if you want so that we feel comfortable about the change.

Re-requesting review now.

pkg/tfgen/docs.go
guineveresaenger: Unfortunately, it's also a bit of a regression. I generated pulumi-gcp with these changes and all of the IAM pages now have multiple `pulumi import....` statements, but all for the same token.

guineveresaenger:
Finally returned to this.

Unfortunately it turns out that sibling token resolution doesn't play well with `docsPointElsewhere`.

I would _love_ to keep the `isHCL` and the inline prose rewrite. Those are real improvements to providers elsewhere - just not GCP.

Token detection remains the sticking point for rewriting these, even when attempting to do it in the GCP provider. I haven't completely exhausted all options for an in-provider docs tweak, but in the meantime can we keep `isHCL` and the inline rewrite as a separate fix?

jkodroff:
Closing in favor of #3626, which carries the two provider-wide fixes from this branch: the `tf`/`hcl` fence handling (#3585) and the inline prose rewrite.

The remaining work here — the quoted-ID regex and sibling token resolution for #3584 — doesn't ship as-is. @guineveresaenger's report was right, and the cause is the `aws_alb` fix I pushed in `fd0ee2d2`: all four GCP triad resources set `Docs.Source: "google_project_iam.html.markdown"`, which is exactly what `docsPointElsewhere` gates on, so that guard disables sibling resolution on precisely the pages this PR was meant to fix. Every import line then gets the page's token. CI missed it because `TestParseImports_SpaceDelimitedIDs` builds a `mockResource` with no `Docs` field — a configuration pulumi-gcp doesn't have.

`docsPointElsewhere` can't tell `aws_alb` from the IAM triad; both are "docs sourced from another entity." And the regex can't land on its own — without sibling resolution it makes the quoted IDs parse and take the page's token, which is worse than not parsing. #3584 stays open as the place to settle the approach, in-provider or in-bridge.

Branch kept at `jkodroff/fix-gcp-iam-import-docs` if any of it is useful later.

PR #3590, opened 2026-08-21
info: carry declared resource timeouts in the GetMapping provider schema

iwahbe:
The provider schema section of the GetMapping payload (`MarshallableProviderShim`) serializes each resource's and data source's schema map, but not the operation timeouts it declares (`shim.Resource.Timeouts`). Consumers that evaluate HCL against the mapping need that set: the SDK implies a `timeouts` block whose attributes are exactly the declared operations (`helper/schema/core_schema.go`), and its value is copied back into state, so `<resource>.timeouts.<op>` is readable only for declared operations, and `<resource>.timeouts` is null when no block is configured. Without the declared set, pulumi-hcl can only hardcode a four-key object (https://github.com/pulumi/pulumi-hcl/pull/439).

- `MarshallableProviderShim` gains `resourceTimeouts` / `dataSourceTimeouts`, keyed by Terraform type, `omitempty`. They are sibling maps rather than part of the per-resource entry because `MarshallableResourceShim` *is* the bare schema map: any extra key there reads as a field to older consumers, while sibling maps leave existing payloads byte-identical and older readers unaffected. `Unmarshal` restores them onto `schema.Resource.Timeouts`, so consumers use the existing `shim.Resource.Timeouts()`.
- Plugin Framework schema-only shims (`pf/proto`, `pf/internal/schemashim`) now return nil from `Timeouts()` instead of panicking: marshalling calls it for every resource (including in a PF provider's `GetMapping`), and a PF resource has no SDK-style declared timeouts — its `timeouts` block, if any, is an ordinary part of the schema. The nested pseudo-resources keep their panics; they are never marshalled at that level.
- Round-trip test in `pkg/tfbridge/info`.

Verified end to end with the pulumi-hcl consumer against a local `replace`: `<resource>.timeouts` typed by the declared operations, null when unset, `default` and data sources included, with tfcompat cases comparing against OpenTofu.

### DIFF
--- a/pkg/pf/internal/schemashim/datasource.go
+++ b/pkg/pf/internal/schemashim/datasource.go
@@ -57,9 +57,7 @@ func (*schemaOnlyDataSource) Importer() shim.ImportFunc {
 	panic("schemaOnlyDataSource does not implement runtime operation ImporterFunc")
 }
 
-func (*schemaOnlyDataSource) Timeouts() *shim.ResourceTimeout {
-	panic("schemaOnlyDataSource does not implement runtime operation Timeouts")
-}
+func (*schemaOnlyDataSource) Timeouts() *shim.ResourceTimeout { return nil }
 
 func (*schemaOnlyDataSource) InstanceState(id string, object,
 	meta map[string]interface{},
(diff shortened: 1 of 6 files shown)


iwahbe:
@pulumi-eon re-review

pose (approved):

PR #3615, opened 2026-09-08
rapidgen: generate only maps of primitives

iwahbe:
The SDK panics in MapFieldWriter when a TypeMap element is a set or a
block. Draw a scalar element type for TypeMap schemas, which is the only
shape the SDK supports at runtime.


iwahbe:

- #3615 ◀
    - #3616
        - #3617
            - #3618
                - #3619
                    - #3620
                        - #3621
                            - #3622
                                - #3496

PR #3616, opened 2026-09-08
rapidgen: draw ForceNew on attributes and blocks

iwahbe:
ForceNew marks a diff as RequiresNew. Draw it on every generated
attribute and block so that planning tests cover replacement diffs.


iwahbe:

- #3615
    - #3616 ◀
        - #3617
            - #3618
                - #3619
                    - #3620
                        - #3621
                            - #3622
                                - #3496

PR #3617, opened 2026-09-08
rapidgen: draw Default on optional scalar attributes

iwahbe:
The SDK fills a Default into the diff when the configuration leaves an
optional attribute unset. Draw a typed Default on optional scalar
attributes, which is the only place the SDK permits one.


iwahbe:

- #3615
    - #3616
        - #3617 ◀
            - #3618
                - #3619
                    - #3620
                        - #3621
                            - #3622
                                - #3496

PR #3618, opened 2026-09-08
rapidgen: draw a StateFunc on string attributes

iwahbe:
A StateFunc rewrites the configured value before the diff is computed,
and the SDK stores the original in the diff's NewExtra and in private
state. Draw an upper-casing StateFunc on string attributes so that this
path is generated.

## Testing

No test in this PR reads the new generator output. It is consumed by `TestPlanResourceChangeMatchesUpstream`, the property-based differential test that #3496 adds at the top of this stack. That test runs the bridge's copy of `PlanResourceChange` and the upstream SDK on every generated resource, so an SDK-invalid schema from this generator fails there.

### DIFF
--- a/pkg/tfshim/sdk-v2/internal/rapid/generators.go
+++ b/pkg/tfshim/sdk-v2/internal/rapid/generators.go
@@ -1,6 +1,8 @@
 package rapidgen
 
 import (
+	"strings"
+
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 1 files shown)


iwahbe:

- #3615
    - #3616
        - #3617
            - #3618 ◀
                - #3619
                    - #3620
                        - #3621
                            - #3622
                                - #3496

iwahbe:
@pulumi-eon please re-review

PR #3619, opened 2026-09-08
rapidgen: give list and set attributes a valid element schema

iwahbe:
The SDK rejects a list or set attribute whose Elem is nil or whose Elem
sets any flag other than Type. The generator left lists without an Elem
and gave sets a full attribute schema, so no generated list or set
attribute passed InternalValidate. Draw a scalar element type instead.
The Elem that was drawn for TypeInt had no effect and is removed.


iwahbe:

- #3615
    - #3616
        - #3617
            - #3618
                - #3619 ◀
                    - #3620
                        - #3621
                            - #3622
                                - #3496

PR #3620, opened 2026-09-08
rapidgen: draw WriteOnly on configurable scalar attributes

iwahbe:
The SDK nulls write-only attributes in the planned state. Draw WriteOnly
on required and optional scalar attributes where the SDK permits it: not
with ForceNew or Default, and not anywhere inside a set block or a
computed block.


iwahbe:

- #3615
    - #3616
        - #3617
            - #3618
                - #3619
                    - #3620 ◀
                        - #3621
                            - #3622
                                - #3496

PR #3621, opened 2026-09-08
rapidgen: draw resource timeouts

iwahbe:
A resource with Timeouts gets a "timeouts" block in its core schema, and
the SDK copies configured timeouts into the planned state and private
state. Draw any subset of the timeout operations on generated resources.

PR #3622, opened 2026-09-08
rapidgen: draw a resource identity schema

iwahbe:
A resource with an Identity gets an identity block, and the SDK carries
identity through SimpleDiff and PlanResourceChange. Draw an identity
schema of scalar attributes on generated resources.

## Testing

No test in this PR reads the new generator output. It is consumed by `TestPlanResourceChangeMatchesUpstream`, the property-based differential test that #3496 adds at the top of this stack. That test runs the bridge's copy of `PlanResourceChange` and the upstream SDK on every generated resource, so an SDK-invalid schema from this generator fails there.


iwahbe:

- #3615
    - #3616
        - #3617
            - #3618
                - #3619
                    - #3620
                        - #3621
                            - #3622 ◀
                                - #3496

corymhall (approved):

PR #3626, opened 2026-09-09
tfgen: drop tf/hcl-fenced import blocks and rewrite import commands in prose

### LINKED ISSUE #3585: Docs: `tf`-fenced HCL import blocks are not dropped, and their failed conversion deletes the whole Import subsection

jkodroff:
# Josh Says

- I read all of Claude's output and put as much polish and tightening on it as possible.
- I think printing the wrong type (IAMMember rather than IAMPolicy) is less wrong and easier for a user to recover from than omitting the section altogether (but it's not great).

# Claude Says

Fixes #3585. Carved out of #3586 at review request — both rewrites are provider-wide and independent of the GCP token work, which stays in #3586.

### 1. `tf`/`hcl`-fenced HCL `import {}` blocks (#3585)

`rewriteImportFence` only recognized ` ```terraform ` when dropping a Terraform-only `import {}` block. But `tf` is also in use upstream: terraform-provider-google's `google_project_iam.html.markdown` tags all seven of its `import {}` blocks ` ```tf `, one under every H3 on the page.

Leaving one in does more than render stray HCL. `isHCL` already treats `tf`/`hcl` as convertible, `pulumi convert` rejects import blocks (`Unsupported block type; Blocks of type "import" are not expected here`), and `convertExamplesInner` responds by stripping **the entire enclosing subsection** — heading, prose, and sibling examples. On that page it takes out the whole Import body.

The check now uses the existing `isHCL`, so it covers the same identifiers as the rest of the file. The `""` case falls through to `guessIsHCL`, which requires a `resource "x" "y" {` header and so never fires on an `import {}` block.

Invisible in local repros that set `PULUMI_CONVERT=0`.

### 2. Import commands embedded in prose

Some upstream docs put `terraform import` in a backticked span outside any fence. Two examples, both already in-tree:

- terraform-provider-random's `string.md.markdown` — "Then importing the resource using `` `terraform import random_string.test test` ``" and again under *Avoiding Replacement* (`test_data/parse-imports/random-string.md`).
- `google_project_iam.html.markdown` — the conditional-bindings note: "make sure to include the title of condition, e.g. `` `terraform import google_project_iam_binding.my_project "…"` ``".

Those lines passed through verbatim, leaving the Terraform CLI and an upstream resource name on a rendered Pulumi page. They now get the page's own token, same as the fenced examples. Spans that aren't import commands are left alone — each is run through `parseImportCode` and only replaced on a match.

### This exposes a wrong token on GCP IAM pages

Fixing #3585 has a side effect on GCP's shared IAM pages: it makes a wrong `pulumi import` line visible that #3584 is responsible for. With the subsection no longer stripped, the IAM policy example renders as:

```
$ pulumi import gcp:projects/iAMMember:IAMMember default {{project_id}}
```

It should read:

```
$ pulumi import gcp:projects/iAMPolicy:IAMPolicy default {{project_id}}
```

The example imports `google_project_iam_policy`, but upstream documents the whole IAM family on one page, and the rewrite stamps every example with the token of the page being generated — here `google_project_iam_member`. The ID is unquoted, so the line already parsed and already took the wrong token; the subsection deletion was hiding it. The other three examples have quoted IDs, still don't parse, and stay as `terraform import`. Resolving each example to its own resource's token is #3584.

### Tests

- `TestParseImportCode` — 23 cases pinning `importCodePattern`. The regex is **unchanged** here; this is a pure regression guard, including 10 "must NOT parse" cases.
- `TestParseImports_DropsHCLImportBlocksForAllTerraformLanguageIdentifiers` — one subtest per fence tag. `terraform` passes as the guard; `tf` and `hcl` were red.
- `TestParseImports_RewritesImportCommandsInProse` — the rewrite, plus four cases that must stay untouched.

One golden moved, and it's an improvement: `random-string-full-expected.md`, the two random prose lines above become `pulumi import random:index/string:String test test`.

`pkg/tfgen` has 6 failures on `main` today (`TestConvertExamples/aws_lambda_function`, `TestTypeSharing`, `TestPropertyDocumentationEdits`, `TestPlainDocsParser` ×3). This branch produces the same 6 — verified by running the suite with the diff stashed in the same checkout.

### Note

Dropping a `tf` fence leaves its introducing sentence dangling — "An `import` block (Terraform v1.5.0 and later) can be used…" now leads into the `pulumi import` example. Fixing that needs heuristics for detecting a preamble to a dropped block; happy to file separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0162nSt3J4Zfk7e3tmrgRzud


jkodroff:
@pulumi-eon You requested changes. Please re-review.

guineveresaenger (changes requested):
Requesting changes because eon's review shouldn't count.

jkodroff:
Note: Guin requested changes at my request. I didn't like it looking like the PR was ready to be merged in my GH inbox.

